PR #1, opened 2018-07-31
*: add initial code and build scripts

crawford:


cmd/main.go
```diff
@@ -0,0 +1,129 @@
...
+	flags struct {
+		kubeconfig string
+		clusterID  string
+		version bool
+	}
+)
+
+func init() {
+	flag.StringVar(&flags.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster. Warning: For testing only, do not use in production.")
+	flag.StringVar(&flags.clusterID, "cluster-id", "", "UUID of the cluster that the channel operator is managing, MUST be set")
```
abhinavdahiya: Should be part of the config that CVO uses?

abhinavdahiya: Should be part of a CVO config ?

pkg/apis/clusterversion.openshift.io/v1/register.go
```diff
@@ -0,0 +1,44 @@
...
+// SchemeGroupVersion is group version used to register these objects
+var SchemeGroupVersion = schema.GroupVersion{Group: apis.GroupName, Version: "v1"}
+
+// Resource takes an unqualified resource and returns a Group qualified GroupResource
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
+
+var (
+	// SchemeBuilder is the scheme builder for MachineConfigPools
```
abhinavdahiya: TODO(later): `MachineConfigPools`

pkg/cincinnati/cincinnati.go
```diff
@@ -0,0 +1,120 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/blang/semver"
+	"github.com/google/uuid"
+)
+
+const (
+	ContentTypeGraphV1 = "application/vnd.redhat.cincinnati.graph+json; version=1.0"
```
abhinavdahiya: TODO(later): all exported fields must have a comment https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

pkg/cincinnati/cincinnati.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type Client struct {
```
abhinavdahiya: same as above.

LICENSE
```diff
@@ -0,0 +1,191 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   Copyright 2014 Red Hat, Inc.
```
wking: nit: I don't think this section is intended to be replaced (e.g. [Kubernetes does not alter it][1]).  It's just a template which you should alter when constructing your own boilerplate (which Kubernetes keeps in per-file headers like [this][2]).  If we don't want per-file headers, we may want the boilerplate statement in a `NOTICE` (like [the installer][3]) or in `README.md`.  Although I'm new to Red Hat, maybe there's a company position on this that I'm not aware of.

[1]: https://github.com/kubernetes/kubernetes/blob/v1.11.1/LICENSE#L190
[2]: https://github.com/kubernetes/kubernetes/blob/v1.11.1/pkg/api/endpoints/util.go#L2-L14
[3]: https://github.com/openshift/installer/blob/154d3b5aa0b9f66aea7106a348dc0f553e1ae273/NOTICE

pkg/cincinnati/cincinnati.go
```diff
@@ -0,0 +1,120 @@
...
+	Version semver.Version
+	Payload string
+}
+
+type edge struct {
+	Origin      int
+	Destination int
+}
+
+func (e *edge) UnmarshalJSON(data []byte) error {
```
abhinavdahiya: Why special unmarshal?

pkg/cvo/cvo.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+func StartWorkers(stopCh <-chan struct{}, config *rest.Config, client cincinnati.Client) {
+	go wait.Until(ensureOperatorStatusExists(config), updateCheckPeriod, stopCh)
+	go wait.Until(checkForUpdate(config, client), updateCheckPeriod, stopCh)
+}
+
+func ensureOperatorStatusExists(config *rest.Config) func() {
+	return func() {
+		client, err := clientset.NewForConfig(config)
```
abhinavdahiya: move this out of returned func, wait.Until will repeatedly call `ensureOperatorStatusExists`. No use of recreating client everytime.

pkg/cvo/cvo.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			glog.Errorf("Failed to create Kubernetes client (API Extensions): %v", err)
+			return
+		}
+
+		_, err = client.ApiextensionsV1beta1().CustomResourceDefinitions().Create(&v1beta1.CustomResourceDefinition{
```
abhinavdahiya: Doesn't handle a case where somebody updates the CRD rather than deleting it.

pkg/cvo/cvo.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil && !errors.IsAlreadyExists(err) {
+			glog.Errorf("Failed to create OperatorStatus CRD: %v", err)
+			return
+		}
+	}
+}
+
+func checkForUpdate(config *rest.Config, cc cincinnati.Client) func() {
+	return func() {
+		payloads, err := cc.GetUpdate("http://localhost:8080/graph", semver.MustParse("1.8.9-tectonic.3"), "fast")
```
abhinavdahiya: TODO(later): remove harded coded logic here.

pkg/cvo/cvo.go
```diff
@@ -0,0 +1,136 @@
...
+		glog.V(4).Infof("Found update payloads: %v", payloads)
+
+		if updateStatus(config, payloads) != nil {
+			glog.Errorf("Failed to update OperatorStatus for ClusterVersionOperator")
+		}
+	}
+}
+
+func updateStatus(config *rest.Config, payloads []string) error {
+	client, err := versioned.NewForConfig(config)
```
abhinavdahiya: same as client in `ensureOperatorStatusExists`

LICENSE
crawford: Huh, I hadn't even noticed that. You're right that the template shouldn't be modified. I'll actually download the license from the web this time.

crawford: @abhinavdahiya I had to rebase the "first" commit (the license). _That_ is why you should always start a repo with an empty commit.

cmd/main.go
crawford: Eventually, yes. For this initial commit, I'm going to leave it as-is.

pkg/cincinnati/cincinnati.go
crawford: Because we are going from an array of indices to a struct.

For example, given the following response:

```
{
  nodes: {...},
  edges: [[0, 1]]
}
```

This needs to be unmarshalled into a struct:

```go
{
  Origin: 0,
  Destination: 1,
}
```

pkg/cvo/cvo.go
abhinavdahiya: You cannot test this function by using `fakeClient`
eg. https://github.com/openshift/machine-config-operator/blob/e34eeb6f06a3e290f349c0450361f3dafccf6ba8/pkg/controller/render/controller_test.go#L77-L97

PR #2, opened 2018-08-01
*: read runtime parameters from CVOConfig

crawford:
Instead of hard-coding parameters and passing others as flags,
periodically read this information from the CVOConfig.


pkg/apis/clusterversion.openshift.io/v1/cvoconfig.go
```diff
@@ -0,0 +1,27 @@
...
+	out := *c
+	c.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
+	return &out
+}
+
+// DeepCopyInto copies the CVOConfig into another CVOConfig. This doesn't
+// actually require a deep copy, but the code generator (and Go itself) isn't
+// advanced enough to determine that.
+func (c *CVOConfig) DeepCopyInto(out *CVOConfig) {
+	*out = *c
```
abhinavdahiya: Does explictly call deepcopy on objectmeta.

PR #6, opened 2018-08-07
cvo: add ability to apply updates

crawford:
When an operator wants to trigger an update, they can write the update
payload and version into the CVOConfig. This will cause the CVO to
update its deployment to use that image instead, rolling the CVO to the
specified version.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -8,6 +8,7 @@ import (
 	"github.com/golang/glog"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/rest"
 )
 
@@ -31,6 +32,12 @@ const (
 //   - checking for updates and writing the available updates into the CVO's
 //     OperatorStatus.
 func StartWorkers(stopCh <-chan struct{}, config *rest.Config) {
+	kubeClient, err := v1.NewForConfig(config)
+	if err != nil {
+		glog.Errorf("Failed to create Kubernetes client (Apps): %v", err)
+		return
+	}
+
 	apiExtClient, err := clientset.NewForConfig(config)
 	if err != nil {
 		glog.Errorf("Failed to create Kubernetes client (API Extensions): %v", err)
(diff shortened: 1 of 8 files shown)


pkg/cvo/cvo.go
```diff
@@ -31,6 +32,12 @@ const (
 //   - checking for updates and writing the available updates into the CVO's
 //     OperatorStatus.
 func StartWorkers(stopCh <-chan struct{}, config *rest.Config) {
+	kubeClient, err := v1.NewForConfig(config)
+	if err != nil {
+		glog.Errorf("Failed to create Kubernetes client: %v", err)
```
abhinavdahiya: nit: This is an `apps/v1` client. so maybe `Failed to create Kubernetes client (apps/v1)` like ones below

pkg/cvo/update.go
```diff
@@ -139,3 +144,71 @@ func updateStatus(cvoClient versioned.Interface, updates []cincinnati.Update) er
...
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxUnavailable: intOrStringPtr(intstr.FromInt(0)),
+					MaxSurge:       intOrStringPtr(intstr.FromInt(1)),
+				},
+			},
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"openshift-core": "cluster-version-operator",
```
abhinavdahiya: nit: previously we used "k8s-app: cluster-version-operator" to tie pods to deployment.

pkg/cvo/update.go
```diff
@@ -139,3 +144,71 @@ func updateStatus(cvoClient versioned.Interface, updates []cincinnati.Update) er
...
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:  "cluster-version-operator",
+						Image: config.DesiredUpdate.Payload,
+					}},
+					NodeSelector: map[string]string{
+						"node-role.kubernetes.io/master": "",
+					},
+					Tolerations: []corev1.Toleration{{
```
abhinavdahiya:
```yaml
tolerations:
- effect: NoSchedule
   key: node-role.kubernetes.io/master
   operator: Exists
```

crawford: I explicitly tolerate everything, just in case.

PR #7, opened 2018-08-14
Use informers for CVO

abhinavdahiya:
Uses informer pattern for CVO
/assign @crawford


cmd/start.go
```diff
@@ -0,0 +1,219 @@
...
+	rootCmd.AddCommand(startCmd)
+	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
+}
+
+func runStartCmd(cmd *cobra.Command, args []string) {
+	flag.Set("logtostderr", "true")
+	flag.Parse()
+
+	// To help debugging, immediately log version
+	glog.Infof("Version: %+v", version.Version)
```
crawford: Use `version.String` instead.

abhinavdahiya: :+1: done

PR #10, opened 2018-08-21
Add CVO sync from payload

abhinavdahiya:
* adds funcs to load manifests from disk
* add resource builder that allows Do on any **lib.Manifest**
* adds syncUpdatePayloadContents that ensures <prefix>/<version>
  contains the <version> update payload
* adds syncUpdatePayload that uses lib.resourcebuilder to generically
  apply the update payload in order.

/assign @crawford @yifan-gu

### DIFF
--- a/lib/manifest.go
+++ b/lib/manifest.go
@@ -1,0 +1,198 @@
+package lib
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// Manifest stores Kubernetes object in Raw from a file.
+// It stores the GroupVersionKind for the manifest.
+type Manifest struct {
+	Raw []byte
+	GVK schema.GroupVersionKind
+
+	obj *unstructured.Unstructured
+}
+
+// UnmarshalJSON unmarshals bytes of single kubernetes object to Manifest.
+func (m *Manifest) UnmarshalJSON(in []byte) error {
+	if m == nil {
+		return errors.New("Manifest: UnmarshalJSON on nil pointer")
+	}
+
+	// This happens when marshalling
+	// <yaml>
+	// ---	(this between two `---`)
+	// ---
+	// <yaml>
+	if bytes.Equal(in, []byte("null")) {
+		m.Raw = nil
+		return nil
+	}
+
+	m.Raw = append(m.Raw[0:0], in...)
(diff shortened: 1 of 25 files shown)


lib/manifest.go
```diff
@@ -0,0 +1,178 @@
...
+
+	obj *unstructured.Unstructured
+}
+
+// UnmarshalJSON unmarshals bytea of single kubernetes object to Manifest.
+func (m *Manifest) UnmarshalJSON(in []byte) error {
+	if m == nil {
+		return errors.New("Manifest: UnmarshalJSON on nil pointer")
+	}
+	if !bytes.Equal(in, []byte("null")) {
```
yifan-gu: nit: why need to special case this? won't this be handled by the Decode()?

lib/manifest.go
```diff
@@ -0,0 +1,178 @@
...
+	m.GVK = ud.GroupVersionKind()
+	m.obj = ud.DeepCopy()
+	return nil
+}
+
+// Object returns underlying metav1.Object
+func (m *Manifest) Object() metav1.Object { return m.obj }
+
+const (
+	rootDirKey = "000"
```
yifan-gu: @abhinavdahiya Could you comment why we need this?

lib/manifest.go
```diff
@@ -0,0 +1,178 @@
...
+//         manifest0
+//         manifest1
+//     01_subdir1/
+//         manifest0
+//         manifest1
+// LoadManifests(<abs path to>/root):
+// returns map
+// 000: [manifest0, manifest1]
+// 00_subdir0: [manifest0, manifest1]
+// 00_subdir0: [manifest0, manifest1]
```
yifan-gu: this might be 01_subdir_1?

lib/manifest.go
abhinavdahiya: update it add comment. look [here](https://github.com/openshift/cluster-version-operator/pull/10/commits/109df8379ae64172ca9ecf4ce2061f71b2a8009c#diff-d1961d4f655d836d50969836a0f51785R96)

lib/manifest.go
abhinavdahiya: done

lib/manifest.go
abhinavdahiya: :+1:

lib/manifest.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Object returns underlying metav1.Object
+func (m *Manifest) Object() metav1.Object { return m.obj }
+
+const (
+	// rootDirKey is used as key for the manifest files in root dir
+	// passed to LoadManifests
+	// It is set to `000` to give it more priority if the actor sorts
+	// based on keys.
+	rootDirKey = "000"
```
smarterclayton: Alternatively we could just force the root dir to be a real dir and ignore files in the root (treat as not manifests).  The only content we had in there in the payload right now was the image mapping and the Cincinnati file.

abhinavdahiya: If the update payload is 
```console
/cincinnati.json
/images.json
manifests/
   ....
   ....
```
The root here would mean manifests dir

abhinavdahiya:
@yifan-gu @smarterclayton any updates ??

lib/manifest.go
yifan-gu: +1 on keeping the manifests dir, seems pretty clean to me and not much overhead.

yifan-gu:
/lgtm
Let go with it!

smarterclayton:
I'm going to test the prototype release build tool against this image today.

PR #14, opened 2018-09-19
Update CVO to flat updatepayload structure

abhinavdahiya:
* updates the `lib/resourcebuilder` to be allow for adding custom resource builders
* add operatorstatus custom resource builder
* update syncs

/cc @yifan-gu


lib/resourcebuilder/interface.go
```diff
@@ -2,37 +2,70 @@ package resourcebuilder
...
+var (
+	// Mapper is default ResourceMapper.
+	Mapper = NewResourceMapper()
 )
 
+// ResourceMapper maps {Group, Version} to a function that returns Interface and an error.
+type ResourceMapper struct {
+	l *sync.Mutex
+
+	gvkToNew map[schema.GroupVersionKind]NewInteraceFunc
```
yifan-gu: what does `gvk` mean?

yifan-gu:
@abhinavdahiya code wise lgtm, but I'm kinda lost at this point on what's going on with the current cvo state... Maybe need another pair of eyes?

lib/resourcebuilder/interface.go
abhinavdahiya: `gvk` : `GroupVersionKind`

PR #15, opened 2018-09-19
Add `image <short-name>` command to return image from updatepayload

abhinavdahiya:
xref: https://jira.coreos.com/browse/CORS-841

/cc @yifan-gu


cmd/image.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/openshift/cluster-version-operator/pkg/cvo"
+)
+
+var (
+	imageCmd = &cobra.Command{
+		Use:     "image",
+		Short:   "Returns image for requested short-name from UpdatePayload",
+		Long:    "",
+		Example: "%[1] image <short-name>",
```
yifan-gu: Does the `%[1]` prints the parent command?

abhinavdahiya: > Does the %[1] print the parent command?

No, will fix that in follow up PR.

PR #24, opened 2018-09-26
fixes CVO dependency on openshift-apiserver

abhinavdahiya:
/cc @deads2k


abhinavdahiya:
this should fix #24 
@mo suggested it is not required as it is already `cluster-admin`

PR #26, opened 2018-09-27
add generic 2-way merge handler.

### LINKED ISSUE #25: CVO needs to run without the service network being functional
The CVO needs to be able to install operators that take care of creating a viable service network.  To do this, it could run with hostnetworking=true and cleverly setting `KUBERNETES_SERVICE_PORT` and `KUBERNETES_SERVICE_HOST` to point to the local kube-apiserver.

If we did this, I think it could come up after the apiserver, controller, scheduler and before anything else.  I bumped into this while trying to get running from a kube control plane.

@mfojtik @knobunc @ironcladlou @abhinavdahiya @smarterclayton @derekwaynecarr

abhinavdahiya:
* Add handler in CVO to apply random `CustomResources` for operators in a way that it stomps all changes, (2-way merge).

/cc @crawford @yifan-gu


pkg/cvo/internal/dynamicclient/client.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// getResourceClient returns the dynamic client for the resource specified by the gvk.
+func (c *resourceClientFactory) getResourceClient(gvk schema.GroupVersionKind, namespace string) (dynamic.ResourceInterface, error) {
+	gvr, namespaced, err := gvkToGVR(gvk, c.restMapper)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get resource type: %v", err)
+	}
+
+	// sometimes manifests of non-namespaced resources
```
crawford: This is formatted strangely.

pkg/cvo/internal/dynamicclient/client.go
```diff
@@ -0,0 +1,90 @@
...
+func gvkToGVR(gvk schema.GroupVersionKind, restMapper *restmapper.DeferredDiscoveryRESTMapper) (*schema.GroupVersionResource, bool, error) {
+	mapping, err := restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to get the resource REST mapping for GroupVersionKind(%s): %v", gvk.String(), err)
+	}
+
+	return &mapping.Resource, mapping.Scope.Name() == meta.RESTScopeNameNamespace, nil
+}
+
+// runBackgroundCacheReset - Starts the rest mapper cache reseting
```
yifan-gu: nit: resetting?

PR #29, opened 2018-10-02
Drop CVO from release manifests

abhinavdahiya:
CVO loads `manifests` dir for manifests that define CVO.
CVO loads `release-manifests` dir for manifests that release-image contains.

CVO still keeps the order `cvo manifests` > `release-image` manifests

/cc @smarterclayton


smarterclayton:
Looks good to me structurally but defer to yifan

PR #30, opened 2018-10-04
Allow overrides for components

abhinavdahiya:
1. update cvoconfig's clusterid 
Making CLusterID a string type allows makes the config better in
terms of human readble.

Added the UnmarshalJSON for string CLusterID with validations that
ensure that uuid type.

This does not prevent users from setting wrong uuid type, but CRDs
don't allow high level validation server side. CVO will not use invalid
uuid.

2. add overrides per component basis
component override allow user to make one of the CVO's owned component
unmanaged.

### DIFF
--- a/pkg/apis/clusterversion.openshift.io/v1/cluster_id.go
+++ b/pkg/apis/clusterversion.openshift.io/v1/cluster_id.go
@@ -1,0 +1,30 @@
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+// UnmarshalJSON unmarshals RFC4122 uuid from string.
+func (cid *ClusterID) UnmarshalJSON(b []byte) error {
+	var strid string
+	if err := json.Unmarshal(b, &strid); err != nil {
+		return err
+	}
+
+	uid, err := uuid.Parse(strid)
+	if err != nil {
+		return err
+	}
+	if uid.Variant() != uuid.RFC4122 {
+		return fmt.Errorf("invalid ClusterID %q, must be an RFC4122-variant UUID: found %s", strid, uid.Variant())
+	}
+	if uid.Version() != 4 {
+		return fmt.Errorf("Invalid ClusterID %q, must be a version-4 UUID: found %s", strid, uid.Version())
+	}
+
+	*cid = ClusterID(uid.String())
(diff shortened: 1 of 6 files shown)


pkg/apis/clusterversion.openshift.io/v1/cluster_id.go
```diff
@@ -0,0 +1,30 @@
...
+
+	uid, err := uuid.Parse(strid)
+	if err != nil {
+		return err
+	}
+	if uid.Variant() != uuid.RFC4122 {
+		return fmt.Errorf("invalid ClusterID %q, must be an RFC4122-variant UUID: found %s", strid, uid.Variant())
+	}
+	if uid.Version() != 4 {
+		return fmt.Errorf("Invalid ClusterID %q, must be a version-4 UUID: found %s", strid, uid.Version())
```
wking: Wait, why do we care about the UUID version?

pkg/cvo/sync.go
```diff
@@ -55,6 +62,19 @@ func (optr *Operator) syncUpdatePayload(config *cvv1.CVOConfig, payload *updateP
...
 }
 
+// getOverrideForManifest returns the override and true when override exists for manifest.
+func getOverrideForManifest(overrides []cvv1.ComponentOverride, manifest lib.Manifest) (cvv1.ComponentOverride, bool) {
+	for idx, ov := range overrides {
+		kind, namespace, name := manifest.GVK.Kind, manifest.Object().GetNamespace(), manifest.Object().GetName()
+		if ov.Kind == kind &&
+			(namespace == "" || ov.Namespace == namespace) && // cluster-scoped objects don't have namespace.
+			ov.Name == name {
+			return overrides[idx], true
```
wking: Why `overrides[idx]` instead of `ov`?

pkg/apis/clusterversion.openshift.io/v1/cluster_id.go
abhinavdahiya: cincinnati cares about it. And CVO needs to only use that specific version of uuid to successfully take to cincinnati

pkg/cvo/sync.go
abhinavdahiya: Don't like using the value from range for anything other than reading....

PR #31, opened 2018-10-05
make operator status more kube-like

deads2k:
@crawford @smarterclayton @abhinavdahiya @derekwaynecarr  @openshift/api-review 

Let's figure out the shape of the status reporting API.  This is WIP where we can work out fine details.  The story so far.....

Kube and OpenShift workload and core resources have a dedicate slice of conditions under `.status.conditions` that individually track `True`, `False`, `Unknown` states.  This avoids multiplying states in a state machine and various ecosystem features have built around them like `kubectl wait` and the CRD spec/status permissions split.  Along with the concept of status being a thing that is completely re-derivable from cluster state.

Since the operatorstatus described here is closely related to the CVO (only consumer), this pull moves the `OperatorStatus` type into `clusterversion.openshift.io`.  Also, the spec/status split for customresources provides a kube-like experience, so I've renamed it to `Operator` with filled in `.status` and an empty `.spec`.  

One key question I had is why the `OperatorStatus` is namespace scoped.  It's fairly easy to keep it there, but in looking at the operators we manage by the CVO, it seems that they are largely cluster-scoped, not largely namespace scoped.

1. kube-apiserver
2. kube-controller-manager
3. kube-scheduler
4. openshift-apiserver
5. openshift-controller-manager
6. dns
7. sdn
8. image registry
9. templates

Given that, naming at a cluster scope seems to make more sense since they do not scale with the number of namespaces available.


abhinavdahiya:
I think we have have previous discussions that there should not be a `Operator` type that has spec that operator might have to read.

OperatorStatus is designed to be write only for operators and read only for CVO and users. Also operatorstatus is separate type because we don;t want operators to vendor any CVO releasted code only the operator status related api and clients.

deads2k:
@crawford @smarterclayton comments?  Milestone 2 requires operators to set this, so we should try to get the API consistent early.

crawford:
> Since the operatorstatus described here is closely related to the CVO (only consumer), this pull moves the OperatorStatus type into clusterversion.openshift.io.

The OperatorStatus is designed to be the common API between all of the operators. Right now, CVO is the only one who cares, but in the future, operators almost certainly will have dependencies between one another. I'd like to keep this in a separate API.

pkg/apis/clusterversion.openshift.io/v1/types.go
```diff
@@ -71,3 +72,102 @@ type Update struct {
...
+	// status holds the information about the state of an operator.  It is consistent with status information across
+	// the kube ecosystem
+	Status OperatorStatus `json:"status"`
+}
+
+// OperatorSpec is empty for now, but you could imagine holding information like, "pause".
+type OperatorSpec struct {
+}
+
+type OperatorStatus struct {
```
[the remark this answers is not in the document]
deads2k: > Why is this being moved back into this API (and leaving the other copy in the other API)?

The copy only exists while we talk about this API.  Before it moves out of WIP, we can remove the other.

deads2k:
> We cannot combine the Status and the Spec into the same object. This was the original design and we pivoted away from it. I don't think there is any need for this parent type.

@crawford `status` is a kubernetes API concept.  Resources have `spec` (intent, empty here) and `status` (actual state that is fully recreate-able based on inspection of cluster or other state).  The concept is so common in kube that you see support for it across multiple components like `kubectl` and `CustomResourceDefinitions`.  Having this API participate in that pattern makes it more consistent with the rest of the stack.

I think the burden is on this type to explain why an empty `spec` cannot work.

deads2k:
> > Since the operatorstatus described here is closely related to the CVO (only consumer), this pull moves the OperatorStatus type into clusterversion.openshift.io.
> 
> The OperatorStatus is designed to be the common API between all of the operators. Right now, CVO is the only one who cares, but in the future, operators almost certainly will have dependencies between one another. I'd like to keep this in a separate API.

I'm not aware of plans to try to have a secondary status object be a means to express and discover dependency relationships.  It seems like we'd want the dependencies to be directly "owned" by producer with consumers being responsible for conforming.

You can see an example of using a common status object not working is the relationship between etcd and all apiservers. Etcd takes care of creating and rotating client cert/keys pairs which are secrets.  They must be consumed by the apiserver operators to keep their etcd connections alive.  If you tried to include that here, you would create a confidential resource that you cannot grant read access to.  If you do not include them here, you have to create the alternative to depending on this resource.  I think we have to go with the latter.

pkg/apis/clusterversion.openshift.io/v1/types.go
```diff
@@ -71,3 +72,102 @@ type Update struct {
...
+// the rest of the cluster.
+// +genclient
+// +k8s:deepcopy-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type Operator struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	// Spec hold the intent of how this operator should behave.
+	Spec OperatorSpec `json:"spec""`
```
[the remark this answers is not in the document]
deads2k: > We cannot combine the Status and the Spec into the same object. This was the original design and we pivoted away from it. I don't think there is any need for this parent type.

[Responded in the main thread with more detail below](https://github.com/openshift/cluster-version-operator/pull/31#issuecomment-428340519).  The tl;dr is that spec/status is so ingrained in kubernetes, the burden is on this API to provide a reason why it cannot be structured like a "normal" resource.  Not the other way around.

crawford:
> I'm not aware of plans to try to have a secondary status object be a means to express and discover dependency relationships.

I wasn't clear. The operator status is designed to allow two operators who have a dependency on one another to communicate; not to declare the dependency. CVO currently enforces ordering in order to work around dependencies, but the eventual goal is to allow operators to coordinate between one another by watching relevant operator statuses.

deads2k:
> I wasn't clear. The operator status is designed to allow two operators who have a dependency on one another to communicate; not to declare the dependency. CVO currently enforces ordering in order to work around dependencies, but the eventual goal is to allow operators to coordinate between one another by watching relevant operator statuses.

If a consumer knows that it has a dependency on a producer, there is no reason to depend on an intermediate object like this.  It could simply go directly to the thing it depends upon.

smarterclayton:
> operators almost certainly will have dependencies between one another. I'd like to keep this in a separate API.

> If a consumer knows that it has a dependency on a producer, there is no reason to depend on an intermediate object like this

So these two statements are pretty close in intent.  Let's work an example:

1. kube-scheduler operator needs to know when kube-apiserver is updated to v1.14 before it allows scheduler to be updated to v1.14.
2. kube-scheduler operator must know two things: that the apiserver is at 1.14, and that the update is "complete".

David, what object were you expecting kube-scheduler operator to look at in a dependency case?

Alex, I assume you are implying that kube-scheduler operator would look at the kube apiserver operator object in the status field and look for a generic "version output"?  So `status.versions = [{"name":"kube-apiserver", "version": "1.14.0"}]`?

deads2k:
> David, what object were you expecting kube-scheduler operator to look at in a dependency case?

I wired up the kube and openshift control planes so that the kube-apiserver-operator publishes version information about itself and the openshift-apiserver-operator reads version information from that location.  It avoids having an intermediate API and the same pattern can be used to collect secret information via secrets.

crawford:
> Alex, I assume you are implying that kube-scheduler operator would look at the kube apiserver operator object in the status field and look for a generic "version output"?

Yep. I'm trying to reduce coupling between the operators. If you know how to read the generic status of one operator, you can read it for all of them. The escape hatch for this scheme is the Extension member of the type.

deads2k:
> Yep. I'm trying to reduce coupling between the operators. If you know how to read the generic status of one operator, you can read it for all of them. The escape hatch for this scheme is the Extension member of the type.

The thing is, if I'm trying to discover that version, I'm already coupled.  This doesn't reduce the coupling, it just makes another object that both operators have to go through.

Ultimately, I'll move the object to a different group to fix the internal schema of said object.  But I think the idea of a "universal operator information sharing space" is going to be poorly used internally and never used externally.

smarterclayton:
So certain collaboration between operators is going to be confusing if we try to make it generic here - the status of certain config flags or whatever belongs in a different object (which it all seems we're on board with).

Re versions - it might be good for us to talk about that in a design doc, but a few thoughts below:

 will agree with Alex here that the whole point of the CVO is to take a top level version and ensure specific, known versions are showing up in other places.  It's reasonable that a human will want to look at operator status and at a glance know whether operator X is at the "right" level (whatever that is).  I don't know that means that the operators have to report that, but who better than the kube operators to know what version of kube exists and report that back to an admin?

The counter argument to that is that most or almost all of the individual versions really need to parrot the top level version (i.e. Openshift 4.2.1) while the versions we're talking about here might be kube (v1.13.1) or prometheus (2.4.0) or etcd (3.2.29).  

I haven't seen a design for this yet, but I am worried about components outside of the master team family having to look at an object geared for the kube-apiserver.  Within master team - sure, since technically the master team has a "cluster-kube-operator" that just happens to be split into 5 processes - but outside the master team I would have expected an operator to ask "is the control plane at v1.13", not "hey kube-apiserver, what version do you think you're at".

smarterclayton:
Re operator spec:

I think the operator spec here isn't the config of the operator - it's the control point that the SLO respects for specific, limited generic operator behavior.  I.e. it's a generic spec that all operators are expected to follow for things like `paused` (which is `managementState` here).  It would not include things like config ever.

Re spec and status:

Given the kube conventions, I argued for this which is the primary goal of the object is to be a generic interface for operator level actions, which includes generic status (human focused) and *possibly* an abstract version thing (which obviously we have disagreement on).  The status is single writer from the operator.  The spec would only be things that were necessary for consistent operational behavior across all SLOs, like pausing the operator.  

Re: status being namespace vs cluster scoped

Just from a purely mechanical level, I'd rather run `oc get operators` and see a list than `oc get operators --all-namespaces`.  I think owner refs could handle out of namespace cleanup, although I agree namespace cleanup is better.   I do see the possibility that we want `oc adm status` that combines fetching operators (in which case `oc get operators` is less of a factor) with also fetching things that might show up on the dashboard including CVO version, firing alerts, etc.

deads2k:
@crawford updated per our conversation on Friday.

1. commit message updated to be more specific
2. OperatorStatus name is untouched, but a status substruct with separate conditions has been added.
3. CRD updated to have a /status subresource

After this merges, I plan to open more pulls to push further (name of resource, possibly scoping, possibly group), but I think we agree this far and it will help make integration from other teams easier and reduce overall change.

pkg/apis/operatorstatus.openshift.io/v1/types.go
```diff
@@ -25,52 +25,81 @@ type OperatorStatus struct {
...
-	// reconciling its operands and is waiting for changes.
-	OperatorStatusConditionTypeDone OperatorStatusConditionType = "Done"
-
-	// OperatorStatusConditionTypeDegraded indicates that the operator has
-	// encountered an error that is preventing it from working properly.
-	OperatorStatusConditionTypeDegraded OperatorStatusConditionType = "Degraded"
+	// OperatorAvailable indicates that the binary supported by the operator is functional and available in the cluster.
+	OperatorAvailable OperatorStatusConditionType = "Available"
+
+	// OperatorProgressing indicates that the operator is actively binary supported by the operator is functional and available in the cluster.
```
smarterclayton: Looks like the wrong godoc.

Is progressing supposed to be "I'm reconciling towards the desired state"?  Or reflect the operator status itself?

I think it's acceptable to have different condition types when the definition is not the same.  For instance, Deployment uses Progressing False to represent "failing" because deployment has no perm fail state.

The removal of "Done" means that a consumer doesn't have an unambiguous signal about when all possible changes have been applied.  I think we should have a positive, affirmative condition on "up to date".

pkg/apis/operatorstatus.openshift.io/v1/types.go
```diff
@@ -25,52 +25,81 @@ type OperatorStatus struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
 
+	// Spec hold the intent of how this operator should behave.
+	Spec OperatorSpec `json:"spec""`
+
```
abhinavdahiya: Why prefix the status with type and not spec?

Also there is an extra `"`

pkg/apis/operatorstatus.openshift.io/v1/types.go
```diff
@@ -25,52 +25,81 @@ type OperatorStatus struct {
...
-	OperatorStatusConditionTypeWorking OperatorStatusConditionType = "Working"
-
-	// OperatorStatusConditionTypeDone indicates that the operator has finished
-	// reconciling its operands and is waiting for changes.
-	OperatorStatusConditionTypeDone OperatorStatusConditionType = "Done"
-
-	// OperatorStatusConditionTypeDegraded indicates that the operator has
-	// encountered an error that is preventing it from working properly.
-	OperatorStatusConditionTypeDegraded OperatorStatusConditionType = "Degraded"
+	// OperatorAvailable indicates that the binary supported by the operator is functional and available in the cluster.
```
abhinavdahiya: Can explain `binary supported by the operator` ?

pkg/cvo/status.go
```diff
@@ -84,3 +128,64 @@ func checkForUpdate(config cvv1.CVOConfig) ([]cincinnati.Update, error) {
 	}
 	return cincinnati.NewClient(uuid).GetUpdates(string(config.Upstream), config.Channel, version.Version)
 }
+
+func SetOperatorStatusCondition(conditions *[]osv1.OperatorStatusCondition, newCondition osv1.OperatorStatusCondition) {
```
abhinavdahiya: Can you move these to `apis`

install/0000_00_cluster-version-operator_01_cvoconfig.crd.yaml
```diff
@@ -15,6 +15,9 @@ spec:
       storage: true
   # either Namespaced or Cluster
   scope: Namespaced
+  subresources:
```
smarterclayton: The larger implication here is a formalization of a more rigid hierarchy of "truth" for cluster configuration:

1. top level configuration (tectonic config -> ImageConfig, NetworkConfig, CloudConfig) representing cluster level truths in functional/component terms rather than operator terms
2. secondary input configurations like secrets or raw config, or lower level component configuration (the pull secret, the cloud provider secret, or the openshift-apiserver config objects)
3. the CVOConfig -> ClusterVersion object (discussed previously), which is the source of truth for the desired version state of the cluster
4. specific API objects for specific operators like ClusterIngress, all the network SDN config, network policy, etc

Those are the configuration inputs to a cluster.  There is a parallel hierarchy of operators:

1. Cluster Version Operator -> ensures all "Cluster Operators" (I like this name better in colloquial use the more I use it) are functional and up to date, based on Cluster Version
2. Second-Level-Operators -> Cluster Operator are responsible for materializing the config hierarchy at a cluster scope
3. All cluster operators are expected to report operatorstatus/operator.status, which is how an administrator / developer performs quick triage of the state of the cluster

By making this change we are implying that an operator is a controllable object that has state that is reconciled with the cluster, and that the controllable object is the common interface to "cluster operators".  I think that having a consistent set of control knobs for operators is the right thing to do (there are zero today, but the first one would be something like "paused"), rather than having to duck type control knobs on config (since the config hierarchy is not operator focused but functional areas focused).  

All "cluster operators" should report status.  It should be possible to list all cluster operators.  It should be possible to build a UI that summarizes the rough relationships between cluster operators (although in most cases that should not be necessary).

It should be the job of the CVO to summarize operator status underneath the state of an upgrade.  Ultimately this is the level of truth:

1. The CVO takes a ClusterVersion and ensures all cluster operators are up to date
2. The CVO summarizes the result of that operation and ensures ClusterVersion reports it
3. An admin or user interfaces with ClusterVersion to effect changes and see their outcome.  
4. A developer or support person looks at operator status.
5. If an operator fails to update, there should be a condition on ClusterVersion which indicates that updates are not progressing.
6. The CVO should be reporting metrics for consumption by telemetry for input to SRE about the state of the update - some of those metrics will be summarizations of the current state like "upgrade to 3.1.5 in progress".  Others will be "failing operator status".

smarterclayton: Since the CVO is a reconciler, I think it's appropriate to talk about CVOConfig/ClusterVersion having a spec "desired outcome" and a status "actual state".  The state the CVO might present to an end user via ClusterVersion would be different than the status passed to OperatorStatus/Operator.status - both are for humans, but one reflects the desired action in cluster version, and the other reflects the desired state the operators themselves are trying to achieve.

pkg/apis/operatorstatus.openshift.io/v1/types.go
deads2k: > Why prefix the status with type and not spec?

`OperatorStatus` conflicted, but sure.

pkg/cvo/status.go
deads2k: > Can you move these to `apis`

No.  We don't put code in api packages.

deads2k:
Initial comments addressed.  Typo, type name, doc.

pkg/cvo/status.go
abhinavdahiya: But these are tied to the api. they are definitely not part of the `pkg/cvo` package api.

abhinavdahiya:
@deads2k https://github.com/openshift/cluster-version-operator/pull/31#discussion_r225261923 these functions don't seem to belong to `pkg/cvo`'s api. 

Where would be rather recommend we keep them. Some Kubernetes api packages do have versioned helper functions in the apis.

deads2k:
> @deads2k [#31 (comment)](https://github.com/openshift/cluster-version-operator/pull/31#discussion_r225261923) these functions don't seem to belong to `pkg/cvo`'s api.
> 
> Where would be rather recommend we keep them. Some Kubernetes api packages do have versioned helper functions in the apis.

I think that in the next month we will move this to openshift/api.  Once we do that, then they will have a home inside of library-go near https://github.com/openshift/library-go/tree/master/pkg/operator/v1alpha1helpers .  We can simulate that structure now if you like.  I didn't think that it mattered either way.

abhinavdahiya:
I am fine if we move these helper functions anywhere out of `pkg/cvo`; `lib` or anything.

abhinavdahiya:
@crawford / @smarterclayton can you take another look and `/lgtm`

install/0000_00_cluster-version-operator_01_cvoconfig.crd.yaml
ericavonb: Are you familiar with how OLM works? It would be much nicer if there was a united view of operators, and OLM already figured a lot of this out already (like having a UI 🙂).

> By making this change we are implying that an operator is a controllable object that has state that is reconciled with the cluster, and that the controllable object is the common interface to "cluster operators". 
Yes, it's called the "ClusterServiceVersion" in OLM.
The operators even report back the status of their components: https://github.com/operator-framework/operator-lifecycle-manager/blob/d99daea0086cf92976617dd9326d0ed2b7ba85cf/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go#L2

I'm not saying OLM solved all of the pieces needed for bootstrapping, but quite a bit of the stuff being discussed here.

smarterclayton: Yes.  These are going to be a deliberate subset of what an OLM operator may support.  The conditions on CSV status are very relevant to CV.  The spec is not.

smarterclayton: The biggest difference is that CSV needs conditions, and it's possible that we'll want to drive a phase for CV or something similar to simplify user input (although I'm leaning towards not having a phase and instead displaying human readable status instead).

PR #35, opened 2018-10-09
switch workload resource to daemonset

deads2k:
This change switches the CVO to be installed as a daemonset instead of a deployment so that it can temporarily bypass the scheduling problem it faces as a deployment.  This allows us to 

1. avoid having the static pod which cannot be managed via `kubectl` 
2. ensures that a CVO rollout to a new image doesn't get backleveled by a static pod
3. ensures that the CVO is automatically balanced on new masters so that should one kube-apiserver go down, you don't lose all three CVO pods (deployments don't do this)

If we later want to switch back to deployments we can, but this change eliminates an unnecessary static pod and places the CVO at the right logical point in the installation flow.

/assign @abhinavdahiya 

@ironcladlou @smarterclayton @mfojtik

### DIFF
--- a/install/0000_00_cluster-version-operator_03_daemonset.yaml
+++ b/install/0000_00_cluster-version-operator_03_daemonset.yaml
@@ -1,9 +1,11 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
   name: cluster-version-operator
   namespace: openshift-cluster-version
 spec:
+  updateStrategy:
+    type: RollingUpdate
   selector:
     matchLabels:
       k8s-app: cluster-version-operator


smarterclayton:
I'm really curious now why this wasn't scheduling - Ryan P was going to look at that as well.

derekwaynecarr:
in k8s 1.12, isnt the scheduler responsible for scheduling daemonset pods since ScheduleDaemonSetPods feature is on by default?

derekwaynecarr:
see: https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#scheduled-by-daemonset-controller-disabled-by-default-since-1-12

abhinavdahiya:
@ravisantoshgudimetla why is the CVO deployment not scheduled? when

https://github.com/openshift/cluster-version-operator/blob/4f5564f43badfd100d998d56f9240b4ca9098d4f/install/0000_00_cluster-version-operator_03_deployment.yaml#L44-L45

tolerates all taints as from doc https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

ravisantoshgudimetla:
@abhinavdahiya - I think we are still in kube1.11 in our installer where DS controller acts as separate scheduler(not exactly, but doesn't rely on default kube-scheduler for scheduling pods onto nodes). Having said that, the toleration that you mentioned should have been good for deployment. I will set up a 4.0 cluster to see why it is not working. As of now, my guess is that it could be because of `unschedulable` field.

abhinavdahiya:
@deads2k i'm going to test this locally.

abhinavdahiya:
seems like all operators were created.
```
oc get pods,nodes --all-namespaces
NAMESPACE                              NAME                                                                  READY     STATUS              RESTARTS   AGE
kube-system                            pod/bootstrap-kube-apiserver-adahiya-0-bootstrap                      1/1       Running             0          3m
kube-system                            pod/bootstrap-kube-controller-manager-adahiya-0-bootstrap             1/1       Running             0          3m
kube-system                            pod/bootstrap-kube-scheduler-adahiya-0-bootstrap                      1/1       Running             0          3m
kube-system                            pod/kube-apiserver-bdttl                                              1/1       Running             0          3m
kube-system                            pod/kube-controller-manager-b6985978d-q87z5                           0/1       ContainerCreating   0          3m
kube-system                            pod/kube-dns-787c975867-knr7l                                         0/3       ContainerCreating   0          3m
kube-system                            pod/kube-flannel-6hmf9                                                2/2       Running             0          43s
kube-system                            pod/kube-flannel-cghcr                                                2/2       Running             0          43s
kube-system                            pod/kube-proxy-7wpmk                                                  1/1       Running             0          3m
kube-system                            pod/kube-proxy-p8dlg                                                  1/1       Running             0          3m
kube-system                            pod/kube-scheduler-78d86f9754-xqbkh                                   0/1       ContainerCreating   0          3m
kube-system                            pod/pod-checkpointer-9l9jp                                            1/1       Running             0          3m
kube-system                            pod/pod-checkpointer-9l9jp-adahiya-0-master-0                         1/1       Running             0          3m
kube-system                            pod/tectonic-network-operator-l6lwc                                   1/1       Running             0          3m
openshift-cluster-api                  pod/machine-api-operator-5d85454676-t8vvn                             0/1       Pending             0          1m
openshift-cluster-version              pod/cluster-version-operator-479tp                                    1/1       Running             0          3m
openshift-core-operators               pod/openshift-cluster-openshift-apiserver-operator-5d86df99d9-j6rlh   0/1       ContainerCreating   0          1m
openshift-core-operators               pod/openshift-cluster-openshift-controller-manager-operator-6cfcf9k   0/1       Pending             0          1m
openshift-core-operators               pod/openshift-service-cert-signer-operator-76d49565c-9bjp8            0/1       ContainerCreating   0          1m
openshift-machine-config-operator      pod/machine-config-operator-756cb6b868-4grfn                          0/1       Pending             0          1m
openshift-operator-lifecycle-manager   pod/catalog-operator-5c47854c6f-p47fs                                 0/1       Pending             0          1m
openshift-operator-lifecycle-manager   pod/olm-operator-5b46cfbc9c-9mwnj                                     0/1       Pending             0          1m
openshift-operator-lifecycle-manager   pod/package-server-65b9bb8db-74c52                                    0/1       Pending             0          1m

NAMESPACE   NAME                       STATUS    ROLES       AGE       VERSION
            node/adahiya-0-bootstrap   Ready     bootstrap   4m        v1.11.0+d4cacc0
            node/adahiya-0-master-0    Ready     master      4m        v1.11.0+d4cacc0
```
/lgtm

abhinavdahiya:
@deads2k I personally like the current guidelines. 
An great example is @wking comments like this https://github.com/openshift/installer/commit/aa013116cfc938d0f24f9739591fd14be11c8b91
The title is short and informative and the descriptions are detailed and useful for future.

ravisantoshgudimetla:
@abhinavdahiya - Just confirmed it on my cluster locally - Seems, the issue is what I was suggesting earlier, DS controller checks for only essential predicates(which doesn't include `CheckNodeConditions` Predicate) in 1.11(https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/controller/daemon/daemon_controller.go#L1410-#L1450)

But when a normal pod or pod created from deployment is scheduled, it goes through `CheckNodeConditions` Predicate which will fail causing the scheduler to throw an error like `default-scheduler  0/2 nodes are available: 1 node(s) didn't match node selector, 1 node(s) were not ready.`

Now, here comes the interesting part, kube 1.12 has by default enabled a feature called `TaintNodesByCondition` which removes this predicate(https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithmprovider/defaults/defaults.go#L184-#L210)

 while scheduling, rather node controller applies a particular taint called `node-not-ready`. Now, if the deployment is able to tolerate any taint this should work. On my latest 4.0 install cluster, I tested this scenario by adding `feature-gates=TaintNodesByCondition=true` to kube-scheduler static pod file and the CVO deployment pods got scheduled. So, I think for now, we can continue having workloads as DS and once the 1.12 rebase lands we can switch to deployments.

abhinavdahiya:
@ravisantoshgudimetla thanks a lot for taking a look at this.

PR #36, opened 2018-10-10
lib: add DaemonSet as managed resource

abhinavdahiya:
Few operators like the network operator and CVO itself are going to have to be
daemonsets to workaround the scheduler bug (not scheduling deployment pods in
hostnetwork when node is reporting Network NotReady)

/cc @crawford


deads2k:
> is this scheduler bug being tracked somewhere? @aveshagarwal @ravisantoshgudimetla

@sjenning ryanj volunteered, but we haven't heard anything back yet.  The bug is here: https://bugzilla.redhat.com/show_bug.cgi?id=1638132 .  We're trying to ensure that we don't end up stuck with sneak dependencies from a static pod CVO and there is a fair amount of difficulty involved in local development when the CVO is a static pod, so this is part of the process of getting unstuck.

PR #38, opened 2018-10-11
Revert "switch workload resource to daemonset"

deads2k:
Reverts openshift/cluster-version-operator#35

In the 4.0 release, we want a deployment.  There is a suspect scheduling bug that must be resolved to switch back: https://bugzilla.redhat.com/show_bug.cgi?id=1638132 .

We made the change so we could remove a static pod that we don't want long term and ensure that bootstrapping sequence doesn't end up with accidental dependencies because of that.  

/hold

hold until the bug is fixed.

/assign @sjenning 
@sjenning I'm not sure who will be driving this change back in, but it's a thing we'll need to fix in 4.0 to avoid management issues related to a daemonset operator.

@abhinavdahiya fyi

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -1,11 +1,9 @@
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   name: cluster-version-operator
   namespace: openshift-cluster-version
 spec:
-  updateStrategy:
-    type: RollingUpdate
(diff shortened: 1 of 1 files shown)


deads2k:
/test all

the scheduler switch to hostnetwork.  This may work.

/hold cancel

/assign @abhinavdahiya

abhinavdahiya:
the installer's bootstrap node is running cvo as static pod. so this might not be actually testing the notschedulable behavior this. but since the CI is green. i'm fine merging this.

/lgtm

PR #39, opened 2018-10-12
WIP: cvo: Allow payloads to retrieve images from an alternate location

smarterclayton:
NOTE: Not a complete implementation, just creating prototype code I can test in an install.  Design coming after I poke around.

JIRA: https://jira.coreos.com/browse/CORS-715

Allow a payload to specify an alternate image repository to use instead of
using the location the payload specifies. When manifests are being
pre-processed, the following algorithm is applied:

1. search the manifest for any string that matches the name of an image
   defined in image-references (with leading and trailing word breaks) 
2. parse the image location, replacing all but the digest with the new
   payloadRepository 
3. replace the string in the manifest with the new location

When the CVO applies the payload, images will be fetched from the override
location rather than the location the payload defined. However, since we
preserve the digest, we ensure that the contents still match.

TODO:

* [ ] explore how installer can accept the mirror location (either accept a mirror location on install or we just force people to rewrite during mirroring - advantages to either)
* [ ] consider moving payload repository to the root of CVOConfig (disadvantage is you can't turn off mirroring by selecting an update from the cloud)

### DIFF
--- a/pkg/apis/clusterversion.openshift.io/v1/types.go
+++ b/pkg/apis/clusterversion.openshift.io/v1/types.go
@@ -70,4 +70,11 @@ type CVOStatus struct {
 type Update struct {
 	Version string `json:"version"`
 	Payload string `json:"payload"`
+
+	// PayloadRepository is an optional field that instructs the ClusterVersionOperator
+	// to load images in the payload from a specific image repository, instead of
+	// from the location the payload specified. The images are still loaded via their
+	// unique identifier (digest) in order to prevent the linked content from being
+	// altered. This value matches the location you mirror an update to.
+	PayloadRepository string `json:"payloadRepository"`
 }
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -190,6 +190,18 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:d7da23741e2be291bda3be66c16887a9bbebf48a1ea133a5fcd210c02b8ccee0"
+  name = "github.com/openshift/library-go"
+  packages = [
+    "pkg/image/internal/digest",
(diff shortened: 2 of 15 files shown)


pkg/apis/clusterversion.openshift.io/v1/types.go
```diff
@@ -70,4 +70,11 @@ type CVOStatus struct {
 type Update struct {
 	Version string `json:"version"`
 	Payload string `json:"payload"`
+
+	// PayloadRepository is an optional field that instructs the ClusterVersionOperator
+	// to load images in the payload from a specific image repository, instead of
+	// from the location the payload specified. The images are still loaded via their
+	// unique identifier (digest) in order to prevent the linked content from being
+	// altered. This value matches the location you mirror an update to.
+	PayloadRepository string `json:"payloadRepository"`
```
abhinavdahiya: I would assume that this was global config in CVOConfig rather than each Update object.

wking:
> When the CVO applies the payload, images will be fetched from the override location rather than the location the payload defined.

This breaks down if for some reason we end up with a partial mirror (e.g. someone forgot to list an image in `image-references` before an `oc adm release mirror ...` call, or the CVO bumps an operator to a new version but doesn't mirror its new image to the override registry).  It would be more resilient (at least for folks who are not air-gapped) if we un-qualified the images instead and set up fallback registries (e.g. "check the local mirror, fall back to `quay.io` and then fall back again to `docker.io`", etc.).  Image digests and CAS will ensure we get the right image regardless of where we find it.  But maybe we're comfortable relying on complete mirroring?

abhinavdahiya:
closing because this is too old now.

/close

PR #40, opened 2018-10-16
cvo/internal: catch up the operator status handler to PR #31

abhinavdahiya:
/cc @wking

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -30,12 +30,12 @@ func init() {
 		panic(err)
 	}
 
-	osMapper.RegisterGVK(osv1.SchemeGroupVersion.WithKind("OperatorStatus"), newOperatorStatusBuilder)
+	osMapper.RegisterGVK(osv1.SchemeGroupVersion.WithKind("ClusterOperator"), newClusterOperatorBuilder)
 	osMapper.AddToMap(resourcebuilder.Mapper)
 }
 
-// readOperatorStatusV1OrDie reads operatorstatus object from bytes. Panics on error.
-func readOperatorStatusV1OrDie(objBytes []byte) *osv1.ClusterOperator {
+// readClusterOperatorV1OrDie reads clusteroperator object from bytes. Panics on error.
+func readClusterOperatorV1OrDie(objBytes []byte) *osv1.ClusterOperator {
 	requiredObj, err := runtime.Decode(osCodecs.UniversalDecoder(osv1.SchemeGroupVersion), objBytes)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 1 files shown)


wking:
Do we really need this?  `OperatorStatus` objects still exist (as I understand it, e.g. see [here][1]).  And `OperatorStatus` seems appropriate for registering `newOperatorStatusBuilder`.  Maybe we need additional logic to register a builder for `ClusterOperator`?

[1]: https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/pkg/cvo/sync.go#L88-L102

abhinavdahiya:
https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/pkg/cvo/sync.go#L88-L102 only exists to not break operators that are using the old CRD. that will be removed very soon.

wking:
> ... only exists to not break operators that are using the old CRD...

So then should this PR be renaming `newOperatorStatusBuilder` -> `newClusterOperatorBuilder` and similar too?  Or if you really want to use `WithKind("ClusterOperator")` to register `newOperatorStatusBuilder`, can you explain why the names should no longer match?  And if all of this is terribly obvious to folks with more experience with CRDs (i.e. not me ;), feel free to ignore ;).

abhinavdahiya:
> So then should this PR be renaming newOperatorStatusBuilder -> newClusterOperatorBuilder and similar too? Or if you really want to use WithKind("ClusterOperator") to register newOperatorStatusBuilder, can you explain why the names should no longer match?

The builder name `newOperatorStatusBuilder` seemed still valid to me because it only operates on the status for `clusteroperator` which is operator's status.

abhinavdahiya:
@wking https://github.com/openshift/cluster-version-operator/pull/40#issuecomment-430327249 WDYT?

wking:
> The builder name `newOperatorStatusBuilder` seemed still valid to me because it only operates on the status for `clusteroperator` which is operator's status.

Sorry for the delay.  I'm trying to wrap my head around `operatorstatus.go`, but here's where I'm at now:

* I think the log messages [here][1] and [here][2] also need `OperatorStatus` -> `ClusterOperator` now that we're getting the information from inside a `ClusterOperator` resource.
* I think `readOperatorStatusV1OrDie` should be renamed to `readClusterOperatorV1OrDie` because it's returning an `*osv1.ClusterOperator`.
* I think `OperatorstatusV1Client` is a strange name for a client whose only methods are `RESTClient` and `ClusterOperators`.
* I am ambivalent about `newOperatorStatusBuilder`.
* I'm fine with `waitForOperatorStatusToBeDone`.

Of the first three points, I'm fine with you picking them up here or punting them to follow-up work.  Thoughts?

[1]: https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/pkg/cvo/internal/operatorstatus.go#L84
[2]: https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/pkg/cvo/internal/operatorstatus.go#L109

abhinavdahiya:
@wking 

> I think the log messages here and here also need OperatorStatus -> ClusterOperator now that we're getting the information from inside a ClusterOperator resource.

Done

> I think readOperatorStatusV1OrDie should be renamed to readClusterOperatorV1OrDie because it's returning an *osv1.ClusterOperator.

Done

> I think OperatorstatusV1Client is a strange name for a client whose only methods are RESTClient and ClusterOperators.

That is generated code as the group for the api is `operatorstatus.openshift.io`, can't do anything...

> I am ambivalent about newOperatorStatusBuilder.

Changed to newClusterOperatorBuilder

> I'm fine with waitForOperatorStatusToBeDone.

Yeah, keeping it as is.

PR #45, opened 2018-10-29
Split ClusterVersion into spec and status, write end-user condition messages, report metrics

smarterclayton:
This continues work in #44 to move fields to spec and status.  Focuses on providing an end user focused view of the cluster version operators work.

* Merge CVOStatus into ClusterVersion.Status
* Move existing fields in ClusterVersion into Spec
* Make ClusterVersion cluster scoped
* Add custom printer columns to the CRDs
* Add godoc to the object
* Improve all condition output
* Add a sync unit test that exercises the core CVO sync loop
* Add prometheus metrics for core logic


smarterclayton:
The commits here need to be broken up more, and I need to do more testing, but this is getting close to rounding out the "CVO reports more accurate status, CV reports desired + actual state, CO is a copy, and users can interact with CV and CO more cleanly" story.

lib/resourceapply/cv.go
```diff
@@ -15,36 +17,46 @@ import (
...
 
-func ApplyOperatorStatus(client osclientv1.ClusterOperatorsGetter, required *osv1.ClusterOperator) (*osv1.ClusterOperator, bool, error) {
+func ApplyClusterOperatorStatus(client osclientv1.ClusterOperatorsGetter, required *osv1.ClusterOperator) (*osv1.ClusterOperator, bool, error) {
 	if required.Status.Extension.Raw != nil && required.Status.Extension.Object != nil {
 		return nil, false, fmt.Errorf("both extension.Raw and extension.Object should not be set")
 	}
 	existing, err := client.ClusterOperators(required.Namespace).Get(required.Name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		actual, err := client.ClusterOperators(required.Namespace).Create(required)
+		if err != nil && !errors.IsAlreadyExists(err) {
```
abhinavdahiya: if we couldn't find the object why do this check... i'm probably missing something here?

pkg/cvo/updatepayload.go
```diff
@@ -42,10 +42,16 @@ const (
...
 )
 
-func loadUpdatePayload(dir, releaseImage string) (*updatePayload, error) {
+type payloadTasks struct {
+	idir       string
+	preprocess func([]byte) ([]byte, error)
+	skipFiles  sets.String
+}
+
+func loadUpdatePayloadMetadata(dir, releaseImage string) (*updatePayload, []payloadTasks, error) {
```
abhinavdahiya: any reason why `loadUpdatePayloadMetadata` returns the payload tasks?

smarterclayton: Right now the CVO doesn't know what version its payload has.  I wanted to share the same code for reading the default payload in the operator init path (where it tries to read the current payload to find its current version) and in when we apply it, so I extracted the first chunk out.  payload tasks encapsulated all the path specific stuff, so it avoided the remainder of loadUpdatePayload from knowing anything about paths.

smarterclayton:
Added a couple more commits that tighten up error messages and reactivity - with all changes applied (ClusterVersion is scoped) you can perform the following get call and see the summary in one line:

```
$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE     STATUS
version   4.0.1     True        False         8m        Cluster version is 4.0.1
```

lib/resourceapply/cv.go
smarterclayton: for status you can't update the status at the same time as creation (creation and update only allow spec changes, you have to call update status).  So this is creating whatever the object is, then applying the status.  I think we could change these methods around to bail out, but since ClusterOperators is primarily a status object this seemed the closest to the intent of apply:  "make sure this object exists with this status".  We always pass in a valid ClusterOperator here.

Also, if someone deletes the ClusterOperator out from under us we just recreate it (we're the object owner).

smarterclayton:
Report prometheus metrics that show the current version state.

```
cluster_version{type="current",version="4.0.1",payload="location/registry/image:tag"}  1
cluster_version{type="update", version="4.0.2",payload="location/registry/image:tag2"} 1
cluster_version{type="failure",version="4.0.1",payload="location/registry/image:tag"}  1
```

abhinavdahiya:
1. ClusterVersionStatus vs ClusterOperatorStatus
Do we need both? I think one is enough.
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version
I would hope we keep just ClusterVersionStatus.

2. All 3 sync loops should not be driven by informers.
`availableUpdatesSync` should be a periodic func
`statusSync` should not be a separate one, we should be updating status as part of main sync
`sync` should be the only one driven by informers.

3. https://github.com/openshift/cluster-version-operator/pull/45/commits/a2cf82888d41f268d9875214a2c3f3b1ee998b4d :heart:

smarterclayton:
For 1. my reasons are:

1. ClusterOperatorStatus is a generic interface.  It can't report information like "what available updates there are".  It can't report structured data response.  ClusterVersionStatus can and should.  I think we could argue that the CVO doesn't have to write a ClusterOperator, but since it will read them it may be worth it.

2. If we change the upstream URL we need to trigger availableUpdatesSync.  If we delete the ClusterVersion we need to trigger the statusSync.

smarterclayton:
To be clear, I don't think CVO *has* to write cluster operator status.  Is there anyone who wants to wait on it?

abhinavdahiya:
> To be clear, I don't think CVO _has_ to write cluster operator status. Is there anyone who wants to wait on it?

can we drop it for now and see in the future if we need to update clusteroperatorstatus, we can revisit then.

lib/resourceapply/cv.go
abhinavdahiya: hmm, i didn't realize that `Create&Update` don't change `.status`. :thinking:

abhinavdahiya:
> If we change the upstream URL we need to trigger availableUpdatesSync.
that's not happening right now.. we are queuing on all events

> If we delete the ClusterVersion we need to trigger the statusSync.
it seems odd that statusSync is syncing crds and clusteroperatorstatus...

cmd/start.go
```diff
@@ -48,12 +50,17 @@ var (
...
+
 		enableAutoUpdate bool
 	}
 )
 
 func init() {
+	startOpts.listenAddr = "0.0.0.0:11345"
+
 	rootCmd.AddCommand(startCmd)
+	startCmd.PersistentFlags().StringVar(&startOpts.listenAddr, "listen", startOpts.listenAddr, "Address to listen on for metrics")
```
abhinavdahiya: any reason why not inline?

pkg/cvo/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
+	current := m.optr.currentVersion()
+	g := m.version.WithLabelValues("current", current.Version, current.Payload)
+	g.Set(1)
+	ch <- g
+	if cv, err := m.optr.cvoConfigLister.Get(m.optr.name); err == nil {
+		if update := cv.Spec.DesiredUpdate; update != nil && update.Payload != current.Payload {
+			g = m.version.WithLabelValues("update", update.Version, update.Payload)
```
abhinavdahiya: it is usual to just reuse the same variable when doing this metric push?

smarterclayton:
Yes.  I will remove those loops.

pkg/cvo/metrics.go
smarterclayton: no, I'll fix that

smarterclayton:
> that's not happening right now.. we are queuing on all events

Yes, available updates needs to be queued by an outside source.

> it seems odd that statusSync is syncing crds and clusteroperatorstatus for crds...

StatusSync is the only bit of code that touches ClusterOperators - and it's the loop that requires cluster operator CRD to exist.  So it's a prerequisite for level driven.  That said, do we need to do CRD reconciliation here or should we move it to manifests/?  Maybe we should be doing cluster version CRD reconciliation here as well?  (the split is so that we can wait for the informers to sync, but it feels wrong).

> can we drop it for now and see in the future if we need to update clusteroperatorstatus, we can revisit then.

Yes.  I will be adding metrics to read ClusterOperator status (to track upgrade progress), but we don't need to write it.

cmd/start.go
smarterclayton: inline how?  localhost you mean?

smarterclayton:
Oh, availableUpdates is being driven by the resync interval.

cmd/start.go
abhinavdahiya:
```golang
startCmd.PersistentFlags().StringVar(&startOpts.listenAddr, "listen", "0.0.0.0:12345", "Address to listen on for metrics")
```

cmd/main.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	componentName      = "cluster-version-operator"
+	componentName      = "version"
```
deads2k: reading top to bottom, this looks like an odd change.

cmd/start.go
```diff
@@ -29,7 +31,7 @@ import (
 )
 
 const (
-	minResyncPeriod = 10 * time.Second
```
deads2k: oh my

cmd/start.go
```diff
@@ -48,12 +50,17 @@ var (
...
 		nodeName   string
 
+		listenAddr string
+
 		enableAutoUpdate bool
 	}
 )
 
 func init() {
+	startOpts.listenAddr = "0.0.0.0:11345"
```
deads2k: pre-existing, but I do prefer cobra these days.

cmd/start.go
deads2k: @abhinavdahiya when using a cobra approach with options structs and flags for customization (like kube-apiserver, generic-apiserver, oc, kubectl, etc), we've found that respecting a struct value allows for flexibility and clean composition.  This matches the upstream projects better.

cmd/start.go
```diff
@@ -78,6 +85,16 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		glog.Fatalf("missing --release-image flag, it is required")
 	}
 
+	if len(startOpts.listenAddr) > 0 {
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
+		go func() {
+			if err := http.ListenAndServe(startOpts.listenAddr, mux); err != nil {
```
deads2k: You know that a lack of TLS makes me twitch.  Follow up issue?

cmd/start.go
deads2k: This is in a container, why not 8443?

deads2k: > This is in a container, why not 8443?

nm, it's hostnetwork

install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -0,0 +1,99 @@
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: clusteroperators.config.openshift.io
+spec:
+  additionalPrinterColumns:
```
deads2k: @mfojtik We need this for our resources too.  Mike or Luis maybe?

install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -0,0 +1,99 @@
...
+    description: The time the operator's Available status last changed.
+    name: Since
+    type: date
+  group: config.openshift.io
+  names:
+    kind: ClusterOperator
+    listKind: ClusterOperatorList
+    plural: clusteroperators
+    singular: clusteroperator
+  scope: Namespaced
```
deads2k: I expected cluster scoped.

deads2k: > I expected cluster scoped.

Oh, we haven't gotten there yet.

lib/resourceapply/cv.go
```diff
@@ -1,107 +1,81 @@
...
+	actual, err := client.ClusterVersions().Update(existing)
 	return actual, true, err
 }
 
-func ApplyOperatorStatusFromCache(lister oslistersv1.ClusterOperatorLister, client osclientv1.ClusterOperatorsGetter, required *osv1.ClusterOperator) (*osv1.ClusterOperator, bool, error) {
-	if required.Status.Extension.Raw != nil && required.Status.Extension.Object != nil {
-		return nil, false, fmt.Errorf("both extension.Raw and extension.Object should not be set")
-	}
-	existing, err := lister.ClusterOperators(required.Namespace).Get(required.Name)
+func ApplyClusterVersionFromCache(lister cvlistersv1.ClusterVersionLister, client cvclientv1.ClusterVersionsGetter, required *cvv1.ClusterVersion) (*cvv1.ClusterVersion, bool, error) {
```
deads2k: @abhinavdahiya as a general statement, is there a reason I can't shim a getter to use a lister for `get` and avoid this?

@mfojtik another mike or Luis?

lib/resourcemerge/cv.go
```diff
@@ -1,32 +1,46 @@
...
+
+	cvv1 "github.com/openshift/cluster-version-operator/pkg/apis/config.openshift.io/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 )
 
-func EnsureCVOConfig(modified *bool, existing *cvv1.CVOConfig, required cvv1.CVOConfig) {
+func EnsureClusterVersion(modified *bool, existing *cvv1.ClusterVersion, required cvv1.ClusterVersion) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
-	if existing.Upstream != required.Upstream {
+	if !reflect.DeepEqual(existing.Spec.Upstream, required.Spec.Upstream) {
```
deads2k: equalities.Semantic?

lib/resourceread/image.go
```diff
@@ -17,11 +17,11 @@ func init() {
 	}
 }
 
-// ReadImageStreamV1OrDie reads imagestream object from bytes. Panics on error.
-func ReadImageStreamV1OrDie(objBytes []byte) *imagev1.ImageStream {
+// ReadImageStreamV1 reads imagestream object from bytes or reports an error.
+func ReadImageStreamV1(objBytes []byte) (*imagev1.ImageStream, error) {
```
deads2k: this is called in a user defined context?

pkg/autoupdate/autoupdate.go
```diff
@@ -188,28 +175,18 @@ func (ctrl *Controller) sync(key string) error {
...
-	obji, _, err := scheme.Codecs.UniversalDecoder().Decode(ops.Status.Extension.Raw, nil, &v1.CVOStatus{})
-	if err != nil {
-		return fmt.Errorf("unable to decode CVOStatus from extension.Raw: %v", err)
-	}
-	cvoststatus, ok := obji.(*v1.CVOStatus)
-	if !ok {
-		return fmt.Errorf("expected *v1.CVOStatus found %T", obji)
-	}
+	config := new(v1.ClusterVersion)
+	clusterversion.DeepCopyInto(config)
```
deads2k: `config := clusterversion.DeepCopy()` ?

pkg/cvo/cvo.go
```diff
@@ -96,32 +113,54 @@ func New(
...
-		releaseImage:  releaseImage,
+		nodename:     nodename,
+		namespace:    namespace,
+		name:         name,
+		releaseImage: releaseImage,
+		payloadDir:   overridePayloadDir,
+
+		minimumUpdateCheckInterval: 2 * time.Minute,
+
+		defaultUpstreamServer: "http://localhost:8080/graph",
```
deads2k: looks suspiciously like a test value

lib/resourceapply/cv.go
abhinavdahiya: > @abhinavdahiya as a general statement, is there a reason I can't shim a getter to use a lister for get and avoid this?

@deads2k can you expand on it a little bit.. :innocent:

pkg/cvo/cvo.go
```diff
@@ -135,119 +174,159 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
...
-	obj, err = optr.cvoConfigLister.CVOConfigs(namespace).Get(name)
-	if apierrors.IsNotFound(err) {
-		obj, err = optr.getConfig()
+	availableCondition := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, osv1.OperatorAvailable)
+
+	// when we're up to date, limit how frequently we check the payload
+	if config.Status.Generation == config.Generation {
+		if availableCondition != nil && availableCondition.Status == osv1.ConditionTrue && availableCondition.LastTransitionTime.Time.After(time.Now().Add(-optr.minimumUpdateCheckInterval)) {
+			glog.V(2).Infof("Waiting for more time to pass before attempting to resync the payload")
+			return nil
```
deads2k: Seems like you should requeue here.  Are you counting on a resync instead?  I would prefer an explicit requeue.

pkg/cvo/updatepayload.go
```diff
@@ -58,19 +68,16 @@ func loadUpdatePayload(dir, releaseImage string) (*updatePayload, error) {
 	irf := filepath.Join(releaseDir, imageReferencesFile)
 	imageRefData, err := ioutil.ReadFile(irf)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	imageRef := resourceread.ReadImageStreamV1OrDie(imageRefData)
-	mrc := manifestRenderConfig{ReleaseImage: releaseImage}
+	imageRef, err := resourceread.ReadImageStreamV1(imageRefData)
```
deads2k: and now I see.  Yeah, much better

lib/resourceapply/cv.go
deads2k: The point of this `FromCache` appears to be to avoid `GET` pressure on the server.  That makes sense, since a cache is already available, but instead of having a separate method, it seems like a clever impl could do the trick with an existing method.  We'll probably knock something together next week (master team operators are GET happy)

deads2k:
no real structural concerns.  From what I read, I see a spec/status split and a desire to reduce resync costs.  I'm a little concerned that the payload check was triggering the overall payload reconcile.  Am I simply missing the payload reconciliation loop that will keep us up to date in between new payloads or have we intentionally reduced that check to every two minutes?

lib/resourceapply/cv.go
abhinavdahiya: Awesome, can you point me to an example in master operators

smarterclayton:
Added more metrics about cluster operators

```
# HELP cluster_operator_conditions Report the conditions for active cluster operators. 0 is False and 1 is True.
# TYPE cluster_operator_conditions gauge
cluster_operator_conditions{condition="Available",name="version",namespace="openshift-cluster-version"} 1
cluster_operator_conditions{condition="Failing",name="version",namespace="openshift-cluster-version"} 0
cluster_operator_conditions{condition="Progressing",name="version",namespace="openshift-cluster-version"} 0
cluster_operator_conditions{condition="RetrievedUpdates",name="version",namespace="openshift-cluster-version"} 0
# HELP cluster_operator_up Reports key highlights of the active cluster operators.
# TYPE cluster_operator_up gauge
cluster_operator_up{name="version",namespace="openshift-cluster-version",version="4.0.1"} 1
# HELP cluster_version Reports the version of the cluster.
# TYPE cluster_version gauge
cluster_version{payload="localhost/test/test:latest",type="current",version="4.0.1"} 1
# HELP cluster_version_available_updates Report the count of available versions for an upstream and channel.
# TYPE cluster_version_available_updates gauge
cluster_version_available_updates{channel="fast",upstream="http://localhost:8080/graph"} 0
```

smarterclayton:
The resync interval (moved from 10s to 2m) will ensure we re-apply a payload.  While reconciling, we don't switch back to Progressing status.

pkg/cvo/cvo.go
smarterclayton: Resync interval at 2m does that for us, yeah.  AvailableUpdates keys off the same thing.  AddAfter isn't enough because need to reconcile every X, not just after the last change.

pkg/cvo/cvo.go
smarterclayton: Carry over from before.  We don't have our prod URLs yet.

pkg/autoupdate/autoupdate.go
smarterclayton: yes

smarterclayton:
I think I got all the review feedback.  I simplified the "have we done X" recently for both sync and available updates to use a local clock instead of using lastTransitionTime.  Also various cleanups.  I am tempted to squash everything but the small changes into a bigger PR to make reading clearer, since the final loops ended up simpler than the intermediate states.

lib/resourceapply/cv.go
```diff
@@ -1,92 +1,66 @@
...
-	existing, err := client.ClusterVersions(required.Namespace).Get(required.Name, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		actual, err := client.ClusterVersions(required.Namespace).Create(required)
-		return actual, true, err
-	}
-	if err != nil {
-		return nil, false, err
-	}
-
+	existing := new(cvv1.ClusterVersion)
```
abhinavdahiya: why not `DeepCopy()` ?

lib/resourceapply/cv.go
```diff
@@ -97,11 +71,11 @@ func ApplyClusterVersionFromCache(lister cvlistersv1.ClusterVersionLister, clien
 	existing := new(cvv1.ClusterVersion)
```
abhinavdahiya: same `DeepCopy`

pkg/autoupdate/autoupdate.go
```diff
@@ -44,15 +44,16 @@ type Controller struct {
 	client        clientset.Interface
 	eventRecorder record.EventRecorder
 
-	syncHandler func(key string) error
+	syncHandler       func(key string) error
+	statusSyncHandler func(key string) error
```
abhinavdahiya: not used.

pkg/cvo/cvo.go
```diff
@@ -62,30 +71,38 @@ type Operator struct {
...
 
-	clusterOperatorLister oslistersv1.ClusterOperatorLister
+	// payloadDir is intended for testing. If unset it will default to '/'
+	payloadDir string
+	// defaultUpstreamServer is intended for testing.
+	defaultUpstreamServer string
 
-	crdLister             apiextlistersv1beta1.CustomResourceDefinitionLister
-	deployLister          appslisterv1.DeploymentLister
+	clusterOperatorLister oslistersv1.ClusterOperatorLister
```
abhinavdahiya: nit: can we group corresponding lister and listersynced

pkg/cvo/cvo.go
```diff
@@ -96,32 +113,48 @@ func New(
...
-		name:          name,
-		releaseImage:  releaseImage,
+		nodename:     nodename,
+		namespace:    namespace,
+		name:         name,
+		releaseImage: releaseImage,
+		payloadDir:   overridePayloadDir,
+
+		minimumUpdateCheckInterval: minimumInterval,
+
```
abhinavdahiya: nit: remove this space both `minimumUpdateCheckInterval` and `defaultUpstreamServer` are related to `syncAvailableUpdates`

pkg/cvo/cvo.go
```diff
@@ -130,124 +163,193 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer optr.queue.ShutDown()
 
-	glog.Info("Starting ClusterVersionOperator")
+	glog.Infof("Starting ClusterVersionOperator with minimum reconcile period %s", optr.minimumUpdateCheckInterval)
```
abhinavdahiya: nit: `glog.Infof("Starting ClusterVersionOperator with minimum reconcile period %s", optr.minimumUpdateCheckInterval)` makes it seem like that is the sync for all actions.

pkg/cvo/status.go
```diff
@@ -2,130 +2,222 @@ package cvo
...
-			Version:   version.Raw,
-			Extension: runtime.RawExtension{},
+			Name: optr.name,
 		},
 	}
-	resourcemerge.SetOperatorStatusCondition(&status.Status.Conditions, osv1.ClusterOperatorStatusCondition{Type: osv1.OperatorAvailable, Status: osv1.ConditionFalse})
-	resourcemerge.SetOperatorStatusCondition(&status.Status.Conditions, osv1.ClusterOperatorStatusCondition{Type: osv1.OperatorProgressing, Status: osv1.ConditionFalse})
-	resourcemerge.SetOperatorStatusCondition(&status.Status.Conditions, osv1.ClusterOperatorStatusCondition{
-		Type: osv1.OperatorFailing, Status: osv1.ConditionTrue,
+	// try to reuse the most recent status if available
```
abhinavdahiya: why is degraded different compared to available, progresing.

smarterclayton: Right now called from two different spots - inside the sync loop, and outside when a generic error happens.

I’m considering a follow up where I add better error output for sync and release access.  I can refactor this there (we might want a structured error set for displaying more succinct errors)

pkg/cvo/cvo.go
smarterclayton: That was kind of my intent - what part isn’t reconciled?

pkg/cvo/cvo.go
smarterclayton: Ok, will tackle in a follow up where I add more progress metrics and explicit error states

pkg/cvo/cvo.go
smarterclayton: Yes

pkg/autoupdate/autoupdate.go
smarterclayton: Yeah, I was debating removing these.  In practice we test successive iterations of calling sync, and testing that informers work hasn’t been a high priority and/or is better covered by e2e tests

lib/resourceapply/cv.go
smarterclayton: This was moved older code, I will update

smarterclayton:
/test e2e-aws

I’ve seen this pass twice in a row, it looks like even if the installer doesn’t set overrides the installer succeeds?

smarterclayton:
https://github.com/openshift/installer/pull/549 was the PR that renames CVO config

smarterclayton:
Nits addressed in #47

abhinavdahiya:
@smarterclayton saw CVO hinge in the CI
```console
I1103 00:49:43.552170       1 cvo.go:287] Updating the cluster to version 0.0.1-2018-11-03T002352Z and image registry.svc.ci.openshift.org/ci-op-2rvhpnvd/release:latest (hash=nSXAm1bk15A=)
I1103 00:49:43.552262       1 request.go:897] Request Body: {"kind":"ClusterVersion","apiVersion":"config.openshift.io/v1","metadata":{"name":"version","selfLink":"/apis/config.openshift.io/v1/clusterversions/version","uid":"0071fa0c-df00-11e8-8940-12e6bb5812e4","resourceVersion":"4259","generation":1,"creationTimestamp":"2018-11-03T00:32:53Z"},"spec":{"clusterID":"9e477bdb-3c5f-47ec-ae46-7d0daf2983fd","desiredUpdate":null,"upstream":"http://localhost:8080/graph","channel":"fast"},"status":{"current":{"version":"0.0.1-2018-11-03T002352Z","payload":"registry.svc.ci.openshift.org/ci-op-2rvhpnvd/release:latest"},"generation":1,"versionHash":"","conditions":[{"type":"RetrievedUpdates","status":"False","lastTransitionTime":"2018-11-03T00:49:42Z","reason":"RemoteFailed","message":"Unable to retrieve available updates: Get http://localhost:8080/graph: dial tcp [::1]:8080: connect: connection refused"},{"type":"Available","status":"False","lastTransitionTime":"2018-11-03T00:49:43Z"},{"type":"Progressing","status":"True","lastTransitionTime":"2018-11-03T00:49:43Z","message":"W [truncated 149 chars]
I1103 00:49:43.653557       1 request.go:485] Throttling request took 389.733091ms, request: PUT:https://ci-op-2rvhpnvd-928f7-api.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version/status
I1103 00:49:43.653624       1 round_trippers.go:383] PUT https://ci-op-2rvhpnvd-928f7-api.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version/status
I1103 00:49:43.653630       1 round_trippers.go:390] Request Headers:
I1103 00:49:43.653635       1 round_trippers.go:393]     Accept: application/json, */*
I1103 00:49:43.653640       1 round_trippers.go:393]     Content-Type: application/json
I1103 00:49:43.653654       1 round_trippers.go:393]     User-Agent: cluster-version-operator/v0.0.0 (linux/amd64) kubernetes/$Format/version
I1103 00:49:43.662971       1 round_trippers.go:408] Response Status: 200 OK in 9 milliseconds
I1103 00:49:43.663035       1 round_trippers.go:411] Response Headers:
I1103 00:49:43.663055       1 round_trippers.go:414]     Content-Length: 799
I1103 00:49:43.663060       1 round_trippers.go:414]     Date: Sat, 03 Nov 2018 00:49:43 GMT
I1103 00:49:43.663064       1 round_trippers.go:414]     Cache-Control: no-store
I1103 00:49:43.663069       1 round_trippers.go:414]     Content-Type: application/json
I1103 00:49:43.663091       1 request.go:897] Response Body: {"apiVersion":"config.openshift.io/v1","kind":"ClusterVersion","metadata":{"creationTimestamp":"2018-11-03T00:32:53Z","generation":1,"name":"version","resourceVersion":"4260","selfLink":"/apis/config.openshift.io/v1/clusterversions/version/status","uid":"0071fa0c-df00-11e8-8940-12e6bb5812e4"},"spec":{"channel":"fast","clusterID":"9e477bdb-3c5f-47ec-ae46-7d0daf2983fd","desiredUpdate":null,"upstream":"http://localhost:8080/graph"},"status":{"availableUpdates":null,"conditions":[{"lastTransitionTime":"2018-11-03T00:49:43Z","message":"Unable to retrieve available updates: Get http://localhost:8080/graph: dial tcp [::1]:8080: connect: connection refused","reason":"RemoteFailed","status":"False","type":"RetrievedUpdates"}],"current":{"payload":"","version":""},"generation":0,"versionHash":""}}
I1103 00:49:43.663157       1 cvo.go:312] Finished syncing available updates "openshift-cluster-version/version" (399.991814ms)
I1103 00:49:43.663186       1 cvo.go:310] Started syncing available updates "openshift-cluster-version/version" (2018-11-03 00:49:43.663183259 +0000 UTC m=+290.902350746)
I1103 00:49:43.663686       1 status.go:99] Upstream server http://localhost:8080/graph could not return available updates: Get http://localhost:8080/graph: dial tcp [::1]:8080: connect: connection refused
I1103 00:49:43.663772       1 request.go:897] Request Body: {"kind":"ClusterVersion","apiVersion":"config.openshift.io/v1","metadata":{"name":"version","selfLink":"/apis/config.openshift.io/v1/clusterversions/version","uid":"0071fa0c-df00-11e8-8940-12e6bb5812e4","resourceVersion":"4260","generation":1,"creationTimestamp":"2018-11-03T00:32:53Z"},"spec":{"clusterID":"9e477bdb-3c5f-47ec-ae46-7d0daf2983fd","desiredUpdate":null,"upstream":"http://localhost:8080/graph","channel":"fast"},"status":{"current":{"version":"","payload":""},"generation":0,"versionHash":"","conditions":[{"type":"RetrievedUpdates","status":"False","lastTransitionTime":"2018-11-03T00:49:43Z","reason":"RemoteFailed","message":"Unable to retrieve available updates: Get http://localhost:8080/graph: dial tcp [::1]:8080: connect: connection refused"}],"availableUpdates":null}}
I1103 00:49:43.853550       1 request.go:485] Throttling request took 301.245517ms, request: PUT:https://ci-op-2rvhpnvd-928f7-api.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version/status
I1103 00:49:43.853610       1 round_trippers.go:383] PUT https://ci-op-2rvhpnvd-928f7-api.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/clusterversions/version/status
I1103 00:49:43.853617       1 round_trippers.go:390] Request Headers:
I1103 00:49:43.853622       1 round_trippers.go:393]     Accept: application/json, */*
I1103 00:49:43.853627       1 round_trippers.go:393]     Content-Type: application/json
I1103 00:49:43.853631       1 round_trippers.go:393]     User-Agent: cluster-version-operator/v0.0.0 (linux/amd64) kubernetes/$Format/version
I1103 00:49:43.861117       1 round_trippers.go:408] Response Status: 409 Conflict in 7 milliseconds
I1103 00:49:43.861135       1 round_trippers.go:411] Response Headers:
I1103 00:49:43.861141       1 round_trippers.go:414]     Date: Sat, 03 Nov 2018 00:49:43 GMT
I1103 00:49:43.861145       1 round_trippers.go:414]     Cache-Control: no-store
I1103 00:49:43.861149       1 round_trippers.go:414]     Content-Type: application/json
I1103 00:49:43.861153       1 round_trippers.go:414]     Content-Length: 370
I1103 00:49:43.861174       1 request.go:897] Response Body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Operation cannot be fulfilled on clusterversions.config.openshift.io \"version\": the object has been modified; please apply your changes to the latest version and try again","reason":"Conflict","details":{"name":"version","group":"config.openshift.io","kind":"clusterversions"},"code":409}
I1103 00:49:43.861255       1 cvo.go:246] Finished syncing cluster version "openshift-cluster-version/version" (400.408075ms)
I1103 00:49:43.861278       1 cvo.go:231] Error syncing operator openshift-cluster-version/version: Operation cannot be fulfilled on clusterversions.config.openshift.io "version": the object has been modified; please apply your changes to the latest version and try again
I1103 00:49:43.861308       1 cvo.go:244] Started syncing cluster version "openshift-cluster-version/version" (2018-11-03 00:49:43.861304815 +0000 UTC m=+291.100472297)
I1103 00:49:43.861330       1 updatepayload.go:58] Loading updatepayload from "/"
I1103 00:49:43.862000       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.862209       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.863068       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.863406       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.863616       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.864421       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.864550       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.864744       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.864899       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.865361       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.865675       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.865845       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.865952       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.866106       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.866271       1 decoder.go:224] decoding stream as YAML
I1103 00:49:43.866490       1 decoder.go:224] decoding stream as YAML
```

^^ this was repeating and CVO never progresses beyond https://github.com/openshift/cluster-version-operator/pull/45/files#diff-85c31beb4341b4c52f892c0581bbb5d6R288

smarterclayton:
oh, I think my PR used the wrong name for the new object in the installer.  Sigh

smarterclayton:
I may also need to return the updated object from apply in the progress sync

smarterclayton:
https://github.com/openshift/installer/pull/605 is one of the changes.  looking for the conflict issue

smarterclayton:
Oh, it's because both loops are dispatching mutations at the same time and neither retries conflicts.  Both try to write, but because latency to master is > x ms they don't see the latest in their cache.  I need to make ApplyClusterVersionStatus more apply like.

smarterclayton:
Good old informer latency.  We have several ways we generally solve this (controller with two independent loops)

1. use expectations to delay until the cache is up to date (track expecting to see a change to an object by having a minimum floor RV)
2. use a critical section in the controller (updating status goes through a lock, hold until new RV shows up)
3. retry on conflict
4. have one of the loops store local state in the controller that the other loop merges on writes
5. patch status

1 doesn't stop race conflicts in all cases (has to uses retries anyway), 2 can block both loops for a long time, 3 performs extra writes, 4 requires more complexity to wake the core loop up, 5 is probably the best but we have to use strategic merge patch for the condition array, but we can't do that on CRDs.  

I need to mull over it a bit, it reminds me that we need to always test with a heavily delayed cache.  I know we have a delaying watch shim somewhere.

abhinavdahiya:
Might be a possibility to move sync available update to separate controller like autoupdater?

smarterclayton:
I'll add that the pattern of ApplyFromCache is part of the issue here - to really benefit from that pattern, you have to exit the sync loop after making a change (since it's not a write through cache).  For example, with a 250ms delay added to the watch, the getOrCreateClusterVersion returns a 409 error 6 times (causing sync backoff) before the item shows up in the cache.  It would be more effective for the core loop (which is a single writer) to write and update the object we hold.

smarterclayton:
Has passed a few times (the other failures looked to be the bootstrap watcher issue Trevor was tracking).

I left a debugging line in there to see the CVO to help debug any near term race condition stuff

smarterclayton:
Last one did not appear to be me

smarterclayton:
This has had 3/5 green e2e-aws runs, I *think* the failures are unrelated / haven’t been able to tie back to this.

smarterclayton:
/hold
found a bug

smarterclayton:
Hold removed, tests weren't good enough

pkg/cvo/status.go
smarterclayton: #47 has most of these fixes, and clarifies naming.

smarterclayton:
This is ready for re-review for last commit

lib/resourceapply/cv.go
```diff
@@ -53,29 +56,33 @@ func ApplyClusterVersionFromCache(lister cvlistersv1.ClusterVersionLister, clien
...
-	if errors.IsNotFound(err) {
-		actual, err := client.ClusterVersions().Create(required)
-		if err != nil && !errors.IsAlreadyExists(err) {
-			return actual, true, err
+// ApplyClusterVersionStatus attempts to overwrite the status subresource of required. If
+// original is provided it is compared to required and no update will be made if the
+// object does not change. The method will retry a conflict by retrieving the latest live
+// version and updating the metadata of required. required is modified if the object on
+// the server is newer.
+func ApplyClusterVersionStatus(client cvclientv1.ClusterVersionsGetter, required, original *cvv1.ClusterVersion) (*cvv1.ClusterVersion, error) {
```
abhinavdahiya: Do you want to move this function out to `pkg/cvo`? It might be confusing be see this as other functions use `exsiting, required.`

pkg/cvo/availableupdates.go
```diff
@@ -0,0 +1,156 @@
...
+
+	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), upstream, channel, optr.releaseVersion)
+	optr.setAvailableUpdates(&availableUpdates{
+		Upstream:  upstream,
+		Channel:   config.Spec.Channel,
+		Updates:   updates,
+		Condition: condition,
+	})
+	// requeue
+	optr.queue.Add(fmt.Sprintf("%s/%s", optr.namespace, optr.name))
```
abhinavdahiya: we are calling this `Sprintf` more than once, so maybe make it a func.

abhinavdahiya:
@smarterclayton just 2 nits. if we want to keep them for future change i'm fine with that too and i'll drop `/lgtm`

lib/resourceapply/cv.go
smarterclayton: yeah

pkg/cvo/availableupdates.go
smarterclayton: yes, `opt.key()`

PR #46, opened 2018-11-02
api: Add stub ClusterVersion CRD so that installer can reference it

smarterclayton:
Create the new API definition for CRD so installer can start creating
the new object. Extracted from PR #45, no changes.


smarterclayton:
/retest

metrics-server didn't come up in kube-system, not sure how this could have caused that.

PR #47, opened 2018-11-03
cvo: Report errors to conditions more consistently

smarterclayton:
Reporting sync errors to the user via the ClusterVersion status is important for usability - this will be the first place user interfaces check when errors occur. Improve the exit paths from the core payload sync path to capture errors.

```
$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE     STATUS
version   4.0.2     False       True          56m       Unable to apply 4.0.3: a required extension is not available to update
```

```
# describe snippet:
Status:
  Available Updates:  <nil>
  Conditions:
    Last Transition Time:  2018-11-05T02:01:24Z
    Status:                False
    Type:                  Available
    Last Transition Time:  2018-11-05T02:01:24Z
    Message:               Unable to apply 4.0.3: a required extension is not available to update
    Reason:                UpdatePayloadResourceTypeMissing
    Status:                True
    Type:                  Progressing
    Last Transition Time:  2018-11-05T02:01:47Z
    Message:               Could not update widget "test" (custom.resource/v1, 1 of 1): the server does not recognize this resource, check extension API servers
    Reason:                UpdatePayloadResourceTypeMissing
    Status:                True
    Type:                  Failing
    Last Transition Time:  2018-11-05T01:54:48Z
    Message:               Unable to retrieve available updates: Get http://localhost:8080/graph: dial tcp [::1]:8080: connect: connection refused
    Reason:                RemoteFailed
    Status:                False
    Type:                  RetrievedUpdates
```

Improve the consistency of error behavior by having the Failing condition contain the error text exactly, and have Progressing summarize the error in the condition in the context of the action the operator was taking (reconcile, upgrade). Attempt to preserve the status of the Progressing and Failing conditions when performing the opposite update so as to keep users from seeing results that strobe.

Follows on to #45 - some stylistic feedback is included here.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -221,6 +221,14 @@
   version = "v2.0.1"
 
 [[projects]]
+  digest = "1:5cf3f025cbee5951a4ee961de067c8a89fc95a5adabead774f82822efabab121"
+  name = "github.com/pkg/errors"
+  packages = ["."]
+  pruneopts = "NUT"
+  revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
+  version = "v0.8.0"
+
+[[projects]]
   digest = "1:aa2da1df3327c3a338bb42f978407c07de74cd0a5bef35e9411881dffd444214"
   name = "github.com/prometheus/client_golang"
   packages = [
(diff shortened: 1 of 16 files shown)


smarterclayton:
This should be ready for review when people have a chance.

abhinavdahiya:
@smarterclayton for all the nested errors, do you think `https://github.com/pkg/errors` might be useful instead of created local `nestedError`s

abhinavdahiya:
Otherwise `/lgtm`

smarterclayton:
Yeah, I actually don't need nested errors as much.  Let me look at retooling.

smarterclayton:
Updated with pkg/errors

PR #48, opened 2018-11-07
Move ClusterOperator to config.openshift.io

dmage:


dmage:
For some reason `informer-gen` wants to use namespaces for the ClusterVersion resource. So `./hack/verify-codegen.sh` fails.

bparees:
@smarterclayton we're in an inconsistent state right now, aren't we?

https://github.com/openshift/cluster-version-operator/blob/master/pkg/apis/apis.go#L7
https://github.com/openshift/cluster-version-operator/blob/master/pkg/apis/operatorstatus.openshift.io/v1/register.go#L11

vs

https://github.com/openshift/cluster-version-operator/blob/master/install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml#L23

smarterclayton:
/hold

I'm going to move this to openshift/api (maybe tomorrow morning), in which case we would then vendor it back.

smarterclayton:
https://github.com/openshift/api/pull/127 is the move PR

PR #49, opened 2018-11-07
manifests: Add a service for CVO so that a ServiceMonitor can scrape it

smarterclayton:
A service is required for the cluster monitoring operator to scrape a target
with prometheus. Create the service and define a metrics port.

### DIFF
--- a/install/0001_00_cluster-version-operator_03_service.yaml
+++ b/install/0001_00_cluster-version-operator_03_service.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: cluster-version-operator
+  namespace: openshift-cluster-version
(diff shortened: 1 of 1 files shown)


install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
wking: Is the `0001_` prefix intentional?  I guess that means "first after all of the ordered objects with [special runlevels][1]"?

[1]: https://github.com/openshift/cluster-version-operator/blob/d2914283326951a6a6035a42b4b9194aa9105bbf/docs/dev/operators.md#how-do-i-get-added-as-a-special-run-level

PR #54, opened 2018-11-16
metrics: Move default metrics port to 9101 in the open internal range

smarterclayton:
OpenShift opens 9000-9999 for internal host network use for services that
must be scraped like node-exporter but must also be on host network, and
which may be on the master. Move CVO to listen on 9101 so that prometheus
can scrape it to get cluster version metrics.


smarterclayton:
Required for openshift/cluster-monitoring-operator#144 now that 9000-9999 is open in the installer.

PR #55, opened 2018-11-16
api: Update to objects from openshift/api

smarterclayton:
ClusterVersion/ClusterOperator moved to the openshift/api repo, with minor
changes that are vendored back here. This is the reaction PR.

The major differences in what went upstream:

* ClusterOperator is now cluster scoped
* ClusterVersion deserialization doesn't check UID or URL deserialization,
  which needs to be handled in by the operator anyway (subsequent commit)
* ClusterVersion URL is not a pointer, suggestion was to create a new field
  which controls update behavior such as `updateMode: <None|Retrieve|Auto>`
  where Retrieve might be the default.

The validation logic checks whether the incoming object is valid, and if it isn't sets the `Invalid` condition on the object.  It then ensures that the config spec has any of the invalid "optional" fields cleared - this is somewhat novel, but it ensures we don't accidentally take action to move the version.  An invalid object reconciles, but doesn't progress.  As part of this change, simplify the initial status sync into something that feels more level driven (ensure we always have conditions populated), which makes the CVO loop a bit easier to read.


abhinavdahiya:
https://github.com/openshift/library-go/pull/97 might come in handy.

smarterclayton:
I still need to add unit test for invalid for this.  David and I talked through the conditions and how they're being used here vs other operators.  We agreed that we were roughly in the same spot:

Available means either "I've reached the desired state you requested" or "my operands are healthy".  In CVO right now we set available false when we start applying a new version, and set it to true after we've synced a payload successfully at least once.  Arguably we don't have to reset available to false when we start applying a new version, because "Progressing = true" means the same thing.  That's something I want to think about more, but `Progressing = false` can also be sufficient to mean `I've reached the desired state and I am going to maintain that state` and is probably fairly intuitive to a user when they think about upgrades (progressing = true when initially setting up or when working towards a desired update, but false while reconciling).

We did talk about how ClusterOperator could be taken into account as a prerequisite for upgrading to a new version.  If not all cluster operators are `Available=true && Progressing=false && Failing=false && Upgradeable!=false`, we should only reconcile to whatever the current CVO version is (treat desired update as if were nil).  That would prevent us from accepting a new update while an existing one is being processed.  At the very end of a payload sync, we would then update `status.current` to be equal to the current CVO object's payload.

1. user is at stable state at 4.0.2, all operators green, no desired update
2. user requests 4.0.3 via desired update
3. CVO checks operators still green before loading the payload - if not all green, reconcile current version rather than go to a new one
4. CVO begins update loop
5. if all updates succeed (including cluster operators going green) update `status.current` to be 4.0.3

We can still init status.current if it's unset, but we shouldn't update current until we've completed a sync (right now the code is updating current to the CVO at steps 3-4 when it updates status).  Also need to think about what type of status report to provide at 3 - are we Progressing=false with an appropriate message?

smarterclayton:
Updated so that this is relatively complete:

1. CVO uses openshift/api types
2. Early in the sync loop CVO verifies the incoming ClusterVersion and flags a condition if the object can't be handled as is (instead of erring out on it before) - that status is reported in the progressing field so that `oc get clusterversion` shows it
3. Reworked how the sync loop sets initial status so that we set all of these at once
4. If the object is invalid, the CVO *only* reconciles (applies its own payload), it will not update.

smarterclayton:
(to be clear, I didn't make any of the changes discussed in my wall of text comment, just added the Invalid condition and cleaned up the sync loop)

lib/resourcemerge/os.go
```diff
@@ -6,15 +6,15 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	osv1 "github.com/openshift/cluster-version-operator/pkg/apis/operatorstatus.openshift.io/v1"
+	configv1 "github.com/openshift/api/config/v1"
 )
 
-func EnsureClusterOperatorStatus(modified *bool, existing *osv1.ClusterOperator, required osv1.ClusterOperator) {
+func EnsureClusterOperatorStatus(modified *bool, existing *configv1.ClusterOperator, required configv1.ClusterOperator) {
```
abhinavdahiya: This file should no longer exist in CVO. https://github.com/openshift/library-go/pull/97 seems like better place for these functions. For now its okay as-is

pkg/cvo/availableupdates.go
```diff
@@ -53,15 +52,15 @@ type availableUpdates struct {
 
 	At time.Time
 
-	Updates   []cvv1.Update
-	Condition osv1.ClusterOperatorStatusCondition
+	Updates   []configv1.Update
```
abhinavdahiya: this `configv1.Update`  is little ambiguous.

abhinavdahiya:
> (to be clear, I didn't make any of the changes discussed in my wall of text comment, just added the Invalid condition and cleaned up the sync loop)

maybe an issue to make sure we don't forget?

pkg/cvo/cvo.go
```diff
@@ -410,24 +407,35 @@ func (optr *Operator) rememberLastUpdate(config *configv1.ClusterVersion) {
...
+		// ensure that the object we do have is valid
+		errs := validation.ValidateClusterVersion(obj)
+		changed, err := optr.syncInitialObjectStatus(obj, errs)
+		if err != nil {
+			return nil, false, err
+		}
+
+		// for fields that have meaning that are incomplete, clear them
+		// prevents us from loading clearly malformed payloads
+		obj = validation.ClearInvalidFields(obj, errs)
```
abhinavdahiya: Is this mutating lister cache?

abhinavdahiya:
This looks good. :+1: 

will `/lgtm` as soon as unit tests pass.

pkg/cvo/availableupdates.go
smarterclayton: Good point - I realized last night too that when we start bringing metadata back from cincinatti (messages and URLs to display pages) we're going to need to split the update.  I was going to suggest `ClusterVersionUpdate` and `AvailableClusterVersionUpdate` as the two structs.

pkg/cvo/cvo.go
smarterclayton: we deep copy before making a change in that method.

smarterclayton:
I updated tests - also made the sync_test use a customizable backoff so we could have the test complete instantly (that's in the api: reaction commit because the package changed).

smarterclayton:
/retest


On Nov 20, 2018, at 9:07 PM, OpenShift CI Robot <notifications@github.com>
wrote:

@smarterclayton <https://github.com/smarterclayton>: The following test
*failed*, say /retest to rerun them all:
Test name Commit Details Rerun command
ci/prow/e2e-aws ab4d84a
<https://github.com/openshift/cluster-version-operator/commit/ab4d84a021c38fea5e5973287e1f5580928907c1>
link
<https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/55/pull-ci-openshift-cluster-version-operator-master-e2e-aws/131>
/test
e2e-aws

Full PR test history
<https://openshift-gce-devel.appspot.com/pr/openshift_cluster-version-operator/55>.
Your PR dashboard
<https://openshift-gce-devel.appspot.com/pr/smarterclayton>. Please help us
cut down on flakes by linking to
<https://github.com/kubernetes/community/blob/master/contributors/devel/flaky-tests.md#filing-issues-for-flaky-tests>
an open issue
<https://github.com/openshift/cluster-version-operator/issues?q=is:issue+is:open>
when you hit one in your PR.

Instructions for interacting with me using PR comments are available here
<https://git.k8s.io/community/contributors/guide/pull-requests.md>. If you
have questions or suggestions related to my behavior, please file an issue
against the kubernetes/test-infra
<https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:>
repository. I understand the commands that are listed here
<https://go.k8s.io/bot-commands>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/cluster-version-operator/pull/55#issuecomment-440501958>,
or mute the thread

PR #56, opened 2018-11-20
metrics: Default metrics port conflicts with node-exporter

smarterclayton:
Change to 9099, I will create a document in openshift/origin to track
claimed ports.


abhinavdahiya:
> I will create a document in openshift/origin to track claimed ports.
that would be super helpful.

/lgtm

smarterclayton:
Documented https://github.com/openshift/origin/pull/21520

PR #57, opened 2018-11-20
 pkg/cvo: allow sync task for a manifest to be requeued on special failures

abhinavdahiya:
Adding `v1.cluster-version-operator.operators.io/RequeueOnError` annotation with
list of supported errros to a manifests will allow cvo to requeue the task once if it encounters one of the specified errors.
For now only `IsNoMatch` is supported.

Also added sync tests

/cc @smarterclayton


pkg/cvo/sync.go
```diff
@@ -12,13 +12,20 @@ import (
...
 	"github.com/openshift/cluster-version-operator/lib"
 	"github.com/openshift/cluster-version-operator/lib/resourcebuilder"
 	cvv1 "github.com/openshift/cluster-version-operator/pkg/apis/config.openshift.io/v1"
 	"github.com/openshift/cluster-version-operator/pkg/cvo/internal"
 )
 
+const (
+	// RequeueOnErrorAnnotationKey is key for annotation on a manifests object that instructs CVO to requeue on specific errors.
+	// The value is comma separated list of causes that forces requeue.
+	RequeueOnErrorAnnotationKey = "v1.cluster-version-operator.operators.openshift.io/RequeueOnError"
```
smarterclayton: Suggest `requeue-on-error` which is the upstream convention we've mostly settled on for new things

smarterclayton: Also, define the set of supported causes here.

abhinavdahiya: :+1:

abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/57/commits/31e4d2957e789ff1251f3c14a5854d48202372b2

PR #59, opened 2018-11-27
docs/dev/operators: Document a dynamic-object approach

wking:
This content is descended from openshift/installer@9c8a77dd (openshift/installer#377), but handling dynamic configuration is a generic issue, so the docs should live in the generic CVO repository. This will make these docs more discoverable.  It will also allow the installer docs to focus on installer-specific issues for the small subset of operators that need direct installer bindings.

There are also two orthogonal copy-edit commits in this PR.  I'm also fine pulling those out into a separate PR (or PRs), or with squashing them all into a single commit, if any of that would make review easier.

### DIFF
--- a/docs/dev/operators.md
+++ b/docs/dev/operators.md
@@ -14,7 +14,7 @@ LABEL io.openshift.release.operator=true
 Ensure your image is published into the cluster release tag by ci-operator
 Wait for a new release payload to be created (usually once you push to master in your operator).
 
-## What do I put in /manifests?
+## What do I put in `/manifests`?
 
 You need the following:
 
@@ -26,7 +26,7 @@ You need the following:
 - Deployment for your operator
 - A ClusterOperator CR [more info here](clusteroperator.md)
 - Any other config objects your operator might need
-- An image-references file (See below)
+- An `image-references` file (see [below](#how-do-i-ensure-the-right-images-get-used-by-my-manifests)).
 
 In your deployment you can reference the latest development version of your operator image (quay.io/openshift/origin-machine-api-operator:latest).  If you have other hard-coded image strings, try to put them as environment variables on your deployment or as a config map.
 
@@ -84,7 +84,7 @@ Your manifests can contain a tag to the latest development image published by Or
 
 Assume you have two images in your manifests - `quay.io/openshift/origin-ingress-operator:latest` and `quay.io/openshift/origin-haproxy-router:latest`.  Those correspond to the following tags `ingress-operator` and `haproxy-router` when the CI runs.
 
-Create a file `image-references` in the /manifests dir with the following contents:
+Create a file `image-references` in the `/manifests` directory with the following contents:
 
 ```yaml
 kind: ImageStream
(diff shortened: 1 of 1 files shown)


docs/dev/operators.md
```diff
@@ -101,13 +101,42 @@ spec:
...
+5. The operator container tries to pull its configuration.
+    * If found, the operator uses that configuration.
+    * If not found, the operator:
+        1. Pulls the dependency resources (e.g. the `cluster-config-v1` config-map from the `kube-system` namespace).
+            Documentation for connecting to the Kubernetes API from pod containers is [here][API-from-pod].
+            Operators written in Go can use [`InClusterConfig`][InClusterConfig] to configure their client.
+        2. Generates the dynamic configuration based on the dependencies.
+        3. Pushes the generated configuration into the cluster, where future operator runs and other consumers will be able to find it.
+
+For an example, see [the machine-API operator's `getOperatorConfig`][getOperatorConfig], although they don't need the "push the generated configuration into the cluster" step.
```
rajatchopra: Maybe we should inline the example in a code block here.

wking: > Maybe we should inline the example in a code block here.

How much of the code?  e24bc29450 currently links to the main portion of this, and I mention `InClusterConfig` further up but don't link to [the operator's `InClusterConfig` call][1] or [client construction][2].  Putting that all together into one example would be something like:

```go
include (
  "k8s.io/client-go/kubernetes"
  corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
  "k8s.io/client-go/rest"
)

func example() {
  config, err := rest.InClusterConfig() // FIXME: error handling
  client := kubernetes.NewForConfigOrDie(rest.AddUserAgent(cb.config, "your-operator")).CoreV1().ConfigMaps("kube-system")
  for {
    config, err := getConfig(client) // FIXME: error handling
    // FIXME: do whatever your operator does
  }
}

func getConfig(client corev1.ConfigMapInterface) (FIXMESomeType, error) {
  clusterConfig, err := client.Get("cluster-config-v1", metav1.GetOptions{}) // FIXME: error handling
  return yourConfigGenerator(clusterConfig)
}
```

That's a lot of FIXMEs for things I've skipped, and it's still fairly involved.  Is it particularly helpful when we're already linking folks to the not-much-more-complicated real-world machine-API operator implementation? 

[1]: https://github.com/openshift/machine-api-operator/blob/v0.1.0/cmd/common/client_builder.go#L56
[2]: https://github.com/openshift/machine-api-operator/blob/v0.1.0/cmd/common/client_builder.go#L32

docs/dev/operators.md
```diff
@@ -14,7 +14,7 @@ LABEL io.openshift.release.operator=true
 Ensure your image is published into the cluster release tag by ci-operator
 Wait for a new release payload to be created (usually once you push to master in your operator).
```
DanyC97: nit: will be nice to add a period - `tag by ci-operator. Wait for a new release...`

PR #63, opened 2018-12-01
cvo: Record history of changes to version and report target version

smarterclayton:
The CVO needs to clearly communicate to users what the recent changes to
the version have been and what version it is targeting right now.

Add a `history` field that tracks transitions in desiredUpdate, and a
status field `desired` that reports when the operator is trying to reach a
stable state.  `current` becomes a pointer - it is set of `desired` is unset and vice versa.

Example:

1. CVO boots at version 4.0.1 - reports history 4.0.1 and state partial, desired as 4.0.1, current nil
2. CVO successfully syncs 4.0.1 - reports history 4.0.1 and state completed, desired nil, current 4.0.1
3. User sets desired 4.0.2 - reports history [4.0.2, 4.0.1], desired 4.0.2, current nil
4. User cancels desired 4.0.2 - reports history [4.0.2, 4.0.1], desired nil, current 4.0.1

Cancel is interesting today - if you catch it before the CVO reboots, it goes back to 4.0.1.  If you catch it after, you stay at 4.0.2.  Since we want to encourage roll forward, it kind of makes sense, just feels ugly to an end user.

As a second change, allow a user to specify only `version` on a desired update and infer the payload from the availableUpdates or history.  If the version doesn't match, report an invalid condition.

Change how directory filenames are created by using an md5 sum of the payload to avoid scenarios where the payload image string changes for the same version but we use the old content.


vendor/github.com/openshift/api/config/v1/types_cluster_version.go
```diff
@@ -73,11 +73,16 @@ type ClusterVersionSpec struct {
 // progress, or is failing.
 // +k8s:deepcopy-gen=true
 type ClusterVersionStatus struct {
-	// current is the version that the cluster will be reconciled to. This
-	// value may be empty during cluster startup, and then will be set whenever
-	// a new update is being applied. Use the conditions array to know whether
-	// the update is complete.
-	Current Update `json:"current"`
+	// desired is the version that the cluster is reconciling towards.
```
abhinavdahiya: i think we might need an `ObservedGeneration` so that we can track `DesiredUpdate` change to `Desired`

abhinavdahiya: Ahh, didn't see https://github.com/openshift/cluster-version-operator/pull/63/files#diff-cf5e673e310efadb9cd9a2cbe7f58fdbR91

vendor/github.com/openshift/api/config/v1/types_cluster_version.go
```diff
@@ -106,6 +111,33 @@ type ClusterVersionStatus struct {
...
+)
+
+type UpdateHistory struct {
+	// state reflects whether the update was fully applied
+	State UpdateState `json:"state"`
+
+	StartedTime    metav1.Time  `json:"startedTime"`
+	CompletionTime *metav1.Time `json:"completionTime"`
+
+	// version is a semantic versioning identifying the update version. When this
```
abhinavdahiya: can we just reuse `Update` or just embed it.. ?

abhinavdahiya:
@smarterclayton except https://github.com/openshift/cluster-version-operator/pull/63#discussion_r245107098 i think this change looks good.

pkg/cvo/cvo.go
```diff
@@ -465,11 +464,14 @@ func (optr *Operator) getOrCreateClusterVersion() (*configv1.ClusterVersion, boo
 
 // versionString returns a string describing the current version.
```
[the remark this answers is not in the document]
smarterclayton: Yes and yes.

vendor/github.com/openshift/api/config/v1/types_cluster_version.go
smarterclayton: We could embed it, although sometimes I dislike embedding when the structs are used in different contexts to avoid accidentally getting a field we don't want.  I don't care too much.

smarterclayton:
Let me create the PR against openshift/api and we can discuss a bit more there.

vendor/github.com/openshift/api/config/v1/types_cluster_version.go
smarterclayton: Best reason to have different is to have different godoc / apidoc.  I think in this case we want different godoc.

smarterclayton:
Updated with latest API changes from upstream.

PR #65, opened 2018-12-05
docs/dev/operators: Document ComponentOverride injection

wking:
To help devs to test one-off object changes.

Builds on #59; review that first.

CC @abhinavdahiya, @runcom

### DIFF
--- a/docs/dev/clusterversion.md
+++ b/docs/dev/clusterversion.md
@@ -1,0 +1,95 @@
+# ClusterVersion Custom Resource
+
+The `ClusterVersion` is a custom resource object which holds the current version of the cluster.
+This object is used by the administrator to declare their target cluster state, which the cluster-version operator (CVO) then works to transition the cluster to that target state.
+
+## Finding your current update payload
+
+You can extract the current update payload from the `ClusterVersion` object:
+
+```console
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
Can you drop #59 changes. They don't seem related

wking:
Ok, I've pushed 636905e -> 98e1d72, addressing all of your review (I think ;), except for [the `jq` vs. JSONPath one][1].  More on why I'm using `jq` [here][2].  I'm happy to switch to JSONPath if there's a way to get it to spit out JSON objects.  And I'll grudgingly switch to JSONPath in the absence of JSON-object output if you really don't want to involve `jq`.  Thoughts?

[1]: https://github.com/openshift/cluster-version-operator/commit/636905e14d256dd50b88966e6a468278f8e177b0#r31563572
[2]: https://github.com/openshift/cluster-version-operator/commit/636905e14d256dd50b88966e6a468278f8e177b0#commitcomment-31634117

wking:
Oops, I just pushed 98e1d72 -> f83e20b adding notes about disabling the CVO completely.  @abhinavdahiya, can you take another look and maybe re-`/lgtm`?

PR #67, opened 2018-12-10
sync: add version annotation to all objects managed by cvo.

abhinavdahiya:
Adds following annotation to all objects.
```
v1.cluster-version-operator.operators.openshift.io/version: <release-payload version>
```

### DIFF
--- a/pkg/cvo/sync.go
+++ b/pkg/cvo/sync.go
@@ -20,6 +20,9 @@ import (
 )
 
 const (
+	// VersionAnnotationKey is used to annotate all objects owned by CVO with the version of the release payload.
+	VersionAnnotationKey = "v1.cluster-version-operator.operators.openshift.io/version"
+
 	// RequeueOnErrorAnnotationKey is key for annotation on a manifests object that instructs CVO to requeue on specific errors.
 	// The value is comma separated list of causes that forces requeue.
 	RequeueOnErrorAnnotationKey = "v1.cluster-version-operator.operators.openshift.io/requeue-on-error"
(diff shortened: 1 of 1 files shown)


pkg/cvo/sync.go
```diff
@@ -20,6 +20,9 @@ import (
 )
 
 const (
+	// VersionAnnotationKey is used to annotate all objects owned by CVO with the version of the release payload.
+	VersionAnnotationKey = "v1.cluster-version-operator.operators.openshift.io/version"
```
smarterclayton: Let's try `cvo.openshift.io/version` for succinctness.  I think I'm ok with this being version less (in practice you can't break annotations without an immense amount of pain).

Re label vs annotation, we have a couple of choices:

1. Do we ever want to query based on this? (I would like to, but it's not required)
2. Other data (besides version) would probably be something we would do in annotations
3. can we have characters in a semantic version that won't work in a label value?

smarterclayton: Also, for the official version stuff (David and Derek's doc) for managing operator versions, we have two open questions:

1. Is downward API of label/annotation into pods sufficient to cover all cases where we need the version from the payload in an operator
2. Do we need payload generation to be able to provide exact version injection (doing this limits a 3rd parties ability to rebuild a payload, if version info suddenly has behavioral meaning).  I.e. someone takes 4.0.1 and they want to make a variant, are they *required* to do 4.0.1-somethingsemantic because otherwise the kube-apiserver-operator horribly breaks.

PR #69, opened 2018-12-12
cvo: record events when syncing payload

abhinavdahiya:


smarterclayton:
Do we need to resurrect this or spend more time on the reconciling loop and making sure it has good messages?

abhinavdahiya:
closing because this is too old now.

/close

PR #73, opened 2019-01-02
docs: update status explanation

deads2k:
Clarify condition meanings and harmonize with the API.

The different dimensions are clarified and the impact of each being in a "bad" state is described.

/assign @smarterclayton 

/cc @bparees

### DIFF
--- a/docs/dev/clusteroperator.md
+++ b/docs/dev/clusteroperator.md
@@ -64,9 +64,15 @@ Refer [the godocs](https://godoc.org/github.com/openshift/api/config/v1#ClusterS
 
 In general, ClusterOperators should contain at least three core conditions:
 
-* `Progressing` which is true if the controller has detected the state is invalid and is working towards it (or if the controller has never reconciled before)
-* `Available` should be true if the controller has successfully reached a working status and no known errors are blocking progress.
-* `Failing` should be true if the controller is blocked from reaching the desired state.
+* `Progressing` must be true if the operator is actually making change to the operand.
+The change may be anything: desired user state, desired user configuration, observed configuration, version update, etc.
+If this is false, it means the operator is not trying to apply any new state.
+If it remains true for an extended period of time, it suggests something is wrong in the cluster.  It can probably wait until Monday.
+* `Available` must be true if the operand is functional and available in the cluster at the level in status.
+If this is false, it means there is an outage.  Someone is probably getting paged.
+* `Failing` should be true if the operator has encountered an error that is preventing it or it's operand from working properly.
+The operand may still be available, but intent may not have been fulfilled.
+If this is true, it means that the operand is at risk of an outage or improper configuration.  It can probably wait until the morning, but someone needs to look at it.
 
 The message reported for each of these conditions is important.  All messages should start with a capital letter (like a sentence) and be written for an end user / admin to debug the problem.  `Failing` should describe in detail (a few sentences at most) why the current controller is blocked. The detail should be sufficient for an engineer or support person to triage the problem. `Available` should convey useful information about what is available, and be a single sentence without punctuation.  `Progressing` is the most important message because it is shown by default in the CLI as a column and should be a terse, human-readable message describing the current state of the object in 5-10 words (the more succinct the better).
 
@@ -85,7 +91,7 @@ If the controller reaches 4.0.1, the conditions might be:
 If an error blocks reaching 4.0.1, the conditions might be:
 
 * `Failing` is true with a detailed message `Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.`
-* `Available` is false with message `Cluster is unable to reach 4.0.1`
+* `Available` is true with message `Cluster has deployed 4.0.0`
 * `Progressing` is true with message `Unable to apply 4.0.1: a required object is missing`
 
 The progressing message is the first message a human will see when debugging an issue, so it should be terse, succinct, and summarize the problem well.  The failing message can be more verbose. Start with simple, easy to understand messages and grow them over time to capture more detail.


docs/dev/clusteroperator.md
```diff
@@ -64,9 +64,15 @@ Refer [the godocs](https://godoc.org/github.com/openshift/api/config/v1#ClusterS
 
 In general, ClusterOperators should contain at least three core conditions:
 
-* `Progressing` which is true if the controller has detected the state is invalid and is working towards it (or if the controller has never reconciled before)
-* `Available` should be true if the controller has successfully reached a working status and no known errors are blocking progress.
-* `Failing` should be true if the controller is blocked from reaching the desired state.
+* `Progressing` must be true if the operator is actually making change to the binary maintained by the operator (operand).
```
smarterclayton: There is no requirement an operator have a binary - use a different term

docs/dev/clusteroperator.md
```diff
@@ -64,9 +64,15 @@ Refer [the godocs](https://godoc.org/github.com/openshift/api/config/v1#ClusterS
...
+The change may be anything: desired user state, desired user configuration, observed configuration, version update, etc.
+If this is false, it means the operator is not trying to apply any new state.
+If it remains true for an extended period of time, it suggests something is wrong in the cluster.  It can probably wait until Monday.
+* `Available` must be true if the binary maintained by the operator (operand) is functional and available in the cluster.
+If this is false, it means there is an outage.  Someone is probably getting paged.
+* `Failing` should be true if the operator has encountered an error that is preventing it or it's operand from working properly.
+The operand may still be available, but intent may not have been fulfilled.
+If this is true, it means that the operand is at risk of an outage or improper configuration.  It can probably wait until the morning, but someone needs to look at it.
```
smarterclayton: i don’t know of a case where available false and failing not true

docs/dev/clusteroperator.md
```diff
@@ -85,7 +91,7 @@ If the controller reaches 4.0.1, the conditions might be:
 If an error blocks reaching 4.0.1, the conditions might be:
 
 * `Failing` is true with a detailed message `Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.`
-* `Available` is false with message `Cluster is unable to reach 4.0.1`
```
smarterclayton: Why did you change my examples?  These are actual failures.  Add new ones.

docs/dev/clusteroperator.md
deads2k: > i don’t know of a case where available false and failing not true

1. initial creation
2. no nodes available
3. controller manager down, so the operator created a workload resource and is waiting for pods
4. apiservice for operand is Available=false, but there was no error creating it.

docs/dev/clusteroperator.md
deads2k: > Why did you change my examples? These are actual failures. Add new ones.

Because I think the examples are currently incorrect.  During an upgrade if you get stuck, the operator status is `Available==true` because there isn't an outage.  Having `Available=false` mean "the operand is unavailable" provides a clear dimension to create a alerts on.  The example in master doesn't have clear distinction between "page me!" and "this is stuck" and in this particular case required a lot of intelligence that doesn't seem tractable.

docs/dev/clusteroperator.md
```diff
@@ -85,7 +91,7 @@ If the controller reaches 4.0.1, the conditions might be:
...
-* `Progressing` is true with message `Unable to apply 4.0.1: a required object is missing`
```
deads2k: this one changed because it logically required an error flow feeding into a progressing message decision which is duplicating information.

docs/dev/clusteroperator.md
smarterclayton: You are available=false because the operand (the nested operators) are also not available during an upgrade that is failing halfway through.

docs/dev/clusteroperator.md
smarterclayton: This is as designed, and it's not duplication, it's summarization.

Progressing tells you what is happening.  Available tells you why it's unavailable.  Failing tells you the details.

smarterclayton:
/hold 

the messages are not correct, we still need to clarify the other issues.  The clarification on available and failing is something that should be added as an example and in the text (an operator that is spinning up is available=false, progressing=true, and failing=false).  I think we need to add better examples of available false and what the operator is allowed to go false on.

docs/dev/clusteroperator.md
bparees: what nested operators?

for a use case like the registry operator, the registry may or may not be available when the registry operator is failing to upgrade, that's up to the registry operator to determine (and in most cases i'd expect us to remain available, at the previous version).

docs/dev/clusteroperator.md
bparees: my interpretation/understanding was:

available tells you if the operand is available or why not if not

progressing tells you if the operand is being modified/reconfigured (and if so, why)

failing tells you the operator has given up on managing/configuring/reconfiguring the operand for some reason (and why).

ie as @deads2k said, they are orthogonal dimensions, not (necessarily) correlated information/explanations.

docs/dev/clusteroperator.md
smarterclayton: These are good examples to have in the doc

docs/dev/clusteroperator.md
```diff
@@ -64,9 +64,15 @@ Refer [the godocs](https://godoc.org/github.com/openshift/api/config/v1#ClusterS
 
 In general, ClusterOperators should contain at least three core conditions:
 
-* `Progressing` which is true if the controller has detected the state is invalid and is working towards it (or if the controller has never reconciled before)
-* `Available` should be true if the controller has successfully reached a working status and no known errors are blocking progress.
-* `Failing` should be true if the controller is blocked from reaching the desired state.
+* `Progressing` must be true if the operator is actually making change to the operand.
```
smarterclayton: We should clarify what to do when we don't have an operator per se.  I'll add that in as a follow up PR.

docs/dev/clusteroperator.md
```diff
@@ -85,7 +91,7 @@ If the controller reaches 4.0.1, the conditions might be:
 If an error blocks reaching 4.0.1, the conditions might be:
 
 * `Failing` is true with a detailed message `Unable to apply 4.0.1: could not update 0000_70_network_deployment.yaml because the resource type NetworkConfig has not been installed on the server.`
-* `Available` is false with message `Cluster is unable to reach 4.0.1`
+* `Available` is true with message `Cluster has deployed 4.0.0`
```
smarterclayton: Of course, now that I really read through this, if NetworkConfig doesn't exist as a CRD we should be paging someone.

smarterclayton:
/unhold

We can do more as a follow up.

PR #76, opened 2019-01-09
WIP: cvo: Label all applied resources with the current version

smarterclayton:
Knowing that an object has changed is very useful when debugging the state
of a cluster, and the version also allows us to perform downward API
actions from deployments to match the desired version out into the outcome
status.

Right now this tags all objects. An alternative would be to only tag:

1. Namespaces and deployments
2. Objects that have the label `cvo.openshift.io/version` already set (i.e. they request it)
3. Objects that don't have the label set

There is other useful metadata that isn't set (but may be more appropriate to annotations):

* Which file in the payload it came from (for debugging)

One unfortunate side effect is that this forces all objects to be updated,
but in practice we do that anyway.

Refactor the internal structure of the resource builder so that we can
inject testing of dynamic client actions. The interface may change in the
future, so this is merely to improve the testing of payload application for
now.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -504,13 +504,14 @@
   version = "kubernetes-1.11.1"
 
 [[projects]]
-  digest = "1:e9232c127196055e966ae56877363f82fb494dd8c7fda0112b477e1339082d05"
+  digest = "1:df623efa281121be190731b5fba78d24f5ceb4fb9345c4ee45a385f8645b7d2c"
   name = "k8s.io/client-go"
   packages = [
     "discovery",
     "discovery/cached",
     "discovery/fake",
     "dynamic",
+    "dynamic/fake",
     "informers",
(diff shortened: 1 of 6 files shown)


smarterclayton:
@abhinavdahiya  @deads2k this is the simplest possible variant of giving operators / administrators insight into what version a resource is "supposed" to be.  There is other metadata that would be useful that I omitted, but this could be a start.

abhinavdahiya:
hmm, is this different from https://github.com/openshift/cluster-version-operator/pull/67 ?

smarterclayton:
Wow, very close.  It is, we can iterate on that one, although I think we should talk about naming and whether we want a label.

abhinavdahiya:
> Wow, very close. It is, we can iterate on that one, although I think we should talk about naming and whether we want a label.

sure, i can reroll it with name and label used in this PR...

Issue #77, opened 2019-01-09
Errors while upgrading and sync loop interactions


smarterclayton:
The sync loop being inline with the status update loop makes upgrades hard to debug (we have to wait for the whole loop to complete to update status).  Also, if a sync is running and you change the desired state, we need to wait for the sync to complete.  There's some other errors that then show up to users that are non obvious.

Instead, we should decouple the CV status from the sync loop, and make the sync loop cancellable.

To do that, I think we should:

1. Create a new sync loop, similar to available updates
2. Have the main loop (the one that reacts to CV changes) issue instructions to the sync loop like: Sync to version X
3. The sync loop would handle reconciling payloads, rate limiting how often it starts a resync (vs a change to a new version), and reporting completion.  On completion, the sync loop would requeue the CV (like available updates loop does).
4. The main loop would observe the state of the sync loop and update status (like available updates).
5. During a sync, have the CV status get updated, although no more often than every 30s or so
6. Use context signalling to stop an update in progress (the code should be structured around a context)

End outcomes:

1. When a user sets a desired update and we are syncing the previous version, we stop immediately
2. When the user sets a desired update and we are still processing a previous update, we communicate that in status to the user immediately but continue processing if we aren't hitting a failure
3. When the user clears desiredUpdate we immediately go back to syncing our current state
4. Status updates from CVO feel immediate

Part of https://jira.coreos.com/browse/CORS-954

smarterclayton:
@abhinavdahiya I think this is a 4.0 blocker for upgrades because the experience of waiting for the sync loop to react now that our upgrades take 5-10 minutes is very painful.   I'm going to sketch out what the code might look like and open a prototype PR.

smarterclayton:
Fixed in #82.

PR #82, opened 2019-01-11
sync: Do config syncing in the background

smarterclayton:
Implement #77

* Refactor payload syncing into its own loop, and the main loop tells it what to do and gathers status on where it is
   * Take subtle behavior of syncing and put it into the worker, not the sync loop (retries, timeouts, remembering status)
* Writing CV status is much simpler and is based only on inputs (what the sync loop is doing + available updates + validation errors)
* Add integration and more extensive core testing

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"math/rand"
 	"net/http"
 	"os"
@@ -16,12 +17,10 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/version"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	apiext "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	apiextinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -47,6 +46,11 @@ var (
 	}
 
 	startOpts struct {
+		// name is provided for testing only to allow multiple CVO's to be running at once
+		name string
+		// namespace is provided for testing only
+		namespace string
+
 		kubeconfig string
 		nodeName   string
 		listenAddr string
@@ -78,6 +82,16 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		startOpts.nodeName = name
 	}
 
+	// exposed for end-to-end testing only
+	startOpts.name = os.Getenv("CVO_NAME")
+	if len(startOpts.name) == 0 {
+		startOpts.name = componentName
+	}
+	startOpts.namespace = os.Getenv("CVO_NAMESPACE")
+	if len(startOpts.name) == 0 {
+		startOpts.namespace = componentNamespace
+	}
+
 	if rootOpts.releaseImage == "" {
 		glog.Fatalf("missing --release-image flag, it is required")
 	}
@@ -99,14 +113,13 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	stopCh := make(chan struct{})
 	run := func(stop <-chan struct{}) {
 
-		ctx := createControllerContext(cb, stopCh)
+		ctx := createControllerContext(cb, startOpts.name, stopCh)
 		if err := startControllers(ctx); err != nil {
 			glog.Fatalf("error starting controllers: %v", err)
 		}
 
+		ctx.CVInformerFactory.Start(ctx.Stop)
 		ctx.InformerFactory.Start(ctx.Stop)
-		ctx.KubeInformerFactory.Start(ctx.Stop)
-		ctx.APIExtInformerFactory.Start(ctx.Stop)
 		close(ctx.InformersStarted)
 
 		select {}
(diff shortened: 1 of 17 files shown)


smarterclayton:
I've got a working prototype of the new structure, although I'm still adding scenario unit tests.

See the comment on the commit, but basically:

1. Create a SyncWorker background loop
2. Have the Operator.sync() method tell the SyncWorker what the desired state is
3. Take whatever the current SyncWorker state is and use that to craft the ClusterVersion status
4. Whenever the SyncWorker records a change, requeue the sync() loop and we'll perform a much simpler status calculation.
5. To prevent excessive writes during update, we rate limit SyncWorker notifications.

I *think* cvo_scenarios_test.go is a lot easier to read and touches all the main code paths for how the core controller works - still need to add a lot more but wanted to checkpoint from you before I moved further.

pkg/cvo/sync_worker.go
```diff
@@ -0,0 +1,488 @@
...
+	retriever   PayloadRetriever
+	builder     ResourceBuilder
+	reconciling bool
+
+	minimumReconcileInterval time.Duration
+
+	// coordination between the sync loop and external callers
+	notify   chan struct{}
+	report   chan SyncWorkerStatus
+	lock     sync.Mutex
```
abhinavdahiya: can we keep the lock next to fields that it is guarding updates... ?

smarterclayton:
There are now two integration tests:

1. Test clean state -> reconcile -> new version -> reconcile
  1. Launch a CVO
  2. Verify it applies the initial payload
  3. Set desired update, verify the upgrade completes
  4. Verify reconcile doesn't update status
2. Test clean state -> reconcile -> failing upgrade -> abort upgrade to previous version
  1. Launch a CVO
  2. Verify it applies the initial payload
  3. Set desired update to a failing payload
  4. Verify it halts and reports errors and the correct status
  5. Set desired update to previous version
  6. Ensure we reconcile back to previous version

abhinavdahiya:
other wise the test refactror and `sync_worker.go` look :100: 
looking at the `cvo_integration_tests.go` right now.

abhinavdahiya:
@smarterclayton 
Ques: Is there a specific reason why `cvo_integration_test.go` is not in separate `e2e` package like [this](https://github.com/openshift/cluster-image-registry-operator/tree/master/test/e2e)? That will force us to not use any internal structs... ?

smarterclayton:
The integration test uses the internal structs to get better testing.  I think there would be a separate e2e test eventually that is more blackbox.  But in this case I can verify things like resync and status info that isn't really available outside.  I can also force predictable payloads (while an e2e test might verify the job part of a payload).

smarterclayton:
> Ques: looking at Update and run interactions, I might be missing something... for a run that takes a long time, each Update is going to backoff the run call exponentially as this as we will try to read notify on next iteration of syncOnce loop ?

That only gets called if the desired state changed, so the sync loop won't be interrupted unless a new desired state is sent.

abhinavdahiya:
> The integration test uses the internal structs to get better testing. I think there would be a separate e2e test eventually that is more blackbox. But in this case I can verify things like resync and status info that isn't really available outside. I can also force predictable payloads (while an e2e test might verify the job part of a payload).

:+1:

abhinavdahiya:
> > Ques: looking at Update and run interactions, I might be missing something... for a run that takes a long time, each Update is going to backoff the run call exponentially as this as we will try to read notify on next iteration of syncOnce loop ?
> 
> That only gets called if the desired state changed, so the sync loop won't be interrupted unless a new desired state is sent.

:+1:

abhinavdahiya:
/approve

This looks good.

smarterclayton:
I'm debugging the new integration job, let me get that up to snuff.  I also want to think through what tests I am missing in integration.

abhinavdahiya:
> I also want to think through what tests I am missing in integration.

- reconcile changes to objects it owns?

smarterclayton:
Integration test is now wired up, but I won't merge it until this merges.

smarterclayton:
job enablement in 
https://github.com/openshift/release/pull/2583

smarterclayton:
> reconcile changes to objects it owns?

yes, will do that

smarterclayton:
Added that test

pkg/cvo/cvo_integration_test.go
```diff
@@ -0,0 +1,712 @@
...
+	lastCV, err = waitForAvailableUpdate(t, client, ns, false, "0.0.1", "0.0.2")
+	if err != nil {
+		t.Logf("latest version:\n%s", printCV(lastCV))
+		t.Fatalf("cluster version never reached available at 0.0.2: %v", err)
+	}
+	t.Logf("Upgraded version:\n%s", printCV(lastCV))
+	verifyClusterVersionStatus(t, lastCV, configv1.Update{Payload: payloadImage2, Version: "0.0.2"}, 2)
+	verifyReleasePayload(t, kc, ns, "0.0.2", payloadImage2)
+
+	t.Logf("delete an object so that the next resync will recover it")
```
abhinavdahiya: This is fine for now, but we should move this to separate test if we can later on.

smarterclayton: so the integration tests are still sort of scenario driven, because it does take non-zero time to do the payload.  I'm halfway in between granular and full scenario tests here.

smarterclayton:
Integration won't run, I have to kick it since the PR didn't merge yet

smarterclayton:
Kicked it

abhinavdahiya:
this change looks good to merge. is there anything else you are testing right now or should i drop a `/lgtm` ?

smarterclayton:
It's ready for the big lgtm

smarterclayton:
There's a lot of follow up tests I want to add

cmd/start.go
```diff
@@ -16,12 +17,10 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/version"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
pruan-rht: typo

abhinavdahiya:
> There's a lot of follow up tests I want to add

let me know if i can help with some :innocent:

PR #83, opened 2019-01-15
pkg/cincinnati: scope HTTP request to a context

lucab:
This adds a context to remote cincinnati calls. The topmost context is
left floating as the caller is not context-aware at the moment.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -1,6 +1,7 @@
 package cvo
 
 import (
+	"context"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -127,7 +128,7 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
 		}
 	}
 
-	updates, err := checkForUpdate(clusterID, upstream, channel, currentVersion)
```
abhinavdahiya: I would assume this is the only place that needs the ctx, why do any body up the chain care about ctx for checking updates?

lucab: I can cut the chain at any arbitrary point that you prefer, but please do note that the topmost `Until` layer already takes a stop channel, but doesn't propagate that down. I'd rather suggest to leave the TODO here and later fix the caller to unify the cancellation chain.

lucab: FYI I landed a fix in k8s master to propagate the context through `Until`: https://github.com/kubernetes/kubernetes/pull/72928

lucab:
This was on hold due to the freeze being in place at the time, it should be fine now.

/hold cancel

lucab:
Rebased, bump for a review.

As stated, I can cut the context wherever you prefer right now, but I personally think that longer term it makes sense to wire all contexts to the top-level one.

lucab:
Gentle ping for a review.

wking:
Carried in #410.

/close

PR #84, opened 2019-01-15
pkg/cincinnati: fix media-type

lucab:
This aligns client-accepted media type to cincinnati spec.

Ref: https://github.com/openshift/cincinnati/blob/06c9ce64367b55f453295b2765a9aab61791dc76/docs/design/cincinnati.md#request

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -11,9 +11,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
/lgtm

/hold

Will let @crawford decide if we can merge this during freeze.

crawford:
Yep, this is a bug that we should fix despite the freeze. Thanks for double checking.

/hold cancel

PR #85, opened 2019-01-16
pkg/cincinnati: add required parameters to requests

lucab:
This adds three mandatory parameters to cincinnati requests, as per
cincinnati-openshift specs.

Ref: https://github.com/openshift/cincinnati/blob/06c9ce64367b55f453295b2765a9aab61791dc76/docs/design/openshift.md#request

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 
 	"github.com/blang/semver"
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


smarterclayton:
Can you guys add a Cincinnati integration test to pkg/start/start_integration_test.go (moved in #87) with a shim Cincinnati endpoint that verifies the behavior?  We have functional tests in cvo_test but an extra higher level test will be good.  Doesn't have to be here, just was looking at the bug and realized it would be a good candidate.

steveeJ:
@smarterclayton is this blocked on the additional test you've asked for?

smarterclayton:
I would like a test showing the url being constructed in cvo_test.go

smarterclayton:
Or scenarios test or thenintegrstion test (the integration test is probably the best place because we can run a local server and report

lucab:
Superseded by https://github.com/openshift/cluster-version-operator/pull/121.

PR #86, opened 2019-01-16
set cpu request for operator

sjenning:
@smarterclayton @derekwaynecarr


smarterclayton:
I'm ok with 20 because this is the root of the tree

/lgtm

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -22,6 +22,10 @@ spec:
           - "--release-image={{.ReleaseImage}}"
           - "--enable-auto-update=false"
           - "--v=4"
+        resources:
+          requests:
+            cpu: 20m
```
derekwaynecarr: Why 20?

PR #78, opened 2019-01-09
restart CVO when credentials rotate causing serviceaccount files to change

abhinavdahiya:
/cc @deads2k @smarterclayton

PR #87, opened 2019-01-17
cvo: Release our leader lease when we are gracefully terminated

smarterclayton:
On an upgrade the CVO is always restarted but we don't release our
lease, which causes the new CVO pod to have to wait. We should
release our lease on exit.

Kube 1.14 (pr 71490) contains a new flag on leader lease that allows
the caller to step down gracefully. Because backporting that change
to client-go is complicated, we instead emulate the logic. When that
code is available we can simplify down.

This also makes install about 90s faster

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,284 +2,36 @@ package main
 
 import (
 	"flag"
-	"fmt"
-	"math/rand"
-	"net/http"
-	"os"
-	"time"
 
 	"github.com/golang/glog"
-	"github.com/google/uuid"
-	clientset "github.com/openshift/client-go/config/clientset/versioned"
-	informers "github.com/openshift/client-go/config/informers/externalversions"
-	"github.com/openshift/cluster-version-operator/pkg/autoupdate"
-	"github.com/openshift/cluster-version-operator/pkg/cvo"
+	"github.com/openshift/cluster-version-operator/pkg/start"
 	"github.com/openshift/cluster-version-operator/pkg/version"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
-	v1 "k8s.io/api/core/v1"
-	apiext "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/tools/leaderelection"
-	"k8s.io/client-go/tools/leaderelection/resourcelock"
-	"k8s.io/client-go/tools/record"
-)
-
-const (
-	minResyncPeriod = 2 * time.Minute
-
-	leaseDuration = 90 * time.Second
-	renewDeadline = 45 * time.Second
-	retryPeriod   = 30 * time.Second
-)
-
-var (
-	startCmd = &cobra.Command{
-		Use:   "start",
-		Short: "Starts Cluster Version Operator",
-		Long:  "",
-		Run:   runStartCmd,
-	}
-
-	startOpts struct {
-		// name is provided for testing only to allow multiple CVO's to be running at once
-		name string
-		// namespace is provided for testing only
-		namespace string
-
-		kubeconfig string
-		nodeName   string
-		listenAddr string
-
-		enableAutoUpdate bool
-	}
 )
 
 func init() {
-	rootCmd.AddCommand(startCmd)
-	startCmd.PersistentFlags().StringVar(&startOpts.listenAddr, "listen", "0.0.0.0:9099", "Address to listen on for metrics")
-	startCmd.PersistentFlags().StringVar(&startOpts.kubeconfig, "kubeconfig", "", "Kubeconfig file to access a remote cluster (testing only)")
-	startCmd.PersistentFlags().StringVar(&startOpts.nodeName, "node-name", "", "kubernetes node name CVO is scheduled on.")
-	startCmd.PersistentFlags().BoolVar(&startOpts.enableAutoUpdate, "enable-auto-update", true, "Enables the autoupdate controller.")
-}
+	opts := start.NewOptions()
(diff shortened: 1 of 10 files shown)


cmd/start.go
```diff
@@ -110,28 +113,70 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+					// WARNING: this is not completely safe until we have Kube 1.14 and ReleaseOnCancel
+					//   and client-go ContextCancelable, which allows us to block new API requests before
+					//   we step down. However, the CVO isn't that sensitive to races and can tolerate
+					//   brief overlap.
+					glog.Infof("Stepping down as leader")
+					// give the controllers some time to shut down
+					time.Sleep(100 * time.Millisecond)
+					// if we still hold the leader lease, clear the owner identity (other lease watchers
+					// still have to wait for expiration) like the new ReleaseOnCancel code will do.
+					if err := lock.Update(resourcelock.LeaderElectionRecord{}); err == nil {
```
abhinavdahiya: reseting to empty looses the https://github.com/kubernetes/client-go/blob/v9.0.0/tools/leaderelection/resourcelock/interface.go#L42 is there a way we can step down without completely loosing that information...?

abhinavdahiya: > other lease watchers still have to wait for expiration

if we reset the https://github.com/kubernetes/client-go/blob/v9.0.0/tools/leaderelection/resourcelock/interface.go#L39-L41 wouldn't the next candidate immediately try to reelect?

cmd/start.go
```diff
@@ -110,28 +113,70 @@ func runStartCmd(cmd *cobra.Command, args []string) {
...
+						// if we successfully clear the owner identity, we can safely delete the record
+						if err := lock.Client.ConfigMaps(lock.ConfigMapMeta.Namespace).Delete(lock.ConfigMapMeta.Name, nil); err != nil {
```
abhinavdahiya: Why delete the config map at all?

cmd/start.go
smarterclayton: No, see the upstream PR - clients wait until the time is clear.  They don't step down today until my PR lands in 1.14.

cmd/start.go
smarterclayton: Deleting the config map is the only way to trigger a new election today.

smarterclayton:
/hold

I need to refactor this to be testable in an integration test (cmd/ is too full anyway)

cmd/start.go
abhinavdahiya: Hmm, okay

smarterclayton:
I took your feedback from previous PR and refactored the start command to have reusable code, and then had the integration test use it and cut the internal dependencies.  Should be cleaner, and start is noticeably simpler (well, to me).

smarterclayton:
I also added an integration test for step down.

pkg/cvo/cvo.go
```diff
@@ -452,3 +452,8 @@ func (optr *Operator) currentVersion() configv1.Update {
 		Payload: optr.releaseImage,
 	}
 }
+
+// SetSyncWorkerForTesting updates the sync worker for whitebox testing.
+func (optr *Operator) SetSyncWorkerForTesting(worker ConfigSyncWorker) {
```
abhinavdahiya: Hmm...

smarterclayton: I contemplated having start lazily configure the worker, but this is a reasonable taint.

cmd/start.go
```diff
@@ -2,284 +2,36 @@ package main
...
-	"os"
-	"time"
 
 	"github.com/golang/glog"
-	"github.com/google/uuid"
-	clientset "github.com/openshift/client-go/config/clientset/versioned"
-	informers "github.com/openshift/client-go/config/informers/externalversions"
-	"github.com/openshift/cluster-version-operator/pkg/autoupdate"
-	"github.com/openshift/cluster-version-operator/pkg/cvo"
+	"github.com/openshift/cluster-version-operator/pkg/start"
```
abhinavdahiya: nit: this could be `pkg/cmd/start` and controller ctx stuff can be `pkg/cmd/common` ?

pkg/start/start.go
```diff
@@ -0,0 +1,316 @@
...
+			},
+		},
+	})
+
+	<-exit
+}
+
+// createResourceLock initializes the lock.
+func createResourceLock(cb *ClientBuilder, namespace, name string) (*resourcelock.ConfigMapLock, error) {
+	recorder := record.NewBroadcaster().NewRecorder(runtime.NewScheme(), v1.EventSource{Component: namespace})
```
abhinavdahiya: ques: shouldn't `runtime.NewScheme()` be `kubernetes.Scheme` ?

pkg/start/start_integration_test.go
```diff
@@ -460,6 +444,107 @@ func TestIntegrationCVO_initializeAndHandleError(t *testing.T) {
...
+			return false, err
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("lock deleted in %s", endTime.Sub(startTime))
+
+	select {
```
abhinavdahiya: I like this ;)

abhinavdahiya:
/approve

i'll rebase https://github.com/openshift/cluster-version-operator/pull/78 on this and add a test for rotation..

abhinavdahiya:
also can you 
- squash this to single commit or 
- 2 commits (move to `pkg/start`) and (leader step down on exit)

?

pkg/start/start.go
smarterclayton: I’m not 100% sure, it doesn’t actually matter for events (we fall back to a generic path if an object isn’t in the scheme)

smarterclayton:
Will squash - teasing them apart is harder now

pkg/start/start.go
smarterclayton: Actually, we didn't have a sink, so we weren't actually recording metrics.  I fixed that here, and made sure each of the places in the CVO we log events we were sending it to our namespace rather than the default namespace (which if you pass "" is where you go).  Added an integration test to verify we receive a leader event, which is a good proxy for us accidentally breaking it in the future.

smarterclayton:
Rebased, squashed, and fixed the event thing.

PR #88, opened 2019-01-18
Parallelize the payload execution

smarterclayton:
Parallelize payload execution using two simple rules:

1. a Job in the payload acts as a serialization barrier (all previous work must complete)
2. two manifests can be run in parallel if:
   a. The name of the manifest matches `0000_<NUMBER>_<COMPONENT>_*`
   b. The number matches
   c. The component does not match

We'd build a graph, then greedily execute sync tasks, starting a node when all prerequisites are satisfied.  We never run manifests in parallel under the same `0000_<NUMBER>_<COMPONENT>` prefix.

The first few rules cut us to about ~120 sequential tasks, instead of 227.

### DIFF
--- a/pkg/cvo/sync_test.go
+++ b/pkg/cvo/sync_test.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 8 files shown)


smarterclayton:
First run... I'm so excited!  Excited = excited to see the flaming pile of rubble after the CVO melts an apiserver.

pkg/cvo/sync_worker.go
```diff
@@ -431,47 +445,56 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
...
+			glog.V(4).Infof("Running sync for %s", task)
+			glog.V(5).Infof("Manifest: %s", string(task.Manifest.Raw))
 
-		ov, ok := getOverrideForManifest(work.Overrides, task.Manifest)
-		if ok && ov.Unmanaged {
-			glog.V(4).Infof("Skipping %s as unmanaged", task)
-			continue
-		}
+			if contextIsCancelled(ctx) {
+				err := fmt.Errorf("update was cancelled at %d/%d", i, len(tasks))
```
smarterclayton: wrong, must fix

smarterclayton:
github/prow/openshift you suck

/test all

smarterclayton:
/assign @abhinavdahiya 

This is ready for review.  Especially the run tests.  I've done a number of long -count runs on the graph traversal, but I'm sure there are tests missing.

pkg/cvo/sync_test.go
```diff
@@ -8,10 +8,10 @@ import (
 	"strings"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/api/meta"
```
abhinavdahiya: nit: moved import..

pkg/cvo/sync_worker.go
```diff
@@ -431,47 +441,42 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
...
-		glog.V(4).Infof("Done syncing for %s", task)
+		return nil
+	})
+	if err != nil {
+		cr.Error(err)
+		return err
 	}
 
-	setAppliedAndPending(version, total, done)
 	work.Completed++
```
abhinavdahiya: why are we tracking completed in both `work` and `consistentReporter` ?

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+// all groups with common XX and different yy-yy-yy may be executed in parallel (0000_99_cluster-monitoring-operator and 0000_99_cluster-samples-operator)
+// within a group:
+//   - all executable or unrecognized types form synchronization barriers (ClusterOperator, Deployment, Pod are all sync barriers)
+//   - all other types may be greedily created (Namespace, RoleBinding, Service, Route)
+// all jobs are global synchronization barriers
+//
+
+// SplitOnJobs enforces the rule that any Job in the payload prevents reordering or parallelism (either before or after)
+func SplitOnJobs(task *Task) bool {
+	return task.Manifest.GVK == schema.GroupVersionKind{Kind: "Job", Version: "v1", Group: "batch"}
```
abhinavdahiya: `k8s.io/api/batch/v1.SchemeGroupVersion.WithKind("Job")`

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+}
+
+var reMatchPattern = regexp.MustCompile(`^0000_(\d+)_([a-zA-Z0-9+]+(-[a-zA-Z0-9+]+)*?)_`)
```
abhinavdahiya: why is `+` allowed in `[a-zA-Z0-9+]` ?

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+
+	var buckets [][]*TaskNode
+	var lastNode *TaskNode
+	for i := 0; i < count; {
+		match := matches[i]
+		j := i + 1
+		var groups []*TaskNode
+		for ; j < count; j++ {
+			next := matches[j]
+			if match == nil || next == nil || match[1] != next[1] {
```
abhinavdahiya: nit: constants like `NN := 0` `NAME := 1` should help readability.

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+//
+
+// SplitOnJobs enforces the rule that any Job in the payload prevents reordering or parallelism (either before or after)
+func SplitOnJobs(task *Task) bool {
+	return task.Manifest.GVK == schema.GroupVersionKind{Kind: "Job", Version: "v1", Group: "batch"}
+}
+
+var reMatchPattern = regexp.MustCompile(`^0000_(\d+)_([a-zA-Z0-9+]+(-[a-zA-Z0-9+]+)*?)_`)
+
+// ByNumberAndComponent creates parallelization for tasks whose original filenames are of the form
```
abhinavdahiya: nit: we need the tasks to be sorted for this, right?

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+	// sends them to workCh. It uses completeCh to know that a previously dispatched item is complete.
+	completeCh := make(chan taskStatus, maxParallelism)
+	defer close(completeCh)
+
+	workCh := make(chan runTasks, maxParallelism)
+	go func() {
+		defer close(workCh)
+
+		// visited tracks nodes we have not sent (0), are currently
+		// waiting for completion (1), or have completed (2)
```
abhinavdahiya: `or have completed (2,3)`

pkg/payload/task_graph.go
```diff
@@ -0,0 +1,521 @@
...
+				visited[finished.index] = nodeFailed
+			}
+			remaining--
+			inflight--
+		}
+
+		// take everything remaining and process in order
+		var unreachable []*Task
+		for i := 0; i < len(visited); i++ {
+			if visited[i] == 0 && canVisit(graph.Nodes[i]) {
```
abhinavdahiya: nit: use `nodeNotVisited`

pkg/cvo/sync_worker.go
smarterclayton: One is updating the status we send back to the main CVO loop (via reporter), one is used internally by the sync_worker to report for the next generation.  I moved this line up to where it belongs, in Start()

pkg/payload/task_graph.go
smarterclayton: I'd prefer not to take the dependency on the actual types when this is a constant that can never change anyway.

pkg/payload/task_graph.go
smarterclayton: him, this was a typo

pkg/payload/task_graph.go
smarterclayton: Yes, I can comment.  If they aren't sorted (perhaps because an earlier step decided the order was incorrect based on characteristics of the workload) then we would not parallelize those steps.

smarterclayton:
responded or fixed everything

PR #89, opened 2019-01-18
cvo: Don't wake the CVO on a cluster operator change

smarterclayton:
We don't actually react to cluster operator objects today. We can
safely ignore this edge since we already wait inside sync loop.


abhinavdahiya:
makes sense.

/lgtm

PR #90, opened 2019-01-18
Increase the QPS limits on clients slightly and use protobuf for sync

smarterclayton:
For the kube apis, protobuf is more efficient on both client and server
side which reduces the base load of the cluster. We can't set protobuf for
custom resource clients or unstructured at the current time.

We are getting a fair amount of throttling during deployment and updates.
Increase QPS and Burst to be 4x the default. Refactor the start client
slightly to allow modifying the client in case we need to do this in the
future.

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -1,3 +1,5 @@
+// package start initializes and launches the core cluster version operator
+// loops.
 package start
 
 import (
@@ -41,6 +43,7 @@ const (
 	retryPeriod   = 30 * time.Second
 )
 
+// Options are the valid inputs to starting the CVO.
 type Options struct {
 	ReleaseImage string
 
@@ -66,6 +69,8 @@ func defaultEnv(name, defaultValue string) string {
 	return env
 }
 
+// NewOptions creates the default options for the CVO and loads any environment
+// variable overrides.
 func NewOptions() *Options {
 	return &Options{
 		ListenAddr: "0.0.0.0:9099",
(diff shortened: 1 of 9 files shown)


pkg/start/start.go
```diff
@@ -231,21 +231,24 @@ type ClientBuilder struct {
 	config *rest.Config
 }
 
-func (cb *ClientBuilder) RestConfig() *rest.Config {
+func (cb *ClientBuilder) RestConfig(fns ...func(*rest.Config)) *rest.Config {
```
abhinavdahiya:
```console
$ godoc ./pkg/start
PACKAGE DOCUMENTATION

package start
    import "./pkg/start"


TYPES

type ClientBuilder struct {
    // contains filtered or unexported fields
}


func (cb *ClientBuilder) ClientOrDie(name string, fns ...func(*rest.Config)) clientset.Interface

func (cb *ClientBuilder) KubeClientOrDie(name string, fns ...func(*rest.Config)) kubernetes.Interface

func (cb *ClientBuilder) RestConfig(fns ...func(*rest.Config)) *rest.Config
```

`fns ...` isn't the best `godoc`, maybe either add comments / name the variable that it is evident that these are config modifiers...

pkg/start/start.go
```diff
@@ -291,7 +299,7 @@ func (o *Options) NewControllerContext(cb *ClientBuilder) *Context {
 			resyncPeriod(o.ResyncInterval)(),
 			cvInformer.Config().V1().ClusterVersions(),
 			sharedInformers.Config().V1().ClusterOperators(),
-			cb.RestConfig(),
+			cb.RestConfig(increaseQPS),
```
abhinavdahiya: this will only affect the generic resource builders. what about the `ClientOrDie, KubeClientOrDie` atleast KubeClientOrDie ?

abhinavdahiya:
@smarterclayton few nits, lgtm otherwise.

pkg/start/start.go
smarterclayton: oh, sorry

pkg/start/start.go
smarterclayton: I left those at the defaults, those have low qps and low actions (rest config is what is used for sync loop).  I'd like to do a shared qps pool eventually though for everything that isn't status updates to CV (which is effectively what this PR is).

smarterclayton:
updated with better godoc

PR #92, opened 2019-01-19
metrics: Report the age of the current payload for alerting

smarterclayton:
One attack vector for clusters is to deny access to new updates or
to deliberately send an older update with vulnerabilities. This commit ensures the date the
payload was created is reported back through metrics and up via
telemetry. It also adds a new `completed` type for the cluster_version
metric that reports the last successfully applied cluster version and
the date it was successfully applied the first time (from the new
status history field). Finally, we correct the naming for the update
type to match the name of the status field to `desired`.

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -2,13 +2,23 @@
 
 The Cluster Version Operator reports the following metrics:
 
-Core metrics for the version
+Core metrics for the version. Types:
+
+* `current` - the version the operator is applying right now (the running CVO version)
+* `failure` - if the failure condition is set, reported for desired and current versions
+* `desired` - reported if different from current
+* `completed` - returns 1 and the info of the most recent `Completed` entry in the history, or 0 otherwise
+
+`current` has the `age` label set to the unix timestamp of the payload age, `completed` has the `age` label
+set to the time at which the update was completed. All other variants have empty age labels.
 
 ```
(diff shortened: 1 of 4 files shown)


smarterclayton:
Ping on this

docs/dev/metrics.md
```diff
@@ -2,13 +2,23 @@
...
 
-Core metrics for the version
+Core metrics for the version. Types:
+
+* `current` - the version the operator is applying right now (the running CVO version)
+* `failure` - if the failure condition is set, reported for desired and current versions
+* `desired` - reported if different from current
+* `completed` - returns 1 and the info of the most recent `Completed` entry in the history, or 0 otherwise
+
+`current` has the `age` label set to the unix timestamp of the payload age, `completed` has the `age` label
```
abhinavdahiya: User seeing the metrics current with these others is going to assume that the value has probably something to do with cvo applying it.. while this value is only related to the release image?

smarterclayton: I agree it's a reasonable concern.  Right now these two things are kind of coupled - I could split them, but it's not clear what the split is.  I could add another label.  Not sure.

smarterclayton: I could imagine us having `cluster_version_image` and `cluster_version`, with `type` matching across them.  I need to see what the actual data is though

PR #95, opened 2019-01-22
make it clear that everyone needs a clusteroperator

deads2k:
Everyone needs a clusteroperator so cluster-admins know the state of their cluster and so that the CVO can be sure that upgrades are successful before trying to apply the next payload.

/assign @smarterclayton @eparis


abhinavdahiya:
https://github.com/openshift/cluster-version-operator/pull/95#event-2089564505

feel free to ping any one of us https://github.com/openshift/cluster-version-operator/blob/master/OWNERS

/approve

PR #97, opened 2019-01-23
refactor: Extract payload related functions for clarity

smarterclayton:
The cvo package is getting a bit full. In preparation for parallel payload processing where we want to have a readable graph processor, move payload related functions into their own package. The syncTask struct is renamed to payload.Task.  Code movement, a few structs are made public purely for the ability to use diff.ObjectReflectDiff on them in testing.

Extracted from #88 to make it easier to change the applier function.


pkg/cvo/sync_worker.go
```diff
@@ -453,11 +459,102 @@ func (w *SyncWorker) apply(ctx context.Context, payload *updatePayload, work *Sy
 
 	setAppliedAndPending(version, total, done)
 	work.Completed++
-	reporter.Report(SyncWorkerStatus{Fraction: 1, Completed: work.Completed, Reconciling: true, VersionHash: payload.ManifestHash, Actual: update})
+	reporter.Report(SyncWorkerStatus{Fraction: 1, Completed: work.Completed, Reconciling: true, VersionHash: payloadUpdate.ManifestHash, Actual: update})
 
 	return nil
 }
 
+var (
```
smarterclayton: These will only reside in sync_worker until the graph iterator (pulling chunks of work from the payload graph and executing in parallel) is added.

smarterclayton:
@abhinavdahiya I know this looks scary, wanted to split it now (mostly) before I play with a different parallel execution

abhinavdahiya:
> @abhinavdahiya I know this looks scary, wanted to split it now (mostly) before I play with a different parallel execution

👻 I just couldn't get to it today, was busy with some other work. I'll take a look at this tomorrow first thing

smarterclayton:
Sorry to keep hitting you with these large reviews

lib/manifest.go
```diff
@@ -76,6 +79,9 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
 			errs = append(errs, errors.Wrapf(err, "error parsing %s", file.Name()))
 			continue
 		}
+		for _, m := range ms {
```
abhinavdahiya: nit: so a bunch of manifests will have the same name... can we add that as comment to `Name` field. please do not use it to dedup or as a map key.

pkg/cvo/sync_worker.go
```diff
@@ -371,47 +377,47 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, reporter Stat
 	glog.V(4).Infof("Running sync %s", versionString(work.Desired))
 	update := work.Desired
 
-	// cache the image until the release image changes
-	image := w.image
-	if image == nil || !equalUpdate(configv1.Update{Image: image.ReleaseImage}, update) {
-		glog.V(4).Infof("Loading image")
+	// cache the payload until the release image changes
+	validPayload := w.payload
```
abhinavdahiya: nit: this change `s/image/validPayload/` seems to reverse the idea from https://github.com/openshift/cluster-version-operator/pull/98

abhinavdahiya: nvm, i think it is to keep up with package name `pkg/payload`.

abhinavdahiya:
The refactor in https://github.com/openshift/cluster-version-operator/pull/97/commits/d14c4018e576d5ab3b48ff6074c91007703de993 looks a good one-to-one move with no major changes to behavior.

/lgtm

abhinavdahiya:
/hold

for https://github.com/openshift/cluster-version-operator/pull/97#discussion_r251062549

pkg/cvo/sync_worker.go
smarterclayton: Yeah, I did a find and replace before, but went back to being it here to have payload be the internals.

smarterclayton:
Updated with comment (move to OriginalFilename and add a comment).

smarterclayton:
Integration test was broken by rebase, I'll disable it until there is a working fix.

PR #98, opened 2019-01-23
api: Rename Payload to Image in ClusterVersion status

smarterclayton:
The payload is delivered as an image, and we have clarified our terminology in multiple spots to say:

* An `update` is delivered in an `image`
* An image carrying an OpenShift update is a `release image`
* A `release image` is associated with a `version` of the product
* The contents of the `release image` include the CVO binary and an `update payload`
* The `update payload` is a set of on disk manifests and associated metadata

This PR includes the openshift/api changes and reacts to those updates.


abhinavdahiya:
both https://github.com/openshift/cluster-version-operator/pull/98/commits/abd3ec0e71403629aa495266c2882346e7f519bd and https://github.com/openshift/cluster-version-operator/pull/98/commits/dd266e160646afa9e1f986c9f224ef2e2cfc4b5f are both changing `vendor/github.com/openshift/api/config/v1/types_cluster_version.go` ;)

pkg/cvo/internal/operatorstatus.go
```diff
@@ -106,9 +107,8 @@ func waitForOperatorStatusToBeDone(client configclientv1.ClusterOperatorsGetter,
 		if available && !progressing && !failing {
 			return true, nil
 		}
-		glog.V(3).Infof("ClusterOperator %s/%s is not done for version %s; it is version=%v, available=%v, progressing=%v, failing=%v",
-			eos.Namespace, eos.Name, os.Status.Version,
-			eos.Status.Version, available, progressing, failing)
+		glog.V(3).Infof("ClusterOperator %s/%s is not done for version %s; it is available=%v, progressing=%v, failing=%v",
```
abhinavdahiya: `version %s` is extra

pkg/cvo/metrics.go
```diff
@@ -133,7 +133,12 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 	// output cluster operator version and condition info
 	operators, _ := m.optr.clusterOperatorLister.List(labels.Everything())
 	for _, op := range operators {
-		g := m.clusterOperatorUp.WithLabelValues(op.Namespace, op.Name, op.Status.Version)
+		var firstVersion string
```
abhinavdahiya: are we assuming the operator version is going to be the first?

smarterclayton: No, I just didn't want to remove the metric value.  We're going to need to define it, I'll add a todo

smarterclayton:
Fixed up version info, added two, rebased, also fixed a metric panic (previous failure for e2e-aws) where describe wasn't right.

I'm going to look at trying to add a metric integration test (but it has to be serial, which sucks)

PR #101, opened 2019-01-24
Add test cases to detect invalid version histories

smarterclayton:
The CVO is reporting history entries back to back that have the same
image and version. Add an integration test to check the status of the
history.


smarterclayton:
/hold

while I debug

PR #102, opened 2019-01-25
status: Generation should be set by what the sync worker is doing

smarterclayton:
The primary mechanism where by a reconciling controller communicates to a client is its status, and the client has to wait until the controller observes its write before assuming the controller is ready.

When writing a client that is triggering an upgrade, we need to:

1. Set desiredUpdate and read resulting generation 
2. Wait until status.generation == generation from 1 
3. Wait until the update that is marked partial in status history is complete

This fixes the sync worker to be the one who propagates generation, so that a client doesn't see 2 as true when syncing a previous generation.


abhinavdahiya:
https://github.com/openshift/cluster-version-operator/pull/102/commits/8ea7884f3f87c9d94879cfa23cd2e43022d0aee4 is a good catch.

/approve

do you want to wait for https://github.com/openshift/api/pull/185/files and bring that in as vendor update ?

smarterclayton:
Yes (that's where I caught this)

PR #103, opened 2019-01-25
metrics: Report cluster_version as seconds since the epoch per type

smarterclayton:
Building on feedback from the first round, report `cluster_version`
for each type as seconds since the epoch. The value allows relevant
data to be extracted such as the age of the cluster, the age of the
current payload, when the most recent successful update was applied,
and when the most recent failure started occurring.

Example queries:

    # the oldest clusters and what versions they are running
    sort(cluster_version{type="cluster"})

    # the most recently successfully updated clusters
    sort_desc(cluster_version{type="completed"})

    # the 20 clusters that recently started failing
    topk(20, cluster_version{type="failure"})

    # the clusters running the oldest payloads
    sort(cluster_version{type="current"})

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -1,7 +1,7 @@
 package cvo
 
 import (
-	"strconv"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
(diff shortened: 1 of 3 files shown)


pkg/cvo/metrics.go
```diff
@@ -162,3 +177,29 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
...
+				latest = t.Time
+			}
+		}
+	}
+	for _, condition := range cv.Status.Conditions {
+		if condition.LastTransitionTime.After(latest) {
+			latest = condition.LastTransitionTime.Time
+		}
+	}
+	if cv.CreationTimestamp.After(latest) {
```
abhinavdahiya: nit:
hmm, the placement for this check is odd, i think we are pretty sure that creation time is less than history and conditions...

abhinavdahiya:
This looks good, Thanks!

/lgtm

PR #104, opened 2019-01-29
docs: catch up to https://github.com/openshift/api/pull/176

abhinavdahiya:
/cc @deads2k

### DIFF
--- a/docs/dev/clusteroperator.md
+++ b/docs/dev/clusteroperator.md
@@ -56,9 +56,33 @@ status:
 
 The operator should ensure that all the fields of `.status` in ClusterOperator are atomic changes. This means that all the fields in the `.status` are only valid together and do not partially represent the status of the operator.
 
-### Version
+### Versions
 
-The operator should report a version which indicates the components that it is applying to the cluster.
+The operator should report a list of versions which indicate the components that it is applying to the cluster.
+
+An example of the `.status.versions` reported by the operator are:
+
+```yaml
+versions:
+- name: operator
+ version: v1.0.0
+- name: operand-1
+ version: v0.12.0
+...
+```
+
+Or,
+
+```yaml
+versions:
+- name: operator
+ version: image/release-image@sha256:12345
+- name: operand-1
+ version: image/release-image@sha256:67890
+...
+```
+
+Each version entry must match the `Available` condition.  If `1.0.0` is `Available`, then this must indicate `1.0.0` even if the operator is trying to rollout `1.1.0`. And the versions list must always include `operator` version.
 
 ### Conditions
 


docs/dev/clusteroperator.md
```diff
@@ -56,9 +56,33 @@ status:
...
+ version: v0.12.0
+...
+```
+
+Or,
+
+```yaml
+versions:
+- name: operator
+ version: image/release-image@sha256:12345
```
deads2k: indicate that this is considered a stepping stone to the real version while we work out substitution values.

docs/dev/clusteroperator.md
```diff
@@ -56,9 +56,33 @@ status:
...
+The operator should report a list of versions which indicate the components that it is applying to the cluster.
+
+An example of the `.status.versions` reported by the operator are:
+
+```yaml
+versions:
+- name: operator
+ version: v1.0.0
+- name: operand-1
+ version: v0.12.0
```
bparees: is there a recommended way for an operator to deduce the version of its operand?  Obviously it depends somewhat on the nature of the operand, but at least for operands that have images?

wking: > ... but at least for operands that have images?

Inspect the image to extract [`org.opencontainers.image.version`][1]?  And/or whatever the OpenShift equivalent is, since I haven't had much luck advocating for OCI annotations ;).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.1/annotations.md#L25-L27

bparees: Expecting operators to pull+inspect images doesn't seem reasonable to me.

From what i've heard they're going to be supplied this information in a similar way to how image-references works, so we can ultimately create an env var that will be populated w/ version info for the components we care about.

wking: > From what i've heard they're going to be supplied this information in a similar way to how image-references works, so we can ultimately create an env var that will be populated w/ version info for the components we care about.

This assumes we control the operand images so we can always put the version information in the same place for the update-payload scraper to find it.  I guess that's a safe assumption.  Spot-checking a recent image, these look like our choices for "the usual place":

```console
$ oc adm release info --pullspecs quay.io/openshift-release-dev/ocp-release:4.0.0-0.3 | grep hyperkube
  hyperkube                                     quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:34475b6e00284fc41a21696b6148a979bbb6de926ab0bef2361b066020e082f7
$ oc image info quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:34475b6e00284fc41a21696b6148a979bbb6de926ab0bef2361b066020e082f7 | grep 'release=\|version='
             release=0.150.0.0
             version=v4.0.0
```

PR #105, opened 2019-01-29
wait for COs versions to be correct

abhinavdahiya:
/cc @smarterclayton

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -1,12 +1,13 @@
 package internal
 
 import (
+	"errors"
 	"fmt"
+	"sort"
+	"strings"
 	"time"
+	"unicode"
 
-	"github.com/golang/glog"
-
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -71,7 +72,7 @@ func (b *clusterOperatorBuilder) Do() error {
 		b.modifier(os)
 	}
 
-	return waitForOperatorStatusToBeDone(b.client, os)
+	return waitForOperatorStatusToBeDone(osPollInternal, osPollTimeout, b.client, os)
 }
 
 const (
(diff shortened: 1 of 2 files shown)


abhinavdahiya:
added unit test with https://github.com/openshift/cluster-version-operator/commit/8e565c4ccb06ff1d70eeaa1f3714246d8f2dc263 -> https://github.com/openshift/cluster-version-operator/commit/afff83fb2ea4d48ded548db6ab693777e7b6380b

pkg/cvo/internal/operatorstatus.go
```diff
@@ -67,27 +68,38 @@ func (b *clusterOperatorBuilder) Do() error {
...
+			return false, nil
 		}
 		glog.V(4).Infof("ClusterOperator %s/%s is reporting %v",
 			eos.Namespace, eos.Name, spew.Sdump(eos.Status))
 
-		// TODO: temporarily disabled while the design for version tracking is finalized
-		// if eos.Status.Version != os.Status.Version {
-		// 	return false, nil
-		// }
+		for _, ve := range os.Status.Versions {
```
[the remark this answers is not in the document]
abhinavdahiya: @smarterclayton pushed changes with detail reports on version, with https://github.com/openshift/cluster-version-operator/commit/afff83fb2ea4d48ded548db6ab693777e7b6380b -> https://github.com/openshift/cluster-version-operator/commit/5b470b8762e2f084e1a410bdcd9750b945cb352a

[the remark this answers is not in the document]
abhinavdahiya: The tests have some examples of how the errors look. PTAL

pkg/cvo/internal/operatorstatus_test.go
```diff
@@ -0,0 +1,217 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: "test-co"},
+			Status: configv1.ClusterOperatorStatus{
+				Versions: []configv1.OperandVersion{{
+					Name: "operator", Version: "v0",
+				}, {
+					Name: "operand-1", Version: "v0",
+				}},
+			},
+		},
+		wantErr: "cluster operator test-co has not completely rolled its operands to desired version: still rolling operand-1 from v0 to v1, still rolling operator from v0 to v1",
```
smarterclayton: Struggling with the wording here.  operands is fairly obscure, and the "rolling" is also potentially the wrong verb

A couple of possible rewordings:

> cluster operator test-co is still updating: operand-1 from 4.0.1 to 4.0.2, operator from 4.0.1 to 4.02

or treat operator special if it exists:

> cluster operator test-co is still updating to 4.0.2: upgrading kubernetes from 1.12.4 to 1.12.5

or just skip the operator completely, assuming that it is already known to be updating:

> cluster operator test-co is still updating: upgrading kubernetes from 1.12.4 to 1.12.5

smarterclayton: or maybe even special case the non operator, single operand case

> cluster operator kube-apiserver is still upgrading kubernetes from 1.12.4 to 1.12.5

abhinavdahiya: @smarterclayton went with this option:
> or just skip the operator completely, assuming that it is already known to be updating:
> 
> > cluster operator kube-apiserver is still updating: upgrading kubernetes from 1.12.4 to 1.12.5

PTAL

abhinavdahiya:
friendly ping @smarterclayton

pkg/cvo/internal/operatorstatus.go
```diff
@@ -113,46 +158,36 @@ func waitForOperatorStatusToBeDone(client configclientv1.ClusterOperatorsGetter,
...
-	if c := resourcemerge.FindOperatorStatusCondition(co.Status.Conditions, configv1.OperatorFailing); c != nil && c.Status == configv1.ConditionTrue {
-		if len(c.Message) > 0 {
-			return &payload.UpdateError{
+		if c := resourcemerge.FindOperatorStatusCondition(actual.Status.Conditions, configv1.OperatorFailing); c != nil && c.Status == configv1.ConditionTrue {
+			message := fmt.Sprintf("Cluster operator %s is reporting a failure", actual.Name)
+			if len(c.Message) > 0 {
+				message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, c.Message)
+			}
+			lastErr = &payload.UpdateError{
+				Nested:  errors.New(strings.ToLower(message)),
```
smarterclayton: We don't want to ToLower the nested message (from c.Message) if capitalization is valid - for instance, if the message said `check the API version` we wouldn't want it to be `check the api version`

pkg/cvo/internal/operatorstatus_test.go
```diff
@@ -0,0 +1,418 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: "test-co"},
+			Status: configv1.ClusterOperatorStatus{
+				Versions: []configv1.OperandVersion{{
+					Name: "operator", Version: "v1",
+				}, {
+					Name: "operand-1", Version: "v1",
+				}},
+			},
+		},
+		expErr: "cluster operator test-co is not done; it is available=false, progressing=true, failing=true",
```
smarterclayton: I think this is too verbose - the previous message was more succinct (next user action can be to view the cluster operator).

smarterclayton:
Thought of a problem with versions - in the event that we have to re-release the same version image (due to accident or error) we can't guarantee the update is rolled out.  That speaks also to the issue we had with deciding which operators to report - that we're attempting to infer the intent of what *must* be upgraded (whether operator or operand) by looking at a side effect value.  While versions are excellent for communicating with humans, they're worse as an interlock between the CVO and the payload.

What if we use versions to summarize (human focused), but the actual value we block on is the payload hash - substituting it into the deployment and expecting the operator to report it back as a field on cluster operator?  I.e. propagate `cluster version.status.versionHash`, and expect the CO to echo it in its status (by adding versionHash).  We control the hash, and the CO simply reports it along with versions.  We advance when the CO reports it in status?

E.g.

1. Write `versionHash` into the deployment in the CVO (based on some signal)
2. Expect the CO to write versionHash to `ClusterOperator.status`
3. Wait until we see versionHash and no failing conditions

?

smarterclayton:
After thinking about it more I can live without the extra concept of versionHash.

We need to define the version that all operators set, whether “operator” or “component” or “*”.  That’s the only one we need to check.  The rest are intended for humans.  Operators must report at least one and cvo checks only that.

pkg/cvo/internal/operatorstatus.go
abhinavdahiya: Ah, Yeah.. will fix it. :innocent:

pkg/cvo/internal/operatorstatus_test.go
abhinavdahiya: So to if the CO conditions do not match, we report just `cluster operator test-co is not done` .. ?

abhinavdahiya: The message reported is `fmt.Sprintf("Cluster operator %s has not yet reported success", actual.Name)` [here](https://github.com/openshift/cluster-version-operator/pull/105/files#diff-e4dff677d52fee4f39f3ba3c602f5659R182) and only the error is more detailed [here](https://github.com/openshift/cluster-version-operator/pull/105/files#diff-e4dff677d52fee4f39f3ba3c602f5659R178)

IMO this should be fine.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -71,35 +72,80 @@ func (b *clusterOperatorBuilder) Do() error {
...
 
-func waitForOperatorStatusToBeDone(client configclientv1.ClusterOperatorsGetter, co *configv1.ClusterOperator) error {
-	var lastCO *configv1.ClusterOperator
-	err := wait.Poll(osPollInternal, osPollTimeout, func() (bool, error) {
-		eos, err := client.ClusterOperators().Get(co.Name, metav1.GetOptions{})
+func waitForOperatorStatusToBeDone(interval, timeout time.Duration, client configclientv1.ClusterOperatorsGetter, expected *configv1.ClusterOperator) error {
+	var lastErr error
+	err := wait.Poll(interval, timeout, func() (bool, error) {
+		actual, err := client.ClusterOperators().Get(expected.Name, metav1.GetOptions{})
 		if err != nil {
```
smarterclayton: Hrm.  I guess this makes sense.

However, this highlighted that we need to be passing Context into this, and we need to abort the poll loop if context is closed.

Also, we need to use `PollImmediate`.

pkg/cvo/internal/operatorstatus_test.go
smarterclayton: Then we should check that in the test case too (table driven tests are great examples to reviewers)

abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/commit/d13c13903c1b92d11240fc628e0b6538b874f29b -> https://github.com/openshift/cluster-version-operator/commit/7098e9e7be4ee0a455c0aee6bc898284d5158bf4

smarterclayton:
/lgtm

This is fine, I'll nitpick later

abhinavdahiya:
Last failure was known router flake

```
Failing tests:

[Conformance][Area:Networking][Feature:Router] The HAProxy router should enable openshift-monitoring to pull metrics [Suite:openshift/conformance/parallel/minimal]
```

/retest

PR #106, opened 2019-02-01
status: Report a percentage when performing the initial sync

smarterclayton:
During both initial cluster init and upgrade it is valuable to show
the user how far we have gotten. Use the fraction reported by status
to show an approximate percentage of our progress through the payload.

During reconcilation, we show nothing.

### DIFF
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -211,9 +211,14 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
 				LastTransitionTime: now,
 			})
 		} else {
-			message := fmt.Sprintf("Working towards %s", version)
-			if len(validationErrs) > 0 {
+			var message string
+			switch {
+			case len(validationErrs) > 0:
 				message = fmt.Sprintf("Reconciling %s: the cluster version is invalid", version)
+			case status.Fraction > 0:
+				message = fmt.Sprintf("Working towards %s: %.0f%% complete", version, status.Fraction*100)
+			default:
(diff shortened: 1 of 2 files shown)


smarterclayton:
@abhinavdahiya

pkg/cvo/status.go
```diff
@@ -211,9 +211,14 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
...
 			})
 		} else {
-			message := fmt.Sprintf("Working towards %s", version)
-			if len(validationErrs) > 0 {
+			var message string
+			switch {
+			case len(validationErrs) > 0:
 				message = fmt.Sprintf("Reconciling %s: the cluster version is invalid", version)
+			case status.Fraction > 0:
+				message = fmt.Sprintf("Working towards %s: %.0f%%", version, status.Fraction*100)
```
abhinavdahiya: just a nit: 
"Working towards %s: %.0f%% complete" or something like that ?
because `Working towards 4.0.1: 45%` seems incomplete.

smarterclayton: Agree, needs something.  Will play with complete and some other variants

smarterclayton: Complete seemed to work fine, we can tweak later if we come up with something better.  Updated.

smarterclayton:
/hold cancel

Had a bad sort condition in the unit test, fixed, passes with -count=1000

smarterclayton:
Fixed in #108

PR #108, opened 2019-02-02
metrics: Fix flaky unit test from transition

smarterclayton:
I wasn't sorting sufficiently to prevent flakes in #107. Tested with -count=1000.


smarterclayton:
Flake merged (I didn't catch hold in time).

Tagging

PR #109, opened 2019-02-03
wip: sync: Perform an initial flood fill of the objects in the payload

smarterclayton:
During install the normal "safe" path for updates works against us by
serializing a bunch of operations that should never fail, or if they do
fail are likely bugs anyway. Exploit the idempotent create behavior of
Kube-like APIs and attempt to create as much of the payload as possible in
a single create pass, halting if we hit any errors. We then perform a
second pass and wait wait for operators to report status.

The most common reason to halt the initial pass is when the kube-apiservers
are restarted during configuration, or when we reach a payload object that
depends on an aggregated API which is unlikely to have succeeded.

There is a tiny amount of unsafety around this optimization. One,
deployments that depend on earlier deployments may fail or go into backoff
because those prerequisites have not healed, which might actually
(unlikely, but possible) delay startup more. Two, this puts more load on
the initial startup and may peg CPU or cause excessive memory use - however
that's more likely to benefit us than hurt. Finally, when updates are
delivered that create objects earlier in the payload (a new operator is
added at runlevel 0000_04_X, before any other operators) AND the previously
first operator adds a new object that depends on that new first operator,
the new object may fail because the first operator hasn't yet completed. In
that case, tests should catch it, and we should fix the operator if
necessary.


smarterclayton:
Got a failure because grafana wedges because it tries to read well-known-authorization too early, can't get it (because openshift apiserver isn't up), and then never retries.  The grafana-proxy then just crash loops forever and CMO doesn't seem to detect it.  That's a real bug, but does highlight how this could break ordering assumptions.  The CMO is currently one of the slower operators to start, so we definitely benefit from getting a start on it.  We could also add an annotation that defeats the flood fill optimization which allows us to stop right before CMO (selective patching).  Need to think about this some more.

PR #110, opened 2019-02-04
sync: Report a specific cluster operator error when not available

smarterclayton:
Cluster operators block most of upgrade, so we should report the information the cluster operator has available back to the CV status object via an UpdateError.

Reduce the amount of debug info printed by cluster operator status waiting to simplify the logs.

While reviewing CV logs from startup, we run the resync loop more frequently waiting for the CO to be created because we exit the poll loop on any error. When we see "CO not found", continue the loop to avoid triggering the top level sync worker backoff loop.  Shaves a few tens of seconds of startup time.

Finally, remove the apiVersion info from the status log message, it didn't make much sense and the end user is going to have to look at the payload anyway if we get into a wierd error case where it matters (the apiVersion was put in in case we had to debug a typo in a version, which really shouldn't happen).

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -1,11 +1,12 @@
 package internal
 
 import (
+	"fmt"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
-
 	"github.com/golang/glog"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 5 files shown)


smarterclayton:
Flake is #111

pkg/cvo/internal/operatorstatus.go
```diff
@@ -102,14 +109,40 @@ func waitForOperatorStatusToBeDone(client configclientv1.ClusterOperatorsGetter,
...
+			return &payload.UpdateError{
+				Reason:  "ClusterOperatorFailing",
+				Message: fmt.Sprintf("Cluster operator %s is reporting a failure", lastCO.Name),
+				Name:    lastCO.Name,
+			}
+		}
+	}
+	return &payload.UpdateError{
+		Reason:  "ClusterOperatorNotAvailable",
+		Message: fmt.Sprintf("Cluster operator %s has not yet reported success", lastCO.Name),
```
abhinavdahiya: lastCO can be nil here...

smarterclayton:
Trivial rebase, reapplying label

PR #112, opened 2019-02-04
[POC] refactor CVO to check for CO status last

deads2k:
This updates the CVO to
1.  avoid knowledge of special API semantics and controllers for particular types (no more waiting on resource)
2. always retry resources (at least I think that's what removing the `shouldRetry` method does)
3. move the CO.status check last

This effectively means the the CVO does nothing but create all the resource in order as quickly as it can.  It may fail and retry to create a particular resource, but it doesn't understand how the controllers manage various conditions on them.  This means that we will avoid unnecessary waiting.

The CVO still looks at clusteroperator status before deciding that it's complete, but that is contained and only done after the operators have all had a chance to run.  This avoids dependency management that will not happen at runtime on a cluster as components go up and down.

@smarterclayton @abhinavdahiya @derekwaynecarr @bparees


bparees:
I assume we'll make sure this passes e2e-aws 5 times before merging it :)

smarterclayton:
/hold

I don't know that we need to do this, and I don't see a benefit in 4.0.

smarterclayton:
There are several bugs in operators right now that prevent them from doing this - I'm fine with using this to chase them down and then reassessing later.

smarterclayton:
Note I do like the simplicity of this, I just don't know that I'm ready to
take the risk.  If we can get a week's worth of soak in this PR fixing bugs
in operators that have ordering (grafana currently gets wedged waiting for
apiserver, a few others race and flake) and we're green and all the teams
are ready, I'd be willing to consider it.

abhinavdahiya:
> This updates the CVO to
> 
>     1. avoid knowledge of special API semantics and controllers for particular types (no more waiting on resource)

why is this bad? also we do `other` controllers care?
for example, 
- not waiting for CRD to be accepted is going to create errors when we try to immediately create CR... so why waste...
- deployments rollout is useful as it ensures that no old version of operator is running...

>     2. always retry resources (at least I think that's what removing the `shouldRetry` method does)

sure, #88 removes this too.

>     3. move the CO.status check last

So you want to make CVO move from using COs as means of `controlling the dependencies` to only using COs as only `update done indicators`

Are the operators ready to resolve dependencies like an operator that requires specific apiserver version etc... ?

> 
> This effectively means the the CVO does nothing but create all the resource in order as quickly as it can. It may fail and retry to create a particular resource, but it doesn't understand how the controllers manage various conditions on them. This means that we will avoid unnecessary waiting.

`unnecessary` ?? It is only unnecessary when all operators do internal dependency resolution on their requirements.

we have currently failed to do the internal resolutions in several operators... openshift-apiserver etc..

> The CVO still looks at clusteroperator status before deciding that it's complete, but that is contained and only done after the operators have all had a chance to run. This avoids dependency management that will not happen at runtime on a cluster as components go up and down.
> 
> @smarterclayton @abhinavdahiya @derekwaynecarr @bparees

deads2k:
> `unnecessary` ?? It is only unnecessary when all operators do internal dependency resolution on their requirements.

If they don't, they are flawed.  While running there is no guarantee of any order or any preconditions "guaranteed" by fencing in the CVO.  It's promoting bad habits that do not conform to steady-state clusters.



> > ```
> > 
> > ```
> 
> why is this bad? also we do `other` controllers care?
> for example,
> 
> * not waiting for CRD to be accepted is going to create errors when we try to immediately create CR... so why waste...
> * deployments rollout is useful as it ensures that no old version of operator is running...

The CVO doesn't logically care about those conditions and when you look at what we're considering for things like CRDs (adding a new state), you're just asking for more rippling pain here.  The end goal for the CVO is successful application of all manifests and settling of the clusteroperator state.  How we do that, what status is needed in between, and what levels of an operator are running are all extraneous to that goal.



> So you want to make CVO move from using COs as means of `controlling the dependencies` to only using COs as only `update done indicators`
> 
> Are the operators ready to resolve dependencies like an operator that requires specific apiserver version etc... ?
> 
> >

This statement here is worrying.  As I noted above, the "guarantee" from the CVO is false.  It's not actually controlling dependencies.  It makes people think that the CVO does, but the CVO does nothing to preserve or maintain those guarantees in a running cluster.  At the most basic level, clusteroperators can become unavailable any time for any reason.  The CVO pretending that it enforced an ordering guarantee makes people think the issue is solved, but it isn't.  We need to avoid this language.  Operators must correctly handle and resolve their dependencies.

deads2k:
wow, it mostly installed.  That's awesome.  Looks like the MCO had trouble with masters.

/retest

deads2k:
> Right now, partially or fully serializing the payload is a mitigation I can live with.

That's just pushing it to a field problem.  Serializing the payload isn't *actually* making sure dependencies are met.  Only that they seem to be once.

PR #117, opened 2019-02-08
hack: change override env for version

abhinavdahiya:
Similar to https://github.com/openshift/machine-config-operator/pull/191.
Turns out `registry.svc.ci.openshift.org/openshift/release:golang-1.10` has `VERSION` env set to golang's version.
So switching to `VERSION_OVERRIDE` prevents conflicts.

/cc @crawford


crawford:
/lgtm

Prometheus was crash looping 
/retest

PR #118, opened 2019-02-10
operators: Retry connection refused errors while waiting for CO

smarterclayton:
During install and update the kube-apiserver may be updated one or
more times, and since the CVO is configured to talk to localhost the
server may be unavailable. If we see connection refused, treat it
as retriable at this level vs propagating it up, since we can't
talk to another server.

A more long term fix is for the rest client to accept multiple
possible servers and try all of them, which requires significant
changes to client-go.

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -2,6 +2,10 @@ package internal
 
 import (
 	"fmt"
+	"net"
+	"net/url"
+	"os"
+	"syscall"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


smarterclayton:
@abhinavdahiya this can come after #105 if you want - it prevents us from seeing errors like `cannot update clusteroperator "monitor"` instead of our new, improved `waiting for cluster operator "monitor"` errors.

abhinavdahiya:
> @abhinavdahiya this can come after #105 if you want - it prevents us from seeing errors like `cannot update clusteroperator "monitor"` instead of our new, improved `waiting for cluster operator "monitor"` errors.

https://github.com/openshift/cluster-version-operator/pull/105/files#diff-e4dff677d52fee4f39f3ba3c602f5659R87 makes all errors retryable... so not sure if this will be needed after #105

PR #120, opened 2019-02-15
Claim runlevel 29 for cloud-credentials-operator.

dgoodwin:
Needs to be running before the machine-api so we can mint credentials it
can then use to bring up the compute nodes.


smarterclayton:
The question in the PR is whether machine config should be properly waiting for creds vs having to run first

dgoodwin:
This PR was clearly forgotten. We've been running at 50 for some time now.

/close

PR #125, opened 2019-03-01
clusteroperator: Skip waiting for progressing when versions expected

smarterclayton:
As operators begin waiting for versions, we no longer need to wait for progressing status before continuing to the payload as long as we aren't failing or available. For operators not yet setting versions, continue to wait for progressing.

Allowing this will allow the core operators to complete faster even if they are restarting to process config changes.  This may cause other operators to be flaky, but that's ok because they need to tolerate those anyway, or the kube-apiserver needs to be better at mitigating, or likely, both.

@derekwaynecarr @abhinavdahiya @bparees as per discussion today as we were switching image-registry operator over.  There's no reason to hold for progressing once an operator has told us it's available at the version we requested.  Will allow kube-apiserver to declare done earlier.

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -155,8 +155,10 @@ func waitForOperatorStatusToBeDone(interval, timeout time.Duration, client confi
 				failing = false
 			}
 		}
-		// if we're at the correct version, and available, not progressing, and not failing, we are done
-		if available && !progressing && !failing {
+		// if we're at the correct version, and available, and not failing, we are done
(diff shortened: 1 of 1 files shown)


bparees:
i'm wary that we may have operators today that are reporting their version before they are really done/ready.  Are we going to hold this until we've audited for that?

smarterclayton:
I reviewed the operators who already request to wait and the only ones that do are the ones Derek and I are changing.  So it can't cause flakes here.

smarterclayton:
Any other opinions?

bparees:
it's ok with me.

smarterclayton:
Tagging

PR #128, opened 2019-03-05
lib/resourcebuilder: update interface to include context when doing work

abhinavdahiya:
all resourcebuilders must respect the context when doing work. This allows slow resource builders to
cancel their task mid-way.

/cc @smarterclayton

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -1,6 +1,7 @@
 package cvo
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 	"sync"
@@ -464,7 +465,7 @@ type resourceBuilder struct {
 }
 
 // NewResourceBuilder creates the default resource builder implementation.
-func NewResourceBuilder(config *rest.Config) ResourceBuilder {
+func NewResourceBuilder(config *rest.Config) payload.ResourceBuilder {
 	return &resourceBuilder{config: config}
 }
 
@@ -479,7 +480,7 @@ func (b *resourceBuilder) BuilderFor(m *lib.Manifest) (resourcebuilder.Interface
 	return internal.NewGenericBuilder(client, *m)
 }
 
-func (b *resourceBuilder) Apply(m *lib.Manifest) error {
+func (b *resourceBuilder) Apply(ctx context.Context, m *lib.Manifest) error {
 	builder, err := b.BuilderFor(m)
 	if err != nil {
 		return err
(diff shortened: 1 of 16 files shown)


pkg/cvo/cvo.go
```diff
@@ -479,13 +480,16 @@ func (b *resourceBuilder) BuilderFor(m *lib.Manifest) (resourcebuilder.Interface
...
+func (b *resourceBuilder) Apply(ctx context.Context, m *lib.Manifest) error {
 	builder, err := b.BuilderFor(m)
 	if err != nil {
 		return err
 	}
 	if b.modifier != nil {
 		builder = builder.WithModifier(b.modifier)
 	}
+	if builder, ok := builder.(resourcebuilder.CancelableInterface); ok {
+		return builder.DoWithContext(ctx)
```
smarterclayton: Any reason not to pass Context to Do?  That would be more consistent with context style.

abhinavdahiya:
```console
$ rg 'Do\(\)' lib
lib/resourcebuilder/apiext.go
37:func (b *crdBuilder) Do() error {

lib/resourcebuilder/apps.go
38:func (b *deploymentBuilder) Do() error {
102:func (b *daemonsetBuilder) Do() error {

lib/resourcebuilder/apireg.go
29:func (b *apiServiceBuilder) Do() error {

lib/resourcebuilder/rbac.go
29:func (b *clusterRoleBuilder) Do() error {
56:func (b *clusterRoleBindingBuilder) Do() error {
83:func (b *roleBuilder) Do() error {
110:func (b *roleBindingBuilder) Do() error {

lib/resourcebuilder/core.go
29:func (b *serviceAccountBuilder) Do() error {
56:func (b *configMapBuilder) Do() error {
83:func (b *namespaceBuilder) Do() error {
110:func (b *serviceBuilder) Do() error {

lib/resourcebuilder/security.go
29:func (b *securityBuilder) Do() error {

lib/resourcebuilder/interface.go
68:     Do() error

lib/resourcebuilder/batch.go
37:func (b *jobBuilder) Do() error {
```
I was hoping to not change all the builders to include ctx with Do. But i can change them too if you think that's a better api.

abhinavdahiya: Also most of the builders (except crd, daemonset, deployment) do not gain much from context.

smarterclayton: I mean, in general the rule for context is that you always pass it as the first argument of a function unless you have a really good reason not to.  So just for fitting in with context convention I'd say it's more normal to do that.  Agree most don't have a use, but that's also true in other context use cases.

abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/compare/5b8906b10087634b5b73fe42380ebac347ac6931..37fc9b62930a441e386f0bfbe1451109c3611a3e

abhinavdahiya:
ping @smarterclayton

PR #131, opened 2019-03-08
WIP: sync: Distinguish between reconcile, initial, and sync states

smarterclayton:
Identify what sort of change we are making to the cluster in order
to better handle the different states.

TODO:

* [ ] Sync differently depending on state

PR #132, opened 2019-03-09
sync: Serve cluster operator gets from the cache lister

smarterclayton:
Reduces the amount of API calls we make significantly when polling,
which reduces load on API server and also doesn't consume as much of
our client rate limit.

This does not impact correctness because we're level driven for
continuity and a subsequent sync loop will already handle this.

Extracted from #131


smarterclayton:
suspicious flake - saw this in the other Pr as well

/retest

PR #133, opened 2019-03-09
sync: Distinguish between reconcile, initial, and sync states

smarterclayton:
Identify what sort of change we are making to the cluster in order
to better handle the different states. Does not change status reporting
or the internal sync which will be in a follow up commit.

The three states are:

* Initializing - we haven't yet completed a payload
* Updating - we're moving between two payloads of different versions
* Reconciling - we're level on a completed payload

The states will allow us to more efficiently reconcile, start faster,
and be safer during upgrades.

Extracted from #131

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -493,7 +493,7 @@ func (b *resourceBuilder) BuilderFor(m *lib.Manifest) (resourcebuilder.Interface
 	return internal.NewGenericBuilder(client, *m)
 }
 
-func (b *resourceBuilder) Apply(ctx context.Context, m *lib.Manifest) error {
+func (b *resourceBuilder) Apply(ctx context.Context, m *lib.Manifest, state payload.State) error {
 	builder, err := b.BuilderFor(m)
 	if err != nil {
 		return err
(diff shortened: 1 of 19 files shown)


smarterclayton:
Made the resource builder aware of the mode without actually changing any of our logic.

pkg/cvo/cvo.go
```diff
@@ -495,3 +503,23 @@ func (b *resourceBuilder) Apply(ctx context.Context, m *lib.Manifest) error {
 	}
 	return builder.Do(ctx)
 }
+
+func hasNeverReachedLevel(cv *configv1.ClusterVersion) bool {
```
abhinavdahiya: This is a stretch :P (sorry my brain just went there)
but, what if somebody overrides the desired more than 10 times, so there will be all partial and we will truncate to all partial because https://github.com/openshift/cluster-version-operator/blob/2a78a6cedd73f5f0bbfbeb8778e23d66490d7ea7/pkg/cvo/status.go#L74-L76

smarterclayton: Hrm, we should leave the last completed.  We should never prune to no completed.  Good catch, will change

PR #138, opened 2019-03-15
Bump CRD manifest support to 1.12 + nullable support

sttts:
The CVO uses typed clients for manifest creation. This means that it only sees (and creates) fields it known about (up to Kube 1.11 before this PR). For OpenShift resources we need nullable support as implemented in origin in https://github.com/openshift/origin/pull/22202.

This PR:
- bumps kube dependencies to 1.12
- switches k8s.io/apiextensions-apiserver to use the origin fork with https://github.com/openshift/origin/pull/22202 included.


Gopkg.toml
```diff
@@ -21,35 +21,37 @@ required = [ "k8s.io/code-generator/cmd/client-gen", "k8s.io/gengo/types" ]
...
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "8.0.0"
+  version = "9.0.0"
 
 [[constraint]]
   name = "k8s.io/apiextensions-apiserver"
-  version = "kubernetes-1.11.1"
+  # use origin fork with nullable support. We can switch back to the upstream version in 1.14.
```
deads2k: @sttts you need to sign your name and sig-master to this 

`# sttts, sig-master`

so they can chase us down if they need to.  I've been forcing other teams to do the same when mess about in our pond.

smarterclayton:
What's the leader election failure?  Oh, we have to step down.

Gopkg.toml
sttts: fixed

pkg/start/start.go
```diff
@@ -153,13 +153,14 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 
 	// TODO: when we switch to graceful lock shutdown, this can be
 	// moved back inside RunOrDie
-	go leaderelection.RunOrDie(leaderelection.LeaderElectionConfig{
+	// TODO: properly wire ctx here
+	go leaderelection.RunOrDie(context.TODO(), leaderelection.LeaderElectionConfig{
```
smarterclayton: Was this not working when you passed ctx to RunOrDie?

sttts: we have some dirty exit channel logic. Prefered to keep that "rewrite" out of the compile fixes.

smarterclayton:
/lgtm

I'll take the TODO to use the new context code.

PR #141, opened 2019-03-25
Bug 1691513: CVO should retry initialization until cancelled to avoid wedging because of failing dependencies

smarterclayton:
The sync loop currently mixes a combination of:

1. Try forever (top level sync loop) 
2. Cancel sync when a timeout is reached 
3. Retry individual Apply actions in the sync workers up to a max
retry 
4. Retry within Apply actions

This led to bugs in initialization where transient errors cause a prereq
step to fail and never complete, leading to other operators never going
live.

This commit simplifies the structure of our retry logic to be:

1. Try to sync forever 
2. Cancel sync after a certain period of failure and go into backoff 
3. Retry individual Applies forever with capped backoff 
4. Retry applies as long as context is not cancelled

This implies all top level sync has a context with a deadline which is the
case today - some test cases are changed for it.

This should also reduce the lag between cancel (on user action, for 
instance) and the sync loop terminating, since everything is context gated.

Also adds a Makefile because I like consistency.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1691513

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+build:
+	hack/build-go.sh
+.PHONY: build
+
+test:
+	go test ./...
+.PHONY: test
--- a/lib/resourcebuilder/apiext.go
+++ b/lib/resourcebuilder/apiext.go
@@ -2,7 +2,6 @@ package resourcebuilder
 
 import (
 	"context"
-	"time"
 
 	"github.com/golang/glog"
 
@@ -49,15 +48,13 @@ func (b *crdBuilder) Do(ctx context.Context) error {
 		return err
 	}
 	if updated {
-		ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
-		defer cancel()
-		return waitForCustomResourceDefinitionCompletion(ctxWithTimeout, 1*time.Second, b.client, crd)
+		return waitForCustomResourceDefinitionCompletion(ctx, b.client, crd)
 	}
 	return nil
 }
 
-func waitForCustomResourceDefinitionCompletion(ctx context.Context, interval time.Duration, client apiextclientv1beta1.CustomResourceDefinitionsGetter, crd *apiextv1beta1.CustomResourceDefinition) error {
-	return wait.PollImmediateUntil(interval, func() (bool, error) {
+func waitForCustomResourceDefinitionCompletion(ctx context.Context, client apiextclientv1beta1.CustomResourceDefinitionsGetter, crd *apiextv1beta1.CustomResourceDefinition) error {
+	return wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
 		c, err := client.CustomResourceDefinitions().Get(crd.Name, metav1.GetOptions{})
 		if errors.IsNotFound(err) {
 			// exit early to recreate the crd.
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -3,7 +3,6 @@ package resourcebuilder
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/golang/glog"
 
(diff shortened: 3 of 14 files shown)


smarterclayton:
@abhinavdahiya with this the sync loop is context aware all the way down and retries until context is cancelled. Retry and bail didn't have much value for us anyway except for status updates and we *should* gather those on cancel (I need to verify that and we can make apply report partial status if necessary).

wking:
[unit][1]:

```
E0325 19:56:04.262847    4132 task.go:68] error running apply for test "file-yml" (3 of 3): unable to proceed
SIGQUIT: quit
```

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/141/pull-ci-openshift-cluster-version-operator-master-unit/355/build-log.txt

abhinavdahiya:
9221dec `Also adds a Makefile because I like consistency.` :( but ok!
1212873 LGTM

> @abhinavdahiya with this the sync loop is context aware all the way down and retries until context is cancelled. Retry and bail didn't have much value for us anyway except for status updates and we _should_ gather those on cancel

The context we have is the global one (leader election ctx), I didn't see any deadlines.
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/cvo.go#L219
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/sync_worker.go#L232
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/sync_worker.go#L242
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/sync_worker.go#L402
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/sync_worker.go#L455
https://github.com/openshift/cluster-version-operator/blob/3d7d6bc08fbe4062ebf222ceffe2f93c6d149830/pkg/cvo/sync_worker.go#L471

>  (I need to verify that and we can make apply report partial status if necessary).

Yeah, we will only see status updates on success or global cancel...

smarterclayton:
> The context we have is the global one (leader election ctx), I didn't see any deadlines.

There should have been one, I'm going to add that.  Was meaningless while we looped, but is meaningful now.

> Also adds a Makefile because I like consistency. :( but ok!

Every other repo has a simple one, it helps orient and define some minimum carry over between them.  I don't expect it to grow or be heavy, but it saves time for folks working everywhere.

smarterclayton:
> Yeah, we will only see status updates on success or global cancel...

We'll see a progress percentage for every group that increases.  It's only crash loops that will be delayed.  I'll tune the sync loop timeout based on that progress upgrade distinct from sync distinct from initializing.

smarterclayton:
For updates we may want to keep going until we see repeated errors, and then report those up.  One thing I think we need is summarizing the errors from multiple nodes in the graph into a stable output message that isn't going back and forth.  Reconcile could also accumulate all possible errors that occur during a pass without retries and then summarize any failures in the progressing message, potentially moving to failing if two successive syncs have any errors.

smarterclayton:
We don't need to exit sync in order to report status, but we should keep in mind that we want to guarantee two things during reconcile:

1. we eventually try every objects
2. we want to guarantee an upper bound on the interval between any two reconcile attempts of an object

abhinavdahiya:
https://github.com/openshift/cluster-version-operator/pull/141/commits/838794aa10bc00685006f4f7dcc66e0194f015b0 LGTM

/retest

wking:
[e2e-aws][1]:

```
Failing tests:

[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Dynamic PV (default fs)] subPath should be able to unmount after the subpath directory is deleted [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Inline-volume (default fs)] subPath should be able to unmount after the subpath directory is deleted [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Pre-provisioned PV (default fs)] subPath should be able to unmount after the subpath directory is deleted [Suite:openshift/conformance/parallel] [Suite:k8s]
```

/retest

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/141/pull-ci-openshift-cluster-version-operator-master-e2e-aws/502

wking:
[e2e-aws][1] hit openshift/origin#22412 again.

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/141/pull-ci-openshift-cluster-version-operator-master-e2e-aws/510

wking:
openshift/origin#22412 landed.

/retest

wking:
:tada:

Issue #142, opened 2019-03-26
Able to trigger a leader election panic, use of closed channel


smarterclayton:
TEST_INTEGRATION=1 go test ./pkg/start/ -tags integration -count=4

```
E0325 22:45:48.713617   10713 sync_worker.go:276] unable to synchronize image (waiting 625ms): Could not update configmap "e2e-cvo-ff4l/config2" (2 of 2): the object is invalid, possibly due to local cluster configuration
E0325 22:45:48.922205   10713 leaderelection.go:256] error initially creating leader election record: namespaces "e2e-cvo-mlm6zv" not found
E0325 22:45:54.301108   10713 event.go:259] Could not construct reference to: '&v1.ConfigMap{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"", GenerateName:"", Namespace:"", SelfLink:"", UID:"", ResourceVersion:"", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:""}, Data:map[string]string(nil), BinaryData:map[string][]uint8(nil)}' due to: 'selfLink was empty, can't make reference'. Will not report event: 'Normal' 'LeaderElection' 'claytonc-mbp.local_4bb2f5bb-5f1d-4543-90dd-75de07986a26 stopped leading'
panic: close of closed channel

goroutine 63 [running]:
github.com/openshift/cluster-version-operator/pkg/start.(*Options).run.func3()
	/Users/clayton/projects/origin/src/github.com/openshift/cluster-version-operator/pkg/start/start.go:190 +0x74
github.com/openshift/cluster-version-operator/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run.func1(0xc0006e4240)
	/Users/clayton/projects/origin/src/github.com/openshift/cluster-version-operator/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:148 +0x40
[...]

PR #144, opened 2019-03-26
leaderelect: Prevent double close channel panic when shutting down lease

smarterclayton:
While running integration tests in loops I was able to trigger a
double close panic. Leader election doesn't guarantee exclusion
in the handlers, and until we have step down from 1.13 client-go
we need to guard.


smarterclayton:
Fixes #142

PR #146, opened 2019-03-26
status: Correctly increment history and don't overwrite controller cache objects

smarterclayton:
When we refactored to the newer status style syncFailingStatus was
left updating the cached map, which can cause corruption of the
in memory state.

Also, we shouldn't add an entry to the history when images are equal but the new version is empty because we can get into that state if the CVO is rescheduled onto another node during an upgrade - while it's loading the payload the version will be empty.  We only add an entry if the version is set and has changed (indicating possibly that the tagged image changed).

### DIFF
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -278,7 +278,7 @@ func TestOperator_sync(t *testing.T) {
 					Status: configv1.ClusterVersionStatus{
 						History: []configv1.UpdateHistory{
 							{State: configv1.PartialUpdate, Version: "0.0.1-abc", Image: "image/image:v4.0.1", StartedTime: defaultStartedTime},
-							{Version: "4.0.1", Image: "image/image:v4.0.1", StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
+							{State: configv1.PartialUpdate, Version: "4.0.1", Image: "image/image:v4.0.1", StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
 						},
 						Desired:     configv1.Update{Version: "0.0.1-abc", Image: "image/image:v4.0.1"},
 						VersionHash: "",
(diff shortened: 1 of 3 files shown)


smarterclayton:
The tests failed but not because of the panic.  Going to retest, but @abhinavdahiya I think this fixes the partial history overwrite bug and is clearly incorrect.  If the next tests pass I'm going to drop the panic commit and repush.  PTAL

pkg/cvo/cvo_test.go
```diff
@@ -13,7 +13,6 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang/glog"
 	"github.com/google/uuid"
-
```
abhinavdahiya: nit: why this drop?

pkg/cvo/status.go
```diff
@@ -274,24 +274,24 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
 // from the cache (instead of clearing the status).
 // if ierr is nil, return nil
 // if ierr is not nil, update OperatorStatus as Failing and return ierr
-func (optr *Operator) syncFailingStatus(config *configv1.ClusterVersion, ierr error) error {
+func (optr *Operator) syncFailingStatus(original *configv1.ClusterVersion, ierr error) error {
```
abhinavdahiya: this change just renaming the variable... ?

abhinavdahiya: Ah, I see https://github.com/openshift/cluster-version-operator/pull/146/commits/1da7d0754c6797878db0edde23ec0f24bb299951#diff-6aba50cdad3ec2c1202f7ba49ed0314bR294

abhinavdahiya:
1da7d0754c6797878db0edde23ec0f24bb299951 LGTM.

pkg/cvo/cvo_test.go
smarterclayton: it's vscode go tools being stupid, sorry

abhinavdahiya:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/146/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/12/artifacts/e2e-aws-upgrade/clusterversion.json
```json
[
  {
    "completionTime": null,
    "image": "registry.svc.ci.openshift.org/ci-op-5897hskp/release@sha256:2b45dd851c11863a5a8e13918b2acc55d5be39a33f3f0000c4dcc300d6c5974e",
    "startedTime": "2019-03-26T21:00:29Z",
    "state": "Partial",
    "version": ""
  },
  {
    "completionTime": "2019-03-26T21:00:29Z",
    "image": "registry.svc.ci.openshift.org/ci-op-5897hskp/release@sha256:2b45dd851c11863a5a8e13918b2acc55d5be39a33f3f0000c4dcc300d6c5974e",
    "startedTime": "2019-03-26T20:26:26Z",
    "state": "Partial",
    "version": "0.0.1-2019-03-26-195653"
  },
  {
    "completionTime": "2019-03-26T20:26:26Z",
    "image": "registry.svc.ci.openshift.org/ci-op-5897hskp/release@sha256:6a209a806f0172d5cdf8cbd5c7966ac3f2360067794f77b7e612d8bec711a06d",
    "startedTime": "2019-03-26T20:10:53Z",
    "state": "Completed",
    "version": "0.0.1-2019-03-26-195555"
  }
]
```
that seems like the reason why we are seeing multiple `Partial`.

smarterclayton:
Updated with the fix (empty version shouldn't trigger new entry).

PR #147, opened 2019-03-27
Remove hardcoded 'amd64' from Dockerfile to support multi-arch builds.

jeffdyoung:
Remove hardcoded 'amd64' from Dockerfile to support multi-arch builds.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,10 +1,12 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-version-operator
 COPY . .
-RUN hack/build-go.sh
+RUN hack/build-go.sh; \
+    mkdir -p /tmp/build; \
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel7
```diff
@@ -1,10 +1,12 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-version-operator
 COPY . .
-RUN hack/build-go.sh
+RUN hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/cluster-version-operator/_output/linux/$(go env GOARCH)/cluster-version-operator /tmp/build/cluster-version-operator
```
abhinavdahiya: nit:
https://github.com/openshift/cluster-version-operator/pull/147/files#diff-b0556bed298b3a4fb269eb10d69c91a5R2
already sets the work directory to `/go/src/github.com/openshift/cluster-version-operator` no need to write again..

jeffdyoung: @abhinavdahiya fixed! :)

PR #150, opened 2019-03-29
Ensure channel isn't empty on update check

steveeJ:
Checking for an empty channel locally prevents unnecessary calls to the
upstream update server.

Partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1679901.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -118,6 +118,13 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
 		}
 	}
 
+	if len(channel) == 0 {
+		return nil, configv1.ClusterOperatorStatusCondition{
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoChannel",
+			Message: "The update channel has not been configured.",
+		}
+	}
+
 	currentVersion, err := semver.Parse(version)
 	if err != nil {
 		glog.V(2).Infof("Unable to parse current semantic version %q: %v", version, err)
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -2052,6 +2052,45 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 			},
 		},
 		{
+			name: "report an error condition when channel isn't set",
+			handler: func(w http.ResponseWriter, req *http.Request) {
+				http.Error(w, "bad things", http.StatusInternalServerError)
+			},
+			optr: Operator{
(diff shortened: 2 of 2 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -118,6 +118,13 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
 		}
 	}
 
+	if len(channel) == 0 {
+		return nil, configv1.ClusterOperatorStatusCondition{
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoChannel",
+			Message: "The cluster version does not have a channel assigned and cannot calculate valid upgrades.",
```
lucab: I think this should be "The cluster does not" or "The cluster version object does not", as the current form is a bit strange to read.

pkg/cvo/cvo_test.go
```diff
@@ -2051,6 +2051,45 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "report an error condition when channel jis set",
```
lucab: typo and missing negation: "when no channel is set".

steveeJ:
Thanks for catching the message typos @lucab :+1:

steveeJ:
@abhinavdahiya can I bother you for a review here? Please :-)

smarterclayton:
I'm confused.  Why does the CVO care about a channel?

smarterclayton:
I would expect if Cincinnati wants to reject this it should return a wellformed error and we should display that error, instead of the CVO doing it.

steveeJ:
> I'm confused. Why does the CVO care about a channel?

I agree that it shouldn't care about a specific channel being set and only ensure that it's set at all. This is unrelated to any checks that Cincinnati would do.

pkg/cvo/availableupdates.go
```diff
@@ -118,6 +118,13 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
 		}
 	}
 
+	if len(channel) == 0 {
+		return nil, configv1.ClusterOperatorStatusCondition{
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoChannel",
+			Message: "The update channel has not been assigned.",
```
wking: nit: "assigned" -> "configured"?  To make it more clear that this is up to the admin to fix this.

steveeJ:
ping @smarterclayton @wking 

I'm still proposing this change despite what does or does not happen on Cincinnati

PR #153, opened 2019-04-02
SetOperatorStatusCondition update transition time for existing

michaelgugino:
This commit ensures whenever there are differences between
existing and new status, we update the transition time with
current time.

### DIFF
--- a/lib/resourcemerge/os.go
+++ b/lib/resourcemerge/os.go
@@ -51,7 +51,7 @@ func SetOperatorStatusCondition(conditions *[]configv1.ClusterOperatorStatusCond
 
 	if existingCondition.Status != newCondition.Status {
 		existingCondition.Status = newCondition.Status
-		existingCondition.LastTransitionTime = newCondition.LastTransitionTime
+		existingCondition.LastTransitionTime = metav1.NewTime(time.Now())
 	}
 
 	existingCondition.Reason = newCondition.Reason


lib/resourcemerge/os.go
```diff
@@ -51,7 +51,7 @@ func SetOperatorStatusCondition(conditions *[]configv1.ClusterOperatorStatusCond
 
 	if existingCondition.Status != newCondition.Status {
 		existingCondition.Status = newCondition.Status
-		existingCondition.LastTransitionTime = newCondition.LastTransitionTime
+		existingCondition.LastTransitionTime = metav1.NewTime(time.Now())
```
runcom: `metav1.Now()`?

michaelgugino: https://github.com/openshift/cluster-version-operator/pull/153/files#diff-7b07a5cf0629a7b019ac699efb7cfe94R47

This is the same method used to generate time elsewhere in this file.

runcom: yeah, just pointing out that there's a direct one, just a nit

abhinavdahiya:
Do you mind explaining why using the last transition time from new condition is not correct. The user must have set the time correctly already right?

michaelgugino:
> Do you mind explaining why using the last transition time from new condition is not correct. The user must have set the time correctly already right?

@abhinavdahiya 
No, we're not setting the last update time.  I suspect nobody else is either.  I don't see much use for this method if it's not going to set the transition time on transitions.

PR #158, opened 2019-04-09
status: Report the operators that have not yet deployed

smarterclayton:
The current single error return strategy from the CVO sync loop predates
parallel payload execution and limited retries. Instead, collect all
errors from the task execution graph and attempt to synthesize better
messages that describe what is actually happening.

1. Filter out cancellation error messages - they aren't useful and are
   a normal part of execution
2. When multiple errors are reported, display a reasonable multi-line
   error that summarizes any blockers
3. Treat ClusterOperatorNotAvailable as a special case - if all errors
   reported are of that type convert it to ClusterOperatorsNotAvailable
   and synthesize a better message
4. In the sync loop, if we are still making progress towards the update
   goal and we haven't waited too long for an update, and if the error
   is the specific cluster operator not available types, display the
   condition Progressing=True instead of Failing=true with a synthetic
   message.

This also passes along the task with the UpdateError so that we can do
more selective error messages for specific error cases.

Needs a few more tests, but for Progressing reports

```
Working towards 1.0.0-abc: 33% complete, waiting on operator-1, operator-2
```

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -238,9 +238,10 @@ func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
 				var syncTimeout time.Duration
 				switch work.State {
 				case payload.InitializingPayload:
-					// during initialization we expect things to fail due to ordering
-					// dependencies, so give it extra time
-					syncTimeout = w.minimumReconcileInterval * 5
+					// during initialization we want to show what operators are being
+					// created, so time out syncs more often to show a snapshot of progress
+					// TODO: allow status outside of sync
+					syncTimeout = w.minimumReconcileInterval
 				case payload.UpdatingPayload:
 					// during updates we want to flag failures on any resources that -
 					// for cluster operators that are not reporting failing the error
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -12,6 +14,7 @@ import (
 	"golang.org/x/time/rate"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 
@@ -66,6 +69,8 @@ type SyncWorkerStatus struct {
 	Reconciling bool
 	Initial     bool
 	VersionHash string
+
+	LastProgress time.Time
 
 	Actual configv1.Update
 }
@@ -304,6 +309,9 @@ func (w *statusWrapper) Report(status SyncWorkerStatus) {
 			}
 		}
 	}
+	if status.Fraction > p.Fraction || status.Completed > p.Completed || (status.Failure == nil && status.Actual != p.Actual) {
+		status.LastProgress = time.Now()
+	}
 	w.w.updateStatus(status)
 }
 
@@ -471,7 +479,7 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 	}
 
 	// update each object
-	err := payload.RunGraph(ctx, graph, maxWorkers, func(ctx context.Context, tasks []*payload.Task) error {
+	errs := payload.RunGraph(ctx, graph, maxWorkers, func(ctx context.Context, tasks []*payload.Task) error {
 		for _, task := range tasks {
 			if contextIsCancelled(ctx) {
 				return cr.CancelError()
@@ -495,8 +503,8 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 		}
 		return nil
 	})
-	if err != nil {
-		cr.Error(err)
+	if len(errs) > 0 {
+		err := cr.Errors(errs)
 		return err
 	}
 
@@ -517,6 +525,12 @@ func init() {
 		metricPayload,
 	)
 }
+
+type errCanceled struct {
+	err error
+}
+
+func (e errCanceled) Error() string { return e.err.Error() }
 
 // consistentReporter hides the details of calculating the status based on the progress
 // of the graph runner.
@@ -553,14 +567,31 @@ func (r *consistentReporter) Error(err error) {
 	copied := r.status
 	copied.Step = "ApplyResources"
 	copied.Fraction = float32(r.done) / float32(r.total)
-	copied.Failure = err
+	if !isCancelledError(err) {
+		copied.Failure = err
+	}
 	r.reporter.Report(copied)
+}
+
+func (r *consistentReporter) Errors(errs []error) error {
+	err := summarizeTaskGraphErrors(errs)
+
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	copied := r.status
+	copied.Step = "ApplyResources"
+	copied.Fraction = float32(r.done) / float32(r.total)
+	if err != nil {
+		copied.Failure = err
+	}
+	r.reporter.Report(copied)
+	return err
 }
 
 func (r *consistentReporter) CancelError() error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
-	return fmt.Errorf("update was cancelled at %d/%d", r.done, r.total)
+	return errCanceled{fmt.Errorf("update was cancelled at %d/%d", r.done, r.total)}
 }
 
 func (r *consistentReporter) Complete() {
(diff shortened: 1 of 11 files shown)


wking:
The additional information looks good, but it's not always there:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/158/pull-ci-openshift-cluster-version-operator-master-e2e-aws/548/artifacts/e2e-aws/installer/.openshift_install.log | grep 'luster .* initialize'
time="2019-04-09T05:33:50Z" level=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-sh9t14qk-928f7.origin-ci-int-aws.dev.rhcloud.com:6443 to initialize..."
time="2019-04-09T05:33:50Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 83% complete"
time="2019-04-09T05:35:35Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 85% complete"
time="2019-04-09T05:35:48Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 90% complete"
time="2019-04-09T05:36:20Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 91% complete"
time="2019-04-09T05:37:21Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 91% complete"
time="2019-04-09T05:39:35Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 92% complete"
time="2019-04-09T05:40:51Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 92% complete"
time="2019-04-09T05:49:05Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 98% complete"
time="2019-04-09T05:49:20Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 98% complete, waiting on authentication, image-registry, ingress, marketplace, monitoring"
time="2019-04-09T05:50:35Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 98% complete"
time="2019-04-09T05:53:35Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 0.0.1-2019-04-09-051341: 99% complete"
time="2019-04-09T05:54:05Z" level=debug msg="Cluster is initialized"
```

If the installer had decided to timeout at 5:53:35, we wouldn't have been able to point at any slow operators.

smarterclayton:
> If the installer had decided to timeout at 5:53:35, we wouldn't have been able to point at any slow operators.

For that to happen we'd have to have been going for 30m.  For install with a timeout that's fine, because right now anything taking longer than ~15 is wedged and would always report.  We reset because we started making progress.  I think if we're making progress we don't want to report.

We can cut the sync interval on install - will add that as a new commit.

wking:
> For that to happen we'd have to have been going for 30m. For install with a timeout that's fine, because right now anything taking longer than ~15 is wedged and would always report. We reset because we started making progress. I think if we're making progress we don't want to report.

So is that "by the time we hit our 30m timeout, we will have stabilized in a wedged state"?  Previous experience like [this][1] shows that sometimes things stick for a while and then eventually shake loose, but not soon enough to get in under the installer timeout.  That particular case predates #141, so maybe we are confident that any cluster that isn't ready in 30 minutes will still be hung and reporting via this PR's code?

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1691513#c1

smarterclayton:
The installer failing at 30m but then eventually recovering is only a problem for CI, and since it's almost always a bug in an operator and we already gather cluster operator status into artifacts, I'm not sure the the message is the important part.  It's 10 seconds per failure to look at the failing operator.  I'd say this is addressing the experiential "an admin eventually sees the failing job show up once we settle on the wedge".

smarterclayton:
other feedback?  @abhinavdahiya ?

sdodson:
> The installer failing at 30m but then eventually recovering is only a problem for CI

I don't understand that. We've had multiple users encounter just that and report bugs on that scenario.

smarterclayton:
Those are caused by bugs.  The bugs are debuggable in CI.  For a user, they should also be debuggable as I noted with this change because we won't be making progress at 30.  If they're timing out because of factors beyond our control such as pull speed, then the installer messaging needs to change to more carefully communicate that we are stopping waiting, but that the install may still complete.  Especially if it isn't marked as failed, but is still progressing.

pkg/cvo/sync_worker.go
```diff
@@ -576,6 +607,136 @@ func (r *consistentReporter) Complete() {
...
+		}
+		if len(uErr.Name) > 0 {
+			names = append(names, uErr.Name)
+		}
+	}
+	if len(names) == 0 {
+		return nil
+	}
+
+	nested := make([]error, 0, len(errs))
```
abhinavdahiya: ques: why create new `nested` error list?

abhinavdahiya:
https://github.com/openshift/cluster-version-operator/pull/158/commits/c2ac20fa172da73a873d3e3728053a30af9cf4dc is LGTM I like the idea of summarizing the errors.

and https://github.com/openshift/cluster-version-operator/pull/158/commits/3c2cf464ef88f7d3f30df4bc66e4767b6e79246a makes sense for now as a stop gap.

abhinavdahiya:
This makes the messages better for CVO and therefore this is:
/approve

but to @wking and @sdodson 's point:
case where cvo hasn't reached level after 30 mins and  is still progressing we (as in installer) would need to inform the user that either cvo is still progressing and you might have to wait longer using `wait-for` or cvo is failing and pass the user the failure reported by the cvo.

wking:
/lgtm

We can file follow-ups if/when we can agree on them ;).

wking:
[integration][1]:

```
start_integration_test.go:244: cluster version never became available: operator listed as failing (MultipleErrors): Multiple errors are preventing progress:
		* Could not update configmap "e2e-cvo-s5m9/config1" (1 of 2): the server has forbidden updates to this resource
		* Could not update configmap "e2e-cvo-s5m9/config2" (2 of 2): the server has forbidden updates to this resource
```

/test integration

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/158/pull-ci-openshift-cluster-version-operator-master-integration/243

smarterclayton:
Heh, that's the actual new error in the integration test.  Fixing.

I think we can file something on installer to double check the output at the end of a release.

smarterclayton:
Hrm... this looks more like the master failed or lost data.

/retest

wking:
> I think we can file something on installer to double check the output at the end of a release.

I'll work that into openshift/installer#1413 tonight.  Should we stop waiting if the CVO ever reports Failing?

smarterclayton:
We should probably just continue until the timeout.  I think we could maybe record the last failing condition we see along the way, and if we are currently not failing but have timed out just report "we failed"

smarterclayton:
/test integration

Delivered a fix in https://github.com/openshift/release/pull/3450 that should have fixed the issue we were seeing

smarterclayton:
nope, that's not it

/test integration

wking:
[e2e-aws][1]:

```
Flaky tests:

[Conformance][Area:Networking][Feature:Router] The HAProxy router should expose the profiling endpoints [Suite:openshift/conformance/parallel/minimal]
[Feature:Builds][Conformance] build without output image  building from templates should create an image from a docker template without an output image reference defined [Suite:openshift/conformance/parallel/minimal]
[Feature:DeploymentConfig] deploymentconfigs generation [Conformance] should deploy based on a status version bump [Suite:openshift/conformance/parallel/minimal]
[Feature:DeploymentConfig] deploymentconfigs viewing rollout history [Conformance] should print the rollout history [Suite:openshift/conformance/parallel/minimal]
[Feature:DeploymentConfig] deploymentconfigs with failing hook [Conformance] should get all logs from retried hooks [Suite:openshift/conformance/parallel/minimal]
[Feature:Platform] OLM should list catalogsources [Suite:openshift/conformance/parallel]
[Feature:Platform][Smoke] Managed cluster should start all core operators [Suite:openshift/conformance/parallel]
[sig-storage] Volume expand Verify if editing PVC allows resize [Suite:openshift/conformance/parallel] [Suite:k8s]

Failing tests:

[Conformance][Area:Networking][Feature:Router] The HAProxy router should set Forwarded headers appropriately [Suite:openshift/conformance/parallel/minimal]
[Conformance][templates] templateservicebroker end-to-end test  should pass an end-to-end test [Suite:openshift/conformance/parallel/minimal]
[Feature:Platform] Managed cluster should have no crashlooping pods in core namespaces over two minutes [Suite:openshift/conformance/parallel]
```

Crashlooping kube-controller-manager is likely one of [rhbz#1691055][2], [rhbz#1698667][3], or [rhbz#1698950][4].

/retest

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/158/pull-ci-openshift-cluster-version-operator-master-e2e-aws/561
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1691055
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1698667
[4]: https://bugzilla.redhat.com/show_bug.cgi?id=1698950

PR #159, opened 2019-04-09
resource-apply: add release.openshift.io/create-only annotation

deads2k:
If a resource to apply specifies this annotation it is created if missing,
but it is never updated.  We want to use this for empty config.openshift.io
and operator.openshift.io resources, but support is uniform.

@smarterclayton I think the name matches our discussion.
/assign @abhinavdahiya

### DIFF
--- a/lib/resourceapply/interface.go
+++ b/lib/resourceapply/interface.go
@@ -1,0 +1,18 @@
+package resourceapply
+
+import (
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CreateOnlyAnnotation means that this resource should be created if it does not exist, but should not be updated
+// if it already exists.  It is uniformly respected across all resources, but the first known use-cases are for
+// empty config.openshift.io and initial low-level operator resources.
+// Set .metadata.annotations["release.openshift.io/create-only"]="true" to have a create-only resource.
+const CreateOnlyAnnotation = "release.openshift.io/create-only"
+
+// IsCreateOnly takes metadata and returns true if the resource should only be created, not updated.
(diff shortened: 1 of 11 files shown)


smarterclayton:
Maybe add a unit test to the sync worker (whomever stresses the generic ones)

lib/resourceapply/interface.go
```diff
@@ -0,0 +1,16 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CreateOnlyAnnotation means that this resource should be created if it does not exist, but should not be updated
+// if it already exists.  It is uniformly respected across all resources, but the first known use-cases are for
+// empty config.openshift.io and initial low-level operator resources.
+const CreateOnlyAnnotation = "release.openshift.io/create-only"
+
+func IsCreateOnly(metadata metav1.Object) bool {
+	return strings.EqualFold(metadata.GetAnnotations()[CreateOnlyAnnotation], "true")
```
abhinavdahiya: is there a guarantee that `GetAnnotations` returns non-nil map?

smarterclayton: Gets on nil maps are allowed

abhinavdahiya:
needs documentation on how to use the annotation.

lib/resourceapply/interface.go
deads2k: > is there a guarantee that `GetAnnotations` returns non-nil map?

retrieval from a nil map is the zero value of the values in that map which works correctly here.

deads2k:
> Maybe add a unit test to the sync worker (whomever stresses the generic ones)

I'll do a unit test on the config.openshift.io case.

deads2k:
unit test created and doc updated

PR #161, opened 2019-04-11
docs: degraded condition

derekwaynecarr:


docs/dev/clusteroperator.md
```diff
@@ -76,7 +76,20 @@ Here are the guarantees components can get when they follow the rules we define:
...
+   pods, and 2 are running, and 1 is crashlooping, your operator service is
+   possibly still `Available`, but its service is in a `Degraded` state.  A
+   component may be `Degraded` while it is `Progressing` to a new desired state;
+   for example, if only 2 of the 3 desired replicas are achieved.  As a result,
+   it may be normal for a operator during upgrade to temporarily report
+   `Degraded`.  A component may be `Progressing` but not `Degraded` if a config
+   change is happening that does not require a roll-out of new pods.  If your
+   rolling deployment bursts above its target replica count, you may be
+   `Progressing` but not `Degraded` because you have your desired number of pods
+   running to meet your service demand.  If  a component remains degraded for an
```
sjenning: I would think another indicator of a persistent Degraded state is Degraded=true and Progressing=false.

I still wish there was a way to indicate the notion of "external intervention is needed to get out of Degraded=true"

docs/dev/clusteroperator.md
```diff
@@ -76,7 +76,22 @@ Here are the guarantees components can get when they follow the rules we define:
...
 1. A component doesn't report the `Available` status condition the first time until they are completely rolled out (or within some reasonable percentage if the component must be installed to all nodes)
-2. A component reports `Failing` when it can't accomplish its task. This might be very broad - the API servers are down so I failed to update - or narrow - I couldn't update the last secret. In either case, Failing communicates something is wrong.
+2. A component reports `Degraded` when it can't accomplish its task or if its
+   current state does not match its desired state resulting in a lower quality
+   of service.  For example, if you desire 3 pods, and 2 are running, but 1 is
+   crash-looping, your operator service is possibly still `Available`, but its
+   service is in a `Degraded` state.  A component may be `Degraded` while it is
+   `Progressing` to a new desired state; for example, if only 2 of the 3 desired
+   replicas are achieved.  As a result, it may be normal for a operator during
+   upgrade to temporarily report `Degraded`.  A component may be `Progressing`
```
smarterclayton: I'm still not 100% on board with this.  I don't want operators flashing between degraded and not degraded.  There needs to be a "within a minute or so, if the condition isn't resolved, an operator should report degraded".  It's acceptable to start by flashing degraded, but I'm going to open bugs to you until during normal upgrades you don't.

I don't think being 2/3 during a rolling upgrade is degraded.  I think being 2/3 because the upgrade can't make progress is degraded.

smarterclayton: We need to be very careful that people don't interpret Degraded as "normal operation but change is happening".  Degraded is "you either have no idea what is going on" (which is something a development team is expected to fix and bugs will be opened against you) or "something is legitimately failing".

I don't want to enshrine "shrug" as degraded.

docs/dev/clusteroperator.md
```diff
@@ -75,11 +75,35 @@ Here are the guarantees components can get when they follow the rules we define:
...
-2. A component reports `Failing` when it can't accomplish its task. This might be very broad - the API servers are down so I failed to update - or narrow - I couldn't update the last secret. In either case, Failing communicates something is wrong.
-3. A component reports `Progressing` when it is rolling out new code, propagating config changes, or otherwise moving from one steady state to another. It should not report progressing when it is reconciling a previously known state. If it is progressing to a new version, it should include the version in the message for the condition like "Moving to v1.0.1".
-4. A component reports `Upgradeable` as `false` when it wishes to prevent an upgrade for an admin-correctable condition. The component should include a message that describes what must be fixed.
-5. A component reports when it has rolled out the new version of its operands
+1. A operator doesn't report the `Available` status condition the first time
+   until they are completely rolled out (or within some reasonable percentage if
+   the component must be installed to all nodes)
+2. An operator reports `Degraded` when its current state does not match its
```
abhinavdahiya: > An operator reports `Degraded` when its current state does not match its
   desired state resulting in a lower quality of service over a period of time.

nit: `An operator reports `Degraded` when its current state does not match its
   desired state over a period of time resulting in a lower quality of service .` is much more clearer that operators mark degraded when they have been trying to achieve desired but haven't achieved it for a period of time....

derekwaynecarr: agreed.  updated phrasing to match.

crawford:
@michaelgugino is your concern that the over-a-period-of-time requirement is going to mask issues? It sounds like you'd rather the operators fail fast than hew and haw. (Just want to understand your position)

eparis:
/retest
/lgtm
we can iterate on this definition over time, but we merged the name change, so we should talk some more about it   :)

eparis:
see: https://github.com/openshift/api/pull/287

abhinavdahiya:
> /retest
> /lgtm
> we can iterate on this definition over time, but we merged the name change, so we should talk some more about it :)

/approve

danwinship:
@eparis So are we changing the definition or not? You implied in the other PRs that this was just a rename for now, but the docs change here also changes the semantics.

eparis:
the name change is required before GA, the semantic change is aspirational.

docs/dev/clusteroperator.md
```diff
@@ -75,11 +75,35 @@ Here are the guarantees components can get when they follow the rules we define:
...
+   desired state over a period of time resulting in a lower quality of service.
+   The period of time may vary by component, but a `Degraded` state represents
+   persistent observation of a condition.  As a result, a component should not
+   oscillate in and out of `Degraded` state. A service may be `Available` even
+   if its degraded.  For example, your service may desire 3 running pods, but 1
+   pod is crash-looping. The service is `Available` but `Degraded` because it
+   may have a lower quality of service.  A component may be `Progressing` but
+   not `Degraded` because the transition from one state to another does not
+   persist over a long enough period to report `Degraded`.  A service should not
```
runcom: Why operator can't be Progressing=True and Degraded=True? Today, if progressing towards a new version fails we flip Failing=True but keep Progressing=True (if for instance, the master pool in MCO don't get ready, and that may be temporary till all nodes roll out or persistent over a period of time which I guess we can try to measure/act on)

PR #162, opened 2019-04-11
pkg/cincinnati: Soften unknown-version message

wking:
Maybe having the current version not in the desired channel's graph is a problem, but maybe you know what you're doing.  And Cincinnati might *know* about your version, and just consider it to be outside your
channel, in which case the solution may be "switch channels" and not "switch versions".  This change hopefully makes it a bit easier for an admin in this condition to figure out the appropriate next step for their cluster.

CC @smarterclayton, @steveeJ

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -90,7 +90,7 @@ func (c Client) GetUpdates(upstream string, channel string, version semver.Versi
 		}
 	}
 	if !found {
-		return nil, fmt.Errorf("unknown version %s", version)
+		return nil, fmt.Errorf("currently installed version %s not found in the %q channel", version, channel)
 	}
 
 	// Find the children of the current version.
--- a/pkg/cincinnati/cincinnati_test.go
+++ b/pkg/cincinnati/cincinnati_test.go
@@ -47,7 +47,7 @@ func TestGetUpdates(t *testing.T) {
 		name:          "unknown version",
 		version:       "4.0.0-3",
 		expectedQuery: "channel=test-channel&id=01234567-0123-0123-0123-0123456789ab&version=4.0.0-3",
-		err:           "unknown version 4.0.0-3",
+		err:           "currently installed version 4.0.0-3 not found in the \"test-channel\" channel",
 	}}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {


pkg/cincinnati/cincinnati.go
```diff
@@ -90,7 +90,7 @@ func (c Client) GetUpdates(upstream string, channel string, version semver.Versi
 		}
 	}
 	if !found {
-		return nil, fmt.Errorf("unknown version %s", version)
+		return nil, fmt.Errorf("version %s is not in the %s channel", version, channel)
```
steveeJ: I'm not sure what the best wording here is, though it should highlight that it's talking about the currently installed version which is not found in the updates available for the requested channel. May be "`currently installed version %q does not exist in channel %q`?

wking: > May be "`currently installed version %q does not exist in channel %q`?

I've pushed 310884a -> 3dca6ac, rebasing on master and adjusting to:

    current version %s not found in the %q channel

Differences vs. your suggestion:

* "currently installed" -> "current".  I think the single word is clear enough.
* `%q` -> `%s` for the version.  We are requiring a `semver.Version` for `version`, so I don't see an opportunity for an empty or whitespace string there.  Quoting the channel makes sense, so I've followed you there.
* "does not exist" -> "not found", to be a bit less opinionated about why we haven't found our version.

Thoughts?

steveeJ: > "currently installed" -> "current". I think the single word is clear enough.

in case we're not trying to save bytes or screenspace I encourage to add the additional word. A user might otherwise misinterpret the error, e.g. thinking that they requested a specific version to be upgraded to and getting confused about it. 

>     * `%q` -> `%s` for the version.  We are requiring a `semver.Version` for `version`, so I don't see an opportunity for an empty or whitespace string there.  Quoting the channel makes sense, so I've followed you there.

:+1: (though I generally prefer single-quotes around variable values for readability; didn't think of escaping)

> "does not exist" -> "not found", to be a bit less opinionated about why we haven't found our version.

:+1:

wking: > in case we're not trying to save bytes or screenspace I encourage to add the additional word. A user might otherwise misinterpret the error, e.g. thinking that they requested a specific version to be upgraded to and getting confused about it.

I think "current version" is clearly not "the version you want to upgrade to", but whatever, I've updated to "currently installed version" with 3dca6ac -> cb4d7b1.

wking:
Fixed [the unit test][1] with cb4d7b1 -> 21e85cd.

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/162/pull-ci-openshift-cluster-version-operator-master-unit/435

wking:
[e2e-aws][1]:

```
Flaky tests:

The bootstrap user should successfully login with password decoded from kubeadmin secret [Suite:openshift/conformance/parallel]
[Conformance][Area:Networking][Feature:Router] The HAProxy router should serve routes that were created from an ingress [Suite:openshift/conformance/parallel/minimal]
[Feature:DeploymentConfig] deploymentconfigs with failing hook [Conformance] should get all logs from retried hooks [Suite:openshift/conformance/parallel/minimal]
[Feature:Platform] OLM should list operatorgroups [Suite:openshift/conformance/parallel]
[sig-network] DNS should provide DNS for the cluster  [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
[sig-storage] Secrets should be able to mount in a volume regardless of a different secret existing with same name in different namespace [NodeConformance] [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]

Failing tests:

[Feature:Builds] forcePull should affect pulling builder images  ForcePull test case execution custom [Suite:openshift/conformance/parallel]
[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Pre-provisioned PV (xfs)] volumes should allow exec of files on the volume [Suite:openshift/conformance/parallel] [Suite:k8s]
```

/retest

[1]: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/162/pull-ci-openshift-cluster-version-operator-master-e2e-aws/597

PR #165, opened 2019-04-17
Change Failed to Degraded

eparis:


eparis:
Based on: https://github.com/openshift/api/pull/287

eparis:
https://github.com/openshift/cluster-version-operator/pull/161

eparis:
see: https://github.com/openshift/api/pull/287

pkg/cvo/internal/operatorstatus.go
```diff
@@ -167,48 +166,71 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
 		available := false
 		progressing := true
 		failing := true
-		for _, condition := range actual.Status.Conditions {
+		var failingCondition *configv1.ClusterOperatorStatusCondition
+		degraded_value := true
```
smarterclayton: `degradedValue` this isn't C :)

pkg/cvo/internal/operatorstatus.go
```diff
@@ -167,48 +166,71 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
...
 			message := fmt.Sprintf("Cluster operator %s is reporting a failure", actual.Name)
-			if len(c.Message) > 0 {
-				message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, c.Message)
+			if len(condition.Message) > 0 {
+				message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, condition.Message)
 			}
 			lastErr = &payload.UpdateError{
 				Nested:  errors.New(lowerFirst(message)),
-				Reason:  "ClusterOperatorFailing",
+				Reason:  "ClusterOperatorDegraded",
```
smarterclayton: Did you change the places this is referenced elsewhere in the CVO?

hack/test-integration.sh
```diff
@@ -5,4 +5,4 @@ set -euo pipefail
 base=$( dirname "${BASH_SOURCE[0]}")
 
 go run "${base}/test-prerequisites.go"
-TEST_INTEGRATION=1 go test ./... -test.run=^TestIntegration -args -alsologtostderr -v=5
\ No newline at end of file
+TEST_INTEGRATION=1 go test ./... -test.run=^TestIntegration
```
smarterclayton: This is a problem because now we can't debug integration logs.  You'll need to have integration explicitly set both alsologtostderr or v5 (although pulling in klog should be sufficient to get the args are back).

pkg/cvo/internal/operatorstatus.go
smarterclayton: Doesn't look like it, change ClusterOperatorFailing in task.go and also change the message.

pkg/cvo/internal/operatorstatus.go
eparis: that'll only last a week or two   :)

eparis:
Updating vendor was via: `dep ensure update github.com/openshift/api github.com/openshift/client-go`
You definitely don't want to just do `dep ensure update`
There doesn't appear to be a `make vendor`

pkg/cvo/internal/operatorstatus.go
eparis: fixed

pkg/cvo/internal/operatorstatus.go
eparis: fixed.

hack/test-integration.sh
eparis: how does one do this given klog?

eparis: ok, hopefully adding an init function will make it work...

smarterclayton:
This looks ok to me but want @abhinavdahiya to take a look if he can tomorrow.

smarterclayton:
Ok, will deal with fallout in separate issues

/lgtm
/approve

PR #166, opened 2019-04-17
sync: During reconcile, try different parts of the payload and don't hang

smarterclayton:
The worst outcome of reconcile would be that a break occurs in a late component and the CVO gets hung trying to reach it to correct the issue (i.e. 0000_30 is broken due to a change in 0000_70).  Because we process in order we never get to 70 and fix it.  The other issue is that a medium to long term break in 30 can prevent entropy repair at 70, so 70 degrades until 30 recovers.

Instead of running the reconcile graph in order, run it in random blocks at a lower parallelization factor to ensure that we will eventually try all blocks, even if something fundamental is broken. The permutation of the graph ordering is stable for a given payload and sync time (so that when we test locally we are testing the same order a customer we would see).

The only gap still left would be a component breaks that can't be recovered due to ordering (to fix a role binding we need to deploy a deployment, but the lack of the deployment prevents us from sending the role binding) which we can potentially investigate later by randomizing the task order.

We track "number of attempts" in a given state / payload to see the input.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -3,6 +3,7 @@ package cvo
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


smarterclayton:
/assign @abhinavdahiya 

I think with this we are in an "ok" state for ship with regards to entropy minimization in the CVO.

smarterclayton:
@wking can you review?

smarterclayton:
/retest

not related

PR #167, opened 2019-04-18
pkg/payload/task: Include the Nested error in UpdateError.Error()

wking:
For example, this would get the underlying reason logged, vs. our current behavior where we don't share this information, even [in our own logs][1]:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-marketplace/151/pull-ci-operator-framework-operator-marketplace-master-e2e-aws/777/artifacts/e2e-aws/pods/openshift-cluster-version_cluster-version-operator-bd6c56bbf-7khsr_cluster-version-operator.log.gz | gunzip
...
E0418 15:45:40.228571       1 task.go:77] error running apply for clusteroperator "network" (39 of 332): Cluster operator network has not yet reported success
...
```

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1701352#c3

### DIFF
--- a/pkg/payload/task.go
+++ b/pkg/payload/task.go
@@ -123,6 +123,9 @@ type UpdateError struct {
 }
 
 func (e *UpdateError) Error() string {
+	if e.Nested != nil {
+		return fmt.Sprintf("%s: %s", e.Message, e.Nested)
+	}
 	return e.Message
 }
 
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -2488,7 +2488,7 @@ func TestCVO_ErrorDuringReconcile(t *testing.T) {
 			},
 			Conditions: []configv1.ClusterOperatorStatusCondition{
 				{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue, Message: "Done applying 1.0.0-abc"},
-				{Type: ClusterStatusFailing, Status: configv1.ConditionTrue, Reason: "UpdatePayloadFailed", Message: "Could not update test \"file-yml\" (3 of 3)"},
+				{Type: ClusterStatusFailing, Status: configv1.ConditionTrue, Reason: "UpdatePayloadFailed", Message: "Could not update test \"file-yml\" (3 of 3): unable to proceed"},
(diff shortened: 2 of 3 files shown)


pkg/payload/task.go
```diff
@@ -119,6 +119,9 @@ type UpdateError struct {
 }
 
 func (e *UpdateError) Error() string {
+	if e.Nested != nil {
```
abhinavdahiya: consumers must call cause to find internal errors right?

wking: > consumers must call cause to find internal errors right?

That's how master works.  This PR is exposing the cause string as part of the the usual `Error()` rendering, because consumers like our logging and Event creation code are currently not bothering to look at `Cause()`.  Should I be updating them to look at `Cause()` instead?  I can't think of a case where someone would call `Error()` and _not_ be interested in the root cause.

wking:
Fixed the unit test and moved from `%#+v` to `%s` for `Nested` with e7d8068 -> 006296c.

pkg/payload/task.go
wking: An alternative approach would be to stick to `Message` for `Error()` but to add a new `Format()` so folks could use `%s`, `%+v`, etc. to easily switch between different renderings.  For an example implementation, see [`pkg/errors`][1].

[1]: https://github.com/pkg/errors/blob/v0.8.1/errors.go#L127-L141

wking:
Or, maybe we do have better logging [a bit later on][1] (thanks, @danwinship):

```
I0418 15:45:40.228896       1 sync_worker.go:645] Update error 39/332: ClusterOperatorNotAvailable Cluster operator network has not yet reported success (*errors.errorString: cluster operator network is not done; it is available=true, progressing=false, failing=true)
```

So I'm not sure why we aren't logging *that* in `task.go` and passing it up to the Event.

While reading around with the above error, see some direct `.Nested` access that this PR should remove if we move forward with it.  But I also see `reasonForPayloadSyncError` and `SummaryForReason`, which seem to be designed to mask/summarize underlying errors.  Ideally error Events would include the error, a high-level summary, and the low-level detail, so folks could get the right level of insight without having to dive into logs.  That's a bigger change though; thoughts on scoping this PR?

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1701352#c5

wking:
Rebased onto master with 285d3221 -> b63d2268.

PR #169, opened 2019-04-20
start: Initialize the payload and fail if it can't be read at startup

smarterclayton:
Avoid the possibility of a bad release image being created that can't
correctly read the payload by failing hard on startup if the payload is
invalid.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -181,20 +181,23 @@ func New(
 			panic(err)
 		}
 	}
+	return optr
+}
 
-	if update, err := payload.LoadUpdate(optr.defaultPayloadDir(), releaseImage); err != nil {
-		glog.Warningf("The local release contents are invalid - no current version can be determined from disk: %v", err)
-	} else {
-		optr.releaseCreated = update.ImageRef.CreationTimestamp.Time
-		// XXX: set this to the cincinnati version in preference
-		if _, err := semver.Parse(update.ImageRef.Name); err != nil {
-			glog.Warningf("The local release contents name %q is not a valid semantic version - no current version will be reported: %v", update.ImageRef.Name, err)
-		} else {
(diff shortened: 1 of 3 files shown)


pkg/cvo/cvo.go
```diff
@@ -181,20 +181,23 @@ func New(
...
-		if _, err := semver.Parse(update.ImageRef.Name); err != nil {
-			glog.Warningf("The local release contents name %q is not a valid semantic version - no current version will be reported: %v", update.ImageRef.Name, err)
-		} else {
-			optr.releaseVersion = update.ImageRef.Name
-		}
+// InitializeFromPayload retrieves the payload contents and verifies the initial state.
+func (optr *Operator) InitializeFromPayload() error {
+	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.releaseImage)
+	if err != nil {
+		return fmt.Errorf("the local release contents are invalid - no current version can be determined from disk: %v", err)
```
abhinavdahiya: nit can we use already imported `github.com/pkg/errors` for wrapping the error?
`return errors.Wrap(err, "failed to release image - no current version can be determined from disk")`

pkg/cvo/cvo.go
```diff
@@ -181,20 +181,23 @@ func New(
...
+	}
+	// XXX: set this to the cincinnati version in preference
+	if _, err := semver.Parse(update.ImageRef.Name); err != nil {
```
abhinavdahiya: are we adding requirement for all release images must be semver ?

abhinavdahiya:
Can you point include in the commit message why we are moving to requirement that all release image versions must be semver? this is a requirement for CVO will Cincinnati support this constraint?

smarterclayton:
Cincinnati already enforces this, as does release creation.  I just
realized in here that we don't enforce.

smarterclayton:
Updated commit message

PR #170, opened 2019-04-20
Verify a GPG signature on the release image digest from a root of trust within the payload

smarterclayton:
Implement:

1. A payload can contain a root of trust config map that describes GPG public keys and one or more sources of signatures
2. A verifier that checks the root of trust config map if present in the payload and fails if a signature cannot be found for all public keys from any of the stores
3. Change the CVO to default to "don't allow unverified images", block update to images that haven't been verified, and allow an admin to override that with a field in desiredUpdate `allowUnverifiedImages`
4. Report unverified errors in CVO status as well as the verify bit on the history

In CI, the payload would have a signing key public key mixed in to the payload (OKD, OCP CI).  In OCP, we would mix in an official signing public key and the location of one or more stores to read from.

Questions identified during prototyping:

1. The sync loop does nothing when desiredUpdate is failing to retrieve a payload - we need to "under sync" in that case (continue to sync whatever the last payload was) until the payload is retrieved
2. We need to allow additional stores to be specified on a cluster that define where else source can be located from.  That could be a second config map we mix in that is create only, or it could be global config.
3. What failure modes are not covered here?   The upstream stores going down can be handled by background retry and allowing override.  
4. Update isn't calling `oc adm release info --verify`, which it should (prevents the payload from being unreachable or inconsistent).  However, we'd have to wire through the pull secret into the container that runs oc.
5. Is this too paranoid?  Escape hatch of `--force` makes me feel better, especially when we do automated updates. I would rather have all of our testing using a signature rather than only what end users see.
6. Use something besides atomic container image?  Would prefer not to because a) it's reasonably well reviewed, b) there is already red hat detached signature infra that can publish signatures (we have to talk with RCM more).

TODO:

* [x] Clients have to change (oc adm upgrade and web console)
* [x] Get the CI release signer and verifier up, and public signatures to GCS or other public location
* [ ] Additional testing
* [x] Decide whether the payload should consider its own payload (the innate one) as verified implicitly.  Answer is probably not.


pkg/cvo/cvo.go
```diff
@@ -197,6 +204,20 @@ func (optr *Operator) InitializeFromPayload() error {
...
 	optr.releaseVersion = update.ImageRef.Name
+
+	// attempt to load a verifier as defined in the payload
+	verifier, err := verify.LoadFromPayload(update)
+	if err != nil {
+		return err
+	}
+	if verifier != nil {
+		glog.Infof("Verifying release authenticity: %v", verifier)
+	} else {
```
abhinavdahiya: would it be better if decided to use Reject based on error type vs nil verifier?

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+type rejectVerifier struct{}
+
+func (rejectVerifier) Verify(ctx context.Context, releaseDigest string) error {
+	return fmt.Errorf("verification is not possible")
+}
+
+// Reject fails always fails verification.
+var Reject Interface = rejectVerifier{}
+
+// LoadFromPayload looks for a config map in the v1 API group within within the provided update
```
abhinavdahiya: dup `within within`

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+// entry with the annotation is used.
+//
+// The presence of this config map instructs the CVO to verify updates before they are downloaded.
+//
+// The keys within the config map define how verification is performed:
+//
+// verifier-public-key-*: One or more GPG public keys in ASCII form that must have signed the
+//                        release image by digest.
+//
+// store-*: A URL (scheme file://, http://, or https://) location that contains signatures. These
```
abhinavdahiya: nit: we should not allow file:// right? there is not way we can make that happen...

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+//          GPG signed and encrypted JSON message.
+//
+// See https://github.com/containers/image/blob/ab49b0a48428c623a8f03b41b9083d48966b34a9/docs/signature-protocols.md
+// for a description of the signature store
+//
+// The returned verifier will require that any new release image will only be considered verified
+// if each provided public key has signed the release image digest. The signature may be in any
+// store and the lookup order is internally defined.
+//
+func LoadFromPayload(update *payload.Update) (Interface, error) {
```
abhinavdahiya: nit it would be better if we did not the errors, and use the `github.com/pkg/errors` for wrapping the errors.

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+	transport, err := transport.New(&transport.Config{})
```
abhinavdahiya: nit: this looks like not the right place to initialize the transport

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+				}
+				verifiers[k] = keyring
+			case strings.HasPrefix(k, "store-"):
+				u, err := url.Parse(v)
+				if err != nil || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "file") {
+					return nil, fmt.Errorf("%s has an invalid key %q: must be a valid URL with scheme file://, http://, or https://", src, k)
+				}
+				stores = append(stores, u)
+			default:
+				glog.Warningf("An unexpected key was found in %s (expected store-* or verifier-public-key-*): %s", src, k)
```
abhinavdahiya: nit: include the action that was taken because of this invalid configuration.

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+		return keyring, nil
+	}
+	return openpgp.ReadKeyRing(bytes.NewReader(data))
+}
+
+const maxSignatureSearch = 10
+
+var validReleaseDigest = regexp.MustCompile(`^[a-zA-Z0-9:]+$`)
+
+type releaseVerifier struct {
```
abhinavdahiya: This needs description.

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+	base := filepath.Join(dir, "signature-")
+	for i := 1; i < maxSignatureSearch; i++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		path := base + strconv.Itoa(i)
+		data, err := ioutil.ReadFile(path)
+		if err != nil {
+			if os.IsNotExist(err) {
```
abhinavdahiya: nit: this can be moved to before `err != nil` check.

pkg/verify/verify.go
```diff
@@ -0,0 +1,469 @@
...
+				if i == 1 {
```
abhinavdahiya: why this check?

pkg/cvo/cvo.go
```diff
@@ -218,6 +205,18 @@ func (optr *Operator) InitializeFromPayload() error {
 	}
 	optr.verifier = verifier
 
+	// set up the config sync worker
```
abhinavdahiya: can you add reasoning for this change to the commit?

pkg/verify/verify.go
smarterclayton: I'll update the note - I allow it for testing both paths and it's possible we could mount it in to the CVO in the future.  Testing only for now.

pkg/verify/verify.go
smarterclayton: Why not?  Transport initialization is cheap?

pkg/cvo/cvo.go
smarterclayton: I only did it for maximum paranoia (easier to review the code block here that guarantees that a verifier is set).

pkg/verify/verify.go
smarterclayton: not truly necessary, I can remove it now that there are tests

pkg/verify/verify.go
smarterclayton: oh, I see what you mean.  Yes can go in the loop.

smarterclayton:
Simplified down to fix two problems:

1. Don't try to reverify unverified payloads - the fact that local doesn't need to be verified complicates that, and in the future we can implement caching and other niceties around the verifier and have it be invoked before the retriever
2. Remove the negation of UnverifiedImage and weirdness of the signature of RetrievePayload by having it return a struct with the data needed by the caller.

Added another cvo_scenario test that showed different logic.

pkg/cvo/status.go
```diff
@@ -377,7 +382,7 @@ func (optr *Operator) syncFailingStatus(original *configv1.ClusterVersion, ierr
 		LastTransitionTime: now,
 	})
 
-	mergeOperatorHistory(config, optr.currentVersion(), now, false)
+	mergeOperatorHistory(config, optr.currentVersion(), false, now, false)
```
abhinavdahiya: why set image no verified when failing ?

smarterclayton: So verified is "level" which false won't reset verified.  We just don't know here, so we don't change the state of the previous history.

smarterclayton:
Are we ok with API changes?  If so #293 needs to be merged before this and I need to bump the API.

smarterclayton:
Weird, I can't recreate the unit test flake locally.  Trying more

smarterclayton:
Pushed a change to tolerate trailing and leading whitespace on config map keys for stores

smarterclayton:
https://github.com/openshift/cluster-update-keys is created and CI will be in place shortly

smarterclayton:
It is not payload enabled so we can test PRs to it

smarterclayton:
Ok, that run passed except for the auth flake against the registry.

If there's no concerns remaining, I think this is ok to merge.  It will be inert until the payload gets an image with a verifier config map.

smarterclayton:
Weird, it should have picked up the change.

/test e2e-aws-upgrade

smarterclayton:
Hrm.  The cvo in the payload will be the previous version.  So running upgrade test here only tests whether the new operator works, not whether it blocks upgrade.

Next retry we should see “force: true” in desired update.

smarterclayton:
This has definitely exacerbated the observedGeneration bug, or made it happen almost consistently.

smarterclayton:
Will rebase on #176 when it merges and then I think this is ready to go (assuming next upgrade goes green)

smarterclayton:
WTH

/retest

abhinavdahiya:
> WTH
> 
> /retest

Looks like there is a merge conflict @smarterclayton

smarterclayton:
Known flake, force merging.

PR #171, opened 2019-04-22
observability: emit events when waiting for clusteroperator

deads2k:
Knowing when clusteroperators start and stop working is valuable for timing
and general visibility.

Found the lack of these while debugging long upgrade timings.  Since we don't create these, my first attempt disconnects them from the overall registration functions to ease integration.


deads2k:
wow, that doesn't work how I thought it would work

/hold

PR #172, opened 2019-04-22
Bug 1702087: install/0000_00_cluster-version-operator_01_clusteroperator.crd: Failing -> Degraded

wking:
[Fix][1]:

```console
$ oc get clusteroperators kube-apiserver
NAME             VERSION                             AVAILABLE   PROGRESSING   FAILING   SINCE
kube-apiserver   4.1.0-0.nightly-2019-04-22-005054   True        False                   20m
```

when the ClusterOperator only sets `Degraded`.  This should have happened in 94c4e576d1 (#165).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1702087

### DIFF
--- a/install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
+++ b/install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
@@ -16,9 +16,9 @@ spec:
     description: Whether the operator is processing changes.
     name: Progressing
     type: string
-  - JSONPath: .status.conditions[?(@.type=="Failing")].status
-    description: Whether the operator is failing changes.
-    name: Failing
+  - JSONPath: .status.conditions[?(@.type=="Degraded")].status
+    description: Whether the operator is degraded.
+    name: Degraded
     type: string
   - JSONPath: .status.conditions[?(@.type=="Available")].lastTransitionTime
     description: The time the operator's Available status last changed.


wking:
CC @eparis, @smarterclayton, since I'm blaming #165 ;)

smarterclayton:
/lgtm

We need to add a cli e2e for this

wking:
If we want to support both Failing and Degraded during the transition, I'm fine with that.  I'm not quite clear on how though.  We can certainly retain two columns here, and I can update the PR accordingly.  But if we wanted to collapse both into the same column (with the collapsing logic like we have in Go in #165), I'll need some pointers for JSONPath syntax.  Since I have a pod listing handy, I've been experimenting there without success to try to find a boolean or expression supported by `oc`:

```console
$ oc get -o jsonpath='{.items[*].status.conditions[?(@.type == "Initialized" or @.Type == "Ready")]}' pods
error: error parsing jsonpath {.items[*].status.conditions[?(@.type == "Initialized" or @.Type == "Ready")]}, unrecognized character in action: U+003D '='
$ oc get -o jsonpath='{.items[*].status.conditions[?(@.type == "Initialized"), ?(@.Type == "Ready")]}' pods
error: error parsing jsonpath {.items[*].status.conditions[?(@.type == "Initialized"), ?(@.Type == "Ready")]}, unclosed array expect ]
$ oc get -o jsonpath='{.items[*].status.conditions[?(@.type == "Initialized" || @.Type == "Ready")]}' pods
error: error parsing jsonpath {.items[*].status.conditions[?(@.type == "Initialized" || @.Type == "Ready")]}, unrecognized character in action: U+007C '|'
```

wking:
I've pushed 7f743ef -> df68d7b to go with the two-column approach until I get more feedback ;).

smarterclayton:
I don't see a ton of value in two columns.  We won't ship with failing and
it's dead code.  It would be better to open PRs against other components
and fix their "Failing" strings :)

wking:
Ok, back to Failing -> Degraded (instead of supporting both) with df68d7b -> 3142181.

eparis:
/lgtm
/hold cancel
Agree, one column, bugs for anything not reporting

eparis:
Apparently the hold was not just for the 1-2 column question

PR #175, opened 2019-04-26
test: Fix flake in RunGraph, comparing the wrong list (errs vs messages)

smarterclayton:
This caused flakes because we weren't checking against the deterministic
list like we should have been.


smarterclayton:
/retest

>

PR #176, opened 2019-04-26
BUG 1703158 cvo: update the status reporter to preserve the previous Generation 

abhinavdahiya:
Previously the reported passed the Generation from its consumer directly to the update channel, therefore consumers like
syncOnce Downloading release image which were passing 0 Generation a do-not-care would end up updating clusterverison observed generation to 0

consistent reporter now preserves previous generation if consumer passed 0 ie do-not-care generation.
/cc @smarterclayton


smarterclayton:
/override ci/prow/e2e-aws

It has passed multiple times

smarterclayton:
Overriding to unblock the queue

smarterclayton:
Not your fault.  Force merging.

PR #177, opened 2019-04-26
Preserve up to 50 version history entries

smarterclayton:
A future change will perform selective pruning and preserve the initial
version of the code.

@wking to get us out the door, as discussed.

### DIFF
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -110,7 +110,8 @@ func mergeOperatorHistory(config *configv1.ClusterVersion, desired configv1.Upda
 		}
 	}
 
-	pruneStatusHistory(config, 10)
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
I'm okay with this change but I'm not sure what's the reasoning behind this change.
/approve

As a stopgap 10 should be enough until we we ship next Y release. And we'll have some soak time and more time to think about how we want the history to evolve in the future.
Do you anticipate more than 50 upgrades before we move to next Y release?

smarterclayton:
If a customer stayed on 4.1 and we decided not to back port they could have
50 versions by the time they upgraded (we've done 130 versions of some
Y's).  10 was arbitrarily low, 50 is more comfortable, and still not a perf
issue.

PR #178, opened 2019-04-26
Bug 1710172: lib/resourcemerge: Remove containers from existing based on required

vikaschoudhary16:
lib does not remove containers from existing based on `Required` while merging. 
Not sure, if this is by design or it is a bug. This PR is more for a discussion/clarification.

### DIFF
--- a/lib/resourcemerge/core.go
+++ b/lib/resourcemerge/core.go
@@ -24,38 +24,8 @@ func ensurePodTemplateSpec(modified *bool, existing *corev1.PodTemplateSpec, req
 }
 
 func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.PodSpec) {
-	// any container we specify, we require.
-	for _, required := range required.InitContainers {
-		var existingCurr *corev1.Container
-		for j, curr := range existing.InitContainers {
-			if curr.Name == required.Name {
-				existingCurr = &existing.InitContainers[j]
-				break
-			}
-		}
-		if existingCurr == nil {
-			*modified = true
-			existing.Containers = append(existing.InitContainers, corev1.Container{})
-			existingCurr = &existing.InitContainers[len(existing.InitContainers)-1]
-		}
-		ensureContainer(modified, existingCurr, required)
-	}
-
-	for _, required := range required.Containers {
-		var existingCurr *corev1.Container
-		for j, curr := range existing.Containers {
-			if curr.Name == required.Name {
-				existingCurr = &existing.Containers[j]
-				break
-			}
-		}
-		if existingCurr == nil {
-			*modified = true
-			existing.Containers = append(existing.Containers, corev1.Container{})
-			existingCurr = &existing.Containers[len(existing.Containers)-1]
-		}
-		ensureContainer(modified, existingCurr, required)
-	}
+	ensureContainers(modified, &existing.InitContainers, required.InitContainers)
+	ensureContainers(modified, &existing.Containers, required.Containers)
 
 	// any volume we specify, we require.
 	for _, required := range required.Volumes {
@@ -89,6 +59,37 @@ func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.Pod
 	ensureTolerations(modified, &existing.Tolerations, required.Tolerations)
 	setStringIfSet(modified, &existing.PriorityClassName, required.PriorityClassName)
 	setInt32Ptr(modified, &existing.Priority, required.Priority)
+}
+
+func ensureContainers(modified *bool, existing *[]corev1.Container, required []corev1.Container) {
+	var existingCurr *corev1.Container
+	for i, existingContainer := range *existing {
+		for _, requiredContainer := range required {
+			if existingContainer.Name == requiredContainer.Name {
+				existingCurr = &(*existing)[i]
+				ensureContainer(modified, existingCurr, requiredContainer)
+				break
+			}
+		}
+		if existingCurr == nil {
+			*modified = true
+			*existing = append((*existing)[:i], (*existing)[i+1:]...)
+		}
+	}
+
+	for _, requiredContainer := range required {
+		match := false
+		for _, existingContainer := range *existing {
+			if existingContainer.Name == requiredContainer.Name {
+				match = true
+				break
+			}
+		}
+		if !match {
+			*modified = true
+			*existing = append(*existing, requiredContainer)
+		}
+	}
 }
 
 func ensureContainer(modified *bool, existing *corev1.Container, required corev1.Container) {
--- a/lib/resourcemerge/core_test.go
+++ b/lib/resourcemerge/core_test.go
@@ -5,6 +5,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/utils/pointer"
 )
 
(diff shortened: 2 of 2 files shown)


smarterclayton:
This surprises me that we *dont* do this

vikaschoudhary16:
ping @abhinavdahiya @crawford

lib/resourcemerge/core.go
```diff
@@ -56,6 +56,15 @@ func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.Pod
 		}
 		ensureContainer(modified, existingCurr, required)
 	}
+	var requiredContainers []corev1.Container
+	for _, existing := range existing.Containers {
+		for _, required := range required.Containers {
+			if existing.Name == required.Name {
+				requiredContainers = append(requiredContainers, existing)
```
ingvagabund: You might also `continue` here since you already found the match.

lib/resourcemerge/core.go
```diff
@@ -56,6 +56,16 @@ func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.Pod
...
+				continue
+			}
+		}
+	}
+	existing.Containers = requiredContainers
```
wking: How does this (and the existing logic) square with [the PodSpec docs'][1]:

> Containers cannot currently be added or removed.

is there wrapping logic that removes the previous pod if `modified` and adds a new Pod with the new PodSpec?

[1]: https://github.com/openshift/cluster-version-operator/blob/40ba3b6bd45553b6118e250e8a0bdbc02e386ac5/vendor/k8s.io/api/core/v1/types.go#L2767

vikaschoudhary16: A new pod will get created as part of deployment redeploy. So container will not be deleted from existing pod.

abhinavdahiya: The entire work of ensure containers done https://github.com/openshift/cluster-version-operator/pull/178/files#diff-f7675c7b39553e431cdc6c3ba8f5c482R44 will be unused due to this.

Also this will keep forcing update of containers because required never has any server side defaults.

What you really want is remove any containers that are not part of required.

abhinavdahiya: Ahh, nvm it sets the existing correctly. I missed the inner loop

abhinavdahiya:
Can would add a unit test for this new behavior. After that we can merge this

lib/resourcemerge/core.go
vikaschoudhary16: > The entire work of ensure containers done https://github.com/openshift/cluster-version-operator/pull/178/files#diff-f7675c7b39553e431cdc6c3ba8f5c482R44 will be unused due to this.

Thanks for pointing out. Will move this logic before ensureContainers to save cycles.

> Also this will keep forcing update of containers because required never has any server side defaults.

I am assuming you meant `server side default init containers`. If correct, we are removing init containers based on required. Is this common to have server side default regular containers?

wking: > If correct, we are not removing init containers based on required.

Should we have a single `ensureContainers` that we run on both `InitContainers` and `Containers`?

lib/resourcemerge/core_test.go
```diff
@@ -16,14 +16,54 @@ func TestEnsurePodSpec(t *testing.T) {
...
+			existing: corev1.PodSpec{
+				Containers: []corev1.Container{
+					corev1.Container{Name: "test"}}},
+			input: corev1.PodSpec{},
+
+			expectedModified: true,
+			expected:         corev1.PodSpec{},
+		},
+		{
+			name: "remove regular containers from existing but dont remove init containers",
```
wking: I'm [still][1] not clear on why we want this.  If "modified" means "[launch a replacement Pod][2]", then I'd expect we want to set `modified` and remove entries from `InitContainers` that weren't in the new required PodSpec.  I've floated 8c83b0c9199414128a2e8e63b095be4ec0099639 with an implementation that you can squash into your branch if that makes sense.

[1]: https://github.com/openshift/cluster-version-operator/pull/178#discussion_r291668007
[2]: https://github.com/openshift/cluster-version-operator/pull/178#discussion_r291457813

vikaschoudhary16: Was trying to play safe and do minimal changes which are required for the issue that i have been trying to fix.
Updated with changes from your commit. Thanks a lot!

wking:
/approve
/hold

I'll wait for @abhinavdahiya to confirm that we want the `InitContainers` merging too, but  4f4a65a492 looks good to me.

lib/resourcemerge/core_test.go
```diff
@@ -34,8 +119,16 @@ func TestEnsurePodSpec(t *testing.T) {
 			}
 
 			if !equality.Semantic.DeepEqual(test.existing, test.expected) {
-				t.Errorf("mismatch PodSpec got: %v want: %v", test.existing, test.expected)
+				t.Errorf("mismatch PodSpec got:\n%s\nwant:\n%s", yamlOrDie(test.existing), yamlOrDie(test.expected))
```
abhinavdahiya: You can use https://godoc.org/k8s.io/apimachinery/pkg/util/diff

wking:
/hold

Trying to figure out why this is failing.

lib/resourcemerge/core_test.go
wking: +1 to using the k8s lib (which we already vendor).

vikaschoudhary16:
@wking @abhinavdahiya Following seems to be the problem:
```
  "type": "RetrievedUpdates",
        "status": "False",
        "lastTransitionTime": "2019-06-13T14:13:27Z",
        "reason": "RemoteFailed",
        "message": "Unable to retrieve available updates: currently installed version 0.0.1-2019-06-13-135753 not found in the \"stable-4.2\" channel"
```
Does not seem to be related to the changes. Thoughts?

vikaschoudhary16:
integration job is failing unrelated to the PR changes, https://github.com/openshift/cluster-version-operator/issues/211

lib/resourcemerge/core_test.go
```diff
@@ -34,8 +120,16 @@ func TestEnsurePodSpec(t *testing.T) {
...
 			if !equality.Semantic.DeepEqual(test.existing, test.expected) {
-				t.Errorf("mismatch PodSpec got: %v want: %v", test.existing, test.expected)
+				t.Errorf("unexpected: %s", diff.ObjectReflectDiff(test.expected, test.existing))
 			}
 		})
 	}
 }
+
+func yamlOrDie(data interface{}) string {
+	bytes, err := yaml.Marshal(data)
```
abhinavdahiya: This function is no longer required. Right?

lib/resourcemerge/core.go
```diff
@@ -91,6 +61,38 @@ func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.Pod
...
+			*existing = append((*existing)[:i], (*existing)[i+1:]...)
+		}
+	}
+
+	for _, requiredContainer := range required {
+		var existingCurr *corev1.Container
+		for j, existingContainer := range *existing {
+			if existingContainer.Name == requiredContainer.Name {
+				existingCurr = &(*existing)[j]
+				ensureContainer(modified, existingCurr, requiredContainer)
```
abhinavdahiya: Hmm, Won't these containers already be ensured from previous loop??

vikaschoudhary16: you are right. But above it dint work because of golang behavior where `requireContainer` is a local variable and changes to that does not update the slice `required`. Will fix, thanks!

abhinavdahiya:
/approve

Working on getting integration tests fixed. so keeping the hold.

PR #181, opened 2019-04-30
Update version-patch.yaml in clusterversion.md

sallyom:
Now that we have no default overrides, the version-patch has to add the
overrides field.  According to patch docs (that I found http://jsonpatch.com/#add) the Add Op
1) If the path includes an array index, the new value is added to the entity at the specified index
2) If the path doesn't exist, a new member is created
3) If the path does exist, the value is replaced

~This worked for me~ Nope a second add squashed my first... 
Now that I've opened, I'm holding until I fix that ^^

### DIFF
--- a/docs/dev/clusterversion.md
+++ b/docs/dev/clusterversion.md
@@ -50,15 +50,32 @@ kind: DaemonSet
 metadata:
   name: cluster-network-operator
   namespace: openshift-cluster-network-operator
-$ cat <<EOF >version-patch.yaml
-> - op: add
->   path: /spec/overrides/-
->   value:
->     kind: DaemonSet
->     name: cluster-network-operator
->     namespace: openshift-cluster-network-operator
->     unmanaged: true
-> EOF
+```
+If there are currently no other overrides configured:
+```console
+$ cat <<EOF >version-patch-first-override.yaml
+- op: add
+  path: /spec/overrides
+  value:
+  - kind: DaemonSet
+    name: cluster-network-operator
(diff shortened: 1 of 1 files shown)


docs/dev/clusterversion.md
```diff
@@ -52,7 +52,7 @@ metadata:
   namespace: openshift-cluster-network-operator
 $ cat <<EOF >version-patch.yaml
 > - op: add
->   path: /spec/overrides/-
+>   path: /spec/overrides
```
sallyom: This allows addition of overrides field with the value below, however, any addition of elements to the overrides overwrites rather than appends.  Unless I'm doing something wrong, there's a bug in json-patch/jsonpatch?

docs/dev/clusterversion.md
```diff
@@ -50,15 +50,32 @@ kind: DaemonSet
...
+- op: add
+  path: /spec/overrides/-
+  value:
+    kind: DaemonSet
+    name: cluster-network-operator
+    namespace: openshift-cluster-network-operator
+    unmanaged: true
+EOF
+```
+```console
```
sallyom: From my experience, this documentation will lead a user to successfully adding unmanaged objects.  This is best documentation I can do.  Note, when adding an element to already existing array, no hyphen before `kind` and add the `/-` to path.  Both are necessary.  When adding the overrides array, you need the hyphen and can't provide the `/-` to path. The path I understand, but the hyphen? :woman_shrugging:

PR #182, opened 2019-05-03
Fix tolerations

ravisantoshgudimetla:
In 4.1, we have taint-based evictions enabled. This changes logic in the node lifecycle controller in the kube-controller-manager significantly.

Historically, the node lifecycle controller would directly evict pods from node that had a Ready condition of False or Unknown after a pod eviction timeout set by the --pod-eviction-timeout flag on the kube-controller-manager. This setting applied to all pods cluster-wide. The default was 5m.

With taint-based evictions, all the node lifecycle controller does is taint the node with node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taint with NoExecute effect. This would normally result in the immediate eviction of all pods that don't tolerate those taints, breaking the old behavior with pod eviction timeouts. Enter DefaultTolerationSeconds mutating admission plugin.

https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go

This plugin will add a default NoExecute toleration for node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taints with a tolerationSeconds of 5m (300s) as long as no such toleration is already specified in the pod spec. This restores the old pod eviction tiemout behavior.

One of the intended effects of this change is the make the pod eviction timeout a pod-level property. Different applications require different timeouts depending on their design and having it controlled at a cluster level before was not optimal. The side-effect is that we won't allow pods to be scheduled onto nodes that have disk, memory, cpu pressure.

The DefaultTolerationSeconds plugin has a flag that allows adjusting the defaults for tolerationSeconds
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/admission/defaulttolerationseconds/admission.go#L34-L40

We might expose this tunable for user control in the future and do not want the cluster control plane components to be subject to it.

Thus, this PR explictly defines a NoExecute toleration for node.kubernetes.io/unreachable and/or node.kubernetes.io/not-ready taints with a tolerationSeconds generically appropriate for cluster components.

Once these changes are in across all components, this e2e will enforce it in the future
openshift/origin#22752

Please refer to this doc, if you have questions around what tolerations you can have:

https://docs.google.com/document/d/1W449BfB5la9NC7pcDxkovgzlBlHMy5Lkj0-WXWjdiTU/edit#

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -58,7 +58,10 @@ spec:
         effect: "NoSchedule"
       - key: "node.kubernetes.io/network-unavailable"
         operator: Exists
-        effect: "NoSchedule"  
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoSchedule" 
       - key: "node.kubernetes.io/unreachable"
         operator: "Exists"
         effect: "NoExecute"
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -50,7 +50,17 @@ spec:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
       tolerations:
-        - operator: Exists
+      - key: node-role.kubernetes.io/master  
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120 
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120 
       volumes:
         - name: etc-ssl-certs
           hostPath:
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -50,9 +50,15 @@ spec:
         node-role.kubernetes.io/master: ""
       priorityClassName: "system-cluster-critical"
       tolerations:
-      - key: node-role.kubernetes.io/master  
+      - key: "node-role.kubernetes.io/master"
         operator: Exists
         effect: "NoSchedule"
+      - key: "node.kubernetes.io/unschedulable"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/network-unavailable"
+        operator: Exists
+        effect: "NoSchedule"  
       - key: "node.kubernetes.io/unreachable"
         operator: "Exists"
         effect: "NoExecute"


install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -50,7 +50,17 @@ spec:
...
       tolerations:
-        - operator: Exists
+      - key: node-role.kubernetes.io/master  
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120 
+      - key: "node.kubernetes.io/not-ready"
```
abhinavdahiya: CVO must be scheduled on master so that it can drive installation even when it is `NotReady` ie not schedulable because it creates operators that create network and apiserver.

ravisantoshgudimetla: To be clear, are you telling we should tolerate both `unschedulable` and `NotReady` taints?

abhinavdahiya: I want to say yes. ;)

We want to make sure CVO gets scheduled even when node is not ready because networking isn't setup yet.

smarterclayton: The cvo must always get a chance to land on any master.

ravisantoshgudimetla: Including `memory` and `disk` pressures? I assume not..

ravisantoshgudimetla: Updated with `unschedulable` and `network-unavailable` tolerations

abhinavdahiya: whay about not-ready NoSchedule ?

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -50,7 +50,26 @@ spec:
...
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/unschedulable"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/network-unavailable"
+        operator: Exists
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/not-ready"
```
ravisantoshgudimetla: Added it here.

abhinavdahiya:
/approve

will let @smarterclayton take a look too.

ravisantoshgudimetla:
Flake:

`May 07 02:07:38.234 W ns/openshift-machine-config-operator pod/etcd-quorum-guard-855976c44d-mgm6x 0/6 nodes are available: 2 node(s) didn't match node selector, 2 node(s) didn't match pod affinity/anti-affinity, 2 node(s) didn't satisfy existing pods anti-affinity rules, 2 node(s) were unschedulable. (109 times)`

 /cc @RobertKrawitz

ravisantoshgudimetla:
`7f4-jkjht Stopping container packageserver
May 07 04:08:28.391 W ns/openshift-operator-lifecycle-manager pod/packageserver-78f48b54d6-mx9dw Readiness probe failed: Get https://10.130.0.86:5443/healthz: dial tcp 10.130.0.86:5443: connect: connection refused (2 times)
May 07 04:08:29.157 W ns/openshift-etcd pod/etcd-member-ip-10-0-160-97.ec2.internal node/ip-10-0-160-97.ec2.internal container=etcd-metrics container restarted
May 07 04:08:29.598 I ns/openshift-kube-apiserver-operator deployment/kube-apiserver-operator Status for clusteroperator/kube-apiserver changed: Degraded message changed from "StaticPodsDegraded: nodes/ip-10-0-160-97.ec2.internal pods/kube-apiserver-ip-10-0-160-97.ec2.internal container=\"kube-apiserver-7\" is not ready" to "" (2 times)
`

Seems an issue with etcd communication for apiserver.

/retest

PR #183, opened 2019-05-03
Bug 1706202: Use klog directly or no line info is printed

smarterclayton:
The glog wrapper of klog has a fatal flaw - it doesn't print the
correct line number of a message. This makes debugging almost impossible
in high level debug.

Fix this by using the method directly.


smarterclayton:
SCC change broke master-sidecar template

/retest

PR #186, opened 2019-05-15
Bug 1665011: updatepayload: prune older jobs to maintain limit.

jstuever:
This change prunes the jobs in the openshift-cluster-version namespace
before creating a new one to download new update payloads. First, it
sorts them in chronological order and then calls delete for all except
the newest two.

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -7,18 +7,21 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
+
+	"github.com/pkg/errors"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	randutil "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 
-	"k8s.io/klog"
 	configv1 "github.com/openshift/api/config/v1"
-
 	"github.com/openshift/cluster-version-operator/lib/resourcebuilder"
 	"github.com/openshift/cluster-version-operator/pkg/payload"
 	"github.com/openshift/cluster-version-operator/pkg/verify"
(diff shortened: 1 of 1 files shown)


pkg/cvo/updatepayload.go
```diff
@@ -189,13 +190,40 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 		},
 	}
 
+	// Prune older jobs while gracefully ignoring errors.
+	r.pruneJobs()
```
abhinavdahiya: I think this call site should be the one ignoring the errors rather than ignoring it in the function itself.

abhinavdahiya: Question:
would be want the function to return a TooManyJobs error type so that we can fail hard when we see that error because that would be indicator of an actual error?

abhinavdahiya:
/cc @smarterclayton 
for sanity check.

pkg/cvo/updatepayload.go
jstuever: Is that not what it is already doing? It returns errors in the function, but doesn't even bother with them when called here.

abhinavdahiya: https://github.com/openshift/cluster-version-operator/pull/186/files/53e5b922c64ac7f534f5df84dcaada10b364db22#diff-d540a41404f678a1c438f4c1e5b92a87R221 is not returning any errors, but warning?

^^ should be `errors.Wrapf(err, "failed to delete job %q", job.Name)`

also we might be in better shape if we try to delete all jobs and aggregate those errors before returning.

pkg/cvo/updatepayload.go
```diff
@@ -189,13 +190,40 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
...
 	return resourcebuilder.WaitForJobCompletion(ctx, r.kubeClient.BatchV1(), job)
 }
 
+// pruneJobs deletes older jobs in the namespace while keeping the latest.
+func (r *payloadRetriever) pruneJobs() error {
+	jobs, err := r.kubeClient.BatchV1().Jobs(r.namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	if len(jobs.Items) <= 2 {
```
smarterclayton: Make this a parameter and use it in both places

pkg/cvo/updatepayload.go
```diff
@@ -189,13 +190,40 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
...
+		return nil
+	}
+
+	// Sort jobs by StartTime to determine the oldest
+	sort.Slice(jobs.Items, func(i, j int) bool {
+		return jobs.Items[i].Status.StartTime.Before(jobs.Items[j].Status.StartTime)
```
smarterclayton: Is start time guaranteed to be non empty and non nil?  Unstarted new jobs should be kept, not deleted

pkg/cvo/updatepayload.go
jstuever: @abhinavdahiya Done, PTAL.

pkg/cvo/updatepayload.go
jstuever: Good catch. New jobs have a nil value immediately after they are created. Code is modified to no longer delete jobs that are new or still active.

pkg/cvo/updatepayload.go
```diff
@@ -189,13 +193,76 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
...
+		return err
+	}
+	if len(jobs.Items) <= retain {
+		return nil
+	}
+
+	// Select jobs to be deleted
+	var deleteJobs []batchv1.Job
+	for _, job := range jobs.Items {
+		switch {
```
abhinavdahiya:
```console
$ gofmt -d -l -s pkg/cvo/updatepayload.go
pkg/cvo/updatepayload.go
diff -u pkg/cvo/updatepayload.go.orig pkg/cvo/updatepayload.go
--- pkg/cvo/updatepayload.go.orig       2019-05-20 13:03:00.391730348 -0700
+++ pkg/cvo/updatepayload.go    2019-05-20 13:03:00.391730348 -0700
@@ -20,8 +20,8 @@
        "k8s.io/client-go/kubernetes"
        "k8s.io/utils/pointer"

-       "k8s.io/klog"
        configv1 "github.com/openshift/api/config/v1"
+       "k8s.io/klog"

        "github.com/openshift/cluster-version-operator/lib/resourcebuilder"
@@ -221,17 +221,17 @@
        var deleteJobs []batchv1.Job
        for _, job := range jobs.Items {
                switch {
-                       // Ignore jobs not begining with operatorName
-                       case !strings.HasPrefix(job.Name, r.operatorName + "-"):
-                               break
-                       // Ignore jobs that have not yet started
-                       case job.Status.StartTime == nil:
-                               break
-                       // Ignore jobs that are still active
-                       case job.Status.Active == 1:
-                               break
-                       default:
-                               deleteJobs = append(deleteJobs, job)
+               // Ignore jobs not begining with operatorName
+               case !strings.HasPrefix(job.Name, r.operatorName+"-"):
+                       break
+               // Ignore jobs that have not yet started
+               case job.Status.StartTime == nil:
+                       break
+               // Ignore jobs that are still active
+               case job.Status.Active == 1:
+                       break
+               default:
+                       deleteJobs = append(deleteJobs, job)
                }
        }
        if len(deleteJobs) <= retain {
```

abhinavdahiya:
ping @smarterclayton for another round of review.

abhinavdahiya:
> ping @smarterclayton for another round of review.

I think 2 days is good enough notice ;) we can iterate on this we people find bugs not obvious right now.

/lgtm

PR #187, opened 2019-05-15
Bug 1702582: Log DeploymentConditions reason and message when Deployment is failing

jcpowermac:
Displays the Deployment Condition reason and message when
the Deployment is not progressing or there is a Replica failure.

Bug 1702582 - https://bugzilla.redhat.com/show_bug.cgi?id=1702582

Example:
```
E0515 20:17:10.936566       1 apps.go:82] Deployment kube-apiserver-operator is not ready. condition: (reason: ProgressDeadlineExceeded, message: ReplicaSet "kube-apiserver-operator-7dfb6ddb9d" has timed out progressing.)
I0515 20:17:10.936620       1 apps.go:86] Deployment kube-apiserver-operator is not ready. status: (replicas: 2, updated: 1, ready: 1, unavailable: 1)
```

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"strings"
+
 	"k8s.io/klog"
 
 	"github.com/openshift/cluster-version-operator/lib"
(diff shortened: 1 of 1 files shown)


lib/resourcebuilder/apps.go
```diff
@@ -74,6 +75,14 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
 		if d.Generation <= d.Status.ObservedGeneration && d.Status.UpdatedReplicas == d.Status.Replicas && d.Status.UnavailableReplicas == 0 {
 			return true, nil
 		}
+		if conditionsLength > 0 {
```
abhinavdahiya: Can we aggregate error from all `False` conditions?

jcpowermac: @abhinavdahiya updated per our conversation earlier today.

[DeploymentConditionType](https://github.com/kubernetes/kubernetes/blob/39fc3c49564aa346a39202c1a5b84a9c150ca43b/pkg/apis/apps/types.go#L454-L467)

abhinavdahiya: nit: Can we aggregate the status of failure to a single error Print.
It can be multi-line.

lib/resourcebuilder/apps.go
```diff
@@ -56,6 +58,7 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.DeploymentsGetter, deployment *appsv1.Deployment) error {
 	return wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
 		d, err := client.Deployments(deployment.Namespace).Get(deployment.Name, metav1.GetOptions{})
+		deploymentConditions := make([]string, 0, len(d.Status.Conditions))
```
abhinavdahiya: d can be nil when err is not nil... move this to https://github.com/openshift/cluster-version-operator/pull/187/files#diff-20dbece9affd4bfefb9de7caca6bc037R80

abhinavdahiya: cc @jcpowermac

jcpowermac: @abhinavdahiya thanks! fixed

PR #189, opened 2019-05-19
ClusterVersion version column should be last completed

smarterclayton:
Use history instead of desired version in the custom columns, since desired state can be set to empty when a user provides invalid input.  The first "Completed" entry in history is our "current" version - if
one doesn't exist, we have "no version".

Reproducer:

1. `oc adm upgrade --to-image=X` (triggering signature verification error)
2. `oc get clusterversion` has an empty VERSION column


wking:
> ... if one doesn't exist, we have "no version".

We don't want to fall back to "attempting `{target}`" or some such if we have no version yet?  I'm worried about making the output less useful during the initial install.

PR #191, opened 2019-05-19
conditions: Use a consistent constant for the Failing condition

smarterclayton:
This needs to be moved back into openshift/api since it is now part
of our public API, but for now ensure it is consistently used.


wking:
I'm still not clear on why the cluster-version operator's needs a `Failing` with different semantics than the second-level operator's `Degraded`.  All the stuff from [the `Degraded` docs][1] seems like it applies here too.

[1]: https://github.com/openshift/cluster-version-operator/blame/2b1eb96bc6c4d4e648e9bcb98183cb7af09e079f/docs/dev/clusteroperator.md#L81-L96

smarterclayton:
/cherry-pick release-4.1

Prerequisite for #197

PR #192, opened 2019-05-19
status: Hide generic operator status in favor of more specific errors

smarterclayton:
When upgrading and a sync loop exits, we often see:

* Operator A not upgraded yet
* Operator B not upgraded yet
* Operator C not upgraded yet
* Some more specific error
* Operator D not upgraded yet

Since the not available messages carry little value and are expected
when we are processing parallel output, if we have one or more errors
during sync that are not the generic error, only report those. This
focuses the user's attention on the actual problem, rather than
distracting them with noise. In the ideal case, the user immediately
sees "router is totally broken" vs "a, b, c, d, e, router is broken, f".

E.g. https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/1724/

```
Multiple errors are preventing progress:
* Cluster operator authentication is still updating: upgrading oauth-openshift from 4.2.0-0.ci-2019-05-18-031713_openshift to 4.2.0-0.ci-2019-05-18-224753_openshift
* Cluster operator cluster-autoscaler is still updating
* Cluster operator monitoring is still updating
* Cluster operator openshift-controller-manager is still updating
* Cluster operator service-catalog-controller-manager is still updating
* Could not update deployment "openshift-cloud-credential-operator/cloud-credential-operator" (93 of 350)
...
```

The "still updating" errors are just noise.

/cherrypick release-4.1

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -10,9 +10,9 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/klog"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/time/rate"
+	"k8s.io/klog"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -751,12 +751,38 @@ func summarizeTaskGraphErrors(errs []error) error {
...
 	}
 	return newMultipleError(errs)
 }
 
+// filterErrors returns only the errors in errs which are false for all fns.
+func filterErrors(errs []error, fns ...func(err error) bool) []error {
+	var filtered []error
+Next:
+	for _, err := range errs {
+		for _, fn := range fns {
```
abhinavdahiya: nit:
can we replace this `break JUMP` with something that's more easier to follow..

like
```
matches := false 
loop fns as fn
    if fn == true matches <= true
if !matches append
```

smarterclayton: Yeah

smarterclayton:
Hrm, this doesn't seem to be working like I expected

smarterclayton:
Oh wait, nm, that was another PR.

pkg/cvo/sync_worker.go
smarterclayton: Fixed

pkg/cvo/sync_worker.go
```diff
@@ -751,12 +751,44 @@ func summarizeTaskGraphErrors(errs []error) error {
...
 	}
+	// hide the generic "not available yet" when there are more specific errors present
+	if filtered := filterErrors(errs, isClusterOperatorNotAvailable); len(filtered) > 0 {
+		return newMultipleError(filtered)
+	}
+	// if we're only waiting for operators, condense the error down to a singleton
 	if err := newClusterOperatorsNotAvailable(errs); err != nil {
 		return err
 	}
 	return newMultipleError(errs)
```
wking: This is the only `newMultipleErrors` consumer.  It feels strange to have some preprocessing locally; can we add this new logic inside `newMultipleErrors` (and drop the local collapse, since `newMultipleErrors` already does that internally)?

smarterclayton: It’s three different flows - I expect there to be more in the future.  MultipleErrors is for when we don’t know anything

smarterclayton:
Reapplying label (here's where we are after fixing a couple of the ugly):

```
May 21 09:48:03.655: INFO: Cluster version operator acknowledged upgrade request
May 21 09:55:13.785: INFO: cluster upgrade is failing: Multiple errors are preventing progress:
* Cluster operator authentication is still updating
* Cluster operator cluster-autoscaler is still updating
* Cluster operator marketplace is still updating
* Cluster operator monitoring is still updating
* Cluster operator openshift-controller-manager is still updating
* Cluster operator service-catalog-apiserver is still updating
* Cluster operator service-catalog-controller-manager is still updating
* Cluster operator storage is still updating
* Could not update deployment "openshift-cluster-samples-operator/cluster-samples-operator" (185 of 350)
* Could not update deployment "openshift-console/downloads" (237 of 350)
* Could not update deployment "openshift-operator-lifecycle-manager/catalog-operator" (254 of 350)
* Could not update deployment "openshift-service-ca-operator/service-ca-operator" (290 of 350)
```

PR #185, opened 2019-05-14
Bug 1825003: pkg/payload/task: Include name/reason in "unknown error" message

wking:
Because this makes it easier to grow the `SummaryForReason` logic.  The previous, generic error message is not very useful (e.g. as seen [here][1]).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1710012#c0

PR #194, opened 2019-05-20
pkg/payload: Initial work on update -> reconciliation wording

wking:
A bunch of small pivots to start moving the package API and error messages away from "update" phrasing and towards generic reconiciliation phrasing.  When complete, this transition will reduce confusion like [rhbz#1711964][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1711964

### DIFF
--- a/pkg/payload/error.go
+++ b/pkg/payload/error.go
@@ -1,0 +1,112 @@
+package payload
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+)
+
+// Error is a wrapper for errors that occur during a payload sync.
+type Error struct {
+	Nested  error
+	Reason  string
+	Message string
+	Name    string
+
+	Task *Task
+}
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
+func (e *Error) Cause() error {
+	return e.Nested
+}
+
+func (e *Error) Summary() string {
+	switch e.Reason {
+
+	// likely temporary errors
+	case "UpdatePayloadResourceNotFound", "UpdatePayloadResourceConflict":
+		return "some resources could not be updated"
+	case "UpdatePayloadClusterDown":
+		return "the control plane is down or not responding"
+	case "UpdatePayloadClusterError":
+		return "the control plane is reporting an internal error"
+	case "UpdatePayloadClusterOverloaded":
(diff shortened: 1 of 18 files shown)


wking:
CC @abhinavdahiya, @smarterclayton 

Not sure if there's anything I can do to help with review, this is a fair bit bigger than #185, and that's been pretty quiet ;).

wking:
Rebased around #190 with 86704fa -> 13630b3.

abhinavdahiya:
I think the CVO reacts to Updates and not Payloads. So I don't completely agree with the move.

like payload.UpdateError - > payload.Error , personally i think former is more clearer.
like payload.LoadUpdate is more analogous to what CVO wants done, loading update not load payload...
adding summary etc function to Error type, personally i think former setup is more clearer.

_haven't looked at the change in depth_

wking:
> I think the CVO reacts to Updates and not Payloads.

The CVO has [three modes][1], all of which run reconciling sync loops, and only one of which is an update.

> like payload.UpdateError - > payload.Error , personally i think former is more clearer

But we are generating these errors outside of updates, as seen in [rhbz#1711964][2].  These issues are payload-specific (which is already covered by the package scoping), but they aren't all upgrade-specific.

> adding summary etc function to Error type, personally i think former setup is more clearer.

Can you put your finger on why?  Getting a short text rendering of an instance of the payload package's error type seems like a good fit for "error type method" to me (vs. "grab a separate function from the payload package and pass it portions of the error type").

[1]: https://github.com/openshift/cluster-version-operator/blob/a17c59a1fd6c03f9af40082c380888523d60820b/pkg/payload/payload.go#L30-L58
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1711964

wking:
Pushed 13630b3 -> 27ee47d fixing the unit tests.

wking:
Rebased around #191 with 27ee47d -> 5242487.

smarterclayton:
We should definitely not use the work payload anywhere - update or version are acceptable, but payload is not part of our public lexicon

pkg/payload/error.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
+func (e *Error) Cause() error {
+	return e.Nested
+}
+
+func (e *Error) Summary() string {
```
smarterclayton: This was separate because it was a specific domain error being applied - the error message for API short status is not the same as the message in other contexts.  Summary isn’t a property of the error, it is an interpretation of it.

So would prefer not to do this part.

PR #195, opened 2019-05-22
Fix Deployments Apply reporting and don't suppress errors

tnozicka:
Should show the real reason we see
`* Could not update deployment "openshift-console/downloads" (237 of 350)`
and similar.

/cc @smarterclayton


lib/resourcebuilder/apps.go
```diff
@@ -77,6 +80,18 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
 	}, ctx.Done())
+
+	if err == wait.ErrWaitTimeout {
+		message := fmt.Sprintf("Deployment %s failed to become ready in time.", d.Name)
+		if d != nil {
+			message +=  fmt.Sprintf("status: (replicas: %d, updated: %d, ready: %d, unavailable: %d)", d.Status.Replicas, d.Status.UpdatedReplicas, d.Status.ReadyReplicas, d.Status.UnavailableReplicas)
+		}
+		klog.Warningf(message)
+
+		return errors.New(message)
```
smarterclayton: unfortunately to do this you need to return an update error and have handling that indicates the why.  you can't use errors.New()

smarterclayton:
/hold

I had some questions about why this was necessary off-line

lib/resourcebuilder/apps.go
```diff
@@ -77,6 +81,21 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
 		klog.V(4).Infof("Deployment %s is not ready. status: (replicas: %d, updated: %d, ready: %d, unavailable: %d)", d.Name, d.Status.Replicas, d.Status.UpdatedReplicas, d.Status.ReadyReplicas, d.Status.UnavailableReplicas)
 		return false, nil
 	}, ctx.Done())
+
+	if err == wait.ErrWaitTimeout {
```
smarterclayton: `d != nil && err == wait.ErrWaitTimeout`

lib/resourcebuilder/apps.go
```diff
@@ -77,6 +81,21 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
+		message := fmt.Sprintf("Deployment %s failed to become ready in time.", d.Name)
```
smarterclayton: We don't use terminology that indicates "past" - we would say "Deployment is not yet available: ..."

lib/resourcebuilder/apps.go
```diff
@@ -77,6 +81,21 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
+		}
+
+		return &payload.UpdateError{
+			Nested:  errors.New(message),
+			Reason:  "UpdatePayloadDeploymentStuck",
+			Message: fmt.Sprintf("Could not update deployment %s: deployment is not yet available”", deployment.Name),
+		}
+	}
+
+	return err
```
smarterclayton: add a unit test on this that tests the different type of errors

lib/resourcebuilder/apps.go
```diff
@@ -54,9 +56,11 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 	return nil
 }
 func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.DeploymentsGetter, deployment *appsv1.Deployment) error {
-	return wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
-		d, err := client.Deployments(deployment.Namespace).Get(deployment.Name, metav1.GetOptions{})
-		if errors.IsNotFound(err) {
+	var d *appsv1.Deployment
```
abhinavdahiya: instead of keeping global d. Please use lastErr that stores the last known error in the Poll loop. keeping the error handling confined... example operatorstatus.go

PR #196, opened 2019-05-31
pkg/verify: Test an armored, multi-key keyring

wking:
New keyring generated with:

```console
$ cat redhat.txt simple.txt | gpg --dearmor | gpg --enarmor >combined.txt
$ sed -i 's/ARMORED FILE/PUBLIC KEY BLOCK/' combined.txt
```

where the `sed` avoids:

```console
$ go test .
--- FAIL: Test_releaseVerifier_Verify (0.00s)
		verify_test.go:45: openpgp: invalid argument: expected public or private key block, got: PGP ARMORED FILE
FAIL
FAIL		github.com/openshift/cluster-version-operator/pkg/verify				0.014s
```

because [Go's parser requires the well-known labels][1].

[1]: https://github.com/golang/crypto/blob/20be4c3c3ed52bfccdb2d59a412ee1a936d175a7/openpgp/keys.go#L244


pkg/verify/testdata/keyrings/combined.txt
```diff
@@ -0,0 +1,56 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
+Version: GnuPG v2.0.22 (GNU/Linux)
```
eparis: I don't see this line in the actual container being created (though I do see the "Comment" line)

wking: Differen't GnuPG versions?  :man_shrugging:

PR #197, opened 2019-06-05
Bug 1717619: metrics: Report reason as a label on the conditions metric

smarterclayton:
Conditions are read by telemetry and the reason incentivizes teams
to have good reasons on their operators. Should not increase
cardinality significantly, but instead provide better insight.

/cherry-pick release-4.1


smarterclayton:
@abhinavdahiya we need this in 4.1.2, could you get someone to backport today?

Issue #99, opened 2019-01-23
Question:  auto update enabled


jatanmalde:
Question: How does the CVO monitors for new image when auto-update is enabled?

How does it looks for new ocp-release image when pushed to the container catalog(which would be used in future for ocp 4.0 if I am not wrong)?

Let me know your thoughts on it.

jatanmalde:
Hello,

Any updates on this?

wking:
> How does it looks for new ocp-release image...

[ClusterVersion's `Upstream` and `Channel` properties][1] tell it where to find [Cincinnati][2].  There's currently no knob for making those automatic outside of [the `--enable-auto-update` option][3], though.  We'll probably need to expand the API to give a ClusterVersion spec knob for that.

[1]: https://github.com/openshift/api/blob/2963b4d8c99701a12f3a18c2d53c28b40408d213/config/v1/types_cluster_version.go#L53-L63
[2]: https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md
[3]: https://github.com/openshift/cluster-version-operator/blob/01adf75393b6e11d3d8c98ecfeeebd3feb998a6c/install/0000_00_cluster-version-operator_03_deployment.yaml#L25

wking:
/lifecycle frozen

Still waiting on buy-in in openshift/enhancements#124.

PR #198, opened 2019-06-06
Update Replica count too while ensuring deployment

vikaschoudhary16:
SSIA

### DIFF
--- a/lib/resourcemerge/apps.go
+++ b/lib/resourcemerge/apps.go
@@ -10,7 +10,12 @@ import (
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if existing.Spec.Selector == nil {
+	if required.Spec.Replicas != nil && required.Spec.Replicas != existing.Spec.Replicas {
+		*modified = true
+		existing.Spec.Replicas = required.Spec.Replicas
+	}
+
+	if existing.Spec.Selector == nil && required.Spec.Selector != nil {
 		*modified = true
 		existing.Spec.Selector = required.Spec.Selector
 	}
--- a/lib/resourcemerge/apps_test.go
+++ b/lib/resourcemerge/apps_test.go
@@ -1,0 +1,79 @@
+package resourcemerge
+
+import (
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+func TestEnsureDeployment(t *testing.T) {
+	replicas := int32(2)
+	expectedReplicas := int32(2)
+	labelSelector := metav1.LabelSelector{}
+	tests := []struct {
+		name     string
+		existing appsv1.Deployment
+		required appsv1.Deployment
+
+		expectedModified bool
+		expected         appsv1.Deployment
+	}{
+		{
+			name: "different replica count",
+			existing: appsv1.Deployment{
+				Spec: appsv1.DeploymentSpec{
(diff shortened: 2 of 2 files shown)


lib/resourcemerge/apps.go
```diff
@@ -10,6 +10,11 @@ import (
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
+	if required.Spec.Replicas != nil && required.Spec.Replicas != existing.Spec.Replicas {
```
abhinavdahiya: Why do you think this is important?

vikaschoudhary16: We are using this lib in machine-api-operator. When MAO reconciles its owned deployments, it is desired that replica count should also be reconciled. https://jira.coreos.com/browse/CLOUD-484

lib/resourcemerge/apps_test.go
```diff
@@ -0,0 +1,78 @@
...
+	expectedReplicas := int32(2)
+	labelSelector := metav1.LabelSelector{}
+	tests := []struct {
+		name     string
+		existing appsv1.Deployment
+		required appsv1.Deployment
+
+		expectedModified bool
+		expected         appsv1.Deployment
+	}{{
```
abhinavdahiya: this is different in format.. than others

wking:
`replicas` seems like something that we'll want to be able to manage via release images.  In fact, the fewer knobs we leave *unmanaged*, the happier I'll be, so I think the burden of proof should be on "why *shouldn't* we manage this?".

/approve

I'll wait for @abhinavdahiya to weigh in on whether that makes sense to him and whether he wants [a different format for the unit tests][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/198#discussion_r292244502

lib/resourcemerge/apps_test.go
vikaschoudhary16: Can you please explain a bit what you mean by "other" format? I tried to follow the format of the tests recently got added in[ #99](https://github.com/openshift/cluster-version-operator/pull/199/files#diff-415c13f11ffc32696c5d69b900b3fe58R11). Also i checked couple more existing unit test files like status_test.go and sync_test.go, those too seem to have different formats.
I think i did not understand correctly what you refering to as "others".

abhinavdahiya:
> `replicas` seems like something that we'll want to be able to manage via release images. In fact, the fewer knobs we leave _unmanaged_, the happier I'll be, so I think the burden of proof should be on "why _shouldn't_ we manage this?".
> 
> /approve
> 
> I'll wait for @abhinavdahiya to weigh in on whether that makes sense to him and whether he wants [a different format for the unit tests](https://github.com/openshift/cluster-version-operator/pull/198#discussion_r292244502).

For second level operators that manage specific operands, it is very justifiable to say i own my operand and no changes are allowed, because the scope is very well defined..

For CVO because we manage the operators, my goal was to make sure we have a way to define _I care about this and I will enforce this, but you can maybe add new things if we feel strongly about it_ And that's why "manage all and prove why i it is un managed" is not a great basis for CVO...

As for replicas, most operators have no replicas defined, ie they get 1 replica using the server-side defaults. But i might be useful for user envs to make more than one operator replica running for more resiliency... So user might scale the operator deployment without requiring them to go to the unsupported bucket of Overrides.

But, an option is that if the operator team decides to put replicas 1 in the manifest, we enforce it. Initially I didn't go this route too, as most operators might define replicas _not really_ wanting that behavior.. And there was no need for CVO to enforce replicas as ClusterOperator check is the backstop for the business logic.

wking:
> Initially I didn't go this route too, as most operators might define replicas *not really* wanting that behavior.

I'm fine making those operator bugs, since "enforce `replicas` when the release-manifest sets it" allows the manifest owner to decide if they care about `replicas` or not.

vikaschoudhary16:
@wking @abhinavdahiya Ping!!

PR #201, opened 2019-06-11
docs/user/reconciliation: Document release-image application

wking:
To make it easier for admins and support staff to troubleshoot when the cluster-version operator gets stuck.

The SVGs were generated with:

```console
$ go build ./cmd/cluster-version-util
$ mkdir /tmp/release
$ oc image extract quay.io/openshift-release-dev/ocp-release:4.1.0[-1] --path /:/tmp/release
$ mkdir /tmp/release/manifests
$ ./cluster-version-util task-graph /tmp/release | dot -Tsvg >docs/user/tasks-by-number-and-component.svg
$ ./cluster-version-util task-graph --parallel flatten-by-number-and-component /tmp/release | dot -Tsvg >docs/user/tasks-flatten-by-number-and-component.svg
```

using:

```console
$ dot -V
dot - graphviz version 2.30.1 (20170916.1124)
```


docs/user/syncronization.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Resource builders
+
+Resource builders synchronize the cluster with a manifest from the release image.
+The general approach is to generates a merged manifest combining critical spec properties from the release-image manifest with data from a preexisting in-cluster object, if any.
+If the merged manifest differs from the in-cluster object, the merged manifest is pushed back into the cluster.
+
+Some types have additional logic, as described in the following subsections.
+Note that this logic only applies to manifests included in the release image itself.
+For example, only ClusterOperator from the release image will have the blocking logic described [below](#clusteroperator]; if an admin or secondary operator pushed a ClusterOperator object, it would not impact the cluster-version operator's graph synchronization.
```
crawford: `[below](#cluster operator)`

wking: > `[below](#cluster operator)`

:man_facepalming: Fixed with dcfcf98 -> c15bd01.

wking:
/hold

Still a few FIXMEs in here.

wking:
Addressed some of the FIXMEs with c15bd01 -> 287ed5c.  The remaining ones are "I don't actually understand why the CVO is doing something" ;).

docs/user/syncronization.md
```diff
@@ -0,0 +1,142 @@
...
+
+### CustomResourceDefinition
+
+After pushing the merged CustomResourceDefinition into the cluster, the builder monitors the in-cluster object and blocks until it is established.
+
+### DaemonSet
+
+After pushing the merged DaemonSet into the cluster, the builder monitors
+
+* FIXME: `UpdatedNumberScheduled == DesiredNumberScheduled` check?
```
wking: This comparison is happening [here][1] since b5837163ec (#36).  @abhinavdahiya, do you remember why you need both that and `NumberUnavailable == 0`?  I'd naively expect either to be sufficient.

[1]: https://github.com/openshift/cluster-version-operator/blob/5f087d4fea766fd563e22f2c07868de87e69ae1a/lib/resourcebuilder/apps.go#L160

docs/user/syncronization.md
```diff
@@ -0,0 +1,142 @@
...
+* Has no unavailable nodes.
+
+### Deployment
+
+After pushing the merged Deployment into the cluster (FIXME: `Generation > 1` check from 14fab0b2?), the builder monitors the in-cluster object and blocks until it:
```
wking: This comparison is happening [here][1] since 14fab0b2 (#26).  @abhinavdahiya, do you remember why you added it?  Is the idea to not wait on Deployments during the initial install phase?

[1]: https://github.com/openshift/cluster-version-operator/blame/5f087d4fea766fd563e22f2c07868de87e69ae1a/lib/resourcebuilder/apps.go#L136

docs/user/syncronization.md
```diff
@@ -0,0 +1,142 @@
...
+
+* FIXME: `UpdatedReplicas == Replicas` check?
```
wking: This comparison is happening [here][1] since 2d334c21901a (#10).  @abhinavdahiya, do you remember why you need both that and `UnavailableReplicas == 0`?  I'd naively expect either to be sufficient.

[1]: https://github.com/openshift/cluster-version-operator/blob/5f087d4fea766fd563e22f2c07868de87e69ae1a/lib/resourcebuilder/apps.go#L76

wking:
Removed the final FIXMEs with 287ed5c -> d29f6ef.

/hold cancel

docs/user/syncronization.md
wking: Talked with @abhinavdahiya out of band; this is checking that all of the scheduled pods have the updated spec (regardless of whether they are available), while `NumberUnavailable == 0` is checking that all of the scheduled pods (regardless of their spec) are available.

docs/user/syncronization.md
wking: > Is the idea to not wait on Deployments during the initial install phase?

Yup.

docs/user/syncronization.md
wking: This is like [the analogous DaemonSet code][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/201#discussion_r292488206

docs/user/reconciliation.md
```diff
@@ -0,0 +1,166 @@
+# Reconciliation
+
+This document describes the cluster-version operator's reconciliation logic and explains how the operator applies a release image to the cluster.
+
+## Release image content
+
+```console
+$ mkdir /tmp/release
+$ oc image extract quay.io/openshift-release-dev/ocp-release:4.1.0[-1] --path /:/tmp/release
```
steveeJ: I would pin the version so the example is reproducible

wking: This is pinned (to the 4.1.0 release).

docs/user/reconciliation.md
```diff
@@ -0,0 +1,166 @@
...
+The general approach is to generates a merged manifest combining critical spec properties from the release-image manifest with data from a preexisting in-cluster object, if any.
+If the merged manifest differs from the in-cluster object, the merged manifest is pushed back into the cluster.
+
+Some types have additional logic, as described in the following subsections.
+Note that this logic only applies to manifests included in the release image itself.
+For example, only ClusterOperator from the release image will have the blocking logic described [below](#clusteroperator); if an admin or secondary operator pushed a ClusterOperator object, it would not impact the cluster-version operator's graph synchronization.
+
+### ClusterOperator
+
+The cluster-version operator does not push ClusterOperator into the cluster.
```
DanyC97: @wking my 0.002£ .

* i'd rename `cluster-version` to `ClusterVersion` as per [the doc](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md) and i'd also add the link to https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md too 
* i'd also insert a link on `ClusterOperator` pointing to https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

reading you doc (much appreciating for caring about adding this types of info) i had to remeber myself the diff between the too ;)

docs/user/reconciliation.md
steveeJ: That's right, I understood this as the image index as opposed to the layer index. However I'm still confused by the usage of the relative index in an example. Are you expecting the layers to change?

wking: > However I'm still confused by the usage of the relative index in an example. Are you expecting the layers to change?

We want the top/last layer, which has the manifests in it.  The total number of layers may or may not change, but `-1` seems like a more straightforward way to say "the last layer" than `4` (or whatever we'd use if we wanted a positive index).

docs/user/reconciliation.md
wking: > * i'd rename cluster-version to ClusterVersion...

I like "cluster-version operator" because it operates on the cluster version, not just the ClusterVersion custom-resource.  But this repo has many names for this operator of which "cluster-version operator" is only one.  So I've left this alone for now, but am ok changing it if we pick a name that we want to make canonical.

> * i'd also insert a link on ClusterOperator...

Done with b097cd0 -> cde8d50.

docs/user/reconciliation.md
```diff
@@ -0,0 +1,166 @@
...
+    status:
+      versions:
+        - name: operator
+          version: "4.1.0"
+    ```
+
+    would block until the in-cluster ClusterOperator reported `operator` at version 4.1.0.
+
+    The progressing check is deprecated and will be removed once all operators are reporting versions.
+* Not degraded (except during initialization, where we ignore the degraded status)
```
abhinavdahiya: there is already some content here https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#conditions-and-installupgrade

wking:
[e2e-aws failure][1]:

```
fail [github.com/openshift/origin/test/extended/operators/cluster.go:114]: Expected
    <[]string | len:2, cap:2>: [
        "Pod openshift-marketplace/csctestlabel-846b994d4b-kflpj is not healthy: container csctestlabel exited with non-zero exit code",
        "Pod openshift-marketplace/opsrctestlabel-d7499fcb8-fmjzc was pending entire time: unknown error",
    ]
to be empty
...
failed: (2m6s) 2019-11-26T22:54:20 "[Feature:Platform] Managed cluster should have no crashlooping pods in core namespaces over two minutes [Suite:openshift/conformance/parallel]"
```

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/201/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1007

wking:
Those pods were no longer running by [must-gather time][1].  Ah well.

[1]: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/201/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1007/artifacts/e2e-aws/must-gather/registry-svc-ci-openshift-org-ci-op-vr88wfw9-stable-sha256-51e78dd04992c2407dafe73e8adc58d3d9a4e7a40d24cefb12d8abebb04f9cf1/namespaces/openshift-marketplace/pods/

docs/user/reconciliation.md
wking: > there is already some content here...

Filed #274 to consolidate.

PR #203, opened 2019-06-11
metrics: Omit `cluster_version{type="completed"}` when not available

smarterclayton:
The cluster_version completed type was inconsistent with other types in that it was always included but the version label was unset when no completed update had been applied. Instead, only include the series when a completed update has been applied.

Discovered while building dashboards for metrics - the `completed` and `updating` metrics are similar in usage (tell me when updating, tell me when completed) and so this makes them consistent.


cben:
We (cc @vkareh) previously thought `type="completed"` is always set, so we tried to use that as "current fully applied version" to show in [cluster manager](https://cloud.redhat.com/openshift). 

I guess we could use latest of `type=~"initial|completed"`, but it feels messy, don't want to special-case clusters that have never upgraded yet.
Should we use some other type instead?  `type="cluster"`?  Is that always same as latest of `type=~"initial|completed"`?

cben:
Is `completed` omitted for fresh normally installed clusters that never upgraded yet,
or are we talking only about clusters whose _initial_ install didn't complete for some reason?

PR #204, opened 2019-06-12
metrics: Give all cluster_version types a `from_version` label

smarterclayton:
The `updating` type is hard to join in PromQL. Instead, report a label for
each type that is the previous completed version. Also change the `cluster`
type to return the original version of the cluster (first history entry).

This answers the following questions:

"what stable version is this cluster updating from"
"what was the first version recorded for this cluster"

Intended to fix bug 1720308 in 4.1

PR #206, opened 2019-06-13
pkg/cvo/metrics: Add a cluster_update gauge

wking:
This gives us a way to track upgrade success, with PromQL like:

```
count(label_join(cluster_update, "upgrade", "->", "from_version", "to_version")) by (state)
```

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -29,6 +29,7 @@ type operatorMetrics struct {
 	conditionTransitions map[conditionKey]int
 
 	version                             *prometheus.GaugeVec
+	update                              *prometheus.GaugeVec
 	availableUpdates                    *prometheus.GaugeVec
 	clusterOperatorUp                   *prometheus.GaugeVec
 	clusterOperatorConditions           *prometheus.GaugeVec
@@ -59,6 +60,10 @@ reached the completed state and is the time the update was
 started.
 .`,
 		}, []string{"type", "version", "image"}),
+		update: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_update",
+			Help: "Report cluster update duration in seconds.",
+		}, []string{"from_version", "from_image", "to_version", "to_image", "state"}),
 		availableUpdates: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "cluster_version_available_updates",
 			Help: "Report the count of available versions for an upstream and channel.",
@@ -192,6 +197,37 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 				g.Set(float64(updating.StartedTime.Unix()))
 			}
 			ch <- g
+		}
+
+		var next configv1.UpdateHistory
+		now := time.Now()
+		for i, history := range cv.Status.History {
+			if i > 0 && history.State == configv1.CompletedUpdate {
+				g := m.update.WithLabelValues(history.Version, history.Image, next.Version, next.Image, string(next.State))
+				var stop time.Time
+				if next.CompletionTime == nil {
+					stop = now
+				} else {
+					stop = next.CompletionTime.Time
+				}
+				g.Set(stop.Sub(next.StartedTime.Time).Seconds())
+				ch <- g
+			}
+			next = history
+		}
+
+		if len(cv.Status.History) > 0 {
+			install := cv.Status.History[len(cv.Status.History)-1]
+			if install.State == configv1.PartialUpdate && install.CompletionTime == nil && len(cv.Status.History) == 1 {
+				g := m.update.WithLabelValues("", "", install.Version, install.Image, string(install.State))
+				g.Set(now.Sub(install.StartedTime.Time).Seconds())
+				ch <- g
+			}
+			if install.State == configv1.CompletedUpdate {
+				g := m.update.WithLabelValues("", "", install.Version, install.Image, string(install.State))
+				g.Set(install.CompletionTime.Time.Sub(install.StartedTime.Time).Seconds())
+				ch <- g
+			}
 		}
 
 		if len(cv.Spec.Upstream) > 0 || len(cv.Status.AvailableUpdates) > 0 || resourcemerge.IsOperatorStatusConditionTrue(cv.Status.Conditions, configv1.RetrievedUpdates) {
--- a/pkg/cvo/metrics_test.go
+++ b/pkg/cvo/metrics_test.go
@@ -15,6 +15,9 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 )
 
+// ignoreValue instructs expectMetric to ignore the gauge value.
+const ignoreValue float64 = -123.456
+
 func Test_operatorMetrics_Collect(t *testing.T) {
 	tests := []struct {
 		name  string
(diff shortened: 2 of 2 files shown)


smarterclayton:
This has pretty high cardinality, which is one significant downside.  It also doesn't help the user find the current update.  What sorts of queries would we do on this?

wking:
Checking [the e2e-aws-upgrade metrics][1]:

```console
$ curl -s 'http://localhost:9090/api/v1/query_range?query=cluster_update&start=1560358271.167&end=1560531071.167&step=691&_=1560422797987' | jq .{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "__name__": "cluster_update",
          "endpoint": "metrics",
          "from_image": "registry.svc.ci.openshift.org/ci-op-pbi7v1sl/release@sha256:7853076fbd0464d8dd7cc20c9de003e41ddf5e49869e0c0e06ab5926586734c6",
          "from_version": "0.0.1-2019-06-13-031720",
          "instance": "10.0.159.157:9099",
          "job": "cluster-version-operator",
          "namespace": "openshift-cluster-version",
          "pod": "cluster-version-operator-857cfb76cd-fggtf",
          "service": "cluster-version-operator",
          "state": "Completed",
          "to_image": "registry.svc.ci.openshift.org/ci-op-pbi7v1sl/release@sha256:e6f6ca2e9850a26792383895926e0da0109f5a0796ff97cdf7b039fabb1d605f",
          "to_version": "0.0.1-2019-06-13-031829"
        },
        "values": [
          [
            1560400422.167,
            "1948"
          ]
        ]
      },
      {
        "metric": {
          "__name__": "cluster_update",
          "endpoint": "metrics",
          "from_image": "registry.svc.ci.openshift.org/ci-op-pbi7v1sl/release@sha256:7853076fbd0464d8dd7cc20c9de003e41ddf5e49869e0c0e06ab5926586734c6",
          "from_version": "0.0.1-2019-06-13-031720",
          "instance": "10.0.163.118:9099",
          "job": "cluster-version-operator",
          "namespace": "openshift-cluster-version",
          "pod": "cluster-version-operator-857cfb76cd-4f5zl",
          "service": "cluster-version-operator",
          "state": "Partial",
          "to_image": "registry.svc.ci.openshift.org/ci-op-pbi7v1sl/release@sha256:e6f6ca2e9850a26792383895926e0da0109f5a0796ff97cdf7b039fabb1d605f",
          "to_version": "0.0.1-2019-06-13-031829"
        },
        "values": [
          [
            1560399731.167,
            "1652.686129026"
          ]
        ]
      }
    ]
  }
}
```

So that seems to be appropriately showing the `Partial` 0.0.1-2019-06-13-031720->0.0.1-2019-06-13-031829 upgrade at 1652.  And then it shows the `Completed` upgrade at 1948 (~5 minutes later).  It's not clear to me why we don't have further entries for the `Completed` upgrade after that; maybe the tests finished before the next 5-minute poll?  Oh, and I should add empty-version->version entries for the initial install...

[1]: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/206/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/128/artifacts/e2e-aws-upgrade/metrics/

wking:
Added an install entry with a2e27c4 -> 5597aa2.

wking:
> This has pretty high cardinality...

I was reluctant to drop the `*_version` with their short, human-recognizable strings or the `*_image` with their unforgeable digests.  But I could probably be talked into either if necessary ;).  Does it help that, excepting forgeries, we expect all entries with a given `from_version` to have the same `from_image`?  I dunno if Prometheus can store that more efficiently or not.

> It also doesn't help the user find the current update.

Why not?  Can't you look for `Partial` upgrades to see what folks have in-flight now?

wking:
I think I got my value order flipped [looking at metrics above][1].  The completed entry's 1560400422.167, 1948 is "as of 2019-06-13T04:33:42Z, the upgrade had completed in 32 minutes".  Comparing with the collected ClusterVersion:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/206/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/128/artifacts/e2e-aws-upgrade/clusterversion.json | jq '.items[].status.history[0] | {startedTime, completionTime}'
{
  "startedTime": "2019-06-13T03:54:23Z",
  "completionTime": "2019-06-13T04:26:51Z"
}
```

that checks out.

[1]: https://github.com/openshift/cluster-version-operator/pull/206#issuecomment-501655289

wking:
So one question this allows you to answer is "how long have upgrades from _a_ to _b_ been taking to complete?":

```
avg(label_join(avg_over_time(cluster_update{from_image!="",state="Completed"}[24h]) / 60, "upgrade", "->", "from_version", "to_version")) by (upgrade)
```

Another question you can almost answer is "how many successful upgrades from _a_ to _b_ have there been?":

```
count(label_join(count_over_time(cluster_update{from_image!="",state="Completed"}[24h]), "upgrade", "->", "from_version", "to_version")) by (upgrade)
```

I think that works, although I'll feel more comfortable once I have more data than the single CI run that I can feed into it ;).  I'm still trying to figure out how to measure failed upgrades.  I expect it will be something like:

```
count(label_join(count_over_time(cluster_update{from_image!="",state="Pending"}[24h] > 45*60), "upgrade", "->", "from_version", "to_version")) by (upgrade)
```

to get things pending longer than 45 minutes, but that's giving me:

> Error executing query: parse error at char 156: binary expression must contain only scalar and instant vector types

wking:
Upgrade completion time:

upgrade completion time

Successful upgrade count (mostly overlapping at 1, unlikely to happen for long in production):

Screenshot from 2019-06-13 05-42-06

smarterclayton:
You know, I forgot to mention this before but remember this morning when Alex asked me another question - the full CV object will come down from support operator, and the rule of thumb should be anything that is high cardinality / retention style will be in support operator, not telemetry.  We should put things in prometheus that we want to live query on.

Some of the things you describe really sound like "post hoc analysis", not "need to know every 5 minutes".  Once a series has been sampled, you should never have to sample it again.  So there may be an argument that since upgrades happen less than every X minutes, you should really be only returning a subset of history upgrades.  If we can control the cardinality by "last 5" or "last 5 or last day" that would be better than unbounded (prometheus is not a data store)

I do think since this creates a new metric that it's complementary to my PR, not conflicting.  So we can do both.

smarterclayton:
We should use `upgrading` or `updating` consistently.  We use `update` everywhere today, so I think the metric type you have should be called `updating`

wking:
> ... the full CV object will come down from support operator...

So lets just land #204 and see if that's sufficient for measuring "we need to yank this edge because it's breaking clusters".  We can revisit edge-specific metrics if we find ourselves missing something.

/close

> We use update everywhere today...

Except in the very user-facing [`oc adm upgrade ...`][1].  I dunno if/how to reconcile that with the CVO's preference for "update".

[1]: https://github.com/openshift/origin/tree/2a8b07639e0b481ed8a77758383efb46b11788e0/staging/src/github.com/openshift/oc/pkg/cli/admin/upgrade

PR #212, opened 2019-06-25
Bug 1723945: Add a metric that combines the current age of the cluster with the current version

smarterclayton:
An important query is "how old are the clusters of a specific version"
and the current "cluster" query does not make this possible. In addition,
we can't get initial + current version in a single query. Split the
use case for "cluster" into "initial" and "cluster", where "initial"
is the same as "cluster" today, and "cluster" now points to the same
image and version as "current", has from_version populated, and has
the same value.

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -6,24 +6,26 @@ The cluster version is reported as seconds since the epoch with labels for `vers
 `image`. The `type` label reports which value is reported:
 
 * `current` - the version the operator is applying right now (the running CVO version) and the age of the payload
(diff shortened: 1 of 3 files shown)


abhinavdahiya:
> /assign @abhinavdahiya allows the "cluster age by version" aggregate view (tells us what the weight of clusters by age is) as well as makes easier to see what the initial version of a cluster was in combo with the current version.

sounds useful.

/lgtm

integration test being fixed by https://github.com/openshift/release/pull/4198

smarterclayton:
/hold

want to see metrics show up

smarterclayton:
/hold cancel

verified these metrics are right, should allow us to do the right queries now

PR #213, opened 2019-06-26
metrics: add cluster_installer series

crawford:
This series reports what installed the cluster. In order to do this, the
openshift-install ConfigMap, which is written by the installer when
doing an IPI install, is read. This ConfigMap is only injected during an
IPI installation and contains the invoker (e.g. username, Hive) as well
as the openshift-install version. The absence of this ConfigMap implies
a UPI installation.

Because the data in this series never changes, it should compress
nicely so we shouldn't have to worry about wasting space.


crawford:
Needs https://github.com/openshift/telemeter/pull/189 before we can see the results.

crawford:
The metrics appear to be working:

image

/retitle metrics: add cluster_installer series

pkg/cvo/metrics.go
```diff
@@ -269,6 +278,25 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
...
+		}
+		if i, ok := installer.Data["invoker"]; ok {
+			invoker = i
+		}
+
+		ch <- m.clusterInstaller.WithLabelValues("openshift-install", version, invoker)
+	} else if apierrors.IsNotFound(err) {
+		ch <- m.clusterInstaller.WithLabelValues("", "", "")
+	} else {
+		klog.Warningf("Unable to get installer ConfigMap: %v", err)
```
smarterclayton: Is this too busy of an error?  We get scraped every 15s, so you'd have a ton of logs in this case.

pkg/cvo/metrics_test.go
```diff
@@ -373,13 +386,72 @@ func Test_operatorMetrics_Collect(t *testing.T) {
...
+						ObjectMeta: metav1.ObjectMeta{Name: "openshift-install"},
+					}},
+				},
+			},
+			wants: func(t *testing.T, metrics []prometheus.Metric) {
+				if len(metrics) != 2 {
+					t.Fatalf("Unexpected metrics %s", spew.Sdump(metrics))
+				}
+				expectMetric(t, metrics[0], 3, map[string]string{"type": "current", "version": "0.0.2", "image": "test/image:1"})
+				expectMetric(t, metrics[1], 0, map[string]string{"type": "openshift-install", "version": "<missing>", "invoker": "<missing>"})
```
smarterclayton: This is somewhat abnormal, usually you'd just return empty values instead of missing.

pkg/cvo/metrics.go
```diff
@@ -78,6 +82,10 @@ version for 'cluster', or empty for 'initial'.
 			Name: "cluster_operator_condition_transitions",
 			Help: "Reports the number of times that a condition on a cluster operator changes status",
 		}, []string{"name", "condition"}),
+		clusterInstaller: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_installer",
+			Help: "Reports the installation method for the cluster",
```
smarterclayton: Is this supposed to have UPI vs IPI?

smarterclayton: I think you'd want to return 1 / 0 if this is about IPI vs UPI, or 1 always if you have a label for UPI.  Recommend documenting your labels in this.

pkg/cvo/metrics.go
```diff
@@ -269,6 +278,25 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
...
+		invoker := "<missing>"
+
+		if v, ok := installer.Data["version"]; ok {
+			version = v
+		}
+		if i, ok := installer.Data["invoker"]; ok {
+			invoker = i
+		}
+
+		ch <- m.clusterInstaller.WithLabelValues("openshift-install", version, invoker)
```
smarterclayton: Would have expected value 1 instead of value 0 here, just by convention.

pkg/cvo/metrics.go
crawford: Well, it's a bit more than that. You can derive UPI vs IPI from it, but you can also look at what invoked the installer (e.g. Hive). I can filled out that help text a bit more though.

pkg/cvo/metrics_test.go
crawford: Hmm, my thinking was that empty values were valid (e.g. a UPI install has the installer type, version, and invoker set to the empty value) whereas `<missing>` indicates an anomalous state. If the openshift-install ConfigMap exists, it will have all of those fields specified unless something were to erroneously remove them.

If we change this to use the empty value, then I need to update what's reported in UPI and it's not clear to me what those should be. I could say something like `m.clusterInstaller.WithLabelValues("human", "", "")`, but I didn't want to complicate things if we ever decide to have the installer explicitly call out when a UPI installation was performed (v.s. inferring it based on the absence of state).

pkg/cvo/metrics.go
smarterclayton: Wait, so now I'm confused.  Is type=UPI the label?  Expected to see a test with that as an example (just so we know).

Might want to update metrics.md too once you finalize.

pkg/cvo/metrics_test.go
smarterclayton: If you have a good reason, that's ok.

crawford: All good stuff for the commit message :wink:

pkg/cvo/metrics.go
crawford: I've updated metrics.md. Can you let me know if it makes sense now?

wking:
Cross-linking openshift/installer#1890 and openshift/installer#2046 which are injecting this information into the cluster.

PR #219, opened 2019-07-05
Add http transport for cincinnati to enable proxy

jcpowermac:
- Modified Cincinnati client to accept the http transport struct
so that a https proxy can be set
- Created new function getHTTPTransportProxy that
retrieves the proxy config, creates the transport and returns the
transport ptr.

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -20,12 +20,13 @@ const (
 // Client is a Cincinnati client which can be used to fetch update graphs from
 // an upstream Cincinnati stack.
 type Client struct {
-	id uuid.UUID
+	id       uuid.UUID
+	proxyURL *url.URL
 }
 
 // NewClient creates a new Cincinnati client with the given client identifier.
-func NewClient(id uuid.UUID) Client {
-	return Client{id: id}
+func NewClient(id uuid.UUID, proxyURL *url.URL) Client {
+	return Client{id: id, proxyURL: proxyURL}
 }
 
 // Update is a single node from the update graph.
@@ -39,6 +40,7 @@ type Update node
 // the current version and their payloads indicate from where the actual update
 // image can be downloaded.
 func (c Client) GetUpdates(upstream string, channel string, version semver.Version) ([]Update, error) {
+	transport := http.Transport{}
 	// Prepare parametrized cincinnati query.
 	cincinnatiURL, err := url.Parse(upstream)
 	if err != nil {
@@ -57,7 +59,11 @@ func (c Client) GetUpdates(upstream string, channel string, version semver.Versi
 	}
 	req.Header.Add("Accept", GraphMediaType)
 
-	client := http.Client{}
+	if c.proxyURL != nil {
+		transport.Proxy = http.ProxyURL(c.proxyURL)
+	}
+
+	client := http.Client{Transport: &transport}
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, err
--- a/pkg/cincinnati/cincinnati_test.go
+++ b/pkg/cincinnati/cincinnati_test.go
@@ -10,8 +10,8 @@ import (
 	"testing"
 
 	"github.com/blang/semver"
-	_ "k8s.io/klog" // integration tests set glog flags.
 	"github.com/google/uuid"
+	_ "k8s.io/klog" // integration tests set glog flags.
 )
 
 func TestGetUpdates(t *testing.T) {
@@ -120,8 +120,9 @@ func TestGetUpdates(t *testing.T) {
 
 			ts := httptest.NewServer(http.HandlerFunc(handler))
 			defer ts.Close()
+			var proxyURL *url.URL
 
-			c := NewClient(clientID)
+			c := NewClient(clientID, proxyURL)
 
 			updates, err := c.GetUpdates(ts.URL, channelName, semver.MustParse(test.version))
 			if test.err == "" {
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -4,9 +4,12 @@ import (
 	"fmt"
 	"time"
 
+	"net/url"
+
 	"github.com/blang/semver"
-	"k8s.io/klog"
 	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -36,7 +39,12 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
 		return nil
 	}
 
-	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), upstream, channel, optr.releaseVersion)
+	proxyURL, err := optr.getHTTPSProxyURL()
+	if err != nil {
+		return err
+	}
+
+	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), proxyURL, upstream, channel, optr.releaseVersion)
 
 	if usedDefaultUpstream {
 		upstream = ""
@@ -103,7 +111,7 @@ func (optr *Operator) getAvailableUpdates() *availableUpdates {
 	return optr.availableUpdates
 }
 
-func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version string) ([]configv1.Update, configv1.ClusterOperatorStatusCondition) {
+func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, upstream, channel, version string) ([]configv1.Update, configv1.ClusterOperatorStatusCondition) {
 	if len(upstream) == 0 {
 		return nil, configv1.ClusterOperatorStatusCondition{
 			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoUpstream",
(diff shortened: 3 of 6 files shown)


jcpowermac:
https://github.com/openshift/cluster-version-operator/blob/master/vendor/github.com/openshift/api/config/v1/types_proxy.go
Does not have ProxyStatus.  Either update or switch to using `proxy.Spec` instead of `proxy.Status`.  

```
# github.com/openshift/cluster-version-operator/pkg/cvo pkg/cvo/availableupdates.go:204:10: proxy.Status undefined...enshift/api/config/v1".Proxy has no field or method Status)
```

jcpowermac:
@abhinavdahiya when you have a moment to take a look for a initial review, thanks!

pkg/cvo/availableupdates.go
```diff
@@ -164,8 +171,44 @@ func checkForUpdate(clusterID, upstream, channel string, currentVersion semver.V
...
 	if len(upstream) == 0 {
 		return nil, fmt.Errorf("no upstream URL set for cluster version")
 	}
-	return cincinnati.NewClient(uuid).GetUpdates(upstream, channel, currentVersion)
+	return cincinnati.NewClient(uuid, transport).GetUpdates(upstream, channel, currentVersion)
+}
+
+func getHTTPTransportProxy() (*http.Transport, error) {
+	transport := http.Transport{}
+	config, err := rest.InClusterConfig()
```
abhinavdahiya: You cannot create a new config here, please reuse the one in the operator struct already.

Or add proxy type client /listers to the operator struct and use here...

That would help in integration tests

pkg/cincinnati/cincinnati.go
```diff
@@ -20,12 +20,13 @@ const (
...
 type Client struct {
-	id uuid.UUID
+	id        uuid.UUID
+	transport *http.Transport
 }
 
 // NewClient creates a new Cincinnati client with the given client identifier.
-func NewClient(id uuid.UUID) Client {
-	return Client{id: id}
+func NewClient(id uuid.UUID, transport *http.Transport) Client {
```
abhinavdahiya: This is pretty broad additon.. 
I think a restricted proxy information and creating a transport with that is more ideal.

pkg/cincinnati/cincinnati_test.go
```diff
@@ -10,8 +10,8 @@ import (
 	"testing"
 
 	"github.com/blang/semver"
-	_ "k8s.io/klog" // integration tests set glog flags.
 	"github.com/google/uuid"
+	_ "k8s.io/klog" // integration tests set glog flags.
 )
```
abhinavdahiya: Would also like to see proxy tests here or in integration tests

jcpowermac:
@abhinavdahiya updated PTAL.  I will work on tests once everything else looks good.  Thanks!

pkg/cincinnati/cincinnati.go
```diff
@@ -20,12 +20,13 @@ const (
...
 type Client struct {
-	id uuid.UUID
+	id  uuid.UUID
+	url *url.URL
 }
 
 // NewClient creates a new Cincinnati client with the given client identifier.
-func NewClient(id uuid.UUID) Client {
-	return Client{id: id}
+func NewClient(id uuid.UUID, url *url.URL) Client {
```
abhinavdahiya: what's the url, comments will be useful

pkg/cvo/availableupdates.go
```diff
@@ -4,9 +4,13 @@ import (
...
 	"time"
 
+	"net/url"
+
 	"github.com/blang/semver"
-	"k8s.io/klog"
 	"github.com/google/uuid"
+	"k8s.io/klog"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
abhinavdahiya: extra line

pkg/cvo/cvo.go
```diff
@@ -163,6 +165,10 @@ func New(
 	}
 
 	cvInformer.Informer().AddEventHandler(optr.eventHandler())
+	proxyInformer.Informer().AddEventHandler(optr.eventHandler())
+
+	optr.proxyLister = proxyInformer.Lister()
+	optr.cacheSynced = append(optr.cacheSynced, proxyInformer.Informer().HasSynced)
```
abhinavdahiya: 4.2 the type exists for proxy.
but 4.1 the type doesn't exit for proxy.

So when user goes from 4.1.z to 4.2.z ; the CVO updates itself to a version that expects proxy type to exist, but the proxy type will be created by the cluster-config-operator. So this cache will never sync and CVO will be stuck.

So it's important our changes keep in mind that the type might not exist.

pkg/cvo/availableupdates.go
```diff
@@ -159,13 +169,35 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
...
 		return nil, fmt.Errorf("no upstream URL set for cluster version")
 	}
-	return cincinnati.NewClient(uuid).GetUpdates(upstream, channel, currentVersion)
+	return cincinnati.NewClient(uuid, url).GetUpdates(upstream, channel, currentVersion)
+}
+
+func (optr *Operator) getHTTPSProxyURL() (*url.URL, error) {
+	proxy, err := optr.proxyLister.Get("cluster")
+
+	if errors.IsNotFound(err) {
```
abhinavdahiya: see https://github.com/openshift/cluster-version-operator/pull/219/files#r301772882

we should do a no-op when the type doesn't exist.

pkg/cincinnati/cincinnati.go
```diff
@@ -20,12 +20,13 @@ const (
 // Client is a Cincinnati client which can be used to fetch update graphs from
 // an upstream Cincinnati stack.
 type Client struct {
-	id uuid.UUID
+	id  uuid.UUID
+	url *url.URL
```
abhinavdahiya: hmm, how is url separate from upstream in `GetUpdates`

pkg/cvo/cvo.go
```diff
@@ -160,6 +163,25 @@ func New(
 
 		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "clusterversion"),
 		availableUpdatesQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "availableupdates"),
+		proxyExists:           false,
+	}
+
+	discover := kubeClient.Discovery()
+	apiResourcesList, err := discover.ServerResourcesForGroupVersion("config.openshift.io/v1")
```
abhinavdahiya: The discovery is not useful.. this will always be true for CVO start...

the proxy CRD is created by an top-level operator cluster-config-operator... that is created by cluster-version-operator.

jcpowermac: If that is the case how would we check if Proxy kind exists?  It seems like any check would fail based on the above description.

abhinavdahiya: when you go to `GET` proxy object and you get not found, you don't use it..

abhinavdahiya: if the proxy api exists and the resource exists, you will be able to GET successfully and then we should start using the Proxy codepath.

pkg/cvo/availableupdates.go
```diff
@@ -159,13 +171,37 @@ func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version strin
...
+}
+
+// getHTTPSProxyURL returns a url.URL object for the configured
+// https proxy only. It can be nil if does not exist or there is an error.
+func (optr *Operator) getHTTPSProxyURL() (*url.URL, error) {
+	proxy, err := optr.proxyLister.Get("cluster")
+
+	if errors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
```
abhinavdahiya: why do you need the `else if` ?

pkg/cvo/cvo.go
```diff
@@ -160,6 +163,16 @@ func New(
 
 		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "clusterversion"),
 		availableUpdatesQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "availableupdates"),
+		proxyExists:           false,
+	}
+
+	_, err := client.ConfigV1().Proxies().Get("cluster", metav1.GetOptions{})
```
abhinavdahiya: there's no requirement for this conditional setup of the informer.

pkg/cvo/cvo.go
```diff
@@ -160,6 +163,16 @@ func New(
...
+
+	if err == nil {
+		optr.proxyLister = proxyInformer.Lister()
+		proxyInformer.Informer().AddEventHandler(optr.eventHandler())
+		optr.cacheSynced = append(optr.cacheSynced, proxyInformer.Informer().HasSynced)
```
abhinavdahiya: just drop the proxy from cachedSynced... it's goal is to make sure we don't start action on inconsistent cache... but there's no such requirement for proxy informer.

pkg/cvo/availableupdates.go
```diff
@@ -103,7 +115,7 @@ func (optr *Operator) getAvailableUpdates() *availableUpdates {
 	return optr.availableUpdates
 }
 
-func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version string) ([]configv1.Update, configv1.ClusterOperatorStatusCondition) {
+func calculateAvailableUpdatesStatus(clusterID, upstream, channel, version string, proxyURL *url.URL) ([]configv1.Update, configv1.ClusterOperatorStatusCondition) {
```
abhinavdahiya: proxyURL should be next to clusterID.

pkg/cvo/availableupdates.go
jcpowermac: I was thinking we would want to know what the error was if it was anything except `IsNotFound`

abhinavdahiya: errros.IsNotFound only returns true when the err is not-found..
ie if err is nil, it will not be true.. and if err is not-nil and but not not-found, it will not be true..

so
```
if errors.IsNotFound(err) {
  return nil, nil
}
if err != nil {
  return nil, err
}
```

cover's all the cases...

PR #220, opened 2019-07-05
Update openshift client-go and api

jcpowermac:
Updates to pull in proxy changes


jcpowermac:
Need for proxy changes:
https://github.com/openshift/cluster-version-operator/pull/219

Not sure I did this right only updated api and client-go.  If not let me know and I will fix.

danehans:
@abhinavdahiya PTAL

PR #221, opened 2019-07-11
payload/render: update the bootstrap render to skip servicemonitor

abhinavdahiya:
the `manifests` directory on the bootstrap is used by the cluster-bootstrap to push to the cluster.
`servicemonitor` for cvo was added by https://github.com/openshift/cluster-version-operator/pull/214
`servicemonitor` api is created by the cluster-monitoring-operator and therefore this causes the bootstrapping to get stuck until we get the monitoring operator running.

This skips the `servicemonitor` in the bootstrap render as it is not required for the bootstrap cvo pod.

/cc @smarterclayton @brancz

### DIFF
--- a/pkg/payload/render.go
+++ b/pkg/payload/render.go
@@ -30,9 +30,12 @@ func Render(outputDir, releaseImage string) error {
 		odir      string
 		skipFiles sets.String
 	}{{
-		idir:      manifestsDir,
-		odir:      oManifestsDir,
-		skipFiles: sets.NewString("image-references"),
+		idir: manifestsDir,
+		odir: oManifestsDir,
+		skipFiles: sets.NewString(
+			"image-references",
+			"0000_90_cluster-version-operator_00_servicemonitor.yaml",
+		),
 	}, {
 		idir:      bootstrapDir,
 		odir:      oBootstrapDir,


smarterclayton:
Seems ok to me

brancz:
We were under the assumption that because of the run level this would already be applied at a later point. I’ll lgtm this and put hold on it, feel free to remove the hold, if that assumption was incorrect.

/lgtm
/hold

smarterclayton:
Bootstrap is special.  Maybe we need an annotation that excludes things from bootstrap render?

abhinavdahiya:
> Bootstrap is special. Maybe we need an annotation that excludes things from bootstrap render?
@smarterclayton 
the list of things right now is small (2) and if this get's bigger we can think about that. and k8s annotations would mean parsing the file... _not thrilled_

smarterclayton:
Don’t we parse after templating before we render?  Or is bootstrap
sufficiently different?

Not a huge deal


On Jul 12, 2019, at 5:47 PM, Abhinav Dahiya <notifications@github.com>
wrote:

Bootstrap is special. Maybe we need an annotation that excludes things from
bootstrap render?
@smarterclayton <https://github.com/smarterclayton>
the list of things right now is small (2) and if this get's bigger we can
think about that. and k8s annotations would mean parsing the file... *not
thrilled*

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/cluster-version-operator/pull/221?email_source=notifications&email_token=AAI37J52AKDTNLSIJKNAL43P7D3YRA5CNFSM4IBRD56KYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODZ26WKA#issuecomment-511044392>,
or mute the thread

abhinavdahiya:
/hold cancel

the bootstrap render is reading, templating (text based) and copying,, it treats them as text file and not k8s manifests. So the annotation part would be a change from it being simple..

for now, I don't think the marshalling into k8s objects to make skipping on render is of great value.

ping @smarterclayton 
if you agree this is a acceptable solution for now, lgtm (`/lgtm`) would be nice :)

smarterclayton:
/lgtm

Sorry

PR #222, opened 2019-07-12
document CVO behavior w/ respect to cluster operator conditions

bparees:


bparees:
/assign @abhinavdahiya 

(should update this w/ the upgradeable condition when that is a thing)

docs/dev/clusteroperator.md
```diff
@@ -97,3 +97,22 @@ If an error blocks reaching 4.0.1, the conditions might be:
...
+
+
+| operation | version | available | degraded | progressing | 
+|-----------|---------|-----------|----------|-------------|
+| Install completion[1] | current(whatever was being installed) | true | any | any
+| Begin upgrade | any | any | any | any
+| Begin upgrade (w/ force) | any | any | any | any 
+| Upgrade completion[2]| newVersion(target version for the upgrade) | true | false | false
+
+[1] Install cannot proceed to the next cluster operator until the previous cluster operator achieves these conditions
```
abhinavdahiya: during install we allow all operators to be created at once _ same level_
cc @smarterclayton

bparees: ah yes, i'll clarify.

docs/dev/clusteroperator.md
```diff
@@ -97,3 +97,22 @@ If an error blocks reaching 4.0.1, the conditions might be:
...
+
+
+| operation | version | available | degraded | progressing | 
+|-----------|---------|-----------|----------|-------------|
+| Install completion[1] | current(whatever was being installed) | true | any | any
+| Begin upgrade | any | any | any | any
+| Begin upgrade (w/ force) | any | any | any | any 
+| Upgrade completion[2]| newVersion(target version for the upgrade) | true | false | false
+
+[1] Install cannot proceed to cluster operators in the next runlevel until all cluster operators in the current runlevel achieve these conditions
```
bparees: @abhinavdahiya updated ^^

bparees:
@abhinavdahiya i think this is ready to go?

docs/dev/clusteroperator.md
smarterclayton: I'm confused what you mean here - ordering only matters during upgrade, it's ignored on install and sync?  Maybe I'm misinterpreting what you mean by run level

bparees: > ordering only matters during upgrade, it's ignored on install and sync

that's what i was trying to say, yes.  By runlevel i meant like "70_blah".  My understanding was everything at run level 70 runs in parallel during install.  then everything at runlevel 71.  etc.

and then during upgrade I was trying to say it's fully ordered, no parallelization.

I wasn't trying to say anything about sync.

docs/dev/clusteroperator.md
```diff
@@ -97,3 +97,22 @@ If an error blocks reaching 4.0.1, the conditions might be:
...
+| operation | version | available | degraded | progressing | 
+|-----------|---------|-----------|----------|-------------|
+| Install completion[1] | current(whatever was being installed) | true | any | any
+| Begin upgrade | any | any | any | any
+| Begin upgrade (w/ force) | any | any | any | any 
+| Upgrade completion[2]| newVersion(target version for the upgrade) | true | false | false
+
+[1] Install works on all components in parallel, it does not wait for any component to complete before starting another one.
+
+[2] Upgrade will not proceed with upgrading components in the next runlevel until the previous runlevel completes.
```
bparees: @smarterclayton @abhinavdahiya updated these descriptions.

abhinavdahiya:
/approve

will let @smarterclayton lgtm

docs/dev/clusteroperator.md
bparees: @abhinavdahiya fyi when I talked to @smarterclayton he said we created all operators at once, period (regardless of runlevel), so i updated the doc to reflect that.

docs/dev/clusteroperator.md
smarterclayton: fine with me

PR #50, opened 2018-11-11
ADd a RHEL7 dockerfile and standardize format

smarterclayton:

PR #223, opened 2019-07-18
Adding support for other archs than amd64

Pensu:
This modifies the Dockerfile to add the build support for archs other than amd64, based on the commit: https://github.com/openshift/cluster-version-operator/commit/bd1c181d38016f5f840353cbd7ce45fdce1f546a

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,12 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-version-operator
 COPY . .
-RUN hack/build-go.sh
+RUN hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp _output/linux/$(go env GOARCH)/cluster-version-operator /tmp/build/cluster-version-operator
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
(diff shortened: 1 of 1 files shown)


Pensu:
Hi @wking @abhinavdahiya, can you please review this?

wking:
Do we actually need two Dockerfiles?  Would be nice to collapse to one.

Pensu:
@wking Yeah, makes sense. Though I am not sure about the usability of Dockerfile.rhel7. It comes from this commit: https://github.com/openshift/cluster-version-operator/pull/50. Do we really need a separate Dockerfile for rhel7? 

cc: @smarterclayton @abhinavdahiya

smarterclayton:
Most go program images do not need a separate dockerfile for rhel since there is no variance.  However, before it is removed you need to guarantee that the ocp build and CI builds are using the stock Dockerfile and that no drift has occurred

Pensu:
@smarterclayton I have checked out the directory: https://github.com/openshift/release/tree/master/ci-operator/config/openshift/cluster-version-operator and couldn't find a reference to Dockerfile.rhel7. 

cc: @stevekuznetsov

abhinavdahiya:
For sake of not breaking _unknown_ consumers could we just keep the rhel Dockerfile for now and later on we can triage if the rhel one is needed at all. getting blocked on that for multi-arch is not ideal i would guess...

Pensu:
@abhinavdahiya Yeah, sure, makes sense. I have reverted back my changes. PTAL.

PR #224, opened 2019-07-18
inject proxy env vars into requesting workload resources

deads2k:
set .metadata.annotations["config.openshift.io/inject-proxy"]="container1,initcontainer2"
to inject into specified container names

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -3,11 +3,9 @@ package resourcebuilder
 import (
 	"context"
 	"fmt"
-
 	"strings"
 
-	"k8s.io/klog"
-
+	configv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/cluster-version-operator/lib"
 	"github.com/openshift/cluster-version-operator/lib/resourceapply"
 	"github.com/openshift/cluster-version-operator/lib/resourceread"
@@ -17,18 +15,21 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog"
 )
 
 type deploymentBuilder struct {
-	client   *appsclientv1.AppsV1Client
-	raw      []byte
-	modifier MetaV1ObjectModifierFunc
+	client      *appsclientv1.AppsV1Client
+	proxyGetter configv1.ProxiesGetter
+	raw         []byte
+	modifier    MetaV1ObjectModifierFunc
 }
 
 func newDeploymentBuilder(config *rest.Config, m lib.Manifest) Interface {
 	return &deploymentBuilder{
-		client: appsclientv1.NewForConfigOrDie(withProtobuf(config)),
-		raw:    m.Raw,
+		client:      appsclientv1.NewForConfigOrDie(withProtobuf(config)),
+		proxyGetter: configv1.NewForConfigOrDie(config),
+		raw:         m.Raw,
 	}
 }
 
@@ -46,6 +47,26 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 	if b.modifier != nil {
 		b.modifier(deployment)
 	}
+
+	// if proxy injection is requested, get the proxy values and use them
+	if containerNamesString := deployment.Annotations["config.openshift.io/inject-proxy"]; len(containerNamesString) > 0 {
+		proxyConfig, err := b.proxyGetter.Proxies().Get("cluster", metav1.GetOptions{})
+		// not found just means that we don't have proxy configuration, so we should tolerate and fill in empty
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		err = updatePodSpecWithProxy(
+			&deployment.Spec.Template.Spec,
+			strings.Split(containerNamesString, ","),
+			proxyConfig.Status.HTTPProxy,
+			proxyConfig.Status.HTTPSProxy,
+			proxyConfig.Status.NoProxy,
+		)
+		if err != nil {
+			return err
+		}
+	}
+
 	actual, updated, err := resourceapply.ApplyDeployment(b.client, deployment)
 	if err != nil {
 		return err
@@ -103,15 +124,17 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
 }
 
 type daemonsetBuilder struct {
-	client   *appsclientv1.AppsV1Client
-	raw      []byte
-	modifier MetaV1ObjectModifierFunc
+	client      *appsclientv1.AppsV1Client
+	proxyGetter configv1.ProxiesGetter
+	raw         []byte
+	modifier    MetaV1ObjectModifierFunc
 }
 
 func newDaemonsetBuilder(config *rest.Config, m lib.Manifest) Interface {
 	return &daemonsetBuilder{
-		client: appsclientv1.NewForConfigOrDie(withProtobuf(config)),
-		raw:    m.Raw,
+		client:      appsclientv1.NewForConfigOrDie(withProtobuf(config)),
+		proxyGetter: configv1.NewForConfigOrDie(config),
+		raw:         m.Raw,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


deads2k:
@ironcladlou @bparees fyi

bparees:
also /cc @adambkaplan @dmage @cdaley as this would be relevant to the registry operator if it were implemented, since it is currently managing its own proxy env config (inconveniently).

lib/resourcebuilder/apps.go
```diff
@@ -46,6 +49,57 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 	if b.modifier != nil {
 		b.modifier(deployment)
 	}
+
+	// if proxy injection is requested, get the proxy values and use them
+	// TODO You'll want a mutation helper that takes a pointer to a pod spec so you can wire it to other workload resources
+	if containerNamesString := deployment.Annotations["config.openshift.io/inject-proxy"]; len(containerNamesString) > 0 {
```
smarterclayton: Not terrible.

smarterclayton: Which operators would use this?  The ones that need all of their traffic to go through the proxy, including service network (I assume status.noProxy must ALWAYS contain service network or you cause an outage?)

mfojtik: auth operator for example....

the noProxy will be provided by network operator which should set these (I asked him same question during arch call ;-)

dmage: The image registry operator may use this. We already have similar logic inside of our reconciliation loop. The expectation is that noProxy covers the API server. And our operator needs to interact with external services like S3.

danehans: > the noProxy will be provided by network operator which should set these

That's correct. I plan to enforce the default noProxy values as part of https://github.com/openshift/cluster-network-operator/pull/245

danehans: > The expectation is that noProxy covers the API server. And our operator needs to interact with external services like S3.

@dmage team leads were updated recently that proxy support is not required if the operator only communicates in-cluster and to hosting cloud provider api's.

danehans: > I assume status.noProxy must ALWAYS contain service network or you cause an outage?

@smarterclayton https://github.com/openshift/cluster-network-operator/pull/245 (WIP) will validate `spec.noProxy`, merge `spec.noProxy` values with the following values required by the cluster and update `status.noProxy`:

```
127.0.0.1
localhost
installConfig.ServiceNetwork
installConfig.ClusterNetwork
installConfig.MachineCIDR
apiServerURL.Hostname
internalAPIServer.Hostname
etcd-{REPL_NUM}.{CLUSTER_DOMAIN}
169.254.169.254 // cloud metadata
```

bparees: @danehans the registry operator communicates to cloud apis even when the cluster is not running on that particular cloud.

danehans:
@deads2k @sttts I'm in favor of this approach.

lib/resourcebuilder/apps.go
danehans: @bparees the registry operator is then responsible for augmenting the cluster-wide `noProxy` with additional no proxy values, correct? [This](https://github.com/openshift/cluster-network-operator/pull/245/files#diff-3ab954bacdc2df09a1ece5c1bd61c752R58-R60) is the WIP implementation of the cloud metadata ip.

adambkaplan: @danehans I don't know if that is desirable, particularly if a customer wants their baremetal cluster to have all outbound traffic go through their proxy (and yet still use s3 or other cloud provider storage as their backend).

bparees: > @bparees the registry operator is then responsible for augmenting the cluster-wide noProxy with additional no proxy values, correct? 

?  My point is that the registry operator has a need to use the proxy configuration when it is talking to the cloudapis that exist outside the cloudprovider the cluster is running on.  NoProxy is irrelevant.

@dmage was observing that the registry operator has a use for this proxy env injection.  You said they didn't need it because they are talking to cloud apis, but that is incorrect.  They are talking to cloudapis, but not cloudapis from the cluster's cloud provider, therefore they will need to go through the proxy (assuming the customer's network topology requires it, which would be our assumption since it's external traffic like any other).

If the customer does not want their AWS api calls going through the proxy (and assuming their network topology doesn't require it), the admin would have to add the AWS api endpoints to their NoProxy value.

danehans:
@deads2k @sttts is this PR moving forward or just a reference implementation that operators can follow?

smarterclayton:
@abhinavdahiya have you looked at this PR yet?

lib/resourcebuilder/apps.go
danehans: > I don't know if that is desirable, particularly if a customer wants their baremetal cluster to have all outbound traffic go through their proxy (and yet still use s3 or other cloud provider storage as their backend).

@adambkaplan the cloud metadata ip will not be set for vsphere, baremetal and non platform types.

deads2k:
We've agreed to go this route.

/hold cancel

lib/resourcebuilder/apps.go
```diff
@@ -46,6 +48,26 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+		proxyConfig, err := b.proxyGetter.Proxies().Get("cluster", metav1.GetOptions{})
+		// not found just means that we don't have proxy configuration, so we should tolerate and fill in empty
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		err = updatePodSpecWithProxy(
+			&deployment.Spec.Template.Spec,
+			strings.Split(containerNamesString, ","),
+			proxyConfig.Status.HTTPProxy,
```
abhinavdahiya: In case of 404, this will probably panic as proxy config could be nil...

abhinavdahiya:
How do we handle a case where an operator wants to remove the envs from it's containers as part of the update??

lib/resourcebuilder/apps.go
```diff
@@ -6,8 +6,7 @@ import (
 
 	"strings"
 
-	"k8s.io/klog"
-
+	v1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
```
abhinavdahiya: nit `configv1`

lib/resourcebuilder/podspec.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// updatePodSpecWithProxy mutates the input podspec with proxy env vars for all init containers and containers
+// matching the container names.
+func updatePodSpecWithProxy(podSpec *corev1.PodSpec, containerNames []string, httpProxy, httpsProxy, noProxy string) error {
+	hasProxy := len(httpsProxy) > 0 || len(httpProxy) > 0 || len(noProxy) > 0
```
abhinavdahiya: If only NoProxy was set, I'm not sure that's going to work correctly...

lib/resourcebuilder/apps.go
deads2k: > In case of 404, this will probably panic as proxy config could be nil...

clients always return an object with it. See https://github.com/openshift/client-go/blob/master/config/clientset/versioned/typed/config/v1/proxy.go#L49-L58

deads2k:
> How do we handle a case where an operator wants to remove the envs from it's containers as part of the update??

It's hard to imagine where that's useful.  We can build a special value into the annotation if it comes up.

lib/resourcebuilder/podspec.go
deads2k: > If only NoProxy was set, I'm not sure that's going to work correctly...

Empty is the same as unset.  See https://github.com/golang/net/blob/master/http/httpproxy/proxy.go#L91-L107

lib/resourcebuilder/apps.go
abhinavdahiya: +1

lib/resourcebuilder/podspec.go
abhinavdahiya: +1

abhinavdahiya:
> > How do we handle a case where an operator wants to remove the envs from it's containers as part of the update??
> 
> It's hard to imagine where that's useful. We can build a special value into the annotation if it comes up.

sure.

I think this get's other operators unblocked for time being. i'll take a todo to add some docs and any fixes needed for 4.1 to 4.2 upgrade.

Thanks!

/lgtm

lib/resourcebuilder/podspec.go
```diff
@@ -0,0 +1,47 @@
...
+
+	for _, containerName := range containerNames {
+		found := false
+		for i := range podSpec.Containers {
+			if podSpec.Containers[i].Name != containerName {
+				continue
+			}
+			found = true
+
+			podSpec.Containers[i].Env = append(podSpec.Containers[i].Env, corev1.EnvVar{Name: "HTTP_PROXY", Value: httpProxy})
```
abhinavdahiya: Wouldn't this keep appending to the env list each sync cycle..?

deads2k: You're reading from raw bytes your read from a file, right?  How would that be cyclical?

abhinavdahiya: Oh, yeah missed it was right after read..

PR #168, opened 2019-04-19
Add CRD openapi definitions

soltysh:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1688638

/assign @deads2k @smarterclayton @sttts

PR #225, opened 2019-07-22
manifests: Update the CRD definitions with more accurate docs

smarterclayton:
This refreshes our CRD schema with the latest copies from the 4.2
streams. Docs are regenerated from tooling in cluster-config-operator
and merged with the current definition.

1. Go to openshift/cluster-config-operator
2. Build crd-schema-gen, invoke it like `crd-schema-gen --apis-dir vendor/github.com/openshift/api/config/v1/ -output-dir manifests`
3. Copy the generated clusterversion/operator manifests over the CRDS in repo
4. Merge the schema portion only in (dropping other changes)

### DIFF
--- a/install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
+++ b/install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
@@ -42,6 +42,9 @@ spec:
     storage: true
   validation:
     openAPIV3Schema:
+      description: ClusterOperator is the Custom Resource object which holds the current
+        state of an operator. This object is used by operators to convey their state
+        to the rest of the cluster.
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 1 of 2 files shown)


smarterclayton:
/assign @abhinavdahiya 

Bug incoming.  I figure we'll potentially want to script this like cluster-config-operator in the future, it's not quite there yet.

smarterclayton:
Hrm.  I need to check whether the schema is too restrictive too.  It looks like some of the optional fields are optional.

abhinavdahiya:
@smarterclayton 
https://github.com/openshift/cluster-version-operator/pull/168 added the wiring to call update-crd-gen ??
that's not sufficient??

smarterclayton:
No it doesn't, but I'll have to dig in why later.  I think the version is too old.  Suggest we merge this, then I'll dig in once we get closer to feature freeze.

/retest

abhinavdahiya:
> No it doesn't, but I'll have to dig in why later. I think the version is too old. Suggest we merge this, then I'll dig in once we get closer to feature freeze.
> 
> /retest

/lgtm

PR #226, opened 2019-07-25
Add certificate option to cincinnati client

jcpowermac:
Adds support to use a ConfigMap that contains a certificate bundle if a MITM proxy is used with cincinnati http client.

- Modify cincinnati client struct to included tlsConfig
- Modify cincinnati NewClient for tls change
- Modify getHTTPSProxyURL to return ConfigMapNameReference
- Add new method getTLSConfig that uses ConfigMap to extract the CA bundle and creates a tlsConfig


pkg/cvo/availableupdates.go
```diff
@@ -39,12 +41,16 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
 		return nil
 	}
 
-	proxyURL, err := optr.getHTTPSProxyURL()
+	proxyURL, cmNameRef, err := optr.getHTTPSProxyURL()
+	if err != nil {
+		return err
+	}
+	tlsConfig, err := optr.getTLSConfig(cmNameRef)
```
abhinavdahiya: the trustedCA for proxies is optional.. so we should keep that in mind.

pkg/cvo/availableupdates.go
```diff
@@ -176,29 +182,59 @@ func checkForUpdate(clusterID string, proxyURL *url.URL, upstream, channel strin
...
 			}
-			return proxyURL, nil
+			return proxyURL, proxy.Spec.TrustedCA, nil
 		}
 	}
-	return nil, nil
+	return nil, configv1.ConfigMapNameReference{}, nil
+}
+
+func (optr *Operator) getTLSConfig(cmNameRef configv1.ConfigMapNameReference) (*tls.Config, error) {
```
abhinavdahiya: this function should assume that the reference is required.

pkg/cvo/availableupdates.go
```diff
@@ -176,29 +182,59 @@ func checkForUpdate(clusterID string, proxyURL *url.URL, upstream, channel strin
...
+	cm, err := optr.kubeClient.CoreV1().ConfigMaps("openshift-config-managed").Get(cmNameRef.Name, metav1.GetOptions{})
```
abhinavdahiya: https://github.com/openshift/cluster-version-operator/pull/213/commits/c14773282d0cdd2c5bcce6d5a29c7e2e274337e5

added an informer/lister for configmaps for `openshif-config` namespaces.. could we do somethings like that do that we are calling the api and using the cache if possible..

jcpowermac:
> I think we should fix the name of previous informer and create a new one based on namespace rather the name for proxy ca

@abhinavdahiya 
Could we create a shared informer for both?  Or do we want to limit the namespaces?

jcpowermac:
> > I think we should fix the name of previous informer and create a new one based on namespace rather the name for proxy ca
> 
> @abhinavdahiya
> Could we create a shared informer for both? Or do we want to limit the namespaces?

Nevermind.

pkg/start/start.go
```diff
@@ -313,14 +313,18 @@ func (o *Options) NewControllerContext(cb *ClientBuilder) *Context {
 	cvInformer := externalversions.NewFilteredSharedInformerFactory(client, resyncPeriod(o.ResyncInterval)(), "", func(opts *metav1.ListOptions) {
 		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", o.Name)
 	})
-	cmInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigNamespace, func(opts *metav1.ListOptions) {
+	cmConfigInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigNamespace, func(opts *metav1.ListOptions) {
 		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", internal.InstallerConfigMap)
 	})
+	cmManagedInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigManagedNamespace, func(opts *metav1.ListOptions) {
+		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", internal.InstallerConfigMap)
```
abhinavdahiya: this is make it so that the informer will be restricted to only one object (named `internal.InstallerConfigMap`) in managed namespace. this will not cache the proxy-ca.

jcpowermac: ack

pkg/start/start.go
```diff
@@ -313,14 +313,18 @@ func (o *Options) NewControllerContext(cb *ClientBuilder) *Context {
...
+	})
+
 	sharedInformers := externalversions.NewSharedInformerFactory(client, resyncPeriod(o.ResyncInterval)())
 
 	ctx := &Context{
 		CVInformerFactory: cvInformer,
-		CMInformerFactory: cmInformer,
+		CMInformerFactory: cmConfigInformer,
```
abhinavdahiya: if you don't add the informer here, then the informer cannot be started here https://github.com/openshift/cluster-version-operator/blob/f443b93f775aa8d7040f5db2408bc444a4f55cfe/pkg/start/start.go#L366-L368

pkg/cvo/availableupdates.go
```diff
@@ -176,29 +182,59 @@ func checkForUpdate(clusterID string, proxyURL *url.URL, upstream, channel strin
...
 		}
 	}
-	return nil, nil
+	return nil, "", nil
+}
+
+func (optr *Operator) getTLSConfig(cmNameRef string) (*tls.Config, error) {
+	cm, err := optr.cmManagedLister.Get(cmNameRef)
+
+	if errors.IsNotFound(err) {
```
abhinavdahiya: can make this function fail on not found.... ?
what the effects of doing that..?

pkg/start/start.go
```diff
@@ -315,15 +314,20 @@ func (o *Options) NewControllerContext(cb *ClientBuilder) *Context {
 	cvInformer := externalversions.NewFilteredSharedInformerFactory(client, resyncPeriod(o.ResyncInterval)(), "", func(opts *metav1.ListOptions) {
 		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", o.Name)
 	})
-	cmInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigNamespace, func(opts *metav1.ListOptions) {
+	cmConfigInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigNamespace, func(opts *metav1.ListOptions) {
 		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", internal.InstallerConfigMap)
 	})
+	cmManagedInformer := informers.NewFilteredSharedInformerFactory(kubeClient, resyncPeriod(o.ResyncInterval)(), internal.ConfigManagedNamespace, func(opts *metav1.ListOptions) {
+		opts.FieldSelector = fmt.Sprintf("metadata.name=%s", internal.UserCAConfigMap)
```
abhinavdahiya: I don't think we can force ourselves to only objects with name `UserCAConfigMap`.. I think we can create a list for maybe keep it for entire namespace to begin with.

pkg/cvo/availableupdates.go
jcpowermac: I see your point.  Since we are checking if `cmNameRef` is not empty if the ConfigMap isn't found that is probably not a good thing.  

The latest commit would `return err`:
https://github.com/openshift/cluster-version-operator/pull/226/files#diff-78f2af341fa49292dd6930f378018867R49

pkg/start/start.go
jcpowermac: ack...will fix this tomorrow.

pkg/internal/constants.go
```diff
@@ -1,6 +1,8 @@
 package internal
 
 const (
-	ConfigNamespace    = "openshift-config"
-	InstallerConfigMap = "openshift-install"
+	ConfigNamespace        = "openshift-config"
+	ConfigManagedNamespace = "openshift-config-managed"
+	InstallerConfigMap     = "openshift-install"
+	UserCAConfigMap        = "user-ca-bundle"
```
abhinavdahiya: this is no longer use.

PR #228, opened 2019-08-01
cincinnati: add arch parameter to requests

crawford:
OpenShift is going to be gaining support for architectures other than
AMD64 soon. In order to make sure that the updates provided to a cluster
are compatible with the underlying architecture(s), we are adding an
"arch" parameter to the Cincinnati upstream queries. This new parameter
will contain a set of architectures, separated by commas, that need to
be supported by a release image in order for it to be applied to the
cluster. In a homogeneous deployment, the set of architectures will only
contain a single architecture (this is what this patch implements).
Eventually, once heterogeneous deployments are supported, the set may
contain multiple architectures. The details of heterogeneous deployments
and their constraints haven't been fully fleshed out, so this scheme may
change over time.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/url"
+	"runtime"
 	"time"
 
 	"github.com/blang/semver"
@@ -30,6 +31,7 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
 		usedDefaultUpstream = true
 		upstream = optr.defaultUpstreamServer
 	}
+	arch := runtime.GOARCH
 	channel := config.Spec.Channel
 
 	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
@@ -50,7 +52,7 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
 		}
 	}
 
-	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), proxyURL, tlsConfig, upstream, channel, optr.releaseVersion)
+	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), proxyURL, tlsConfig, upstream, arch, channel, optr.releaseVersion)
 
 	if usedDefaultUpstream {
 		upstream = ""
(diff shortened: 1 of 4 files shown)


steveeJ:
> @crawford: The following test **failed**, say `/retest` to rerun them all:

Seem like https://github.com/openshift/cluster-version-operator/blob/a5f5007c17cc14281c558ea363518dcc5b6675c7/pkg/start/start_integration_test.go#L695 needs adapting for the test to pass.

pkg/cvo/availableupdates.go
```diff
@@ -30,6 +31,7 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
 		usedDefaultUpstream = true
 		upstream = optr.defaultUpstreamServer
 	}
+	arch := runtime.GOARCH
```
jwforres: can I see an example of what this actually resolves to

crawford: Here's the full list of architectures: https://github.com/golang/go/blob/release-branch.go1.10/src/go/build/syslist.go#L8

steveeJ: This means we have to adhere to these strings wherever we tag builds and process these tags. Maybe we want to define our own list and map the GOARCH variants to our own?

crawford: I had thought about doing that, but I wanted to keep this client-side code as simple as possible. We can easily change the policies server-side, in the event that a mistake is discovered. Plus, we already have to adhere to those architectures in our build system.

steveeJ: > Plus, we already have to adhere to those architectures in our build system.

In that case I see no reason to put a mapping in between.

lucab:
> This new parameter will contain a set of architectures, separated by commas, that need to
be supported by a release image in order for it to be applied to the cluster.

@crawford is the CVO supposed to be sending this comma-separated list, or is the plan to have a policy-engine in front filtering per-architecture?

crawford:
@lucab The plan is to ignore support for heterogeneous clusters for now (that's why this patch only sends a single architecture), but not paint ourselves into a corner. Once we need to tackle heterogeneous, we'll update CVO again to consult something in the cluster (probably the infrastructure config) to determine the necessary architectures and then send that list in the request.

crawford:
There is general buy-in from an architecture perspective.

/hold cancel

pkg/cvo/availableupdates.go
lucab: There is a known caveat here, which is ARM 32bits. There, GOARCH is not enough to identify the architecture: https://github.com/golang/go/wiki/GoArm.

crawford: I only want to support the ARM64 Server spec. If we want to double back and support the Raspberry PI, we can figure out some mapping scheme.

PR #229, opened 2019-08-05
metrics: add cluster_installer series

crawford:
This series reports what installed the cluster. In order to do this, the
openshift-install ConfigMap, which is written by the installer when
doing an IPI install, is read. This ConfigMap is only injected during an
IPI installation and contains the invoker (e.g. username, Hive) as well
as the openshift-install version. The absence of this ConfigMap implies
a UPI installation. This data will allow us to determine whether a
cluster was installed IPI or UPI and, once CI and Hive have been
updated, whether the cluster was created by CI or Hive.

Because the data in this series never changes, it should compress
nicely so we shouldn't have to worry about wasting space.

If the data in the ConfigMap is missing, those label values are reported
as `<missing>`. That was used instead of an empty value in order to
distinguish between normal operation and an anomalous state. If the
openshift-install ConfigMap exists, it will have all of those fields
specified unless something were to erroneously remove them.

This is a backport of https://github.com/openshift/cluster-version-operator/pull/213.


crawford:
The release architects have agreed that there is no need for a valid Bugzilla for this particular backport due to it being purely telemetry-focused and because QE isn't able to verify.

/bugzilla valid-bug

PR #232, opened 2019-08-08
Bug 1717632: install: add alerts for cluster-version-operator

abhinavdahiya:
Adds alerts for cluster-version-operator and cluster operators

* `ClusterVersionOperatorDown`
  This alert is fired when cluster-version-operator is not providing any metrics. Serverity is critical as upgrades will not work and the clusters can drift from expected state.
* `ClusterOperatorDegraded`
  This alert is fired when the cluster operator is degraded. This is important as the cluster might be in an unacceptable state for produciton cluster, for example, using emptyDir for storage backend for registry. Severity is critical as
  degraded operator implies the operands are in a state that is not correct for the cluster.
* `ClusterOperatorDown`
  This alert fires when a cluster operator is not up ie cluster_operator_up is 0. This means that the operator might not be reconcile the operands.
* `ClusterOperatorFlapping`
  This alert fires when a cluster operator is flapping between up and down continously because of some weird condition.

### DIFF
--- a/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
+++ b/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
@@ -16,3 +16,49 @@ spec:
   selector:
     matchLabels:
       k8s-app: cluster-version-operator
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    k8s-app: cluster-version-operator
+  name: cluster-version-operator
+  namespace: openshift-cluster-version
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: ClusterVersionOperatorDown
+      annotations:
+        message: Cluster version operator has disappeared from Prometheus target discovery. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
+      for: 10m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} has not been available for 10 mins. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
+      expr: |
+        cluster_operator_up{job="cluster-version-operator"} == 0
+      for: 10m
+      labels:
+        severity: critical
+    - alert: ClusterOperatorDegraded
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} has been degraded for 10 mins. Operator is degraded because {{ "{{ $labels.reason }}" }} and cluster upgrades will be unstable.
+      expr: |
+        cluster_operator_conditions{job="cluster-version-operator", condition="Degraded"} == 1
+      for: 10m
+      labels:
+        severity: critical
+    - alert: ClusterOperatorFlapping
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} up status is changing often. This might cause upgrades to be unstable.
+      expr: |
+        changes(cluster_operator_up{job="cluster-version-operator"}[2m]) > 2
+      for: 10m
+      labels:
+        severity: warning
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -3,10 +3,10 @@ package cvo
 import (
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"github.com/prometheus/client_golang/prometheus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -247,7 +247,7 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 			break
 		}
 		g := m.clusterOperatorUp.WithLabelValues(op.Name, firstVersion)
-		failing := resourcemerge.IsOperatorStatusConditionTrue(op.Status.Conditions, ClusterStatusFailing)
+		failing := resourcemerge.IsOperatorStatusConditionTrue(op.Status.Conditions, configv1.OperatorDegraded)
 		available := resourcemerge.IsOperatorStatusConditionTrue(op.Status.Conditions, configv1.OperatorAvailable)
 		if available && !failing {
 			g.Set(1)
--- a/pkg/cvo/metrics_test.go
+++ b/pkg/cvo/metrics_test.go
@@ -174,7 +174,7 @@ func Test_operatorMetrics_Collect(t *testing.T) {
 								},
 								Conditions: []configv1.ClusterOperatorStatusCondition{
 									{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue},
-									{Type: ClusterStatusFailing, Status: configv1.ConditionTrue, LastTransitionTime: metav1.Time{Time: time.Unix(5, 0)}},
+									{Type: configv1.OperatorDegraded, Status: configv1.ConditionTrue, LastTransitionTime: metav1.Time{Time: time.Unix(5, 0)}},
 								},
 							},
 						},
@@ -188,7 +188,7 @@ func Test_operatorMetrics_Collect(t *testing.T) {
 				expectMetric(t, metrics[0], 0, map[string]string{"type": "current", "version": "", "image": "", "from_version": ""})
 				expectMetric(t, metrics[1], 0, map[string]string{"name": "test", "version": "10.1.5-1"})
 				expectMetric(t, metrics[2], 1, map[string]string{"name": "test", "condition": "Available"})
-				expectMetric(t, metrics[3], 1, map[string]string{"name": "test", "condition": "Failing"})
+				expectMetric(t, metrics[3], 1, map[string]string{"name": "test", "condition": "Degraded"})
 				expectMetric(t, metrics[4], 1, map[string]string{"type": ""})
 			},
 		},


install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+    k8s-app: cluster-version-operator
+  name: cluster-version-operator
+  namespace: openshift-cluster-version
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: ClusterVersionOperatorDown
+      annotations:
+        message: ClusterVersionOperator has disappeared from Prometheus target discovery.
```
smarterclayton: Probably better to indicate why someone might care as well as possible causes since this critical.  “Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible”

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
+      for: 15m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: ClusterOperator {{ $labels.name}} has not been available for 10 mins.
```
smarterclayton: I would prefer to check for degraded or not available.  Both are significant.

smarterclayton: Also explain impact here.

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+        severity: critical
+    - alert: ClusterOperatorFlapping
+      annotations:
+        message: ClusterOperator {{ $labels.name}} changing it's up status often
+      expr: |
+        changes(cluster_operator_up{job="cluster-version-operator"}[2m]) > 2
+      for: 2m
+      labels:
+        severity: critical
+
```
wking:
```console
$ git show --check --oneline origin/pr/232
297f6ca isntall: add alerts for cluster-version-operator
install/0000_90_cluster-version-operator_00_servicemonitor.yaml:57: new blank line at EOF.
```

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+      annotations:
+        message: ClusterOperator {{ $labels.name}} has not been available for 10 mins.
+      expr: |
+        cluster_operator_up{job="cluster-version-operator"}
+      for: 10m
+      labels:
+        severity: critical
+    - alert: ClusterOperatorFlapping
+      annotations:
+        message: ClusterOperator {{ $labels.name}} changing it's up status often
```
wking: nit: "it's" -> "its"

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+      for: 15m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: ClusterOperator {{ $labels.name}} has not been available for 10 mins.
+      expr: |
+        cluster_operator_up{job="cluster-version-operator"}
```
wking: Do we actually report this?  From my reading of [this loop][1], it's just for ClusterOperator objects, and that's the operators reporting to the CVO, not the CVO itself (e.g. no `cluster-version-operator`) in [here][2].  Or is this "`cluster_operator_up` reported by the cluster-version-operator job" or some such?  Can you unpack this expression for me?

[1]: https://github.com/openshift/cluster-version-operator/blob/9a6c1125a72bcef276847e489eac8d580ccd1470/pkg/cvo/metrics.go#L241-L242
[2]: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_console/2261/pull-ci-openshift-console-master-e2e-aws/6623/artifacts/e2e-aws/must-gather/cluster-scoped-resources/config.openshift.io/clusteroperators/

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
wking: > I would prefer to check for degraded or not available.

Isn't that what [the current `available && !failing`][1] is doing?  If you're not available, we're not setting `cluster-operator-up` for you.  If you're available but degraded (which backs the `failing` variable), we're not setting it either.

[1]: https://github.com/openshift/cluster-version-operator/commit/72e8f1a9d1c51d2fbb62ce4b61de37b183298489#diff-b301ca75c2e59524b215636eb7a4586aR252

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
```diff
@@ -16,3 +16,42 @@ spec:
...
+  namespace: openshift-cluster-version
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: ClusterVersionOperatorDown
+      annotations:
+        message: ClusterVersionOperator has disappeared from Prometheus target discovery.
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
```
wking: This is likely to be sufficient on its own, but we may want `namespace="openshift-cluster-version"` in the filter as well.

wking:
/title Bug 1717632: install: add alerts for cluster-version-operator

That's what this is fixing, right?

wking:
Oops:

/retitle Bug 1717632: install: add alerts for cluster-version-operator

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/pull/232/commits/4a9719f52fb3a4b5fe44eed7bf8f68adb58ec276

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
abhinavdahiya: added `ClusterOperatorDegraded` in https://github.com/openshift/cluster-version-operator/pull/232/commits/48237bc2395f66194c99017b84406ce70aee8256

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
abhinavdahiya: @wking done with https://github.com/openshift/cluster-version-operator/pull/232/commits/136fb720be157e62cf63b560e65304898c40d989

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
abhinavdahiya: i'm not sure if other actions in the cluster can also emit something with name `cluster_operator_up`. so just want to make sure it's on the time series emitted by cluster-version-operator.

currently there's only one cvo, so namespace should be fine.

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
abhinavdahiya: https://github.com/openshift/cluster-version-operator/pull/232#discussion_r313140916

install/0000_90_cluster-version-operator_00_servicemonitor.yaml
wking: > currently there's only one cvo, so namespace should be fine.

Couldn't end users create jobs with that name?  Can we rely on not scraping user namespaces?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+    k8s-app: cluster-version-operator
+  name: cluster-version-operator
+  namespace: openshift-cluster-version
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: ClusterVersionOperatorDown
+      annotations:
+        message: ClusterVersionOperator has disappeared from Prometheus target discovery. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
```
smarterclayton: Would be tempted to say `Cluster version operator has ...` to make it flow as a message in more contexts.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
+      for: 15m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: ClusterOperator {{ "{{ $labels.name }}" }} has not been available for 10 mins. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
```
smarterclayton: Probably `Cluster operator {{ ... }}`

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      annotations:
+        message: ClusterOperator {{ "{{ $labels.name }}" }} has been degraded for 10 mins. Operator is degraded because {{ "{{ $labels.reason }}" }} and cluster upgrades will be unstable.
+      expr: |
+        cluster_operator_conditions{job="cluster-version-operator", condition="Degraded"} == 1
+      for: 10m
+      labels:
+        severity: critical
+    - alert: ClusterOperatorFlapping
+      annotations:
+        message: ClusterOperator {{ "{{ $labels.name }}" }} changing its up status often. This might cause upgrades to be unstable.
```
smarterclayton: Might suggest `... up status is changing often.  This might ...`.

Do pods already have an alert that fires if they're restarting?  I think this list is probably sufficient.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/pull/232/commits/e7bc61e84d6d0f4f7161644378eed9478f58cc5d

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/pull/232/commits/e7bc61e84d6d0f4f7161644378eed9478f58cc5d

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: @smarterclayton done with https://github.com/openshift/cluster-version-operator/pull/232/commits/e7bc61e84d6d0f4f7161644378eed9478f58cc5d

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
+      for: 15m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} has not been available for 10 mins. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
```
brancz: Why this alert as opposed to an alert like `ClusterVersionOperatorDown` for each cluster operator? As I understand this alert has multiple indirections which we want to avoid at all costs in monitoring.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      labels:
+        severity: critical
+    - alert: ClusterOperatorDegraded
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} has been degraded for 10 mins. Operator is degraded because {{ "{{ $labels.reason }}" }} and cluster upgrades will be unstable.
+      expr: |
+        cluster_operator_conditions{job="cluster-version-operator", condition="Degraded"} == 1
+      for: 10m
+      labels:
+        severity: critical
```
brancz: I understand the intend of this, but a cluster administrator is not going to know what to do here. This goes back to my general critizism about using `cluster_operator_conditions` for any type of alerting. Apart from the indirections it imposes, this is the wrong approach to monitoring. Instead we should be setting SLOs for each component, and alert on the failure to deliver those. Those alerts can then have clear runbooks as to what actionable things can be done by the cluster administrator about this particular component.

I think we can have alerts that go in a similar direction as thi this on the _telemeter_ side, so our engineering teams can act on this, but I don't think this is useful for customers as is. In fact I think it's harmful as it produces alert fatigue.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      for: 15m
+      labels:
+        severity: critical
+  - name: cluster-operators
+    rules:
+    - alert: ClusterOperatorDown
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} has not been available for 10 mins. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
+      expr: |
+        cluster_operator_up{job="cluster-version-operator"} == 0
```
brancz: How is this metric determined? And I assume this metric is somehow synthetically produced by the cluster-version-operator?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      expr: |
+        cluster_operator_conditions{job="cluster-version-operator", condition="Degraded"} == 1
+      for: 10m
+      labels:
+        severity: critical
+    - alert: ClusterOperatorFlapping
+      annotations:
+        message: Cluster operator {{ "{{ $labels.name }}" }} up status is changing often. This might cause upgrades to be unstable.
+      expr: |
+        changes(cluster_operator_up{job="cluster-version-operator"}[2m]) > 2
```
brancz: 2m range selector doesn't seem to cover sufficient failure cases, also this behavior seems highly specific to each operator. Some operators have a very short reconciliation loop, some have a very long one. I don't think we're going to be able to find any good general purpose value for this.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+      for: 2m
```
brancz: This seems a little short.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: Cluster version operator is not a SLO. it does not have a cluster operator. that's why the `ClusterVersionOperatorDown` right above.

abhinavdahiya: > As I understand this alert has multiple indirections which we want to avoid at all costs in monitoring.

hmm, do you mean this alert can point to as many cluster operators.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: > How is this metric determined? And I assume this metric is somehow synthetically produced by the cluster-version-operator?

what do we gain from knowing how this is generated?

it looks at clusteroperator and reports 1 when `available & ! degraded` else 0

https://github.com/openshift/cluster-version-operator/blob/e47c7784427479b738a596f7f633e2448ddf2dac/pkg/cvo/metrics.go#L262-L270

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: I think we want the user to actually be alerted when an cluster operator is degraded for that long.

And can you explain why is there a requirement for a runbook for each alert?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: The clusteroperators should not be flapping between conditions. It has nothing to do with reconcile times.

> https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource

for example,
> An operator reports Progressing when it is rolling out new code, propagating config changes, or otherwise moving from one steady state to another. It should not report progressing when it is reconciling a previously known state

that makes it pretty clear that reconciles of current behavior making clusteroperator progressing is not recommended.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: we want to alert when the operators are flapping as that is not the normal behavior.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
abhinavdahiya: do we have a better suggestion for the severity?? any guidelines that we can follow?

wking: > In fact I think it's harmful as it produces alert fatigue.

I'd rather avoid alert fatigue on these by not hitting these thresholds, not by not monitoring the thresholds :).  All of these things seem like pretty bad "I'd want to know first thing in the morning if this was happening" things.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -16,3 +16,49 @@ spec:
...
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: ClusterVersionOperatorDown
+      annotations:
+        message: Cluster version operator has disappeared from Prometheus target discovery. Operator may be down or disabled, cluster will not be kept up to date and upgrades will not be possible.
+      expr: |
+        absent(up{job="cluster-version-operator"} == 1)
+      for: 15m
```
wking: Why 15m here and 10m for the other operators?  I don't have strong opinions, but I'm curious about the motivation and think we should record why we picked different values in the commit message to inform future devs interested in changing the defaults.  Maybe "Prometheus scrapes every `${whatever}` minutes, so allow the operators to be down for three scrapes before freaking out", or some such.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
wking: There's probably a whole here where a rolling outage across operators could trigger the alert without having a single operator down for the full threshold.  For example:

1. Scrape 1, operator A is down.
2. Scrape 2, operator B is down; A is back up.
3. Scrape 3, operator C is down; B is back up.
4. Alert fires with "there has been at least one down operator for the last three scrapes (over 10m)"

I'm not sure how to adjust this query to avoid that though, although it plays into the "many operators behind one alert" point you raise [here][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/232/files#r315432978

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: No, it's an alert on a metric that is not exposed by the process of origin. As in the cluster-version-operator is exposing a metric on behalf of another operator, and even worse, with information provided by the other operator. This requires that _everything_ is working properly to provide a failure case. This is why metrics should be exposed whereever the origin of any information is.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: It's unclear that that's what you want to test for. Also this sounds like it rather belongs in an e2e test than an alert.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: These are metrics that should be exposed by the operators themselves and not by the CVO. That would remove the indirections, and we can make alerts concrete and actionable. As they are, these alerts, except for the very first one are not actionable, nor do they follow the best practice of metrics coming from the origin.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: Critical severity is for when you page someone and there is something concrete must be fixed ASAP. I would suggest severity warning for something like this, but it's still too vague of an alert that I would send it to anyone, as it's unclear what to do on this alert.

Even if against our instinct, removing/not using flaky alerts is better than causing alert fatigue, alert fatigue causes people to not trust in our alerts and ignore all of them.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: We need to know whether it is happening in the wild.  I think we’ve shown conclusively in 4.1 with telemetry that our internal testing of operators is insuffficient to catch what is happening in the wild.

I want to prove these things aren’t happening in the wild (and we have evidence it is) and maybe we can remove the alert later if it isn’t firing in the wild.  But because this is a critical control loop extra alerts are a good thing.

pkg/cvo/metrics.go
```diff
@@ -247,7 +247,7 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 			break
 		}
 		g := m.clusterOperatorUp.WithLabelValues(op.Name, firstVersion)
-		failing := resourcemerge.IsOperatorStatusConditionTrue(op.Status.Conditions, ClusterStatusFailing)
+		failing := resourcemerge.IsOperatorStatusConditionTrue(op.Status.Conditions, configv1.OperatorDegraded)
```
smarterclayton: Ugh

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: We are monitoring the core control loop of the cluster.  We have to monitor both, not just one or the other.  This alert reflects the “making progress” part of the product - operator A and B coordinate via this api, not via alerts and metrics.  We *should* have per operator monitoring.  We *must* have core product control loop monitoring and alerting.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: See my other comment, but the metrics are part of the product.  An individual operator doesn’t report whether other operators see the result.  We *must* see the core loop view like this.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: The point of this alert is to wake up an engineer because they are broken.  We are increasingly targeting alerts towards our engineering / monitoring team, not just the customers.

Alert fatigue is real but this is a critical failure of a core operator, and we have to react to it.

brancz: > We are increasingly targeting alerts towards our engineering / monitoring team, not just the customers.

Then this should be an alert evaluated by telemetry infrastructure, not in-cluster.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: If we move forward with this PR, then I would want to see *concrete* actions as to how we will properly resolve this in the very next release (4.3), which would be the _requirement_ for each cluster-operator to:

* expose identify and expose metrics to describe their SLIs
* only be allowed to page customers (that includes our own SRE) with alerts that describe error budget is burnt at a high rate
* not include alerts that are meant for our engineering team's, these are to be done via telemetry infrastructure (as in alert on data in telemetry)

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: To prove this isn't happening in the wild severity warning is sufficient. I am ok with adding this to prove that it's not happening.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
brancz: I understand what you mean by these metrics describing the "advancing of the CVO", and I'm not arguing that it's not important, however, it is not clear to an SRE (even our SRE I've talked to them) if they are paged with this alert today what an actionable thing to troubleshoot is, not even what's broken. As is, this is at most actionable for our own engineering. As I said in the other comment, I want us to have a plan for how we will get to a state where this _is_ clear and it _is_ clear what about the product is not working when it's not. This needs to be top priority for the next release(s).

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: Degraded operators are functional indistinguishable from broken clusters.  We have alert because we know something is wrong.

Evidence wise I haven’t seen more than a small percentage of degraded operators that were not error conditions that should result in a customer requesting a bug fix for users.

For disconnected users alerts are mandatory on cluster to inform them to contact red hat.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
smarterclayton: I disagree - we will have customers without telemetry.

Our job is to fix bugs rapidly.  Alerts can be generated by both bugs, user error, and unanticipated states.

It would be irresponsible of us not to alert these conditions on cluster for all disconnected users.

PR #233, opened 2019-08-12
Bug 1739988: Stop double reporting "failure" cluster_version

smarterclayton:
We currently report failure twice. Now that we have from_version,
we can report the previous completed version.

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -3,10 +3,10 @@ package cvo
 import (
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"github.com/prometheus/client_golang/prometheus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


abhinavdahiya:
do we need to update the docs at https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/metrics.md#cvo-metrics for `failure` ??

pkg/cvo/metrics.go
```diff
@@ -194,9 +195,10 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 					g.Set(float64(failing.LastTransitionTime.Unix()))
 				}
 				ch <- g
+				setFailing = true
 			}
 		}
-		if failing != nil && failing.Status == configv1.ConditionTrue {
+		if !setFailing && failing != nil && failing.Status == configv1.ConditionTrue {
```
abhinavdahiya: should we DRY it up with something like:
```go
if failing != nil && failing.Status == configv1.ConditionTrue {
    g := m.version.WithLabelValues("failure", current.Version, current.Image, completed.Version)
    if update := cv.Spec.DesiredUpdate; update != nil && update.Image != current.Image {
       g = m.version.WithLabelValues("failure",update.Version, update.Image, completed.Version)
    }
    if failing.LastTransitionTime.IsZero() {
        g.Set(0)
    } else {
        g.Set(float64(failing.LastTransitionTime.Unix()))
    }
    ch <- g
}
```

smarterclayton: Yeah that might be better

smarterclayton:
Updated with comments addressed

PR #236, opened 2019-08-14
Bug 1741645: pkg/cvo/metrics: Report cluster-version conditions with reasons

wking:
This will allow us to discover upgrade and other failure reasons without having to resort to a must-gather or similar.

Stick this in `cluster_operator_conditions`, since we already have a `reason` slot there.  I don't see a reason to add a new metric to separate cluster-version operator failures from second-level operator failures; the name should be sufficient for that.

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -227,6 +227,19 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 			g.Set(float64(len(cv.Status.AvailableUpdates)))
 			ch <- g
 		}
+
+		for _, condition := range cv.Status.Conditions {
+			if condition.Status == configv1.ConditionUnknown {
+				continue
+			}
+			g := m.clusterOperatorConditions.WithLabelValues("version", string(condition.Type), string(condition.Reason))
+			if condition.Status == configv1.ConditionTrue {
+				g.Set(1)
+			} else {
+				g.Set(0)
+			}
+			ch <- g
+		}
 	}
 
 	g := m.version.WithLabelValues("current", current.Version, current.Image, completed.Version)
--- a/pkg/cvo/metrics_test.go
+++ b/pkg/cvo/metrics_test.go
@@ -269,7 +269,7 @@ func Test_operatorMetrics_Collect(t *testing.T) {
 							},
 							Status: configv1.ClusterVersionStatus{
 								Conditions: []configv1.ClusterOperatorStatusCondition{
-									{Type: configv1.RetrievedUpdates, Status: configv1.ConditionTrue},
+									{Type: configv1.RetrievedUpdates, Status: configv1.ConditionTrue, Reason: "Because stuff"},
 								},
 							},
 						},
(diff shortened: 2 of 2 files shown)


wking:
This may overlap with #232 as a way to get ClusterVersion failure reasons out into Telemetry.  Are the parallel tracks (alerts and `cluster_operator_conditions` metrics) a problem?  Do we want to consolidate on alerts and drop `cluster_operator_conditions` altogether?  Or do we want both for a belt-and-suspenders approach, or just because each channel gives us slightly different information (e.g. `cluster_operator_conditions` may make it easier to get failure rates, because we still push the metrics, with a zero value, when the condition is not active).

pkg/cvo/metrics.go
```diff
@@ -196,12 +196,22 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
...
+				g := m.version.WithLabelValues("failure", current.Version, current.Image, completed.Version)
+				if failing.LastTransitionTime.IsZero() {
+					g.Set(0)
+				} else {
+					g.Set(float64(failing.LastTransitionTime.Unix()))
+				}
+				ch <- g
+			}
+
+			g := m.clusterOperatorConditions.WithLabelValues("cluster-version", "Failing", string(failing.Reason))
```
abhinavdahiya: the name should be `version`. all the clusteroperators are already cluster.. it is redundant. Also the ClusterVersion object is also named version.

abhinavdahiya: Also we should probably just push all the conditions from the clusterversion object like https://github.com/openshift/cluster-version-operator/blob/b69cd4847fab21a74a15544cc66a1d69dab63f31/pkg/cvo/metrics.go#L268-L279. The conditions are same as other clusteroperators.. probably here https://github.com/openshift/cluster-version-operator/blob/b69cd4847fab21a74a15544cc66a1d69dab63f31/pkg/cvo/metrics.go#L281

abhinavdahiya:
we should also create a bug for this.

pkg/cvo/metrics.go
wking: Fixed in b69cd48 -> d881086.

wking: > Also we should probably just push all the conditions...

Done with d881086 -> 240244c.

wking:
/retitle Bug 1741645: pkg/cvo/metrics: Report cluster-version conditions with reasons

wking:
[e2e-aws][1]:

```
level=error msg="Error: NoSuchBucket: The specified bucket does not exist" level=error msg="\tstatus code: 404, request id: CE2EC2492682F231, host id: Hzp7AoarEa//ZVR8XEoblAxIFZ2A8AipR3lyGzeCjGxO69/oSHeCNZap3tbF2g/T+mdcd/xmEzo="
```

Haven't seen that one recently...

/test e2e-aws

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/236/pull-ci-openshift-cluster-version-operator-master-e2e-aws/873

wking:
[e2e-aws][1]:

```
Aug 15 20:25:46.283: INFO: Couldn't delete ns: "e2e-svcaccounts-2403": namespace e2e-svcaccounts-2403 was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed (&errors.errorString{s:"namespace e2e-svcaccounts-2403 was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed"}) 
```

That's [rhbz#1727090][2].

/test e2e-aws

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/236/pull-ci-openshift-cluster-version-operator-master-e2e-aws/876#0:build-log.txt%3A6218
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1727090

PR #238, opened 2019-08-16
Bug 1741786: pkg/cvo: Drop ClusterVersion defaulting during bootstrap

wking:
This avoids a race between:

* The cluster-version operator pushing its internally-generated default and
* The cluster-bootstrap process pushing the installer-generated ClusterVersion into the cluster ([rhbz#1741786][1])

Because cluster-bootstrap does not override already-existing objects, when the CVO's default won that race, the cluster would come up with a bogus `channel` (`fast`, instead of the installer's default `stable-4.2`, etc.) and `clusterID` (causing the reported Telemetry to come in under a different UUID than the one the installer [provided in its `metadata.json` output][2]).

By removing the default, the CVO will continue applying the current version and reporting some metrics like `cluster_version{type="current"}`.  But it will not report metrics that depend on the state stored in the ClusterVersion object (past versions, `cluster_version{type="failure"}`, etc.).  And even the metrics it does provide may not make it up to the Telemetry API because the monitoring operator will not be able to pass [the Telemeter container][3] the [expected `clusterID`][4].  I'm not clear on how the Telemeter config is updated when the ClusterVersion's `clusterID` *does* change because of [this guard][5], but the monitoring operator is [willing to continue without it][6].  The monitoring operator only updates the Telemeter Deployment to [*set* (or update) the cluster ID][7]; it never clears an existing cluster ID.

While we won't get Telemetry out of the cluster until the cluster-bootstrap process (or somebody else) pushes the ClusterVersion in, we can still alert on it locally.  I haven't done that in this PR, because #232 is in flight touching this space.

The defaulting logic dates back to the early CVOConfig work in ea678b1af4 (#2).  @smarterclayton [expressed concern][9] about recovering from ClusterVersion deletion without the default, but:

* Who deletes their ClusterVersion?

* Even if someone wanted to delete their ClusterVersion, we'll want to eventually make the CVO an admission gate for ClusterVersion changes (ab4d84a021).  So (once we set up that admission gate), someone determined to remove ClusterVersion would have to disable that admission config before they could remove the ClusterVersion.  That seems like enough steps that you wouldn't blow ClusterVersion away by accident.

* If you do successfully remove your ClusterVersion, you're going to lose all the status and history it contained.  If you scale down your CVO and then remove your ClusterVersion, it's going to be hard to recover that lost information.  If you left the CVO running, we could presumably cache the whole thing in memory and push it back into the cluster.  That would at least avoid the "compiled-in CVO defaults differ from user's choices" issue.  But we'd still be fighting with the user over the presence of ClusterVersion, and I'd rather not.

* Telemetry will continue pushing reports with the last-seen clusterID, so unless you disabled Telemetry first, future "ClusterVersion is missing" alerts (not added in this commit, see earlier paragraph mentioning alerts) would make it out to let us know that something was being silly.

I'm pushing this up to get feedback on the approach (CC @abhinavdahiya).  The WIP is because I haven't updated the unit test to handle the lack of defaulting.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1741786
[2]: https://github.com/openshift/installer/blob/4e204c5e509de1bd31113b0c0e73af1a35e52c0a/pkg/types/clustermetadata.go#L17-L18
[3]: https://github.com/openshift/cluster-monitoring-operator/blob/14b1093149217a6ab5e7603f19ff5449f1ec12fc/manifests/0000_50_cluster_monitoring_operator_04-deployment.yaml#L62
[4]: https://github.com/openshift/cluster-monitoring-operator/blob/14b1093149217a6ab5e7603f19ff5449f1ec12fc/pkg/manifests/config.go#L217-L229
[5]: https://github.com/openshift/cluster-monitoring-operator/blob/14b1093149217a6ab5e7603f19ff5449f1ec12fc/pkg/operator/operator.go#L369-L370
[6]: https://github.com/openshift/cluster-monitoring-operator/blob/14b1093149217a6ab5e7603f19ff5449f1ec12fc/pkg/operator/operator.go#L376
[7]: https://github.com/openshift/cluster-monitoring-operator/blob/14b1093149217a6ab5e7603f19ff5449f1ec12fc/pkg/manifests/manifests.go#L1762-L1763
[9]: https://bugzilla.redhat.com/show_bug.cgi?id=1708697#c12

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -28,6 +28,7 @@ func init() {
 	cmd.PersistentFlags().StringVar(&opts.Kubeconfig, "kubeconfig", opts.Kubeconfig, "Kubeconfig file to access a remote cluster (testing only)")
 	cmd.PersistentFlags().StringVar(&opts.NodeName, "node-name", opts.NodeName, "kubernetes node name CVO is scheduled on.")
 	cmd.PersistentFlags().BoolVar(&opts.EnableAutoUpdate, "enable-auto-update", opts.EnableAutoUpdate, "Enables the autoupdate controller.")
+	cmd.PersistentFlags().BoolVar(&opts.EnableDefaultClusterVersion, "enable-default-cluster-version", opts.EnableDefaultClusterVersion, "Allows the operator to create a ClusterVersion object if one does not already exist.")
 	cmd.PersistentFlags().StringVar(&opts.ReleaseImage, "release-image", opts.ReleaseImage, "The Openshift release image url.")
 	rootCmd.AddCommand(cmd)
 }
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -23,6 +23,7 @@ spec:
           - "start"
           - "--release-image={{.ReleaseImage}}"
           - "--enable-auto-update=false"
+          - "--enable-default-cluster-version=true"
           - "--v=4"
         resources:
           requests:
--- a/bootstrap/bootstrap-pod.yaml
+++ b/bootstrap/bootstrap-pod.yaml
@@ -14,6 +14,7 @@ spec:
       - "start"
       - "--release-image={{.ReleaseImage}}"
       - "--enable-auto-update=false"
+      - "--enable-default-cluster-version=false"
       - "--v=4"
       - "--kubeconfig=/etc/kubernetes/kubeconfig"
     securityContext:
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -85,6 +85,10 @@ type Operator struct {
 	// releaseCreated, if set, is the timestamp of the current update.
 	releaseCreated time.Time
 
+	// enableDefaultClusterVersion allows the operator to create a
+	// ClusterVersion object if one does not already exist.
+	enableDefaultClusterVersion bool
+
 	client        clientset.Interface
 	kubeClient    kubernetes.Interface
 	eventRecorder record.EventRecorder
@@ -133,6 +137,7 @@ func New(
 	nodename string,
 	namespace, name string,
 	releaseImage string,
+	enableDefaultClusterVersion bool,
 	overridePayloadDir string,
 	minimumInterval time.Duration,
 	cvInformer configinformersv1.ClusterVersionInformer,
@@ -152,6 +157,8 @@ func New(
 		namespace:    namespace,
 		name:         name,
 		releaseImage: releaseImage,
+
+		enableDefaultClusterVersion: enableDefaultClusterVersion,
 
 		statusInterval:             15 * time.Second,
 		minimumUpdateCheckInterval: minimumInterval,
@@ -349,12 +356,16 @@ func (optr *Operator) sync(key string) error {
 
 	// ensure the cluster version exists, that the object is valid, and that
 	// all initial conditions are set.
-	original, changed, err := optr.getOrCreateClusterVersion()
+	original, changed, err := optr.getOrCreateClusterVersion(optr.enableDefaultClusterVersion)
 	if err != nil {
 		return err
 	}
 	if changed {
 		klog.V(4).Infof("Cluster version changed, waiting for newer event")
+		return nil
+	}
+	if original == nil {
+		klog.V(4).Infof("No ClusterVersion object and defaulting not enabled, waiting for one")
 		return nil
 	}
 
@@ -448,7 +459,7 @@ func (optr *Operator) rememberLastUpdate(config *configv1.ClusterVersion) {
 	optr.lastResourceVersion = i
 }
 
-func (optr *Operator) getOrCreateClusterVersion() (*configv1.ClusterVersion, bool, error) {
+func (optr *Operator) getOrCreateClusterVersion(enableDefault bool) (*configv1.ClusterVersion, bool, error) {
 	obj, err := optr.cvLister.Get(optr.name)
 	if err == nil {
 		// if we are waiting to see a newer cached version, just exit
@@ -460,6 +471,10 @@ func (optr *Operator) getOrCreateClusterVersion() (*configv1.ClusterVersion, boo
 
 	if !apierrors.IsNotFound(err) {
 		return nil, false, err
+	}
+
+	if !enableDefault {
+		return nil, false, nil
 	}
 
 	var upstream configv1.URL
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -69,11 +69,12 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 	})
 
 	o := &Operator{
-		namespace: "test",
-		name:      "version",
-		queue:     workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cvo-loop-test"),
-		client:    client,
-		cvLister:  &clientCVLister{client: client},
+		namespace:                   "test",
+		name:                        "version",
+		enableDefaultClusterVersion: true,
+		queue:                       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cvo-loop-test"),
+		client:                      client,
+		cvLister:                    &clientCVLister{client: client},
 	}
 
 	dynamicScheme := runtime.NewScheme()
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -255,11 +255,12 @@ func TestOperator_sync(t *testing.T) {
 		{
 			name: "create version and status",
 			optr: Operator{
-				releaseVersion: "4.0.1",
-				releaseImage:   "image/image:v4.0.1",
-				namespace:      "test",
-				name:           "default",
-				client:         fake.NewSimpleClientset(),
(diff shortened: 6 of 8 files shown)


wking:
Pulled the WIP, because 26e83b2 passes unit tests locally for me.  We also have [a few rounds][1] of green e2e-aws(-upgrade).  I dunno what's going on with [the two `TestIntegrationCVO_*` failures][2] yet. 

[1]: https://prow.svc.ci.openshift.org/pr-history/?org=openshift&repo=cluster-version-operator&pr=238
[2]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/238/pull-ci-openshift-cluster-version-operator-master-integration/515#0:build-log.txt%3A1327

wking:
All green :).  @abhinavdahiya, @smarterclayton, please take a look.

smarterclayton:
/hold

Why are we creating cluster version during install and why do we have to?  Your description alluded to it but didn’t lay out the current justifications.  Originally CV was moving to not installer created

smarterclayton:
To explore the counter, why does channel need to be set by installer?  Metadata.json having it is a bit weird.

The CVO shouldn’t be syncing anything out of the payload until it has written the CV, so the bug here seems to be with bootstrap using idempotent create for objects instead of idenpotent create or update?

smarterclayton:
Also alternatively, perhaps the bootstrap process should be passing intiital cluster id and channel to cvo to create bootstrap deployment,

wking:
> Why are we creating cluster version during install and why do we have to? 

Historically, this was used to override broken components, but we haven't needed to do that since OpenShift/installer#851.  Although obviously we could have been managing that from this repo instead of managing it via the installer repo.

> Metadata.json having it is a bit weird.

Yeah.  In fact, Cluster version having it is a bit weird ;).  If it's just for telemetry, shouldn't it just be in [the Telemetry config][1]?  Is telemetry/monitoring represented by a custom resource?  Nothing stands out to me [in the API][2].  Maybe configuration is via [a ConfigMap][3]?

At the moment, Hive is using the telemetry from `metadata.json` to hold space for "coming OSD cluster".  It could pull the Telemetry ID put of the running cluster instead, but then OSD-ness would race cluster Telemetry reporting vs. Hive's extraction.  Until we get OSD-ness via the `cluster_installer` metrics (openshift/installer#1890 openshift/cluster-version-operator#213), which may be where we want it long-term (we'd still need a way to associate Hive-reported admin creds with the appropriate Telemetry ID).

> The CVO shouldn’t be syncing anything out of the payload until it has written the CV, so the bug here seems to be with bootstrap using idempotent create for objects instead of idenpotent create or update?

I don't like multiple parties having opinions about cluster config specs.  Components that have opinions about their own config could represent their defaults in `status`, but should stay out of `spec`, which should be admin-managed for these top-level configs (and installer-managed as a proxy for the admin at install-time).  I'd be fine with cluster-bootstrap clobbering `spec` (repeatedly?) until all resource `spec`s matched its expectations, but that's a reasonably large change vs. its current behavior.

> Also alternatively, perhaps the bootstrap process should be passing intiital cluster id and channel to cvo to create bootstrap deployment...

As in "but not via ClusterVersion"?  I'm fine just leaving channel/upstream defaulting to the CVO.  And if we adjust Telemetry ID injection, I'd rather move it to an in-cluster Telemetry custom resource.

[1]: https://github.com/openshift/cluster-monitoring-operator/blob/fbd0e89e8a4fc72a993b64fb80c79145c78f8b67/pkg/manifests/config.go#L125
[2]: https://github.com/openshift/api/tree/a94e914914f4228d0bcba6fc8a22614c5f5e2dad/config/v1
[3]: https://github.com/openshift/cluster-monitoring-operator/blob/e64d3f1b595959ae9cd4388d7a4fa5d41b5f3a5a/pkg/operator/operator.go#L338

smarterclayton:
To the last questions, "no", it should be as "tell the bootstrap CVO
command how you want to generate its initial resources via arguments".  Or
pass the default CV object to bootstrap CVO command.  Or inject the CV
immediately after the CRD is created.  There are lots of options for taking
the data the installer has and ensuring the CVO gets to handle the
appropriate ordering.

I think cluster-bootstrap is insufficiently intelligent (or the bootstrap
commands it invokes are insufficiently intelligent).  Passing those flags
to CVO bootstrap seems the closest to actual intent.

abhinavdahiya:
> Yeah. In fact, ClusterVersion having it is a bit weird ;). If it's just for telemetry, shouldn't it just be in [the Telemetry config](https://github.com/openshift/cluster-monitoring-operator/blob/fbd0e89e8a4fc72a993b64fb80c79145c78f8b67/pkg/manifests/config.go#L125)? Is telemetry/monitoring represented by a custom resource? Nothing stands out to me [in the API](https://github.com/openshift/api/tree/a94e914914f4228d0bcba6fc8a22614c5f5e2dad/config/v1). Maybe configuration is via [a ConfigMap](https://github.com/openshift/cluster-monitoring-operator/blob/e64d3f1b595959ae9cd4388d7a4fa5d41b5f3a5a/pkg/operator/operator.go#L338)?

The cluster version ID is the UUID for the cluster. We use it to identify the cluster in telemetry on red hat side, also the cincinatti stacks needs a way to identify the cluster.

So I think the ID definitely belongs to the clusterversion object.

wking:
> Or pass the default CV object to bootstrap CVO command.

I've pushed 0361562 -> e816739 pivoting from "drop all ClusterVersion defaulting" to "drop ClusterVersion defaulting during bootstrap".  That's effectively "pass the installer-provided ClusterVersion as the default", and it does not affect the production (post-bootstrap) operator.  Thoughts?

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -23,6 +23,7 @@ spec:
           - "start"
           - "--release-image={{.ReleaseImage}}"
           - "--enable-auto-update=false"
+          - "--enable-default-cluster-version={{.Bootstrap|not}}"
```
abhinavdahiya: don't change this... rather the https://github.com/openshift/cluster-version-operator/blob/master/bootstrap/bootstrap-pod.yaml

abhinavdahiya:
we should keep the defaulting to true, and only update the bootstrap-cvo pod to use defaulting to false.

wking:
> we should keep the defaulting to true, and only update the bootstrap-cvo pod...

Can we just make it explicit in both `start`-calling locations?  I've pushed e816739 -> d7760ce to do that and fix the integration failures, now that you've [pointed out][1] that we use `bootstrap/bootstrap-pod.yaml` for bootstrapping and not `install/0000_00_cluster-version-operator_03_deployment.yaml`.

[1]: https://github.com/openshift/cluster-version-operator/pull/238#discussion_r315954659

install/0000_00_cluster-version-operator_03_deployment.yaml
wking: > don't change this... rather the https://github.com/openshift/cluster-version-operator/blob/master/bootstrap/bootstrap-pod.yaml

[Done](https://github.com/openshift/cluster-version-operator/pull/238#issuecomment-523274889).

wking:
Upgrade failure was an unrelated flake:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/238/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/278/artifacts/e2e-aws-upgrade/must-gather/namespaces/openshift-cluster-version/pods/cluster-version-operator-7f597b6d9b-rtmr8/cluster-version-operator/cluster-version-operator/logs/current.log | tail -n1
2019-08-21T05:22:44.288426934Z I0821 05:22:44.288364       1 apps.go:115] Deployment marketplace-operator is not ready. status: (replicas: 1, updated: 1, ready: 0, unavailable: 1, reason: MinimumReplicasUnavailable, message: Deployment does not have minimum availability.)
```

/retest

wking:
All green :).

smarterclayton:
/hold cancel

Thanks

cmd/start.go
```diff
@@ -28,6 +28,7 @@ func init() {
 	cmd.PersistentFlags().StringVar(&opts.Kubeconfig, "kubeconfig", opts.Kubeconfig, "Kubeconfig file to access a remote cluster (testing only)")
 	cmd.PersistentFlags().StringVar(&opts.NodeName, "node-name", opts.NodeName, "kubernetes node name CVO is scheduled on.")
 	cmd.PersistentFlags().BoolVar(&opts.EnableAutoUpdate, "enable-auto-update", opts.EnableAutoUpdate, "Enables the autoupdate controller.")
+	cmd.PersistentFlags().BoolVar(&opts.EnableDefaultClusterVersion, "enable-default-cluster-version", opts.EnableDefaultClusterVersion, "Allows the operator to create a ClusterVersion object if one does not already exist.")
```
abhinavdahiya: this makes the default `false`, when the previous behavior was `true`.  So that would be bad in terms of compatibility...

wking: > this makes the default `false`, when the previous behavior was `true`. So that would be bad in terms of compatibility...

Do we care what the default is if we always set this flag explicitly?

abhinavdahiya: I know only openshift uses this with explicit flags, so it doesn't matter that much. But I would prefer we don't make these incompatible changes.

I wouldn't block on it this time.

/lgtm

wking:
Oops, we already had one :p.

PR #239, opened 2019-08-19
WIP: Bug 1741786: cmd: Add --telemetry-id option

wking:
As an alternative to:

* removing the ClusterVersion defaulting (#238), or
* moving the telemetry ID [out into a Telemetry-specific config object][2]

this commit allows us to address the race between the installer-created ClusterVersion and the cluster-version-operator-generated default ClusterVersion by removing the former while still allowing an installer-generated Telemetry ID.

We currently need an installer-generated Telemetry ID to support things like [OpenShift Dedicated (OSD)][3], where we currently use the telemetry ID that [the installer exposes in `metadata.json`][4] to feed [the cluster manager][5].  There is concern that if Hive [6] extracted the Telemetry ID from the running cluster (vs. feeding in an installer-chosen ID), that this would race with the cluster's reported Telemetry and that, for example, Hive-launched OSD Telemetry might be erroneously claimed to be a user-provisioned cluster.

I'm less clear on why we don't prefer pushing the Telemetry ID into a Telemetry-specific config.  It seems like it would be less work than this commit (which has a lot of middleman plumbing).  I'm not entirely clear on the API we're committed to supporting for ClusterVersion's `spec.clusterID` property.  But this CVO plumbing was what we were able to form a good-enough-for-now consensus around, and we can come back later with a Telemetery-config approach if we can build consensus around that.

Also update the default channel from `fast` to `stable-4.2`; if we're going to rely on the CVO-generated default, we need it to have a channel that actually exists.

I've added the WIP subject because there are still two FIXMEs where I'm still working out the plumbing.

[1]: https://github.com/openshift/cluster-version-operator/pull/238
[2]: https://github.com/openshift/cluster-version-operator/pull/238#issuecomment-522350796
[3]: https://www.openshift.com/products/dedicated/
[4]: https://github.com/openshift/installer/blob/4e204c5e509de1bd31113b0c0e73af1a35e52c0a/pkg/types/clustermetadata.go#L17-L18
[5]: https://cloud.redhat.com/openshift/
[6]: https://github.com/openshift/hive/

### DIFF
--- a/cmd/render.go
+++ b/cmd/render.go
@@ -20,6 +20,7 @@ var (
 	renderOpts struct {
 		releaseImage string
 		outputDir    string
+		telemetryID  string
 	}
 )
 
@@ -27,6 +28,7 @@ func init() {
 	rootCmd.AddCommand(renderCmd)
 	renderCmd.PersistentFlags().StringVar(&renderOpts.outputDir, "output-dir", "", "The output directory where the manifests will be rendered.")
 	renderCmd.PersistentFlags().StringVar(&renderOpts.releaseImage, "release-image", "", "The Openshift release image url.")
+	renderCmd.PersistentFlags().StringVar(&renderOpts.telemetryID, "telemetry-id", "", "A telemetry ID which should be used for the default ClusterVersion spec.clusterID.")
 }
 
 func runRenderCmd(cmd *cobra.Command, args []string) {
@@ -39,7 +41,7 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
 	if renderOpts.releaseImage == "" {
 		klog.Fatalf("missing --release-image flag, it is required")
 	}
-	if err := payload.Render(renderOpts.outputDir, renderOpts.releaseImage); err != nil {
+	if err := payload.Render(renderOpts.outputDir, renderOpts.releaseImage, renderOpts.telemetryID); err != nil {
 		klog.Fatalf("Render command failed: %v", err)
 	}
 }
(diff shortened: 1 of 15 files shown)


abhinavdahiya:
The clusterversion.spec.clusterID is the UUID for the cluster and it is used for more things in addition to the telemetry like the Cincinatti stack uses that as a way to identify one cluster from another. So i'm against the telemetry-id option.

Here's what I would rather see.
1) the bootstrap-render find the clusterversion object from the directory... and find the clsuterID that's required for the clsuter.
2) only the bootstrap cluster-version-operator pod be given the clusterID from the flag to ensure the bootstrapping is correct.
3) The in-cluster cluster-version-operator stay intact with the existing behavior.

wking:
> The `clusterversion.spec.clusterID` is the UUID for the  cluster and it is used for more things in addition to the telemetry like  the Cincinatti stack uses that as a way to identify one cluster from  another.

Ah, I'd forgotten about Cincinnati.  Are we aware of any other consumers?

> 2. only the bootstrap cluster-version-operator pod be given the clusterID from the flag to ensure the bootstrapping is correct.

This does not allow us to recover from "I accidentally deleted my ClusterVersion", because the CVO would generate a fresh one with a fresh UUID.  If we store the current cluster ID in memory, we can recover from that unless the user kills their CVO and removes the ClusterVersion while it's down.  Do we care about blocking that hole?  Plumbing the option through would certainly be easier if we leave it open.

wking:
Closing now that #238 has new legs.

PR #240, opened 2019-08-20
Bug 1726455: Reconcile container ResourceRequirements in resource merge

jcpowermac:
- Add ensureResourceRequirements function
If existing and required are not equal will DeepCopy required
into existing.

- Update core_test to include testing container resources limits addition
and removal.

### DIFF
--- a/lib/resourcemerge/core.go
+++ b/lib/resourcemerge/core.go
@@ -102,6 +102,7 @@ func ensureContainer(modified *bool, existing *corev1.Container, required corev1
 	ensureEnvVar(modified, &existing.Env, required.Env)
 	ensureEnvFromSource(modified, &existing.EnvFrom, required.EnvFrom)
 	setStringIfSet(modified, &existing.WorkingDir, required.WorkingDir)
+	ensureResourceRequirements(modified, &existing.Resources, required.Resources)
 
 	// any port we specify, we require
 	for _, required := range required.Ports {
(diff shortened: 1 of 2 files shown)


lib/resourcemerge/core.go
```diff
@@ -445,6 +446,13 @@ func ensureSELinuxOptions(modified *bool, existing *corev1.SELinuxOptions, requi
 	setStringIfSet(modified, &existing.Level, required.Level)
 }
 
+func ensureResourceRequirements(modified *bool, existing *corev1.ResourceRequirements, required corev1.ResourceRequirements) {
```
abhinavdahiya: include both https://godoc.org/k8s.io/api/core/v1#ResourceRequirements `Limits` and `Requests`

jcpowermac: @abhinavdahiya ok will update.

lib/resourcemerge/core_test.go
```diff
@@ -113,6 +114,100 @@ func TestEnsurePodSpec(t *testing.T) {
 				Containers: []corev1.Container{
 					corev1.Container{Name: "test"}}},
 		},
+		{
+			name: "remove limits and requests on container",
```
abhinavdahiya: nit: add a no-op ?

lib/resourcemerge/core_test.go
```diff
@@ -113,6 +114,138 @@ func TestEnsurePodSpec(t *testing.T) {
...
+						Resources: corev1.ResourceRequirements{
+							Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4m")},
+							Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4m")},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "match limits and requests on container",
```
jcpowermac: @abhinavdahiya added matching existing and input last night with `expectedModified: false`

PR #241, opened 2019-08-21
pkg/payload/image: avoid rendering manifests

markmc:
The `image` command currently loads the image-references file and then
renders all manifests. Clearly the manifest rendering can be avoided
when we just want to look up an image reference.

This command is currently run 17 times by bootkube, and my quick test
shows this refactor halving the roughly 10 seconds this takes.

The main refactoring here is to split the loadUpdatePayloadMetadata()
function into getPayloadTasks() and loadImageReferences() so that the
latter can be used standalone in this particular case.

### DIFF
--- a/pkg/payload/image.go
+++ b/pkg/payload/image.go
@@ -2,6 +2,7 @@ package payload
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 )
@@ -9,12 +10,18 @@ import (
 // ImageForShortName returns the image using the updatepayload embedded in
 // the Operator.
 func ImageForShortName(name string) (string, error) {
-	up, err := LoadUpdate(DefaultPayloadDir, "")
-	if err != nil {
-		return "", errors.Wrapf(err, "error loading release manifests from %q", DefaultPayloadDir)
+	if err := ValidateDirectory(DefaultPayloadDir); err != nil {
+		return "", err
 	}
 
-	for _, tag := range up.ImageRef.Spec.Tags {
+	releaseDir := filepath.Join(DefaultPayloadDir, ReleaseManifestDir)
+
+	imageRef, err := loadImageReferences(releaseDir)
+	if err != nil {
+		return "", errors.Wrapf(err, "error loading image references from %q", releaseDir)
+	}
+
+	for _, tag := range imageRef.Spec.Tags {
 		if tag.Name == name {
 			// we found the short name in ImageStream
 			if tag.From != nil && tag.From.Kind == "DockerImage" {
--- a/pkg/payload/payload.go
+++ b/pkg/payload/payload.go
@@ -216,22 +216,23 @@ func loadUpdatePayloadMetadata(dir, releaseImage string) (*Update, []payloadTask
 		releaseDir = filepath.Join(dir, ReleaseManifestDir)
 	)
 
-	// XXX: load cincinnatiJSONFile
-	cjf := filepath.Join(releaseDir, cincinnatiJSONFile)
-	// XXX: load imageReferencesFile
-	irf := filepath.Join(releaseDir, imageReferencesFile)
-	imageRefData, err := ioutil.ReadFile(irf)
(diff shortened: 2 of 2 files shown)


pkg/payload/image.go
```diff
@@ -4,17 +4,23 @@ import (
...
 
 	"github.com/pkg/errors"
+	"k8s.io/klog"
 )
 
 // ImageForShortName returns the image using the updatepayload embedded in
 // the Operator.
 func ImageForShortName(name string) (string, error) {
-	up, err := LoadUpdate(DefaultPayloadDir, "")
+	klog.V(4).Infof("Loading image references from %q", DefaultPayloadDir)
```
abhinavdahiya: why the logs?

pkg/payload/payload.go
```diff
@@ -103,14 +103,23 @@ type Update struct {
 }
 
 func LoadUpdate(dir, releaseImage string) (*Update, error) {
-	payload, tasks, err := loadUpdatePayloadMetadata(dir, releaseImage)
+	klog.V(4).Infof("Loading updatepayload from %q", dir)
```
abhinavdahiya: why the logs?

pkg/payload/payload.go
```diff
@@ -206,39 +215,34 @@ type payloadTasks struct {
...
-		return nil, nil, err
-	}
+func getPayloadTasks(dir, releaseImage string) []payloadTasks {
+	mrc := manifestRenderConfig{ReleaseImage: releaseImage}
 
-	imageRef, err := resourceread.ReadImageStreamV1(imageRefData)
-	if err != nil {
-		return nil, nil, errors.Wrapf(err, "invalid image-references data %s", irf)
-	}
+	cjf := filepath.Join(dir, ReleaseManifestDir, cincinnatiJSONFile)
```
abhinavdahiya: keep the cvoDir and releaseDir.

pkg/payload/payload.go
```diff
@@ -206,39 +215,34 @@ type payloadTasks struct {
...
-	// XXX: load cincinnatiJSONFile
-	cjf := filepath.Join(releaseDir, cincinnatiJSONFile)
-	// XXX: load imageReferencesFile
-	irf := filepath.Join(releaseDir, imageReferencesFile)
-	imageRefData, err := ioutil.ReadFile(irf)
-	if err != nil {
-		return nil, nil, err
-	}
+func getPayloadTasks(dir, releaseImage string) []payloadTasks {
+	mrc := manifestRenderConfig{ReleaseImage: releaseImage}
```
abhinavdahiya: move it to closer to where it's used.

pkg/payload/payload.go
```diff
@@ -103,14 +103,23 @@ type Update struct {
 }
 
 func LoadUpdate(dir, releaseImage string) (*Update, error) {
-	payload, tasks, err := loadUpdatePayloadMetadata(dir, releaseImage)
+	klog.V(4).Infof("Loading updatepayload from %q", dir)
+	if err := ValidateDirectory(dir); err != nil {
```
abhinavdahiya: is there a case where we want to load from unvalidated directory?? If not, why make each caller to load* validate before calling them..?

pkg/payload/payload.go
```diff
@@ -103,14 +103,23 @@ type Update struct {
 }
 
 func LoadUpdate(dir, releaseImage string) (*Update, error) {
```
abhinavdahiya: this function did 1) load metadata 2) start using it.
not this function does 1) validate dir, 2) load image reference 3) create an payload object 4) load tasks 

I think this not the ideal direction.. feel free to split the image referece and tasks, but don't remove the `loadUpdatePayloadMetadata` ??

pkg/payload/image.go
markmc: removed

pkg/payload/payload.go
markmc: ok

PR #243, opened 2019-08-26
Bug 1730401: prevent automatic upgrades for clusters with cluster operators reporting Upgradeable false

abhinavdahiya:
/cc @smarterclayton @wking

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -37,6 +37,8 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/cvo/internal/dynamicclient"
 	"github.com/openshift/cluster-version-operator/pkg/internal"
 	"github.com/openshift/cluster-version-operator/pkg/payload"
+	"github.com/openshift/cluster-version-operator/pkg/payload/precondition"
+	preconditioncv "github.com/openshift/cluster-version-operator/pkg/payload/precondition/clusterversion"
 	"github.com/openshift/cluster-version-operator/pkg/verify"
 )
 
@@ -234,9 +236,10 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 
 	// after the verifier has been loaded, initialize the sync worker with a payload retriever
 	// which will consume the verifier
-	optr.configSync = NewSyncWorker(
+	optr.configSync = NewSyncWorkerWithPreconditions(
 		optr.defaultPayloadRetriever(),
 		NewResourceBuilder(restConfig, burstRestConfig, optr.coLister),
+		optr.defaultPreconditionChecks(),
 		optr.minimumUpdateCheckInterval,
 		wait.Backoff{
 			Duration: time.Second * 10,
@@ -642,4 +645,10 @@ func hasReachedLevel(cv *configv1.ClusterVersion, desired configv1.Update) bool 
 		return false
 	}
 	return desired.Image == cv.Status.History[0].Image
+}
+
+func (optr *Operator) defaultPreconditionChecks() precondition.List {
+	return []precondition.Precondition{
+		preconditioncv.NewUpgradeable(optr.cvLister),
+	}
 }
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -113,10 +113,17 @@ type Operator struct {
 	queue workqueue.RateLimitingInterface
 	// availableUpdatesQueue tracks checking for updates from the update server.
 	availableUpdatesQueue workqueue.RateLimitingInterface
+	// upgradeableQueue tracks checking for upgradeable.
+	upgradeableQueue workqueue.RateLimitingInterface
 
 	// statusLock guards access to modifying available updates
 	statusLock       sync.Mutex
 	availableUpdates *availableUpdates
+
+	// upgradeableStatusLock guards access to modifying Upgradeable conditions
+	upgradeableStatusLock sync.Mutex
+	upgradeable           *upgradeable
+	upgradeableChecks     []upgradeableCheck
 
 	// verifier, if provided, will be used to check an update before it is executed.
 	// Any error will prevent an update payload from being accessed.
@@ -171,10 +178,8 @@ func New(
 
 		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "clusterversion"),
 		availableUpdatesQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "availableupdates"),
+		upgradeableQueue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "upgradeable"),
 	}
-
-	optr.proxyLister = proxyInformer.Lister()
-	proxyInformer.Informer().AddEventHandler(optr.eventHandler())
 
 	cvInformer.Informer().AddEventHandler(optr.eventHandler())
 
@@ -184,7 +189,11 @@ func New(
 	optr.cvLister = cvInformer.Lister()
 	optr.cacheSynced = append(optr.cacheSynced, cvInformer.Informer().HasSynced)
 
+	optr.proxyLister = proxyInformer.Lister()
 	optr.cmConfigLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigNamespace)
+
+	// make sure this is initialized after all the listers are initialized
+	optr.upgradeableChecks = optr.defaultUpgradeableChecks()
 
 	if enableMetrics {
 		if err := optr.registerMetrics(coInformer.Informer()); err != nil {
@@ -263,6 +272,7 @@ func (optr *Operator) Run(ctx context.Context, workers int) {
 	go runThrottledStatusNotifier(stopCh, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
 	go optr.configSync.Start(ctx, 16)
 	go wait.Until(func() { optr.worker(optr.availableUpdatesQueue, optr.availableUpdatesSync) }, time.Second, stopCh)
+	go wait.Until(func() { optr.worker(optr.upgradeableQueue, optr.upgradeableSync) }, time.Second, stopCh)
 	go wait.Until(func() {
 		defer close(workerStopCh)
 
@@ -292,10 +302,12 @@ func (optr *Operator) eventHandler() cache.ResourceEventHandler {
 		AddFunc: func(obj interface{}) {
 			optr.queue.Add(workQueueKey)
 			optr.availableUpdatesQueue.Add(workQueueKey)
+			optr.upgradeableQueue.Add(workQueueKey)
 		},
 		UpdateFunc: func(old, new interface{}) {
 			optr.queue.Add(workQueueKey)
 			optr.availableUpdatesQueue.Add(workQueueKey)
+			optr.upgradeableQueue.Add(workQueueKey)
 		},
 		DeleteFunc: func(obj interface{}) {
 			optr.queue.Add(workQueueKey)
@@ -430,6 +442,29 @@ func (optr *Operator) availableUpdatesSync(key string) error {
 	}
 
 	return optr.syncAvailableUpdates(config)
+}
+
+// upgradeableSync is triggered on cluster version change (and periodic requeues) to
+// sync upgradeableCondition. It only modifies cluster version.
+func (optr *Operator) upgradeableSync(key string) error {
+	startTime := time.Now()
+	klog.V(4).Infof("Started syncing upgradeable %q (%v)", key, startTime)
+	defer func() {
+		klog.V(4).Infof("Finished syncing upgradeable %q (%v)", key, time.Since(startTime))
+	}()
+
+	config, err := optr.cvLister.Get(optr.name)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if errs := validation.ValidateClusterVersion(config); len(errs) > 0 {
+		return nil
+	}
+
+	return optr.syncUpgradeable(config)
 }
 
 // isOlderThanLastUpdate returns true if the cluster version is older than
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/openshift/cluster-version-operator/lib"
 	"github.com/openshift/cluster-version-operator/pkg/payload"
+	"github.com/openshift/cluster-version-operator/pkg/payload/precondition"
 )
 
 func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fake.Clientset, *dynamicfake.FakeDynamicClient, func()) {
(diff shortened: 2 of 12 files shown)


pkg/payload/preflight/preflights.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	msg := ""
+	if len(msgs) == 1 {
+		msg = msgs[0]
+	} else {
+		msg = fmt.Sprintf("Multiple preflight checks failed:\n* %s", strings.Join(msgs, "\n* "))
+	}
+	return &payload.UpdateError{
+		Nested:  nil,
+		Reason:  "PreflightChecksFailed",
```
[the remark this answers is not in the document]
abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/243/commits/a5500d315c8932363108dba5a95259c8a034af3c

pkg/payload/preflight/features/features.go
```diff
@@ -0,0 +1,64 @@
+package features
```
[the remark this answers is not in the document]
abhinavdahiya: So the goal is to prevent upgrades when features are set.. 

the precondition `upgradeable == false` preventing upgrades _personally_ I think needs more thought and testing to make sure the operators are no missusing before we add is important. So I don't think we should be doing it.

KAO already setting a `uprageable = false` on features is great, but I think depending on that wouldn't be ideal as mentioned above, and also depends on KAO always respecting that :P I think it's better if CVO does it.

pkg/cvo/cvo.go
```diff
@@ -37,6 +37,9 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/cvo/internal/dynamicclient"
 	"github.com/openshift/cluster-version-operator/pkg/internal"
 	"github.com/openshift/cluster-version-operator/pkg/payload"
+	"github.com/openshift/cluster-version-operator/pkg/payload/preflight"
```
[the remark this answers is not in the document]
abhinavdahiya: > I expected this PR to summarize an Upgradeable condition - ideally before a user ever upgrades.

That would be a long running syncer, and I don't think it's great to add that at this time.

> The current structure can't easily do that (and there are advantages to having preflight check coming on the new code),

The current structures goal is to prevent upgrades, the long running syncer doesn't do that.. even if we add the long running syncer, we still needs this code to check the precondition before cvo accepts the upgrade. So I think the syncer to provide the Upgradeable condition is _orthogonal_ to this, and perhaps when we add that later on, the precondition could use that as another check??

pkg/payload/preflight/features/features.go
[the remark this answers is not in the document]
abhinavdahiya: ^^ ping @smarterclayton

pkg/cvo/cvo.go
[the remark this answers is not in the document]
abhinavdahiya: ^^ ping @smarterclayton

pkg/payload/preflight/features/features.go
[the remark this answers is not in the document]
smarterclayton: Most of our components ALREADY report Upgradeable false

pkg/cvo/cvo.go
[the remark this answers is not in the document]
smarterclayton: > That would be a long running syncer, and I don't think it's great to add that at this time.

I don't understand why.  You just have to summarize the state.

Blocking without reporting seems unacceptable to me.  I should know up top whether upgrades should even work.

I think the miscommunication / gap here is that we already have an implemented mechanism for reporting "Don't upgrade me" as Upgradable false since 4.1 GA, but this is targeting other checks which are much more specific / tied to implementation.  We may need to have a higher bandwidth conversation on that, but the generic condition is better at isolating CVO from details, and behaves exactly like degraded in that sense.

abhinavdahiya:
@smarterclayton added the long running upgradeable reporting sync https://github.com/openshift/cluster-version-operator/pull/243/commits/35c42999be2f4044fdcca36855034228461bdc33

and updated the preflights to single check if upgradeable...

PTAL if this looks better.

pkg/payload/task.go
```diff
@@ -188,6 +188,9 @@ func SummaryForReason(reason, name string) string {
 	case "ImageVerificationFailed":
 		return "the image may not be safe to use"
 
+	case "PreflightChecksFailed":
```
smarterclayton: Prefer `UpgradePreconditionCheckFailed`

smarterclayton: Nm, saw that was in a later commit

pkg/cvo/cvo.go
```diff
@@ -225,9 +236,10 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 
 	// after the verifier has been loaded, initialize the sync worker with a payload retriever
 	// which will consume the verifier
-	optr.configSync = NewSyncWorker(
+	optr.configSync = NewSyncWorkerWithPreflights(
```
smarterclayton: WithPreconditions

pkg/cvo/status.go
```diff
@@ -158,7 +158,11 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
 	// update the config with the latest available updates
 	if updated := optr.getAvailableUpdates().NeedsUpdate(config); updated != nil {
 		config = updated
-	} else if original == nil || original == config {
```
smarterclayton: Hrm, this isn't the same condition anymore.  If updated is not nil, original should not be deep copied no matter what.

pkg/cvo/sync_test.go
```diff
@@ -442,3 +443,28 @@ func (b *testResourceBuilder) Apply(ctx context.Context, m *lib.Manifest, state
 	}
 	return builder.Do(ctx)
 }
+
+type testPreflight struct {
```
smarterclayton: Let's call these Preconditions globally

pkg/cvo/sync_worker.go
```diff
@@ -493,6 +504,30 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 		payloadUpdate.VerifiedImage = info.Verified
 		payloadUpdate.LoadedAt = time.Now()
 
+		// need to make sure the payload is only set when the preflights have been successfull
+		if !info.Local && len(w.preflights) > 0 {
+			reporter.Report(SyncWorkerStatus{
+				Generation:  work.Generation,
+				Step:        "PreflightChecks",
```
smarterclayton: PreconditionChecks

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+			Reason:             conds[0].Reason,
+			Message:            conds[0].Message,
+			LastTransitionTime: now,
+		})
+	} else if len(conds) > 1 {
+		conds = append(conds, configv1.ClusterOperatorStatusCondition{
+			Type:               configv1.OperatorUpgradeable,
+			Status:             configv1.ConditionFalse,
+			Reason:             "MultipleReasons",
+			Message:            fmt.Sprintf("Cluster is not Upgradeable due to multiple reasons: %s", strings.Join(reasons, ",")),
```
smarterclayton: Should be `Cluster is not upgradeable ...`.  Don't capitalize the condition.

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+			resourcemerge.RemoveOperatorStatusCondition(&config.Status.Conditions, origc.Type)
+		}
+	}
+	return config
+}
+
+func collectUpgradeableConditions(conditions []configv1.ClusterOperatorStatusCondition) []configv1.ClusterOperatorStatusCondition {
+	var ret []configv1.ClusterOperatorStatusCondition
+	for _, c := range conditions {
+		if strings.HasPrefix(string(c.Type), string(configv1.OperatorUpgradeable)) {
```
smarterclayton: Hrm, is this really a good idea?  Do we *need* multiple nested condition types?

pkg/payload/preflight/clusterversion/upgrdeable.go
```diff
@@ -0,0 +1,57 @@
+package clusterversion
```
smarterclayton: name of this file is a typo

pkg/payload/preflight/clusterversion/upgrdeable.go
```diff
@@ -0,0 +1,57 @@
...
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
+	"github.com/openshift/cluster-version-operator/pkg/payload/preflight"
+)
+
+// Upgradeable checks if clusterversion is upgradeable currently.
+type Upgradeable struct {
```
smarterclayton: `Upgradable`, no e

abhinavdahiya: https://github.com/openshift/api/blob/763878ba9fae6d87ab1d8310aa0c01fe7dd2499a/config/v1/types_cluster_operator.go#L140
has `Upgradeable` :P

pkg/cvo/upgradeable.go
abhinavdahiya: _my thought was_
It gives us the opportunity to a provide detailed message for different cases why upgradeable is false, but still allow `Upgradeable` condition to be summarize if only one type of failure exists, and link to individual ones when there are multiple types of reasons.. like https://github.com/openshift/cluster-version-operator/pull/243/files#diff-745b27730a4db8ced457d29d6656b64aR3022-R3037

do you think single condition is more cleaner and enough?

pkg/cvo/cvo.go
abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/243/commits/64249a4e283ddf99631885faedf6297401fb7e88

pkg/cvo/sync_test.go
abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/243/commits/64249a4e283ddf99631885faedf6297401fb7e88

pkg/cvo/sync_worker.go
abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/243/commits/64249a4e283ddf99631885faedf6297401fb7e88

pkg/payload/preflight/clusterversion/upgrdeable.go
abhinavdahiya: done with https://github.com/openshift/cluster-version-operator/pull/243/commits/64249a4e283ddf99631885faedf6297401fb7e88

pkg/cvo/status.go
abhinavdahiya: brought it back to previous state with https://github.com/openshift/cluster-version-operator/pull/243/commits/03e188b3d7fcf3e2e374687186d8c994685d605f

pkg/cvo/upgradeable.go
abhinavdahiya: fixed with https://github.com/openshift/cluster-version-operator/pull/243/commits/03e188b3d7fcf3e2e374687186d8c994685d605f

pkg/cvo/upgradeable.go
abhinavdahiya: also the multiple conditions only happen when there are more than one type of failure... for most cases it will be single condition.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -620,6 +621,266 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
...
+func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
+	o, cvs, client, _, shutdownFn := setupCVOTest("testdata/payloadtest")
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	defer shutdownFn()
+
+	worker := o.configSync.(*SyncWorker)
+	worker.preconditions = []precondition.Precondition{&testPrecondition{SuccessAfter: 100}}
```
smarterclayton: It wasn't clear reading this test which step checks this, could you make that clearer?

pkg/payload/task.go
```diff
@@ -188,6 +188,9 @@ func SummaryForReason(reason, name string) string {
 	case "ImageVerificationFailed":
 		return "the image may not be safe to use"
 
+	case "UpgradePreconditionCheckFailed":
+		return "the update might not be safe to proceed"
```
smarterclayton: Wording on this is a little weird - maybe `it may not be safe to apply this update`

pkg/cvo/cvo_test.go
```diff
@@ -1353,6 +1353,229 @@ func TestOperator_sync(t *testing.T) {
...
+						Channel:   "fast",
+					},
+					Status: configv1.ClusterVersionStatus{
+						ObservedGeneration: 2,
+						Conditions: []configv1.ClusterOperatorStatusCondition{
+							{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse},
+							{Type: configv1.OperatorProgressing, Status: configv1.ConditionFalse, Message: "Cluster version is 0.0.1-abc"},
+							{Type: ClusterStatusFailing, Status: configv1.ConditionFalse},
+							{Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse},
+							{Type: configv1.ClusterStatusConditionType("UpgradeableA"), Status: configv1.ConditionFalse},
```
smarterclayton: Fold this into a constant in pkg/cvo (like ClusterStatusFailing) or move into api (former is faster, can do the others after)

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+		if o.Unmanaged {
+			overrides = true
+		}
+	}
+	if !overrides {
+		return nil
+	}
+
+	cond.Reason = "ClusterVersionOverridesSet"
+	cond.Message = "Cluster Version overrides is being used to unmanage certain owned resources."
```
smarterclayton: Message should probably be `Disabling ownership via cluster version overrides prevents upgrades.  Please remove overrides before continuing.`

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+	if len(notup) == 0 {
+		return nil
+	}
+	msg := ""
+	reason := ""
+	if len(notup) == 1 {
+		msg = notup[0]
+		reason = "ClusterOperatorNotUpgradeable"
+	} else {
+		msg = fmt.Sprintf("Multiple Cluster Operators are not upgradeable:\n* %s", strings.Join(notup, "\n* "))
```
smarterclayton: We should use lower case `Multiple cluster operators cannot be upgraded:`

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		cond.Reason = "FailedToListClusterOperators"
+		cond.Message = errors.Wrap(err, "failed to list cluster operators").Error()
+		return cond
+	}
+
+	var notup []string
+	for _, op := range ops {
+		if up := resourcemerge.FindOperatorStatusCondition(op.Status.Conditions, configv1.OperatorUpgradeable); up != nil && up.Status == configv1.ConditionFalse {
+			notup = append(notup, fmt.Sprintf("Cluster Operator %q is not Upgradeable(%q): %s", op.GetName(), up.Reason, up.Message))
```
smarterclayton: `Cluster operator cannot be upgraded`

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if len(notup) == 0 {
+		return nil
+	}
+	msg := ""
+	reason := ""
+	if len(notup) == 1 {
+		msg = notup[0]
+		reason = "ClusterOperatorNotUpgradeable"
```
smarterclayton: I would prefer to use the reason from the operator directly (like we do for other summaries).

pkg/cvo/sync_worker.go
```diff
@@ -493,6 +504,30 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 		payloadUpdate.VerifiedImage = info.Verified
 		payloadUpdate.LoadedAt = time.Now()
 
+		// need to make sure the payload is only set when the preconditions have been successfull
+		if !info.Local && len(w.preconditions) > 0 {
+			reporter.Report(SyncWorkerStatus{
```
smarterclayton: I'm confused, shouldn't this be an else block?

pkg/cvo/cvo_test.go
smarterclayton: Or just use OperatorUpgradeable (isn't that in api already?)

pkg/cvo/sync_worker.go
abhinavdahiya: was following https://github.com/openshift/cluster-version-operator/pull/243/files/64249a4e283ddf99631885faedf6297401fb7e88#diff-b2099007e7494e37fc4f7a4320f3622bR471-R489

where we send update that we are running the preconditions and and a following failure update if failed to do so..

pkg/cvo/cvo_scenarios_test.go
abhinavdahiya: added a comment, what do think now? any clearer?

pkg/payload/task.go
abhinavdahiya: +1 update to match your suggestion.

pkg/cvo/upgradeable.go
```diff
@@ -0,0 +1,215 @@
...
+			Reason:             conds[0].Reason,
+			Message:            conds[0].Message,
+			LastTransitionTime: now,
+		}}
+	} else if len(conds) > 1 {
+		conds = append(conds, configv1.ClusterOperatorStatusCondition{
+			Type:               configv1.OperatorUpgradeable,
+			Status:             configv1.ConditionFalse,
+			Reason:             "MultipleReasons",
+			Message:            fmt.Sprintf("Cluster is not upgradeable due to multiple reasons: %s", strings.Join(reasons, ",")),
```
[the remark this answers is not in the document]
abhinavdahiya: +1 fixed to match your suggestion.

pkg/cvo/upgradeable.go
abhinavdahiya: +1 fixed to match your suggestion.

pkg/cvo/upgradeable.go
abhinavdahiya: +1 fixed to match your suggestion.

pkg/cvo/upgradeable.go
abhinavdahiya: +1 fixed to match your suggestion.

pkg/cvo/upgradeable.go
abhinavdahiya: +1 fixed to match your suggestion.

pkg/cvo/cvo_test.go
abhinavdahiya: > https://github.com/openshift/cluster-version-operator/pull/243/files#diff-745b27730a4db8ced457d29d6656b64aR1433

So this test is actually testing that the original clusterversion object has a condition `UpgradeableA` set, but this new sync needs `Upgradeable` and `UpgradeableB` (https://github.com/openshift/cluster-version-operator/pull/243/files#diff-745b27730a4db8ced457d29d6656b64aR1446-R1447) to be set.
and the end result was exactly that https://github.com/openshift/cluster-version-operator/pull/243/files#diff-745b27730a4db8ced457d29d6656b64aR1500-R1501

abhinavdahiya:
@smarterclayton updated the PR to address your comments.

PR #249, opened 2019-10-04
Bug 1755383: Bump deps to kubernetes-1.16.0

marun:


sdodson:
@marun does this fix / enable something you're trying to use but can't without the bump?

marun:
@sdodson Yes, as per bz: https://bugzilla.redhat.com/show_bug.cgi?id=1755383

I had backburnered this, will prioritize getting it merged again. Not sure the image build issue is related, it wasn't failing due to that when I initially submitted.

sttts:
@sdodson this is even more critical than before. It blocks finishing the 1.16 rebase of the platform because CVO drops necessary CRD fields which then break OpenAPI publishing of all our 74 CRDs, which breaks `oc explain` and console CRD documentation.

It would be very helpful to get this merged soonish as we have enough follow-up work to get all CRD manifest in order and fix possible issues.

sdodson:
/retitle Bug 1755383: Bump deps to kubernetes-1.16.0

PR #252, opened 2019-10-17
Allow excluding manifests from CVO payload

csrwng:
Allows specifying an identifier with the `EXCLUDE_MANIFESTS` environment variable that will be used to skip manifests that contain an annotation like `exclude.release.openshift.io/<identifier>=true`

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -139,6 +139,10 @@ type Operator struct {
 	// lastAtLock guards access to controller memory about the sync loop
 	lastAtLock          sync.Mutex
 	lastResourceVersion int64
+
+	// exclude is an optional identifier used to exclude certain manifests
+	// via annotation
+	exclude string
 }
 
 // New returns a new cluster version operator.
@@ -156,6 +160,7 @@ func New(
 	client clientset.Interface,
 	kubeClient kubernetes.Interface,
 	enableMetrics bool,
+	exclude string,
 ) *Operator {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(klog.Infof)
@@ -181,6 +186,8 @@ func New(
 		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "clusterversion"),
 		availableUpdatesQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "availableupdates"),
 		upgradeableQueue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "upgradeable"),
+
+		exclude: exclude,
 	}
 
 	cvInformer.Informer().AddEventHandler(optr.eventHandler())
@@ -246,6 +253,7 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 			Factor:   1.3,
 			Steps:    3,
 		},
+		optr.exclude,
 	)
 
 	return nil
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -126,11 +126,11 @@ func (w SyncWorkerStatus) DeepCopy() *SyncWorkerStatus {
 //     syncOnce() returns nil -> Reconciling
 //
 type SyncWorker struct {
-	backoff     wait.Backoff
-	retriever   PayloadRetriever
-	builder     payload.ResourceBuilder
-	preconditions  precondition.List
-	reconciling bool
+	backoff       wait.Backoff
+	retriever     PayloadRetriever
+	builder       payload.ResourceBuilder
+	preconditions precondition.List
+	reconciling   bool
 
 	// minimumReconcileInterval is the minimum time between reconcile attempts, and is
 	// used to define the maximum backoff interval when syncOnce() returns an error.
@@ -148,11 +148,16 @@ type SyncWorker struct {
 
 	// updated by the run method only
 	payload *payload.Update
+
+	// exclude is an identifier used to determine which
+	// manifests should be excluded based on an annotation
+	// of the form exclude.release.openshift.io/<identifier>=true
+	exclude string
 }
 
 // NewSyncWorker initializes a ConfigSyncWorker that will retrieve payloads to disk, apply them via builder
 // to a server, and obey limits about how often to reconcile or retry on errors.
-func NewSyncWorker(retriever PayloadRetriever, builder payload.ResourceBuilder, reconcileInterval time.Duration, backoff wait.Backoff) *SyncWorker {
+func NewSyncWorker(retriever PayloadRetriever, builder payload.ResourceBuilder, reconcileInterval time.Duration, backoff wait.Backoff, exclude string) *SyncWorker {
 	return &SyncWorker{
 		retriever: retriever,
 		builder:   builder,
@@ -165,14 +170,16 @@ func NewSyncWorker(retriever PayloadRetriever, builder payload.ResourceBuilder, 
 		// Status() or use the result of calling Update() instead because the channel can be out of date
 		// if the reader is not fast enough.
 		report: make(chan SyncWorkerStatus, 500),
+
+		exclude: exclude,
 	}
 }
 
 // NewSyncWorkerWithPreconditions initializes a ConfigSyncWorker that will retrieve payloads to disk, apply them via builder
 // to a server, and obey limits about how often to reconcile or retry on errors.
 // It allows providing preconditions for loading payload.
-func NewSyncWorkerWithPreconditions(retriever PayloadRetriever, builder payload.ResourceBuilder, preconditions precondition.List, reconcileInterval time.Duration, backoff wait.Backoff) *SyncWorker {
-	worker := NewSyncWorker(retriever, builder, reconcileInterval, backoff)
+func NewSyncWorkerWithPreconditions(retriever PayloadRetriever, builder payload.ResourceBuilder, preconditions precondition.List, reconcileInterval time.Duration, backoff wait.Backoff, exclude string) *SyncWorker {
+	worker := NewSyncWorker(retriever, builder, reconcileInterval, backoff, exclude)
 	worker.preconditions = preconditions
 	return worker
 }
@@ -614,7 +621,7 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 			klog.V(4).Infof("Running sync for %s", task)
 			klog.V(5).Infof("Manifest: %s", string(task.Manifest.Raw))
 
-			ov, ok := getOverrideForManifest(work.Overrides, task.Manifest)
+			ov, ok := getOverrideForManifest(work.Overrides, w.exclude, task.Manifest)
 			if ok && ov.Unmanaged {
 				klog.V(4).Infof("Skipping %s as unmanaged", task)
 				continue
@@ -906,7 +913,7 @@ func newMultipleError(errs []error) error {
 }
 
 // getOverrideForManifest returns the override and true when override exists for manifest.
-func getOverrideForManifest(overrides []configv1.ComponentOverride, manifest *lib.Manifest) (configv1.ComponentOverride, bool) {
+func getOverrideForManifest(overrides []configv1.ComponentOverride, excludeIdentifier string, manifest *lib.Manifest) (configv1.ComponentOverride, bool) {
 	for idx, ov := range overrides {
 		kind, namespace, name := manifest.GVK.Kind, manifest.Object().GetNamespace(), manifest.Object().GetName()
 		if ov.Kind == kind &&
@@ -914,6 +921,10 @@ func getOverrideForManifest(overrides []configv1.ComponentOverride, manifest *li
 			ov.Name == name {
 			return overrides[idx], true
 		}
+	}
+	excludeAnnotation := fmt.Sprintf("exclude.release.openshift.io/%s", excludeIdentifier)
+	if annotations := manifest.Object().GetAnnotations(); annotations != nil && annotations[excludeAnnotation] == "true" {
+		return configv1.ComponentOverride{Unmanaged: true}, true
 	}
 	return configv1.ComponentOverride{}, false
 }
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -58,6 +58,11 @@ type Options struct {
 	EnableAutoUpdate            bool
 	EnableDefaultClusterVersion bool
 
+	// Exclude is used to determine whether to exclude
+	// certain manifests based on an annotation:
+	// exclude.release.openshift.io/<identifier>=true
+	Exclude string
+
 	// for testing only
 	Name            string
 	Namespace       string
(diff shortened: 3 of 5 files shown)


csrwng:
@crawford @abhinavdahiya this should be ready for a review. This is a first step towards introducing the concept of profiles where each component will declare what files go in a given profile. When that is introduced, these flags will be removed.

abhinavdahiya:
> this should be ready for a review. This is a first step towards introducing the concept of profiles where each component will declare what files go in a given profile.

Is there a backing enhancement that provides some context on how this change is on the path to profiles.

Who can set these cli flags? How do we find out about these flags are being set?
Setting these flags should probably make your clusters Not Upgradeable for now.

Also this change overlaps with what clusterversion.overrides do? Why do we this new feature when we can achieve the same thing with clusterversion API

pkg/cvo/sync_worker.go
```diff
@@ -148,11 +148,14 @@ type SyncWorker struct {
 
 	// updated by the run method only
 	payload *payload.Update
+
+	excludeFilter []string
```
abhinavdahiya: why is this called a filter but is a string slice? this doesn't really provide context on how to use this to filter.

pkg/payload/payload.go
```diff
@@ -103,11 +104,47 @@ type Update struct {
...
+func (f regexFilter) matches(s string) bool {
+	for _, re := range f {
+		if re.MatchString(s) {
+			return true
+		}
+	}
+	return false
+}
+
+func LoadUpdateWithFilters(dir, releaseImage string, excludes, includes []string) (*Update, error) {
```
abhinavdahiya: the name is LoadUpdateWithFilters, but it doen't take any filters though...

abhinavdahiya:
> Who can set these cli flags? How do we find out about these flags are being set?
> Setting these flags should probably make your clusters Not Upgradeable for now.

if these flags are not supposed to be set by any Openshift 4 user, these should be behind a build time flag.

csrwng:
> Is there a backing enhancement that provides some context on how this change is on the path to profiles.

There isn't an enhancement proposal for profiles yet, I will create one. The target for profiles would be 4.4. 

> Who can set these cli flags? How do we find out about these flags are being set?
These flags would be set in the  deployment for the CVO in a hosted environment, where the cvo runs as a pod in a management server but operates on a user cluster control plane (also hosted as pods on the management server)

> Setting these flags should probably make your clusters Not Upgradeable for now.
Ack, will make that change.

> Also this change overlaps with what clusterversion.overrides do? Why do we this new feature when we can achieve the same thing with clusterversion API
One of the benefits of hosted clusters is enabling cluster admin rights for users to their own cluster and making it hard for them to bring down the control plane on their own. Having the clusterversion resource (inside the user cluster) control which manifests get laid down puts too much of the responsibility on the user to keep it in a good working state.

csrwng:
> if these flags are not supposed to be set by any Openshift 4 user, these should be behind a build time flag.

The release is meant to be run unchanged in hosted mode, it won't get its own build.

csrwng:
@abhinavdahiya thinking more about this.... since this is only a temporary enhancement that will be removed once we add a way to specify a profile, maybe instead of adding flags it should just be a set of env vars, wdyt?

pkg/payload/payload.go
csrwng: It does (`excludes`, `includes`)

pkg/cvo/sync_worker.go
csrwng: It's just a slice of regular expressions. I can add a comment explaining what they are.

csrwng:
@abhinavdahiya I've now simplified this some and removed the includes part (less to remove later). We should only need to specify what to exclude. I've also removed the additional command flags and switched to reading environment variables to specify the regular expressions for manifest exclusion. I did not make the change to switch to not Upgradeable when these filters are specified because we will still need the CVO to apply upgrades to  non-control plane components. 

We need to get this in for the current release. Please let me know what else you'd like to see changed so we can get this merged.

/cc @derekwaynecarr

abhinavdahiya:
> 
> > Who can set these cli flags? How do we find out about these flags are being set?
> 
> These flags would be set in the deployment for the CVO in a hosted environment, where the cvo runs as a pod in a management server but operates on a user cluster control plane (also hosted as pods on the management server)

So that means the customer is not allowed to set/effect these values in external-hosted-profile.

but what about self-hosted-profile clusters, there the users have permission and access to modify their cluster-version-operator deployment to add these flags/envs. How do we prevent that or if we can't prevent that, how do we know the customer is doing this.

> 
> > Also this change overlaps with what clusterversion.overrides do? Why do we this new feature when we can achieve the same thing with clusterversion API
> 
> One of the benefits of hosted clusters is enabling cluster admin rights for users to their own cluster and making it hard for them to bring down the control plane on their own. Having the clusterversion resource (inside the user cluster) control which manifests get laid down puts too much of the responsibility on the user to keep it in a good working state.

So the users are allowed to modify the clusterversion object in external-hosted-profile? which will affect how cvo behaves, and affect how it managed the cluster components, but we think using `overrides` section would be too much access to the user?

> @abhinavdahiya I've now simplified this some and removed the includes part (less to remove later). We should only need to specify what to exclude. I've also removed the additional command flags and switched to reading environment variables to specify the regular expressions for manifest exclusion. I did not make the change to switch to not Upgradeable when these filters are specified because we will still need the CVO to apply upgrades to non-control plane components.

As I mentioned above, there is a chance users will set these values in self-hosted cases, and not knowing how/when these are being set is a blocker in my opinion.

> We need to get this in for the current release.

`current`? 4.3 or 4.4 ?

> Please let me know what else you'd like to see changed so we can get this merged.
> 
> /cc @derekwaynecarr

pkg/cvo/cvo.go
```diff
@@ -139,6 +139,10 @@ type Operator struct {
 	// lastAtLock guards access to controller memory about the sync loop
 	lastAtLock          sync.Mutex
 	lastResourceVersion int64
+
+	// excludeManifestRegularExpressions is a set of regular expresions that are used
+	// to exclude manifest file names that match them.
+	excludeManifestRegularExpressions []string
```
abhinavdahiya: why not store regex compiled here? compiling these regex every time is known to cause problems.

csrwng:
> but what about self-hosted-profile clusters, there the users have permission and access to modify their cluster-version-operator deployment to add these flags/envs. How do we prevent that or if we can't prevent that, how do we know the customer is doing this.

Users  can do this today. They can modify the cvo deployment in all kinds of ways. They can specify an alternate image, alternate namespace or a different payload directory (https://github.com/openshift/cluster-version-operator/blob/master/pkg/start/start.go#L85-L87). 
How are these variables different?


> So the users are allowed to modify the clusterversion object in external-hosted-profile? which will affect how cvo behaves, and affect how it managed the cluster components, but we think using `overrides` section would be too much access to the user?

Yes they will be able to change the clusterversion resource and if they like, add additional overrides, but they can't add things that are already excluded on the management side.



> As I mentioned above, there is a chance users will set these values in self-hosted cases, and not knowing how/when these are being set is a blocker in my opinion.

Again, not sure how it changes what you can do today. If you are privileged enough to change the cluster version operator deployment, then there's really not much we can do to stop you from shooting yourself in the foot. 



> `current`? 4.3 or 4.4 ?

4.3 (we have exception approval)

pkg/cvo/sync_worker.go
michaelgugino: maybe plural Filters?

csrwng:
Pushed update to parse regular expressions once on startup

csrwng:
@abhinavdahiya I've submitted a change to make the cvo not upgradable when manifest excludes are set (verified that it doesn't affect upgrades for externally hosted mode). Please let me know if this addresses your concern.

pkg/cvo/upgradeable.go
```diff
@@ -215,9 +216,26 @@ func (check *clusterVersionOverridesUpgradebale) Check() *configv1.ClusterOperat
...
+type clusterManifestExcludesUpgradable struct {
+	filter payload.RegexpFilter
+}
+
+func (check *clusterManifestExcludesUpgradable) Check() *configv1.ClusterOperatorStatusCondition {
+	if len(check.filter) == 0 {
+		return nil
+	}
+	return &configv1.ClusterOperatorStatusCondition{
+		Type:    configv1.ClusterStatusConditionType("UpgradableManifestExcludes"),
```
abhinavdahiya: > https://github.com/openshift/cluster-version-operator/pull/252/files#diff-d9212d08828a9f9d8b54a2e90f31500aR195
the conditions need to start with `Upgradeable`

`UpgradableManifestExcludes` -> `UpgradeableManifestExcludes`

hmm, also this conditions names is not that great.

abhinavdahiya: :thinking:

pkg/cvo/upgradeable.go
```diff
@@ -215,9 +216,26 @@ func (check *clusterVersionOverridesUpgradebale) Check() *configv1.ClusterOperat
...
+		Status:  configv1.ConditionFalse,
+		Reason:  "ManifestExcludesSet",
+		Message: "Certain manifests are excluded from the operator's payload. The cluster cannot be upgraded.",
```
abhinavdahiya: > https://github.com/openshift/cluster-version-operator/pull/252/files#diff-d9212d08828a9f9d8b54a2e90f31500aR215

based on that, something like `Excluding manifests from the release image prevents upgrades. Please remove the exclude filters before continuing.`

pkg/cvo/cvo.go
```diff
@@ -139,6 +139,10 @@ type Operator struct {
 	// lastAtLock guards access to controller memory about the sync loop
 	lastAtLock          sync.Mutex
 	lastResourceVersion int64
+
+	// excludeManifestsFilter is a set of regular expresions that are used
+	// to exclude manifest file names that match them.
+	excludeManifestsFilter payload.RegexpFilter
```
smarterclayton: I think the payload should be required to identify the manifests that match a string, rather than a regex.  I do not like regex and I don't want arbitrary regular expressions.

pkg/cvo/upgradeable.go
smarterclayton: I don't like regex on manifests, I'd like this to be a higher level name that manifests are associated with.  If that's the case, then this shouldn't even be a condition because expectation is that they are tested and we don't need this alert.

pkg/cvo/cvo.go
csrwng: ack, string matching should be sufficient for what we need, will switch to that

wking: > I think the payload should be required to identify the manifests that match a string...

I'm trying to catch up with the previous discussion in this PR, but if the folks setting this are creating release images, can't they just remove the manifests they don't want pushed?  And if they aren't writing to release images, why is [the existing `overrides`][1] insufficient.  Do we have [the enhancement][2] explaining how this is expected to fit together yet?

[1]: https://github.com/openshift/api/blob/a27ff30ebf098b145a702153b2a466a0f9e6e794/config/v1/types_cluster_version.go#L68-L72
[2]: https://github.com/openshift/cluster-version-operator/pull/252#issuecomment-549687836

csrwng: We don't want a separate release image for this case. The release image should be the same we use everywhere.

And no, the enhancement is not ready, since I think the design of profiles needs a little bit more discussion. The purpose of this PR is simply to a stop-gap in 4.3 until we get to profiles.

csrwng:
@abhinavdahiya I've now updated the PR to use an annotation to filter manifests. I've tested this using a custom-built release and the cluster works fine. I have commits ready to go for the affected repositories but first want to make sure this gets reviewed first.

pkg/cvo/sync_worker.go
```diff
@@ -620,6 +627,11 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 				continue
 			}
 
+			if shouldExclude(w.exclude, task.Manifest) {
```
abhinavdahiya: syncwork already has an override here https://github.com/openshift/cluster-version-operator/pull/252/files#diff-b2099007e7494e37fc4f7a4320f3622bR68

and exclude very similar to that, is it possible we can unify that? they seem to have pretty similar fingerprint??

csrwng:
Updated to use the single getOverrides method. Is that what you meant @abhinavdahiya ?

abhinavdahiya:
> Updated to use the single getOverrides method. Is that what you meant @abhinavdahiya ?

So the sync_work already has a list of overrides, https://github.com/openshift/cluster-version-operator/blob/9cb26a332a5ace0310b1268fa62d9b0ba3cc214f/pkg/cvo/sync_worker.go#L68

and this change adds another style of override/skip at the syncworker level. https://github.com/openshift/cluster-version-operator/blob/9cb26a332a5ace0310b1268fa62d9b0ba3cc214f/pkg/cvo/sync_worker.go#L152-L155

And the sync work can change at each sync, but the sync worker is only configured once at startup.
is that acceptable? 

secondly, i think we lost the reporting excludes are set with the current change, until we have supported profiles I would like to make sure we know when users are setting this using env.

abhinavdahiya:
cc @smarterclayton 

does this look more in-line to what you were hoping?

csrwng:
> And the sync work can change at each sync, but the sync worker is only configured once at startup.
> is that acceptable?

Yes I think that is acceptable because the exclude identifier will be statically configured for the cvo deployment, unlike overrides which can be modified via the api on the clusterversion CR. 



> secondly, i think we lost the reporting excludes are set with the current change, until we have supported profiles I would like to make sure we know when users are setting this using env.

I didn't add the condition back because of [this comment](https://github.com/openshift/cluster-version-operator/pull/252#discussion_r344910108) from @smarterclayton

smarterclayton:
Looks fine to me.

pkg/start/start.go
```diff
@@ -87,6 +92,7 @@ func NewOptions() *Options {
 		PayloadOverride: os.Getenv("PAYLOAD_OVERRIDE"),
 		ResyncInterval:  minResyncPeriod,
 		EnableMetrics:   true,
+		Exclude:         os.Getenv("EXCLUDE_MANIFESTS"),
```
smarterclayton: When exclude is set we should log this.

`Excluding manifests for %q`

in the main logs so we can see it if we have bugs.

csrwng: done

csrwng:
```
listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400
```
/retest

wking:
> 4.3 (we have exception approval)

Is there a bug we can link from this PR for this exception?

pkg/cvo/sync_worker.go
```diff
@@ -148,11 +148,16 @@ type SyncWorker struct {
 
 	// updated by the run method only
 	payload *payload.Update
+
+	// exclude is an identifier used to determine which
+	// manifests should be excluded based on an annotation
+	// of the form exclude.release.openshift.io/<identifier>=true
+	exclude string
```
wking: Do we want a single string, here or a `[]string` of identifiers to exclude?  Related to [this thread][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/269#discussion_r346049977

csrwng: @wking the idea is that the identifier would be the equivalent of a single profile. If you have a different exclusion profile, you should use a different identifier. Manifests in the individual operators can contain exclusion annotations for multiple profiles.

PR #254, opened 2019-10-20
Bug 1665828: cmd/main.go: fix cli use to cluster-version-operator instead of version

abhinavdahiya:
/assign @wking

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -7,14 +7,9 @@ import (
 	"k8s.io/klog"
 )
 
-const (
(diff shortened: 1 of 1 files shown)


wking:
Matches the executable filename [from the `Makefile`][1] that builds it.

/lgtm

[1]: https://github.com/openshift/cluster-version-operator/commit/ddeb140857b9c59216d0c55d8167d7e5a0bfe325#diff-9283775b0feecc10455ec28bd08983b0R33

PR #255, opened 2019-10-21
Bug 1763821: pkg/payload/task_graph: Canceling the task graph partway though is an error even if no tasks fail

wking:
Avoid situations [like][1]:

```
2019-10-21T10:34:30.63940461Z I1021 10:34:30.639073       1 start.go:19] ClusterVersionOperator v1.0.0-106-g0725bd53-dirty
...
2019-10-21T10:34:31.132673574Z I1021 10:34:31.132635       1 sync_worker.go:453] Running sync quay.io/runcom/origin-release:v4.2-1196 (force=true) on generation 2 in state Updating at attempt 0
...
2019-10-21T10:40:16.168632703Z I1021 10:40:16.168604       1 sync_worker.go:579] Running sync for customresourcedefinition "baremetalhosts.metal3.io" (101 of 432)
2019-10-21T10:40:16.18425522Z I1021 10:40:16.184220       1 task_graph.go:583] Canceled worker 0
2019-10-21T10:40:16.184381244Z I1021 10:40:16.184360       1 task_graph.go:583] Canceled worker 3
...
2019-10-21T10:40:16.21772875Z I1021 10:40:16.217715       1 task_graph.go:603] Workers finished
2019-10-21T10:40:16.217777479Z I1021 10:40:16.217759       1 task_graph.go:611] Result of work: []
2019-10-21T10:40:16.217864206Z I1021 10:40:16.217846       1 task_graph.go:539] Stopped graph walker due to cancel
...
2019-10-21T10:43:08.743798997Z I1021 10:43:08.743740       1 sync_worker.go:453] Running sync quay.io/runcom/origin-release:v4.2-1196 (force=true) on generation 2 in state Reconciling at attempt 0
...
```

Where the CVO cancels some workers, sees that there are no errors, and decides "upgrade complete" despite never having attempted to push the bulk of its manifests.  With this commit, the result of work will include several worker-canceled errors, and we'll take another upgrade round instead of declaring success and moving into reconciling.

[1]: https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.1/754/artifacts/e2e-aws-upgrade/must-gather/registry-svc-ci-openshift-org-origin-4-1-sha256-f8c863ea08d64eea7b3a9ffbbde9c01ca90501afe6c0707e9c35f0ed7e92a9df/namespaces/openshift-cluster-version/pods/cluster-version-operator-5f5d465967-t57b2/cluster-version-operator/cluster-version-operator/logs/current.log

### DIFF
--- a/pkg/payload/task_graph.go
+++ b/pkg/payload/task_graph.go
@@ -612,5 +612,9 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
 	if len(errs) > 0 {
 		return errs
 	}
+	// if the context was cancelled, we may have unfinished work
+	if err := ctx.Err(); err != nil {
+		return []error{err}
+	}
 	return nil
 }
--- a/pkg/payload/task_graph_test.go
+++ b/pkg/payload/task_graph_test.go
@@ -721,12 +721,14 @@ func TestRunGraph(t *testing.T) {
 		wantErrs   []string
 	}{
 		{
+			name: "tasks executed in order",
 			nodes: []*TaskNode{
 				{Tasks: tasks("a", "b")},
 			},
 			order: []string{"a", "b"},
 		},
 		{
+			name: "nodes executed after dependencies",
 			nodes: []*TaskNode{
 				{Tasks: tasks("c"), In: []int{3}},
 				{Tasks: tasks("d", "e"), In: []int{3}},
@@ -756,6 +758,7 @@ func TestRunGraph(t *testing.T) {
 			},
 		},
 		{
+			name: "task error interrupts node processing",
 			nodes: []*TaskNode{
 				{Tasks: tasks("c"), In: []int{2}},
 				{Tasks: tasks("d"), In: []int{2}, Out: []int{3}},
@@ -781,6 +784,7 @@ func TestRunGraph(t *testing.T) {
 			},
 		},
 		{
+			name: "mid-task cancellation error interrupts node processing",
 			nodes: []*TaskNode{
 				{Tasks: tasks("c"), In: []int{2}},
 				{Tasks: tasks("d"), In: []int{2}, Out: []int{3}},
@@ -796,15 +800,15 @@ func TestRunGraph(t *testing.T) {
 					case <-time.After(time.Second):
 						t.Fatalf("expected context")
 					case <-ctx.Done():
-						t.Logf("got cancelled context")
-						return fmt.Errorf("cancelled")
+						t.Logf("got canceled context")
+						return ctx.Err()
 					}
 					return fmt.Errorf("error A")
 				}
 				return nil
 			},
 			want:     []string{"a", "b", "c"},
-			wantErrs: []string{"cancelled"},
+			wantErrs: []string{"context canceled"},
 			invariants: func(t *testing.T, got []string) {
 				for _, s := range got {
 					if s == "e" {
@@ -814,6 +818,7 @@ func TestRunGraph(t *testing.T) {
 			},
 		},
 		{
+			name: "task errors in parallel nodes both reported",
 			nodes: []*TaskNode{
 				{Tasks: tasks("a"), Out: []int{1}},
 				{Tasks: tasks("b"), In: []int{0}, Out: []int{2, 4, 8}},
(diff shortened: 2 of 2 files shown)


pkg/payload/task_graph.go
```diff
@@ -581,6 +581,7 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
 				select {
 				case <-nestedCtx.Done():
 					klog.V(4).Infof("Canceled worker %d", job)
+					errCh <-nestedCtx.Err()
```
smarterclayton: It would be more correct to, at the end of the function, if the context is cancelled, do

```
if err := ctx.Err(); err != nil && len(errs) == 0 {
  errs = append(errs, err)
}
```

smarterclayton: We don't need one error per worker.

smarterclayton:
you're also going to have to fix all the tests, and I'd like to see enough new tests that verify that this has the correct behavior on the sync worker (unit is fine probably, unless we can get the right behavior via one of the integration).  then I want to see upgrade soaks to verify this doesn't happen anymore.

wking:
> you're also going to have to fix all the tests...

Turned out that the problem was a bit deeper; 40b7e58 -> 5eaad22 brings in a more thorough fix that does not break preexisting tests.

> ...I'd like to see enough new tests that verify that this has the correct behavior on the sync worker (unit is fine probably...

5eaad22 has a new unit tests which fails without my channel/goroutine refactor.  Lots and lots of me trying to explain what was going on before, what was going wrong, and how I've fixed it in the 5eaad22 commit message ;).

pkg/payload/task_graph.go
wking: Fixed in 40b7e58 -> 5eaad22.

wking:
[e2e-aws-upgrade][1]:

```
level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Bootstrap Ignition Config\": failed to fetch dependency of \"Master Machines\": failed to generate asset \"Platform Credentials Check\": validate AWS credentials: mint credentials check: error simulating policy: Throttling: Rate exceeded\n\tstatus code: 400, request id: b03a5cbe-8a90-4a84-a3a7-c1300122210f
```

/test e2e-aws-upgrade

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/255/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/324

wking:
/bugzilla refresh

Also, now that I have a unit test, I think I can pull your hold:

/hold cancel

smarterclayton:
/hold

Hard no on refactoring this entire method.  Please only change the minimal function necessary to close the gap.  Adding a cancel error at the end is the safest minimal fix.  Do not refactor like this when critical changes are required.

wking:
> Please only change the minimal function necessary to close the gap.  Adding a cancel error at the end is the safest minimal fix.

I don't think we want to add the error when we have actual failed jobs.  And I like being able to say "and `${TASK}` was the first thing that didn't get run" to get better fingerprints when we lock up.  So I've pushed 5eaad22 -> f69e58d that leaves the extra goroutines in place and doesn't block the worker under-provisioning race, but does write to `errCh` with those new errors.  Is that acceptably small, or do I need to keep trimming?

wking:
Rerolled again with f69e58d -> 11676c7 to take @smarterclayton's minimal fix from #260.  The new commit message explains why I'm unhappy to drop my `1 incomplete task nodes, beginning with b` error and the potential (not very important) race against late cancels being reported as errors even if they come after we've successfully pushed the whole graph.  But getting this fixed is the most important thing, and we can revisit the quality of error message if we have many clusters stuck with `context canceled`.

wking:
And 11676c7 -> eaa3d19 to drop the no-longer-needed-because-we-don't-produce-an-error ba6f23e8fa24.

smarterclayton:
/hold cancel
/lgtm

We can follow up next week on a slightly better message summarize from syncworker#apply.

PR #256, opened 2019-10-22
bug 1764017: Fix replicas comparison

enxebre:
The CVO wrongly makes an update API call and returns `modified=true` when ensuring deployments with same number of replicas

### DIFF
--- a/lib/resourcemerge/apps.go
+++ b/lib/resourcemerge/apps.go
@@ -10,7 +10,7 @@ import (
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if required.Spec.Replicas != nil && required.Spec.Replicas != existing.Spec.Replicas {
+	if required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas {
 		*modified = true
 		existing.Spec.Replicas = required.Spec.Replicas
 	}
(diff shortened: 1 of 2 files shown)


enxebre:
cc @wking @abhinavdahiya @smarterclayton

lib/resourcemerge/apps_test.go
```diff
@@ -25,29 +23,29 @@ func TestEnsureDeployment(t *testing.T) {
...
 		},
 		{
 			name: "same replica count",
 			existing: appsv1.Deployment{
 				Spec: appsv1.DeploymentSpec{
-					Replicas: &replicas}},
+					Replicas: int32Pointer(2)}},
 			required: appsv1.Deployment{
 				Spec: appsv1.DeploymentSpec{
-					Replicas: &replicas}},
```
wking: Heh, using `replicas` instead of `expectedRepkicas` here should have tipped us off.  Thanks for the fix.

/lgtm

lib/resourcemerge/apps.go
```diff
@@ -10,7 +10,7 @@ import (
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if required.Spec.Replicas != nil && required.Spec.Replicas != existing.Spec.Replicas {
+	if required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas {
```
wking: Just to be extra-conservative, do we want:

```go
required.Spec.Replicas != nil && (required.Spec.Replicas == nil || *required.Spec.Replicas != *existing.Spec.Replicas)
 ```

/hold

PR #257, opened 2019-10-22
Bug 1755383: Bump to 1.16 kube libraries

sttts:


Gopkg.toml
```diff
@@ -15,49 +15,41 @@ required = [ "k8s.io/code-generator/cmd/client-gen", "k8s.io/gengo/types", "gith
...
-  # use origin fork with nullable support.
-  # TODO(sttts, sig-master): switch back to upstream version in 1.14
-  branch = "origin-4.1-kubernetes-1.13.4"
-  source = "https://github.com/openshift/kubernetes-apiextensions-apiserver"
+  version = "kubernetes-1.16.0"
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  version = "kubernetes-1.13.4"
+  version = "kubernetes-1.16.0"
```
soltysh: Don't you want to pin this to latest k8s which is 1.16.2?

sttts: Took the one from library-go. Otherwise, dep is unhappy.

abhinavdahiya:
overridden by https://github.com/openshift/cluster-version-operator/pull/249

/close

PR #258, opened 2019-10-22
Bug 1795412: Print whether CRDs are updated

sttts:
This will help to see hotloops due to missing defaults in manifests, leading invisibly to https://bugzilla.redhat.com/show_bug.cgi?id=1795412.


abhinavdahiya:
Since clients of the Apply are using updated only to print it, I would prefer that updated remain internal to Apply.. so maybe printing in apply itself.

sttts:
@abhinavdahiya added the logging to the apply func, simple twoliner. PTAL. This will help us to diagnose https://bugzilla.redhat.com/show_bug.cgi?id=1795412. Continuous CRD updates increase CPU usage of the apiservers.

PR #259, opened 2019-10-22
Bug 1755383: Add v1 CRD support and stop hotlooping

sttts:


abhinavdahiya:
/lgtm

will take up follow-up to add tests for these.

PR #260, opened 2019-10-22
payload: Task graph might lose some work if cancelled

smarterclayton:
A context cancel is always an error for the task graph


abhinavdahiya:
overridden by https://github.com/openshift/cluster-version-operator/pull/255

also 255 has unit test.

/close

PR #264, opened 2019-10-22
pkg/payload/task_graph: Handle node pushing and result collection without a goroutine

wking:
We want to give #255 time to cook in master before backporting it, so:

/hold

while that happens.  Once it does, I think we want this larger refactor in the master branch, even if we don't end up backporting it to earlier z streams.

The old code launched a goroutine for the pushing/reaping, which was an unecessary, and made error reporting on any outstanding tasks more complicated.  Dropping this goroutine let me get rid of `errCh` which used to be used to pass errors back to the main goroutine.  I also dropped the `wg.Wait()` goroutine, which used to be part of supporting the `range errCh` used in the main goroutine to collect failed jobs. But because `errCh` was filled directly from the worker goroutines, nothing used to be collecting the completeness of graph coverage from the pushing/reaping goroutine.

I've also restructured pushing/reaping to be more resistant to locking and spinning.  The old implementation had a node-pushing loop which attempted non-blocking pushes, then a result-reaping loop, then some condition checking, then a blocking result reaping attempt (after `we did nothing this round, so we have to wait for more`).  The idea with the blocking reap seems to have been that if the pusher/reaper didn't push anything (because `workCh` was full) and the earlier `for len(completeCh) > 0` reaping didn't pull anything in (because workers were busy working), that the pusher/reaper should block until it collected a result, in the hopes that worker which returned the result would have cleared a job out of `workCh` to give the pusher/reaper space to push a new job with the next loop iteration.  But if Go decides to give the pxusher/reaper more time in the scheduler, it might attempt the next `workCh` push before the job gets around to being scheduled and popping from `workCh`.  And then the pusher/reaper might trickle down to the blocking reap and wait for another worker (hopefully `maxParallelism > 1`) to return a result to unblock the pusher/reaper and give it another shot at a `workCh` push.  During this time, the worker that returned earlier is idling with nothing to do.

With this PR, when we have a next node to push, we have a single switch statement that blocks until we are either able to push the node or to reap a result.  So there's no need for a non-blocking push, and no chance at spinning, although it does mean we need to recalculate the next node after each channel action.  When we've been canceled, we stop pushing into `workCh`, but continue to reap from `resultCh` until we have no in-flight jobs left.  And if we have nothing to push, and there's nothing in-flight to reap, we're obviously done, so that choice is a lot simpler now.

I've dropped the `Waiting for workers to complete` log line, because `wg.Wait()` call should block for much less time now.  And because the main `RunGraph` goroutine is doing the waiting, we no longer need the `runTask, ok := <-workCh` read to protect against `workCh` being closed early.  With the `wg.Wait()` now getting called after we have drained all in-flight jobs (previously we launched it immediately after launching workers), there is less call for the `Waiting for...` line.

But the most externally noticeable change is that now, if we run without any failing jobs to give us errors, I'm filling in a new "incomplete task nodes" error so folks don't have to look in the CVO logs to see how far the CVO got before hanging.  It also allows us to not return the "I was canceled" error in cases where the cancellation happened late enough that we were still able to successfully process the full graph.

### DIFF
--- a/pkg/payload/task_graph.go
+++ b/pkg/payload/task_graph.go
@@ -5,14 +5,12 @@ import (
 	"fmt"
 	"math/rand"
 	"regexp"
-	"sort"
 	"strings"
 	"sync"
 
-	"k8s.io/klog"
-
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog"
 )
 
 // SplitOnJobs enforces the rule that any Job in the payload prevents reordering or parallelism (either before or after)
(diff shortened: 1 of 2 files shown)


wking:
#255 has cooked for a week and so has the 4.2 backport #262.  The [upgrade failure][1] was an install-time `error simulating policy: Throttling: Rate exceeded`, which is unrelated to this PR.

/hold cancel
/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/264/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/340

wking:
Rebased around #201 with fe98f76 -> 437c38c.

wking:
Rebased again with c674859 -> 5ee7d5b, because I'd forgotten I'd rebased this last week :facepalm:

pkg/payload/task_graph.go
```diff
@@ -437,169 +435,117 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
...
-					completeCh <- taskStatus{index: runTask.index, success: err == nil}
-					if err != nil {
-						errCh <- err
-					}
+					err := fn(ctx, runTask.tasks)
+					resultCh <- taskStatus{index: runTask.index, error: err}
 				}
 			}
-		}(i)
+		}(nestedCtx, i)
```
abhinavdahiya, line 491: do we need this nextedCtx ?

pkg/payload/task_graph.go
```diff
@@ -437,169 +435,117 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
...
+				case <-ctx.Done():
+					klog.V(4).Infof("Canceled worker %d while waiting for work", job)
 					return
-				case runTask, ok := <-workCh:
-					if !ok {
-						klog.V(4).Infof("No more work for %d", job)
-						return
-					}
+				case runTask := <-workCh:
 					klog.V(4).Infof("Running %d on %d", runTask.index, job)
```
abhinavdahiya, line 486:
```suggestion
					klog.V(4).Infof("Running %d on worker %d", runTask.index, job)
```

pkg/payload/task_graph.go
abhinavdahiya: hmm.. i see we use it to shutdown workers when the Sync is done or canceled.

could we move this nextexCtx definition closer to these worker start instead of top of Sync?

pkg/payload/task_graph.go
```diff
@@ -437,169 +435,117 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
...
+			inflight--
+		default: // no work to push and nothing in flight.  We're done
+			done = true
+		}
+		if !done {
+			nextNode = getNextNode()
+		}
+	}
+
+	cancelFn()
```
abhinavdahiya, line 521: the runner is 
- waiting on ctx
- waiting on work from workCh

the also passes along the ctx to the syn task fn.

maybe what we should do is,
the runner only loops on the workCh for work, and passes along the ctx to task sync fn so that we can terminate in progress work.
here we should just close workCh as there is no longer any work left to done.

that simplifies/compartmentalizes what's used for what.

pkg/payload/task_graph.go
wking: > could we move this nextexCtx definition closer to these worker start instead of top of Sync?

We may have to shift it back in the future if `RunGraph` grows something else it needs to shut down in the future.  But sure, shifted closer to worker start in 7f95e2e -> f6271d0.

pkg/payload/task_graph.go
wking: Fine with me, but that's an orthogonal change to master code.  I've filed a parallel #377 to make the change.

pkg/payload/task_graph.go
wking: From [the Go spec][1]:

> Sending to or closing a closed channel causes a run-time panic.

So if we manually call `close(workCh)` [here][2], then [this deferred call][3] will panic, right?  Maybe Go ignores panics from deferred functions, but still, I personally prefer having the worker watching both `nestedCtx` and `workCh`.

[1]: https://golang.org/ref/spec#Close
[2]: https://github.com/openshift/cluster-version-operator/commit/f6271d01756b8c744c1cbaf57dd9ea79ff8b836e#diff-bf8962ebeca893755f0b6fbd9b0252eaR507
[3]: https://github.com/openshift/cluster-version-operator/commit/f6271d01756b8c744c1cbaf57dd9ea79ff8b836e#diff-bf8962ebeca893755f0b6fbd9b0252eaR463

wking:
Rebased around 30bdaa8d276 (#377) with f6271d0 -> 55ef3d3.

PR #265, opened 2019-10-24
apiext: wait for CRD using v1beta1 for compatibility on upgrades

sttts:
This is softening 4ee7b07 (#259) to use the older v1beta1 API version for waiting for CRDs to be established. This fixes the creation of v1beta1 CRDs before kube-apiserver is updated during a 4.2->4.3 update.

Note: this will not fix the creation of v1 CRDs before kube-apiserver is upgraded. Hence, we are still restricted in the use of v1 CRDs for early manifests.

### DIFF
--- a/lib/resourcebuilder/apiext.go
+++ b/lib/resourcebuilder/apiext.go
@@ -71,12 +71,12 @@ func (b *crdBuilder) Do(ctx context.Context) error {
 	}
 
 	if updated {
-		return waitForCustomResourceDefinitionCompletion(ctx, b.clientV1, name)
+		return waitForCustomResourceDefinitionCompletion(ctx, b.clientV1beta1, name)
 	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


wking:
Context:

This is softening 4ee7b07e992 (#259) to support the older v1beta1 types because 4.2->4.3 upgrades will load the 4.3 cluster-version operator before loading the 4.3 OpenShift API server.  The 4.2 OpenShift API server doesn't support the v1 types, so the 4.3 CVO chokes and dies (example link?).  With this PR, the 4.3 CVO will use the v1beta1 types that both 4.2 and 4.3 OpenShift API servers understand.

Or something like that?

wking:
Looks good to me.  Not sure if there's a way we can talk Prow into running 4.2->PR-tip CI on this...

wking:
All green.  I guess we'll just see about 4.2->4.3 once this gets built into the 4.3 images ;).

/lgtm

PR #266, opened 2019-10-28
Replace ClusterVersions CRD with openshift/api artifact

sttts:


damemi:
This LGTM

Dockerfile
```diff
@@ -6,5 +6,6 @@ RUN hack/build-go.sh
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-version-operator/_output/linux/amd64/cluster-version-operator /usr/bin/
 COPY install /manifests
+COPY vendor/github.com/openshift/api/config/v1/*_cluster-version-operator_*.crd.yaml /manifests/
```
abhinavdahiya: I would rather like this to be explicit, rather than regex.

install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -1,531 +0,0 @@
...
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
-          type: string
-        kind:
-          description: 'Kind is a string value representing the REST resource this
-            object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
-          type: string
-        metadata:
-          description: ObjectMeta is metadata that all persisted resources must have,
```
wking:
```console
$ git diff -M20 origin/pr/266^^..origin/pr/266
...
         metadata:
-          description: ObjectMeta is metadata that all persisted resources must have,
-            which includes all objects users must create.
-          properties:
...
```

Where did the metadata details go?  Don't we want those?

damemi: @wking no, those were removed in https://github.com/openshift/api/pull/478 to conform to structural schema requirements

vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -0,0 +1,162 @@
...
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: spec hold the intent of how this operator should behave.
```
wking: I've filed openshift/api#499 to fix this `spec hold the` typo.

vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -0,0 +1,162 @@
...
+          type: object
+        status:
+          description: status holds the information about the state of an operator.  It
+            is consistent with status information across the kube ecosystem.
```
wking: I've filed openshift/api#500 expanding `kube` -> `Kubernetes`.

vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
```diff
@@ -0,0 +1,162 @@
...
+                  namespace:
+                    description: namespace of the referent.
+                    type: string
+                  resource:
+                    description: resource of the referent.
+                    type: string
+            versions:
+              description: versions is a slice of operand version tuples.  Operators
+                which manage multiple operands will have multiple entries in the array.  If
+                an operator is Available, it must have at least one entry.  You must
```
wking: Even operators with a single operand will have multiple entries, one for `operator` and one for the operand.  And when you're `Available`, you need to have an `operator` entry, not just "an entry".  I've filed openshift/api#501 to wordsmith that sort of thing.

install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
wking: > ... to conform to structural schema requirements

Can you link me to docs on these requirements?  I don't see anything that sounds like it [here][1].

[1]: https://github.com/kubernetes/community/blob/ecea14c0af705635a47d6dbc99dc050ed2373065/contributors/devel/sig-architecture/api-conventions.md#metadata

Dockerfile
sttts: fair enough. changing.

sttts: done

install/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml
sttts: Metadata is exclusively owned by the apiserver and apimachinery. A schema inside a CRD will interfere with apimachinery and we cannot guarantee that a CR from today works tomorrow.

Metadata is validated, but implicitly following the logic inside the kube-apiserver.

wking:
/lgtm

We can always bump the vendored dir again if/when my API PRs land.

PR #267, opened 2019-11-06
Bug 1766907: Add proxy support for CVO.

patrickdillon:
This PR adds proxy support to the CVO for verifying signatures. It centralizes some of the proxy logic used in Cincinnati to be shared. More significantly, it restructures the releaseVerifier struct to be able to use an up-to-date proxy. The releaseVerifier is created before a watch has been established on the cluster proxy. This PR ensures that the releaseVerifier gets the proxy after it is available to the operator and that the proxy settings are updated if there are any changes to the configuration.

Fixes BZ-1766907.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -2,7 +2,9 @@ package cvo
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"net/url"
 	"strconv"
 	"sync"
 	"time"
@@ -651,4 +653,22 @@ func (optr *Operator) defaultPreconditionChecks() precondition.List {
 	return []precondition.Precondition{
 		preconditioncv.NewUpgradeable(optr.cvLister),
 	}
+}
+
+// getTransportOpts retrieves the URL of the cluster proxy and the CA
+// trust, if they exist.
+func (optr *Operator) getTransportOpts() (*url.URL, *tls.Config, error) {
+	proxyURL, cmNameRef, err := optr.getHTTPSProxyURL()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var tlsConfig *tls.Config
+	if cmNameRef != "" {
+		tlsConfig, err = optr.getTLSConfig(cmNameRef)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	return proxyURL, tlsConfig, nil
 }
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strconv"
 	"sync"
@@ -24,6 +25,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/transport"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
@@ -207,6 +209,16 @@ func New(
 	return optr
 }
 
+// verifyClientBuilder is a wrapper around the operator's HTTPClient method.
+// It is used by the releaseVerifier to get an up-to-date http client.
+type verifyClientBuilder struct {
+	builder func() (*http.Client, error)
+}
+
+func (vcb *verifyClientBuilder) HTTPClient() (*http.Client, error) {
+	return vcb.builder()
+}
+
 // InitializeFromPayload retrieves the payload contents and verifies the initial state, then configures the
 // controller that loads and applies content to the cluster. It returns an error if the payload appears to
 // be in error rather than continuing.
@@ -223,8 +235,11 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	optr.releaseCreated = update.ImageRef.CreationTimestamp.Time
 	optr.releaseVersion = update.ImageRef.Name
 
+	// Wraps operator's HTTPClient method to allow releaseVerifier to create http client with up-to-date config.
+	clientBuilder := &verifyClientBuilder{builder: optr.HTTPClient}
+
 	// attempt to load a verifier as defined in the payload
-	verifier, err := verify.LoadFromPayload(update)
+	verifier, err := verify.LoadFromPayload(update, clientBuilder)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/cvo/cvo.go
```diff
@@ -652,3 +661,28 @@ func (optr *Operator) defaultPreconditionChecks() precondition.List {
 		preconditioncv.NewUpgradeable(optr.cvLister),
 	}
 }
+
+func (optr *Operator) getHTTPClientTransport() (http.RoundTripper, error) {
+	proxyURL, cmNameRef, err := optr.getHTTPSProxyURL()
```
abhinavdahiya: https://github.com/openshift/cluster-version-operator/blob/59601a89bb7a65cf2fbaa4697d5142091f68534f/pkg/cvo/availableupdates.go#L44-L53

does similar, so maybe both can use common method.

pkg/cvo/cvo.go
```diff
@@ -652,3 +661,28 @@ func (optr *Operator) defaultPreconditionChecks() precondition.List {
 		preconditioncv.NewUpgradeable(optr.cvLister),
 	}
 }
+
+func (optr *Operator) getHTTPClientTransport() (http.RoundTripper, error) {
```
abhinavdahiya: should all https clients in cvo pkg use this? because the name most definitely implies this.

pkg/verify/verify.go
```diff
@@ -74,7 +72,7 @@ var Reject Interface = rejectVerifier{}
 // if each provided public key has signed the release image digest. The signature may be in any
 // store and the lookup order is internally defined.
 //
-func LoadFromPayload(update *payload.Update) (Interface, error) {
+func LoadFromPayload(update *payload.Update, transport http.RoundTripper) (Interface, error) {
```
abhinavdahiya: i think passing the entire transport is not get an API for this function.

maybe we can think about passing a little more explicit item? transport.Config maybe, or the tlsConfig

jcpowermac: I would figure you would need to pass both the `tlsConfig` and `proxyURL`.
Something like what is in: 
https://github.com/jcpowermac/cluster-version-operator/blob/master/pkg/cvo/availableupdates.go

Then create the http client:
https://github.com/jcpowermac/cluster-version-operator/blob/d4e138fac270b557fb9d2592d000de40fcd9affd/pkg/cincinnati/cincinnati.go#L64-L72

pkg/cvo/cvo.go
wking: This is now centralized in `getTransportOpts`.

patrickdillon:
@abhinavdahiya I have tested this and verified that `storage.googleapis.com` is now read in the proxy logs which is an indication that this PR is working. 

I have also updated the PR description to reflect that we need to take into account that the `releaseVerifier` is created before the operator has a watch on the cluster proxy, so we cannot read the settings for the proxy when creating the releaseVerifier. We also need to make sure to update the client if the proxy changes. 

The current implementation addresses these issues by creating an interface implemented by the operator itself which will pull the current proxy settings and create a new httpClient whenever the releaseVerifier needs to verify signatures.

pkg/verify/verify.go
```diff
@@ -74,7 +72,7 @@ var Reject Interface = rejectVerifier{}
 // if each provided public key has signed the release image digest. The signature may be in any
 // store and the lookup order is internally defined.
 //
-func LoadFromPayload(update *payload.Update) (Interface, error) {
+func LoadFromPayload(update *payload.Update, clientBuilder clientBuilder) (Interface, error) {
```
abhinavdahiya: I don't think an argument of type `package private` is a good pattern for `public` function

pkg/cvo/cvo.go
```diff
@@ -224,7 +226,7 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	optr.releaseVersion = update.ImageRef.Name
 
 	// attempt to load a verifier as defined in the payload
-	verifier, err := verify.LoadFromPayload(update)
+	verifier, err := verify.LoadFromPayload(update, optr)
```
abhinavdahiya: instead of passing the whole optr object, can we pass a small type that just does HTTPClient .. 
you can `optr.verifyClientBuilder` or some such.

pkg/verify/verify.go
```diff
@@ -150,7 +142,19 @@ var validReleaseDigest = regexp.MustCompile(`^[a-zA-Z0-9:]+$`)
 type releaseVerifier struct {
 	verifiers map[string]openpgp.EntityList
 	stores    []*url.URL
-	client    *http.Client
+	client    clientBuilder
+}
+
+type clientBuilder interface {
+	GetHTTPClient() (*http.Client, error)
```
abhinavdahiya: nit: `GetHTTPClient` -> `HTTPClient`

pkg/cvo/cvo.go
patrickdillon: I think we can do that if we store pointers to the listers for proxy and trust bundle in the new type.

pkg/verify/verify.go
patrickdillon: Updated to export ClientBuilder interface type/

pkg/cvo/cvo.go
patrickdillon: Created a wrapper struct to hold the function.

pkg/verify/verify.go
patrickdillon: done

patrickdillon:
This is ready for another round of reviews. Suggestions have been incorporated and tested (I can see requests to storage.googleapis.com in proxy).

pkg/verify/verify.go
```diff
@@ -256,7 +261,11 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
 		case "http", "https":
 			copied := *store
 			copied.Path = path.Join(store.Path, name)
-			if err := checkHTTPSignatures(ctx, v.client, copied, maxSignatureSearch, verifier); err != nil {
+			client, err := v.clientBuilder.HTTPClient()
+			if err != nil {
+				return err
+			}
```
abhinavdahiya: I think we are createing an http client for each store, maybe we can do it once per Verify call ?

patrickdillon: Absolutely. I moved client creation out of the for loop.

patrickdillon:
This is ready for another round of reviews. Let me know if there are certain situations I can test. 

With this patch I can see requests to storage.googleapis.com on the proxy (which aren't there otherwise). Those requests get 404 responses but I believe that is to be expected.

PR #268, opened 2019-11-09
Bug 1685338: pkg/cvo: Reason granularity for RemoteFailed

wking:
We [want][1] to be able to distinguish these conditions, which can be due to internal misconfiguration or external Cincinnati/network errors.  The former can be fixed by cluster admins.  The latter could go either way.

I dropped the `len(upstream)` guard from `checkForUpdate` because there's already an earlier guard in `syncAvailableUpdates`.  The guard I'm removing is from db150e6db4 (#45).  The covering guard is from the later 286641de68 (#55).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1685338

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -58,22 +58,18 @@ func (err *Error) Error() string {
 // finding all of the children. These children are the available updates for
 // the current version and their payloads indicate from where the actual update
 // image can be downloaded.
-func (c Client) GetUpdates(upstream string, arch string, channel string, version semver.Version) ([]Update, error) {
+func (c Client) GetUpdates(uri *url.URL, arch string, channel string, version semver.Version) ([]Update, error) {
 	transport := http.Transport{}
 	// Prepare parametrized cincinnati query.
-	cincinnatiURL, err := url.Parse(upstream)
-	if err != nil {
-		return nil, &Error{Reason: "InvalidURI", Message: fmt.Sprintf("failed to parse upstream URL: %s", err)}
-	}
-	queryParams := cincinnatiURL.Query()
+	queryParams := uri.Query()
 	queryParams.Add("arch", arch)
 	queryParams.Add("channel", channel)
 	queryParams.Add("id", c.id.String())
 	queryParams.Add("version", version.String())
-	cincinnatiURL.RawQuery = queryParams.Encode()
+	uri.RawQuery = queryParams.Encode()
 
 	// Download the update graph.
-	req, err := http.NewRequest("GET", cincinnatiURL.String(), nil)
+	req, err := http.NewRequest("GET", uri.String(), nil)
 	if err != nil {
 		return nil, &Error{Reason: "InvalidRequest", Message: err.Error(), cause: err}
 	}
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -34,6 +34,23 @@ func NewClient(id uuid.UUID, proxyURL *url.URL, tlsConfig *tls.Config) Client {
 // Update is a single node from the update graph.
 type Update node
 
+// Error is returned when are unable to get updates.
+type Error struct {
+	// Reason is the reason suggested for the ClusterOperator status condition.
+	Reason string
+
+	// Message is the message suggested for the ClusterOperator status condition.
+	Message string
+
+	// cause is the upstream error, if any, being wrapped by this error.
+	cause error
+}
+
+// Error serializes the error as a string, to satisfy the error interface.
+func (err *Error) Error() string {
+	return fmt.Sprintf("%s: %s", err.Reason, err.Message)
+}
+
 // GetUpdates fetches the next-applicable update payloads from the specified
 // upstream Cincinnati stack given the current version and channel. The next-
 // applicable updates are determined by downloading the update graph, finding
@@ -46,7 +63,7 @@ func (c Client) GetUpdates(upstream string, arch string, channel string, version
 	// Prepare parametrized cincinnati query.
 	cincinnatiURL, err := url.Parse(upstream)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse upstream URL: %s", err)
+		return nil, &Error{Reason: "InvalidURI", Message: fmt.Sprintf("failed to parse upstream URL: %s", err)}
 	}
 	queryParams := cincinnatiURL.Query()
 	queryParams.Add("arch", arch)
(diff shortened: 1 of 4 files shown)


pkg/cincinnati/cincinnati.go
```diff
@@ -34,19 +34,36 @@ func NewClient(id uuid.UUID, proxyURL *url.URL, tlsConfig *tls.Config) Client {
...
+
 // GetUpdates fetches the next-applicable update payloads from the specified
 // upstream Cincinnati stack given the current version and channel. The next-
 // applicable updates are determined by downloading the update graph, finding
 // the current version within that graph (typically the root node), and then
 // finding all of the children. These children are the available updates for
 // the current version and their payloads indicate from where the actual update
 // image can be downloaded.
-func (c Client) GetUpdates(upstream string, arch string, channel string, version semver.Version) ([]Update, error) {
+func (c Client) GetUpdates(upstream string, arch string, channel string, version semver.Version) ([]Update, *Error) {
```
abhinavdahiya: i don't think we should be returning `*Error`, we should continue to return `error`

pkg/cincinnati/cincinnati.go
```diff
@@ -34,19 +34,36 @@ func NewClient(id uuid.UUID, proxyURL *url.URL, tlsConfig *tls.Config) Client {
 // Update is a single node from the update graph.
 type Update node
 
+// Error is returned when are unable to get updates.
+type Error struct {
```
abhinavdahiya: maybe implement `causer` interface https://github.com/pkg/errors#retrieving-the-cause-of-an-error
so that we can use the `github.com/pk/errors` to get the cause as it's used widely in the repo.

pkg/cvo/availableupdates.go
```diff
@@ -182,15 +182,12 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
 	}
 }
 
-func checkForUpdate(clusterID string, proxyURL *url.URL, tlsConfig *tls.Config, upstream, arch, channel string, currentVersion semver.Version) ([]cincinnati.Update, error) {
+func checkForUpdate(clusterID string, proxyURL *url.URL, tlsConfig *tls.Config, upstream, arch, channel string, currentVersion semver.Version) ([]cincinnati.Update, *cincinnati.Error) {
```
abhinavdahiya: We shouldn't be returning types errors, people who care about the type can check that for themselves.

pkg/cincinnati/cincinnati.go
```diff
@@ -34,19 +34,36 @@ func NewClient(id uuid.UUID, proxyURL *url.URL, tlsConfig *tls.Config) Client {
...
 // the current version and their payloads indicate from where the actual update
 // image can be downloaded.
-func (c Client) GetUpdates(upstream string, arch string, channel string, version semver.Version) ([]Update, error) {
+func (c Client) GetUpdates(upstream string, arch string, channel string, version semver.Version) ([]Update, *Error) {
 	transport := http.Transport{}
 	// Prepare parametrized cincinnati query.
 	cincinnatiURL, err := url.Parse(upstream)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse upstream URL: %s", err)
+		return nil, &Error{Reason: "InvalidURI", Message: fmt.Sprintf("failed to parse upstream URL: %s", err)}
```
abhinavdahiya: empty upstream will create cryptic error message as it's not quoted.

wking: Empty upstream is handled by an earlier guard.  But I can shift the parsing over there to make that more clear.

pkg/cvo/availableupdates.go
wking: returning `*Error` lets the caller avoid a cast attempt and never-gonna-run code for when that cast fails.  You'd still rather me return `error` and cast?

pkg/cincinnati/cincinnati.go
wking: I'd added `Cause()` locally, but didn't commit it because there were no consumers.  You want me to add it just in case?

pkg/cvo/availableupdates.go
abhinavdahiya: Yeah, i don't think the clients need to care if it's this error type, but just `error`, it's fine if some clients care and they can cast imo.

pkg/cincinnati/cincinnati.go
abhinavdahiya: > Empty upstream is handled by an earlier guard.

hmm, that not really the best way to handle this.. making sure the error msg are clear from this function is still important imo.

pkg/cvo/availableupdates.go
wking: Outside of unit tests, this package is the only consumer.  But rerolled to use `error` and a cast with 363a762 -> 8772ab1.

pkg/cincinnati/cincinnati.go
wking: Shifted to live next to the existing `len` guard with bff5982.  Or were you asking me to shift both of those guards down into `GetUpdates` instead of having them live up in `calculateAvailableUpdatesStatus`?

pkg/cincinnati/cincinnati.go
wking: Discussed and (hopefully) resolved in https://github.com/openshift/cluster-version-operator/pull/268#discussion_r344424103

pkg/cvo/availableupdates.go
```diff
@@ -127,6 +127,22 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
...
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "InvalidURI",
+			Message: fmt.Sprintf("failed to parse upstream URL: %s", err),
+		}
+	}
+
+	uuid, err := uuid.Parse(string(clusterID))
+	if err != nil {
+		return nil, configv1.ClusterOperatorStatusCondition{
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "InvalidID",
+			Message: fmt.Sprintf("invalid cluster ID: %v", err),
```
abhinavdahiya: nit:  `%v` vs `%s` ?

abhinavdahiya:
except https://github.com/openshift/cluster-version-operator/pull/268#discussion_r345034940
LGTM

pkg/cincinnati/cincinnati.go
abhinavdahiya: it's fine for now.

pkg/cvo/availableupdates.go
wking: > nit: `%v` vs `%s` ?

I dunno what the difference is for `error` values, but sure, fixed with bff5982 -> dca2686.

PR #269, opened 2019-11-13
Add annotations to manifests to exclude in hosted deployment

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See https://github.com/openshift/cluster-version-operator/pull/252

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: cluster-version-operator
   namespace: openshift-cluster-version
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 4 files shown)


install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: cluster-version-operator
   namespace: openshift-cluster-version
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
wking: Why `internal-openshift-hosted`?  Wouldn't this be `internal-cluster-version-operator` or some such?  Then folks can drop components with #252 and `EXCLUDE_MANIFESTS=internal-cluster-version-operator,.whatever,others,they,need-to-exclude`.  Or if we want a single knob for all the exclusions you need, we probably want docs (at least in the PR/commit message) for the intended components that fall under `internal-openshift-hosted`.

csrwng: `internal-openshift-hosted` is the profile that this exclusion is for ( a hosted control plane deployment as opposed to a traditional self-hosted one ). It doesn't prevent us from excluding the same manifest for other "profiles" such as single-node.

PR #272, opened 2019-11-25
Bug 1769717: Add support to EnsureServicePorts

enxebre:


enxebre:
cc @wking @abhinavdahiya

abhinavdahiya:
The services have a lot of defaulting from API server, can we make sure we have tests that show it works in case of defaults...? like TCP/UDP, http/https etc...

lib/resourceapply/core.go
```diff
@@ -57,6 +57,8 @@ func ApplyService(client coreclientv1.ServicesGetter, required *corev1.Service)
 	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 	selectorSame := equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector)
 	typeSame := equality.Semantic.DeepEqual(existing.Spec.Type, required.Spec.Type)
+	resourcemerge.EnsureServicePorts(modified, &existing.Spec.Ports, required.Spec.Ports)
```
wking: nit: instead of dropping this between the `selectorSame` and `typeSame` definitions and the if-statement consuming it, can we put this right after `EnsureObjectMeta` a few lines up?

lib/resourcemerge/core.go
```diff
@@ -202,6 +202,44 @@ func ensureContainerPort(modified *bool, existing *corev1.ContainerPort, require
 	}
 }
 
+func EnsureServicePorts(modified *bool, existing *[]corev1.ServicePort, required []corev1.ServicePort) {
+	for i, existingServicePort := range *existing {
+		var existingCurr *corev1.ServicePort
+		for _, requiredServicePort := range required {
+			if existingServicePort.Name == requiredServicePort.Name {
```
wking: Matching by name and then taking the union of in-cluster and in-manifest ports means there's no way to remove an entry.  Also, [`name` is optional][1], although I'm not clear on how that squares [with][2]:

> All ports within a ServiceSpec must have unique names.

Maybe it's "If you set a `name`, it must not duplicate an existing `ports` entry in the same Service"?

Do we even need to support folks adding additional `ports` values to Service manifests managed by the CVO?  Can we just clobber the existing value with the required value and set `modified` if that changed anything?

[1]: https://github.com/kubernetes/api/blob/11707872ac1cab52aefea0325f4698eaffeda3c8/core/v1/types.go#L3985

enxebre: >	// Optional if only one ServicePort is defined on this service.
https://github.com/kubernetes/api/blob/11707872ac1cab52aefea0325f4698eaffeda3c8/core/v1/types.go#L3984

>Matching by name and then taking the union of in-cluster and in-manifest ports means there's no way to remove an entry

There's no union, `required` overwrite `existing` via `ensureServicePort`. The loop below ensures unmatched entries are removed https://github.com/openshift/cluster-version-operator/pull/272/files#diff-f7675c7b39553e431cdc6c3ba8f5c482R221. See test cases for each scenario (with/without name) https://github.com/openshift/cluster-version-operator/pull/272/files#diff-415c13f11ffc32696c5d69b900b3fe58R283.

Instead of a nested loop this could get better performance by hashing the port name but I just wanted to replicate the pattern already present for similar funcs, e.g `ensureContainers` https://github.com/openshift/cluster-version-operator/blob/df38364e842d9a3b5caa2a07be92c1cee2d2651a/lib/resourcemerge/core.go#L64

enxebre:
>The services have a lot of defaulting from API server, can we make sure we have tests that show it works in case of defaults...? like TCP/UDP, http/https etc...

@abhinavdahiya not sure how you mean to test cover that?  the cvo does  `Semantic.DeepEqual` comparison in all `ensureX` funcs, if what's required does not have explicitly set the defaults given by api machinery then it'll return `*modified = true`

lib/resourcemerge/core_test.go
```diff
@@ -262,3 +262,284 @@ func TestEnsurePodSpec(t *testing.T) {
...
+							Port: 8080,
+						},
+						{
+							Name: "bar",
+							Port: 8081,
+						},
+					},
+				},
+			},
+		},
```
abhinavdahiya: can you add a unit test wrt 

1) input service has `Protocol` set due to defaulting and required doesn't https://github.com/kubernetes/api/blob/11707872ac1cab52aefea0325f4698eaffeda3c8/core/v1/types.go#L3988-L3991

2)  `TargetPort` https://github.com/kubernetes/api/blob/11707872ac1cab52aefea0325f4698eaffeda3c8/core/v1/types.go#L4005

the API defaults to Port if unset, also it might be a string defaulted if the port is known IANA etc..

3) also `NodePort` https://github.com/kubernetes/api/blob/11707872ac1cab52aefea0325f4698eaffeda3c8/core/v1/types.go#L4013

as this value is assigned some value if none is already provided.

abhinavdahiya:
@enxebre  approved the PR as it seems okay in terms of direction, but would make sure we have unit tests for https://github.com/openshift/cluster-version-operator/pull/272#discussion_r350861427 before we merge.

lib/resourcemerge/core_test.go
enxebre: @abhinavdahiya sorry still not clear what we want to cover. `EnsureServicePorts` as any other ensuring function in this library do not known about defaults. `EnsureServicePorts` is just mutating a struct when !equality.Semantic.DeepEqual and it will overwrite whatever `existing` has regardless this has been previously applied explicitly by a client or a by the api machinery defaulting.
The only think I can think of to increase coverage would be to apply the manifest after running `EnsureServicePorts`, then client.get the object back and validate that defaults are applied back properly, but the fake client does not apply any defaults unless we build a custom Reactor, therefore we'll be actually just covering our custom Reactor logic and this seems out of the scope to unit for `EnsureServicePorts`.
Let me know if this does not make sense to you and still want me to add anything.

abhinavdahiya: > EnsureServicePorts as any other ensuring function in this library do not known about defaults. EnsureServicePorts is just mutating a struct when !equality.Semantic.DeepEqual and it will clobber whatever existing has regardless this has been previously applied explicitly by a client or a by the api machinery defaulting.

I think we should prevent un-necessary updates, for the api machinery defaulting.. because that is known right...?
client changes are more than welcome to be over-written.

abhinavdahiya: > The only think I can think of to increase coverage would be to apply the manifest after running EnsureServicePorts, then client.get the object back and validate that defaults are applied back properly,

the unit tests already have an option of what exists in cluster currently and what we want..

for example, if the Protocol is empty (required), let's not run an update call because we know that it was defaulted. or when TargetPort is empty (required) make sure we are only running update when targetport-port differ because that is client change and not defaulting..

etc.. 

Even if we don't want to change these, unit-test cases that show that we are running update in presence of defaults is useful for future work when we can be smarter..

hopefully that provides more context...

enxebre: @abhinavdahiya thanks! Ok I see, I added defaulting for `Protocol` https://github.com/openshift/cluster-version-operator/pull/272/files#diff-f7675c7b39553e431cdc6c3ba8f5c482R244 and test coverage https://github.com/openshift/cluster-version-operator/pull/272/files#diff-415c13f11ffc32696c5d69b900b3fe58R551-R638
Other values might depend on the serviceType, I don't think we should redo api machinery defaulting logic here any farther.
Please let me know if it looks good now or still wanting any change

PR #274, opened 2019-11-27
docs/dev/clusteroperator: Point at the reconciliation docs

wking:
cde8d50a83 (#201) just landed, so drop some of the reconciliation docs from the parallel 105ea1f5e2 (#222) and similar in favor of a link to the new reconciliation location.

Also move the file-extension note over to the reconciliation side, since it's more generic than ClusterOperator.  It's not really user docs either, but that's the most generic location we have at the moment.

Also fix a few more `Failing` -> `Degraded` bits to catch up with fad0688c34 (#232) and similar.

### DIFF
--- a/docs/dev/clusteroperator.md
+++ b/docs/dev/clusteroperator.md
@@ -12,19 +12,10 @@ Cluster-admins make use of these resources to check the status of their clusters
 
 ## How should I include ClusterOperator Custom Resource in /manifests
 
-### How ClusterVersionOperator handles ClusterOperator in release image
+When ClusterVersionOperator encounters a ClusterOperator Custom Resource manifest, it does not push it into the cluster.
+Instead, the operators create ClusterOperator themselves, and the CVO waits for the ClusterOperator status to [match expected targets](../user/reconciliation.md#clusteroperator).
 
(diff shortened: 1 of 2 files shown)


docs/dev/clusteroperator.md
```diff
@@ -140,55 +131,34 @@ When your operator begins rolling out a new version it must continue to report t
...
-| Install completion[1] | current(whatever was being installed) | true | any | any
-| Begin upgrade | any | any | any | any
-| Begin upgrade (w/ force) | any | any | any | any 
-| Upgrade completion[2]| newVersion(target version for the upgrade) | true | false | false
-
-[1] Install works on all components in parallel, it does not wait for any component to complete before starting another one.
-
-[2] Upgrade will not proceed with upgrading components in the next runlevel until the previous runlevel completes.
-
-See also: https://github.com/openshift/cluster-version-operator/blob/a5f5007c17cc14281c558ea363518dcc5b6675c7/pkg/cvo/internal/operatorstatus.go#L176-L189
```
abhinavdahiya: why did we remove this section ?

wking: Because it's covered in the reconciliation doc.  Was there anything here you think is not reflected there?

PR #276, opened 2019-12-09
Bug 1781283: Set resource requests on the extract job

smarterclayton:
aren't in burstable.


smarterclayton:
Oh, already backported.

PR #277, opened 2019-12-09
Bug 1781284: Set resource requests on the extract job

smarterclayton:
aren't in burstable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## Building and Publishing CVO
 
 ```sh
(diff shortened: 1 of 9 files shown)


abhinavdahiya:
@smarterclayton you have https://github.com/openshift/cluster-version-operator/pull/277/commits/f01bd4342faaa90201e4c41a9af89fc2f69f14e9 and https://github.com/openshift/cluster-version-operator/pull/277/commits/3e6c910baffd6d201c1c3560d89ea43250def801 extra commits?

PR #278, opened 2019-12-09
Bug 1781284: Set resource requests on the extract job

smarterclayton:
aren't in burstable.


sdodson:
Backports #276 
For some reason bugzilla is rejecting links to this pr and the master branch PR. Overiding `bugzilla/valid-bug` label as everything looks proper to me.

PR #279, opened 2019-12-10
Refactor the verify package to not have dependencies on CVO

smarterclayton:
This is in preparation to move the verify code into library-go
for reuse in the admin tooling. verify has no dependencies,
payload specific logic is moved to pkg/cvo.

In addition, add a simple 64 entry cache of recent signatures
so that we can fetch the signatures in the current structure.

Written in a way to allow backport, but not explicitly targeted
at that (4.3 for in cluster verification is still being targeted).


pkg/verify/verify.go
```diff
@@ -151,16 +72,28 @@ type ClientBuilder interface {
 	HTTPClient() (*http.Client, error)
 }
 
-// simpleClientBuilder implements the ClientBuilder interface and should be used for testing.
+// DefaultClient uses the default http.Client for accessing signatures.
+var DefaultClient = simpleClientBuilder{}
+
+// simpleClientBuilder implements the ClientBuilder interface and may be used for testing.
```
abhinavdahiya: why add DefaultClient to package api?

pkg/verify/configmap.go
```diff
@@ -0,0 +1,87 @@
...
+//          the current time, although future versions of the CVO might allow host mounting of
+//          signatures.
+//
+// See https://github.com/containers/image/blob/ab49b0a48428c623a8f03b41b9083d48966b34a9/docs/signature-protocols.md
+// for a description of the signature store
+//
+// The returned verifier will require that any new release image will only be considered verified
+// if each provided public key has signed the release image digest. The signature may be in any
+// store and the lookup order is internally defined.
+func NewFromConfigMap(src string, data map[string]string, clientBuilder ClientBuilder) (*ReleaseVerifier, error) {
```
abhinavdahiya: this function is named ConfigMap but technically this doesn't care is the data came from one..?

pkg/cvo/cvo_test.go
```diff
@@ -3245,3 +3251,189 @@ func fakeClientsetWithUpdates(obj *configv1.ClusterVersion) *fake.Clientset {
...
+	redhatData, err := ioutil.ReadFile(filepath.Join("..", "verify", "testdata", "keyrings", "redhat.txt"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tests := []struct {
+		name          string
+		update        *payload.Update
+		want          bool
+		wantErr       bool
```
wking: Neither `want` nor `wantErr` are very specific about what came back.  Can we make `wantErr` a regexp or string matching the expected error message?  And `want` something about the set of loaded keys?

pkg/verify/configmap.go
wking: Maybe rename to `NewFromMap`?  The docstring also needs to be more generic, because this function doesn't care about looking up the configmap in a release's manifest set, it's just about creating a verifier from an already-found map.

pkg/verify/verify.go
```diff
@@ -151,16 +72,28 @@ type ClientBuilder interface {
 	HTTPClient() (*http.Client, error)
```
wking: Now that this is on the way to becoming cross-repo API, can we add some docstrings around why this is a function?  Is it intended to be run "occasionally"?  Before every network request?  At least every 5m?  What is the client caching model?

pkg/verify/verify.go
```diff
@@ -151,16 +72,28 @@ type ClientBuilder interface {
...
-// simpleClientBuilder implements the ClientBuilder interface and should be used for testing.
+// DefaultClient uses the default http.Client for accessing signatures.
+var DefaultClient = simpleClientBuilder{}
+
+// simpleClientBuilder implements the ClientBuilder interface and may be used for testing.
 type simpleClientBuilder struct{}
 
 // HTTPClient from simpleClientBuilder creates an httpClient with no configuration.
-func (s *simpleClientBuilder) HTTPClient() (*http.Client, error) {
+func (s simpleClientBuilder) HTTPClient() (*http.Client, error) {
```
wking: Why are we dropping the pointer collector?

pkg/verify/verify.go
```diff
@@ -151,16 +72,28 @@ type ClientBuilder interface {
...
 	return &http.Client{}, nil
 }
 
+// Verifiers returns a copy of the verifiers in this payload.
+func (v *ReleaseVerifier) Verifiers() map[string]openpgp.EntityList {
+	out := make(map[string]openpgp.EntityList)
```
wking: `, len(v.verifiers)`?

pkg/verify/verify.go
```diff
@@ -283,9 +223,54 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
 		}
 		return fmt.Errorf("unable to locate a valid signature for one or more sources")
 	}
+
+	v.cacheVerification(releaseDigest, signedWith)
```
wking: Why cache `signedWith`?  Seems like this is a boolean cache "yes, we verified that digest".  Possibly interesting metadata would include the signatures' and signing keys' expiration dates, so we knew when to invalidate the cache.

pkg/verify/verify.go
```diff
@@ -283,9 +223,54 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
...
+
 	return nil
 }
 
+// Signatures returns a copy of any cached signatures that have been validated
+// so far. It may return no signatures.
+func (v *ReleaseVerifier) Signatures() map[string][][]byte {
```
wking: Why would you want all of these?  I think `Verify` is a sufficient API.  If this is "which sigs will `oc` have to pack into a ConfigMap for disconnected folks?", I'd rather have:

```go
 func (v *ReleaseVerifier) Signatures(digest string) [][]byte
```

or some such.

pkg/verify/verify.go
```diff
@@ -245,6 +184,7 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
 				continue
 			}
 			delete(remaining, k)
```
wking: There's some contention between efficiently verifying a digest, where you only need one sig for each keyring, and shipping sigs around, where you may want too keep walking to find the "best" valid sig for each keyring.  Best might be "all valid sigs" or "latest expiration date" or "strongest cypher" or whatever.  Do we need a mode toggle to switch between efficient verification and interesting-sig collection?

pkg/verify/verify.go
```diff
@@ -283,9 +223,54 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
...
+}
+
+const maxSignatureCacheSize = 64
+
+// cacheVerification caches the result of signature check for a digest for later retrieval.
+func (v *ReleaseVerifier) cacheVerification(releaseDigest string, signedWith [][]byte) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	if len(signedWith) == 0 || len(releaseDigest) == 0 || v.signatureCache == nil {
```
wking: `NewReleaseVerifier` initializes the map and `hasVerified` lacks a `nil` guard.  Feels strange to have a `nil` guard here.

pkg/verify/verify.go
```diff
@@ -283,9 +223,54 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
...
+		return
+	}
+	// remove the new entry
+	delete(v.signatureCache, releaseDigest)
+	// ensure the cache doesn't grow beyond our cap
+	for k := range v.signatureCache {
+		if len(v.signatureCache) < maxSignatureCacheSize {
+			break
+		}
+		delete(v.signatureCache, k)
```
wking: This is pretty basic, and could lead to flapping two recent dogests while we hold a bunch of ancient digests in the cache.  Is there some LRU thing we can pull off a shelf?

pkg/verify/verify_test.go
```diff
@@ -162,206 +160,19 @@ func Test_releaseVerifier_Verify(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			v := &releaseVerifier{
+			v := &ReleaseVerifier{
```
wking: `NewReleaseVerifier`?

pkg/verify/verify_test.go
```diff
@@ -397,13 +208,83 @@ func Test_releaseVerifier_String(t *testing.T) {
...
+		DefaultClient,
+	)
+	for i := 0; i < maxSignatureCacheSize*2; i++ {
+		verifier.signatureCache[fmt.Sprintf("test-%d", i)] = [][]byte{[]byte("blah")}
+	}
+
+	if err := verifier.Verify(context.Background(), signedDigest); err != nil {
+		t.Fatal(err)
+	}
+	if sigs := verifier.Signatures(); len(sigs) != 64 || !reflect.DeepEqual(sigs[signedDigest], [][]byte{expectedSignature}) {
```
wking: 64 -> `maxSignatureCacheSize`?

pkg/verify/verify.go
wking: I guess what we want is time aware least recently used (TLRU) eviction, with the time component being min(sig expiration, key expiration).

pkg/verify/verify.go
smarterclayton: So test packages in others can use it.  I had to use it in 2-3 packages that wouldn't be in verify (and might not be in library go).  It's a standard implementation.

pkg/verify/configmap.go
smarterclayton: It uses the schema defined for a config map, even if it's coming from a map.  So the name is matching up with our config map rules, not any other set of rules.

pkg/verify/verify.go
smarterclayton: You don't need pointer on static types (which this is)

pkg/verify/verify.go
smarterclayton: because we're going to start saving these in a config map on cluster in the future so if the signing servers go down we can still upgrade to a version we saw in available versions.

pkg/verify/verify.go
smarterclayton: I want to get all of them in bulk at once for saving.

pkg/verify/verify.go
smarterclayton: If we get into that model, we could do that.  But I think that's cart before horse, and if we need it we would have the cache function potentially manage that.

pkg/verify/verify.go
smarterclayton: hasVerified doesn't need a nil guard.  Read from map doesn't.

pkg/verify/verify.go
smarterclayton: Map walk is random.  If you get flapping you should go play the lottery and retire on it.

I don't think we need anything more than random.

pkg/verify/verify_test.go
smarterclayton: No, I want to set signature cache explicitly.

pkg/verify/verify_test.go
smarterclayton: will change

pkg/verify/configmap.go
smarterclayton: `NewFromConfigMapData` is probably the most accurate.

smarterclayton:
I pushed the two commits that write back objects to the cluster, gives some better context for the existence of those methods.

smarterclayton:
integrated some feedback and responded to others, PTAL (including the newer commits so you can see why I had those methods)

smarterclayton:
Tracking in https://bugzilla.redhat.com/show_bug.cgi?id=1782982

pkg/verify/verify.go
wking: > because we're going to start saving these in a config map on cluster in the future...

I still don't see why that means "cache the sigs in memory".  Can we leave this off now and grow it back when we add the cache-to-ConfigMap functionality?

pkg/verify/verify.go
wking: > I want to get all of them in bulk at once for saving.

How about configuring a helper that gets called on signatures when they are added to the cache?  And another that is called when they are evicted?  Then your mirror case or "protect from the sig store going down" case could be addressed by attaching a collecter to the added-to-cache callback.  And you'd have a way to evict ConfigMaps in the "protect from the sig store going down" case when we dropped them from the in-memory cache.

pkg/verify/verify.go
wking: From [a 2013 Go comment][1]:

> Don't rely on the ordering for ranges over maps.  Relying on it to be random is just as
bad as relying on it to be sorted, or repeatable, or whatever.  Making it random is an
attempt to *prevent* people from relying on the order, not an *invitation* to rely on
the order for randomness.

Is that still current?  [The spec][2] has:

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

without committing to a particular quality of randomness.

[1]: https://github.com/golang/go/issues/5362#issuecomment-66078617
[2]: https://golang.org/ref/spec#RangeClause

pkg/verify/verify.go
```diff
@@ -47,25 +58,37 @@ var validReleaseDigest = regexp.MustCompile(`^[a-zA-Z0-9:]+$`)
 // digest - all verifiers must have at least one valid signature attesting the release
 // digest. If any failure occurs the caller should assume the content is unverified.
 type ReleaseVerifier struct {
-	verifiers     map[string]openpgp.EntityList
-	stores        []*url.URL
+	verifiers map[string]openpgp.EntityList
+
+	locations     []*url.URL
```
wking: Why keep the URI slice?  Can't we create just create `SignatureStore`s for these in `NewReleaseVerifier`?  Or drop `locations` from `NewReleaseVerifier` and have `NewFromConfigMapData` (or whatever we end up calling that) create the stores for each URI and call `WithStores`.

pkg/verify/verify.go
```diff
@@ -47,25 +58,37 @@ var validReleaseDigest = regexp.MustCompile(`^[a-zA-Z0-9:]+$`)
...
-		stores:        stores,
+		locations:     locations,
 		clientBuilder: clientBuilder,
 
 		signatureCache: make(map[string][][]byte),
 	}
 }
 
+// WithStores copies the provided verifier and adds any provided stores to the list.
+func (v *ReleaseVerifier) WithStores(stores []SignatureStore) *ReleaseVerifier {
```
wking: `stores ...SignatureStore` so folks adding a single store don't need to build a slice?

pkg/verify/verify.go
```diff
@@ -30,6 +30,17 @@ type Interface interface {
 	Verify(ctx context.Context, releaseDigest string) error
 }
 
+// SignatureStore retrieves signatures for a digest or returns an error. It requires String()
+// for describing the store.
```
wking: nit: I'd drop the `It requires String...` sentence, since the `String` method is motivated below inside the `interface` block.  `SignatureStore is a store from which signatures can be retrieved.` or something is probably a sufficient comment for the interface as a whole.  Or maybe `SignatureStore is a store selling signatures.` or some such ;).

pkg/verify/verify.go
```diff
@@ -30,6 +30,17 @@ type Interface interface {
...
+type SignatureStore interface {
+	// DigestSignatures returns zero or more signatures for the provided digest, or returns an
+	// error if no signatures could be retrieved.
```
wking: maybe obvious, but probably still worth calling out that the store is not required to perform any internal signature validation.

pkg/verify/verify.go
```diff
@@ -128,24 +151,44 @@ func (v *ReleaseVerifier) String() string {
...
-		fmt.Fprintf(&builder, " %s", store.String())
+	} else {
+		fmt.Fprintf(&builder, " - <ERROR: no locations or stores>")
 	}
 	return builder.String()
 }
 
 // Verify ensures that at least one valid signature exists for an image with digest
-// matching release digest in any of the provided stores for all verifiers, or returns
+// matching release digest in any of the provided locations for all verifiers, or returns
```
wking: `locations` -> `stores`?  Looks like you're checking both, so probably want to keep `stores` anyhow (and I think we want to [fold `locations` into `stores`][1], which would remove the distinction.

[1]: https://github.com/openshift/cluster-version-operator/pull/279/commits/abd6802c5a4c1fd840b28bb5d516d28431fee228#r359092643

pkg/cvo/cvo.go
```diff
@@ -238,17 +242,21 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	}
 	// XXX: set this to the cincinnati version in preference
 	if _, err := semver.Parse(update.ImageRef.Name); err != nil {
-		return fmt.Errorf("The local release contents name %q is not a valid semantic version - no current version will be reported: %v", update.ImageRef.Name, err)
+		return fmt.Errorf("the local release contents name %q is not a valid semantic version - no current version will be reported: %v", update.ImageRef.Name, err)
```
wking: This PR is epic.  Can we pull things like this and [the collector pointer bit][1] out into separate pull requests so they can land while the API refactor and feature extension stuff gets hashed out?

[1]: https://github.com/openshift/cluster-version-operator/pull/279#discussion_r356662055

pkg/cvo/cvo.go
```diff
@@ -294,15 +303,21 @@ func loadConfigMapVerifierDataFromUpdate(update *payload.Update, clientBuilder v
...
-			return nil, err
+			return nil, nil, err
 		}
-		return verifier, nil
+
+		// allow the verifier to consult the cluster for signature data, and also configure
+		// a process that writes signatures back to that store
+		signatureStore := verifyconfigmap.NewStore(configMapClient, nil)
+		verifier = verifier.WithStores(signatureStore)
+		persister := verify.NewSignatureStorePersister(signatureStore, verifier)
```
wking: Having `signatureStore` baked into `verifier` and also living as a sibling here feels strange, but I'm still wrapping my head around these new functions.

pkg/verify/verify.go
```diff
@@ -43,124 +52,78 @@ func (rejectVerifier) Verify(ctx context.Context, releaseDigest string) error {
...
-	}
-	return openpgp.ReadKeyRing(bytes.NewReader(data))
+// HTTPClient from simpleClientBuilder creates an http.Client with no configuration.
+func (s simpleClientBuilder) HTTPClient() (*http.Client, error) {
+	return &http.Client{}, nil
 }
 
+// maxSignatureSearch prevents unbounded recursion on malicious signature stores (if
+// an attacker was able to take ownership of the store to perform DoS on clusters).
 const maxSignatureSearch = 10
```
wking: this should be build into the store implementations, right?  By the time you get down to the verifier, the CVO has already been DoSed by the exploding store trying to fill the `DigestSignatures` request.

pkg/verify/verify.go
```diff
@@ -43,124 +52,78 @@ func (rejectVerifier) Verify(ctx context.Context, releaseDigest string) error {
...
+		signatureCache: make(map[string][][]byte),
+	}
 }
 
-// simpleClientBuilder implements the ClientBuilder interface and should be used for testing.
-type simpleClientBuilder struct{}
+// WithStores copies the provided verifier and adds any provided stores to the list.
+func (v *ReleaseVerifier) WithStores(stores ...SignatureStore) *ReleaseVerifier {
+	return &ReleaseVerifier{
+		verifiers:      v.verifiers,
```
wking: should be `v.Verifiers()` to get a separate copy?

pkg/verify/verify.go
wking: I've pushed up [a verify-store-api][1] branch (currently at 0cc3e41c1) to do this, polish the APIs up a bit more (e.g. using a callback so sig-store consumers can cancel sig iteration), and decompose things into more, smaller subpackages.  Still need to go through and rework the persistence logic to use [my preferred add/evict callbacks][2], but the API changes I already have in my branch are fairly orthogonal to that, so they should be ready for review now.

[1]: https://github.com/wking/cluster-version-operator/tree/verify-store-api
[2]: https://github.com/openshift/cluster-version-operator/pull/279#discussion_r359088967

pkg/verify/verify.go
smarterclayton: a controller feels super heavy weight.  Why do we need something more complex than "get" and "write"?

pkg/verify/verify.go
wking: > I've pushed up a verify-store-api branch...

Spun this out into the parallel #294.

pkg/verify/verify.go
wking: > Why do we need something more complex than "get" and "write"?

Maybe those are fine.  But I like decoupled logic so the verifier can say "I verify sigs, and if you want to bolt a cache or whatever on, here's where you hand over your callbacks".  Then the code that handles verification is all nice and compact, and the code that handles caching can live somewhere else all nice and compact, and the package that wants caching verification can glue the two together to get what they want.

pkg/verify/verify.go
jottofar: Picking up this change, or not, does not appear to be hugely impactful especially if we consider that much of this code may be OBE post 4.4 if we deploy a local cincy. Since this is holding up other 4.4 changes we need to make a decision. I would expect that this new code will be owned by the OTA team. Can we therefore move the final approval to that team and @LalatenduMohanty

wking:
/lgtm

We can refine in follow-up work ;).

PR #284, opened 2019-12-18
pkg/verify: Use a more-specific digest regular expression

wking:
Use the expression [from go-digest][1], based on [this spec][2].  We'd get more robust validation by using go-digest directly, but that would require a vendoring bump while bumping our regexp does not.

[1]: https://github.com/opencontainers/go-digest/blob/v1.0.0-rc1/digest.go#L62
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests


wking:
Rebased onto master (around the comment addition from 1d13e88aa0e, #279) with fac4277 -> a3aa56c.

wking:
Ah, moved to openshift/library-go#1032 to follow the code I'm trying to patch.

/close

PR #285, opened 2019-12-18
[poc] honor UpgradeableMinor condition

deads2k:
Just an experiment to see how much violence is required to support an `UpgradeableMinor` condition.  This is a very ugly example of how to do it, but it looks pretty safe to do.

### DIFF
--- a/pkg/payload/precondition/precondition.go
+++ b/pkg/payload/precondition/precondition.go
@@ -31,7 +31,7 @@ func (e *Error) Cause() error {
 // Precondition defines the precondition check for a payload.
 type Precondition interface {
 	// Run executes the precondition checks ands returns an error when the precondition fails.
-	Run(context.Context) error
+	Run(ctx context.Context, desiredVersion string) error
 
 	// Name returns a human friendly name for the precondition.
 	Name() string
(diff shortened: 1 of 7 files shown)


deads2k:
/hold

this is just a poc

pkg/payload/precondition/precondition.go
```diff
@@ -31,7 +31,7 @@ func (e *Error) Cause() error {
 // Precondition defines the precondition check for a payload.
 type Precondition interface {
 	// Run executes the precondition checks ands returns an error when the precondition fails.
-	Run(context.Context) error
+	Run(ctx context.Context, desiredVersion string) error
```
smarterclayton: If we're doing this can we just unify the loops for preconditions?

smarterclayton: (I don't care that much, but there's a lot of duplicate machinery)

deads2k: I could if that was desired.  Most of this is copy/pasta.  This was mostly about proving that the effort is fairly easy/straightforward, so we can argue about whether there is a need that can be easily met and not get distracted by the amount of work.

deads2k:
Alright, alternate pitch, what if Upgradeable actually meant "cannot upgrade y, but can upgrade z"?  We set it for...

 1.  UnsupportedConfigOverrides - those people are already unsupported.  
 2. TechPreview - can we tolerate having "weird" upgrade behavior sometimes for Y releases that ideally don't fully break things?

I don't know of another case where we set it offhand

shawn-hurley:
That would be helpful for removing items that are deprecated but still in use IIUC. I assume though that this would not help if I was removing something in 4.4 because the 4.3 CVO won’t have this bit?

smarterclayton:
The original argument was to only do minor upgrade blocking (since that’s what actually matters to most end users) - the “no z upgrade for tech preview” was a hardline stronger than our requirements.  I’d be willing to consider redefining upgradeable to mean “minor” given that and the history of why we did this in the first place.  Needs a bit of F2F discussion though about possible downsides.

deads2k:
closed in favor of https://github.com/openshift/cluster-version-operator/pull/291

PR #287, opened 2020-01-01
Fixing example command to fetch desired image from ClusterVersion resource

ravitri:
With reference to [1] and [2], suggesting to change example command and comments to replace _current_ to _desired_. Let me know if any corrections required.

[1]
https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L87

[2] 
```console
$ oc explain clusterversion.status.desired.image
KIND:     ClusterVersion
VERSION:  config.openshift.io/v1

FIELD:    image <string>

DESCRIPTION:
     image is a container image location that contains the update. When this
     field is part of spec, image is optional if version is specified and the
     availableUpdates field contains a matching version.
```

### DIFF
--- a/docs/dev/clusterversion.md
+++ b/docs/dev/clusterversion.md
@@ -1,14 +1,14 @@
 # ClusterVersion Custom Resource
 
-The `ClusterVersion` is a custom resource object which holds the current version of the cluster.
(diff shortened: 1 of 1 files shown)


ravitri:
Closing the PR for now. Will raise separate one if required in future.

PR #289, opened 2020-01-02
Bug 1787422: pkg/cvo/updatepayload: Drop ephemeral-storage request

wking:
Forward-porting #288 to 4.3.  Although 4.3 kubelet capacity reporting works, we still need to drop the 4.3 request, to support flows like:

1. 4.2 cluster running with 4.2 CVO and 4.2 kubelets (so no capacity reporting).
2. Admin requests an update to 4.3.1.
3. 4.2 CVO launches a version pod without requests, because of the 4.2 reversion (#288).  This works fine.
4. Update gets far enough to run a 4.3 CVO.
5. Update hangs on some 4.3.1 bug, while it's still running 4.2 kubelets.
6. Admin requests an update to 4.3.2.
7. 4.3 CVO launches a version pod with an ephemeral-storage request, which hangs because the 4.2 kubelets are still running and not reporting ephemeral-storage capacity.

PR #290, opened 2020-01-02
Bug 1787334: pkg/cvo/updatepayload: Drop ephemeral-storage request

wking:
Reopening #289 on release-4.3, since Prow seemed to not notice [my changing #289's base branch][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/289#event-2918526191


wking:
This partially reverts #278.

sdodson:
/retitle Bug 1787334: pkg/cvo/updatepayload: Drop ephemeral-storage request

wking:
We shouldn't have to force the labels.  The [previous issue][1] was that we had the 4.3 rhbz#1787334 depending on the 4.2 rhbz#1786315.  I've flipped that around to have the 4.2 bug depend on the 4.3 bug, so the bot should be happy now:

/bugzilla refresh

[1]: https://github.com/openshift/cluster-version-operator/pull/290#issuecomment-570323500

PR #291, opened 2020-01-02
Bug 1797624: loosen upgradeable condition to allow z-level upgrades

deads2k:
If current and desired have the same 4.y, then the upgrade is allowed even
if upgradeable==false.  If the 4.y is different, then upgrades are not allowed.
This permits CVE updates and fixes the current service-catalog upgrade problem.

Alternative to https://github.com/openshift/cluster-version-operator/pull/285

### DIFF
--- a/pkg/payload/precondition/clusterversion/upgradeable.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable.go
@@ -2,6 +2,7 @@ package clusterversion
 
 import (
 	"context"
+	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
@@ -29,7 +30,7 @@ func NewUpgradeable(lister configv1listers.ClusterVersionLister) *Upgradeable {
 // Run runs the Upgradeable precondition.
 // If the feature gate `key` is not found, or the api for clusterversion doesn't exist, this check is inert and always returns nil error.
 // Otherwise, if Upgradeable condition is set to false in the object, it returns an PreconditionError when possible.
-func (pf *Upgradeable) Run(ctx context.Context) error {
+func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.ReleaseContext) error {
 	cv, err := pf.lister.Get(pf.key)
 	if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
 		return nil
@@ -42,16 +43,43 @@ func (pf *Upgradeable) Run(ctx context.Context) error {
 			Name:    pf.Name(),
 		}
 	}
-	if up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable); up != nil && up.Status == configv1.ConditionFalse {
-		return &precondition.Error{
-			Nested:  err,
-			Reason:  up.Reason,
-			Message: up.Message,
-			Name:    pf.Name(),
-		}
+
+	// if we are upgradeable==true we can always upgrade
+	up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable)
+	if up == nil || up.Status != configv1.ConditionFalse {
+		return nil
 	}
-	return nil
+
+	// we can always allow the upgrade if there isn't a version already installed
+	if len(cv.Status.History) == 0 {
+		return nil
+	}
+
+	currentMinor := getEffectiveMinor(cv.Status.History[0].Version)
+	desiredMinor := getEffectiveMinor(releaseContext.DesiredVersion)
+
+	// if there is no difference in the minor version (4.y.z where 4.y is the same for current and desired), then we can still upgrade
+	if currentMinor == desiredMinor {
+		return nil
+	}
+
+	return &precondition.Error{
+		Nested:  err,
+		Reason:  up.Reason,
+		Message: up.Message,
+		Name:    pf.Name(),
+	}
 }
 
 // Name returns Name for the precondition.
 func (pf *Upgradeable) Name() string { return "ClusterVersionUpgradeable" }
+
+// getEffectiveMinor attempts to do a simple parse of the version provided.  If it does not parse, the value is considered
+// empty string, which works for the comparison done here for equivalence.
+func getEffectiveMinor(version string) string {
+	splits := strings.Split(version, ".")
+	if len(splits) < 2 {
+		return ""
+	}
+	return splits[1]
+}
--- a/pkg/payload/precondition/precondition.go
+++ b/pkg/payload/precondition/precondition.go
@@ -28,10 +28,20 @@ func (e *Error) Cause() error {
 	return e.Nested
 }
 
+// ReleaseContext holds information about the update being considered
+type ReleaseContext struct {
+	// DesiredVersion is the version of the payload being considered.
+	// While this might be a semantic version, consumers should not
+	// require SemVer validity so they can handle custom releases
+	// where the author decided to use a different naming scheme, or
+	// to leave the version completely unset.
+	DesiredVersion string
+}
+
 // Precondition defines the precondition check for a payload.
 type Precondition interface {
 	// Run executes the precondition checks ands returns an error when the precondition fails.
-	Run(context.Context) error
+	Run(ctx context.Context, releaseContext ReleaseContext) error
 
 	// Name returns a human friendly name for the precondition.
 	Name() string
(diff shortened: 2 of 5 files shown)


pkg/payload/precondition/precondition.go
```diff
@@ -31,7 +31,7 @@ func (e *Error) Cause() error {
 // Precondition defines the precondition check for a payload.
 type Precondition interface {
 	// Run executes the precondition checks ands returns an error when the precondition fails.
-	Run(context.Context) error
+	Run(ctx context.Context, desiredVersion string) error
```
smarterclayton: Probably better to pass a concrete struct here like `ReleaseContext struct { DesiredVersion types.Version }`

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -42,15 +43,33 @@ func (pf *Upgradeable) Run(ctx context.Context) error {
...
-	if up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable); up != nil && up.Status == configv1.ConditionFalse {
-		return &precondition.Error{
-			Nested:  err,
-			Reason:  up.Reason,
-			Message: up.Message,
-			Name:    pf.Name(),
-		}
+
+	// if we are upgradeable==true we can always upgrade
+	up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable)
```
smarterclayton: Update the test cases

shawn-hurley:
This might be slightly off-topic for this PR, but are there alerts that are fired for this condition and do we need to change the text for those alerts?

pkg/payload/precondition/clusterversion/upgradeable.go
deads2k: Cute. this was merged without tests initially :)

deads2k:
updated for comments, tests added.

deads2k:
terraform

/retest

wking:
One issue with this is that currently we block things like 4.2.13 -> 4.2.14 when the registry is set Unmanaged:

```json
{
  "conditions": [
    {
      "type": "Available",
      "status": "True",
      "lastTransitionTime": "2020-01-07T08:19:28Z",
      "reason": "Unmanaged",
      "message": "The registry configuration is set to unmanaged mode"
    },
    {
      "type": "Progressing",
      "status": "False",
      "lastTransitionTime": "2020-01-07T08:19:28Z",
      "reason": "Unmanaged",
      "message": "The registry configuration is set to unmanaged mode"
    },
    {
      "type": "Degraded",
      "status": "False",
      "lastTransitionTime": "2019-08-13T08:54:27Z",
      "reason": "Unmanaged",
      "message": "The registry configuration is set to unmanaged mode"
    }
  ],
  "versions": [
    {
      "name": "operator",
      "version": "4.2.13"
    }
  ],
  ...
}
```

Setting `Upgradeable=False` would allow the registry operator to say "I'm Unmanaged so I'm not going to block any upgrade attempts" and head off updates before they get started.  Currently that 4.2.13->4.2.14 update is hung mid-update with `Cluster operator image-registry is still updating`.  Do we have a recommended approach for this issue if we change the `Updateable` semantics?  Should the registry operator be going along with updates and bumping `version` for z-stream bumps in those cases?

deads2k:
> Setting `Upgradeable=False` would allow the registry operator to say "I'm Unmanaged so I'm not going to block any upgrade attempts" and head off updates before they get started. Currently that 4.2.13->4.2.14 update is hung mid-update with `Cluster operator image-registry is still updating`. Do we have a recommended approach for this issue if we change the `Updateable` semantics? Should the registry operator be going along with updates and bumping `version` for z-stream bumps in those cases?

Having an unmanaged component that blocks upgrades of any kind would really surprise me since it seems like they would always want to encourage updates that would allow the component to go back managed.  z-level upgrades should basically always work or you will be blocking CVE updates.

wking:
> ... are there alerts that are fired for this condition ...

Not yet, but we have an internal ticket to add some.

wking:
> Having an unmanaged component that blocks upgrades...

Sounds like [rhbz#1791934][1] will be resolved by having the operator bump its `version` to not block updates, so it is no longer a concern vs. this PR.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1791934

deads2k:
terraform again

/retest

sdodson:
/approve
This not a code review, but general concept was discussed and agreed upon during arch call. Marking as such.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -42,16 +43,43 @@ func (pf *Upgradeable) Run(ctx context.Context) error {
...
+	if up == nil || up.Status != configv1.ConditionFalse {
+		return nil
+	}
+
+	// we can always allow the upgrade if there isn't a version already installed
+	if len(cv.Status.History) == 0 {
+		return nil
+	}
+
+	currentMinor := getEffectiveMinor(cv.Status.History[0].Version)
```
wking: Instead of pulling this out of the cluster's ClusterVersion history, I'd rather pull it from [the local state][1].  But then you'd have to pipe it through to this precondition code.  Including it in `ReleaseContext` would be fine, although we'd want to rename to something more generic.  Maybe just `Context` (so `precondition.Context` in external packages)?  To get from the `Operator` into the `SyncWorker`... I dunno, seems like we could feed the context into `NewSyncWorkerWithPreconditions` [here][2]?  If that sounds plausible, I can work up a fixup commit that you could squash in here.

[1]: https://github.com/openshift/cluster-version-operator/blob/9281b45f9ab65a92b1dd4fe263683a6d60628d48/pkg/cvo/cvo.go#L628-L630
[2]: https://github.com/openshift/cluster-version-operator/blob/9281b45f9ab65a92b1dd4fe263683a6d60628d48/pkg/cvo/cvo.go#L274

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -42,16 +43,43 @@ func (pf *Upgradeable) Run(ctx context.Context) error {
...
+	desiredMinor := getEffectiveMinor(releaseContext.DesiredVersion)
+
+	// if there is no difference in the minor version (4.y.z where 4.y is the same for current and desired), then we can still upgrade
+	if currentMinor == desiredMinor {
+		return nil
```
wking: We may want to build the `*precondition.Error` above and log the fact that we're ignoring it due to the minor update here.  Or maybe push out an Event, since once we actually start updating, we're going to replace the CVO Pod and lose the old CVO's logs?  Leaving some breadcrumbs around this to make it less surprising would be nice, but is not something we need to block merging if it would be too difficult to implement.

pkg/payload/precondition/precondition.go
```diff
@@ -28,10 +28,17 @@ func (e *Error) Cause() error {
 	return e.Nested
 }
 
+// ReleaseContext holds information about the update being considered
+type ReleaseContext struct {
+	// DesiredVersion is the version of the payload being considered.  Remember, this is not guaranteed to be a version
+	// as a customer normally sees it. CI for instance uses invalid versions to ensure you don't rely on this behavior.
```
wking: What sort of CI names are you thinking of?  We also claim SemVer in the API, e.g. [here][1].  But putting that asside, I think it makes sense to have the CVO not require SemVer.  How about a docstring like:

```go
// DesiredVersion is the version of the payload being considered.
// While this might be a semantic version, consumers should not
// require SemVer validity so they can handle custom releases
// where the author decided to use a different naming scheme, or
// to leave the version completely unset.
```

or something?
[1]: https://github.com/openshift/api/blob/f263157f58d3b209b6215ad0ea3ef9287316132d/config/v1/types_cluster_version.go#L169-L174

pkg/payload/precondition/clusterversion/upgradeable.go
deads2k: It's not the way I would do it since we're already getting the information here (pre-existing gets) and you're already getting cached data, so it seems unnecessary and odd.  Could you do it as a follow-up instead if you really want it after?

pkg/payload/precondition/clusterversion/upgradeable.go
deads2k: that seems a little weird to do, since people are unlikely to wonder why their desired update was allowed and succeeded.

pkg/payload/precondition/precondition.go
deads2k: The versions in CI parse, but they don't mean anything.  Updating comment.

pkg/payload/precondition/clusterversion/upgradeable.go
wking: > that seems a little weird to do, since people are unlikely to wonder why their desired update was allowed and succeeded.

I can see component authors wondering why an update was initiated despite them having set `Upgradeable=False`.  But whatever, obviously not critical.

jottofar: Was discussing bug https://bugzilla.redhat.com/show_bug.cgi?id=1822513 with @wking and he suggested I update here as a follow-up to this thread.
The bug's underlying cause is that currentMinor is always being pulled from cv.Status.History[0].Version (https://github.com/openshift/cluster-version-operator/blob/40ec7e4f90b9fa0992145b926bd5f5bf6bd973a3/pkg/payload/precondition/clusterversion/upgradeable.go#L65) which contains the version being upgraded to and not the current version. In the bug's specific case, when --to-image is used, cv.Status.History[0].Version = "" which then fails the check for a z-level upgrade. Instead we should iterate the version history to find and use the first version with State == configv1.CompletedUpdate, which will yield the current version, and pull currentMinor from it.

jottofar: @deads2k thoughts on the above?

PR #293, opened 2020-01-07
pkg/verify: Ensure signature name matches release name

wking:
Guard against attackers who present a target image that has been signed by a trusted key, but which is not the expected release image.  For example, it could be a bare RHEL image and not an OpenShift release image.

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -88,7 +88,7 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
 	if index := strings.LastIndex(update.Image, "@"); index != -1 {
 		releaseDigest = update.Image[index+1:]
 	}
-	if err := r.verifier.Verify(ctx, releaseDigest); err != nil {
+	if err := r.verifier.Verify(ctx, fmt.Sprintf("quay.io/openshift-release-dev/ocp-release:%s", update.Version), releaseDigest); err != nil {
 		vErr := &payload.UpdateError{
 			Reason:  "ImageVerificationFailed",
 			Message: fmt.Sprintf("The update cannot be verified: %v", err),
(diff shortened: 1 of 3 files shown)


smarterclayton:
/hold

I don't understand the attack here.  We explicitly chose not to verify against the arbitrary from spec (which might change in the future arbitrarily).

wking:
If a malicious Cincinnati served a RHEL image as a release image update target, the CVO would validate the signature (which would pass).  Then it [calls `targetUpdatePayloadDir`][1], which [calls `ValidateDirectory`][2], which [looks for expected release-image stuff][3].  So we don't need this change to guard against that.

[1]: https://github.com/openshift/cluster-version-operator/blob/a45fa12c42047b24a70d8edaa17b85a7eaf6ad38/pkg/cvo/updatepayload.go#L106-L115
[2]: https://github.com/openshift/cluster-version-operator/blob/a45fa12c42047b24a70d8edaa17b85a7eaf6ad38/pkg/cvo/updatepayload.go#L124
[3]: https://github.com/openshift/cluster-version-operator/blob/a45fa12c42047b24a70d8edaa17b85a7eaf6ad38/pkg/payload/payload.go#L184-L196

wking:
There is no existing protection for downgrade attacks (e.g. 4.2.13 -> 4.0.0), but sometimes we want downgrades (e.g. if we recommend 4.y.z -> 4.y.(z+1), but 4.y.(z+1) proves unstable, so we recommend 4.y.(z+1) -> 4.y.z to get folks back to a safe place until we can cut 4.y.(z+2)).  I'm not worried about that in general.

I am worried that a malicious Cincinnati could attach a name (e.g. 4.2.13) to a different, signed-by-Red-Hat pullspec (e.g. the one for 4.0.0).  But @smarterclayton points out that you could guard against that by pulling and inspecting signed metadata before adding releases to ClusterVersion's `status.AvailableUpdates`, so you don't need this signature-name guard for that.

So I still think we want this PR to ensure we're using images that the signer expected to be OpenShift release images, but I'll survive with us saying "the signer thought the image was good for something, and on layer inspection it seems to be an OpenShift release image".

/close

PR #297, opened 2020-01-09
Migrate to go modules and k8s 1.17

vrutkovs:
Use go mod to handle dependencies and update to kube 1.17


vrutkovs:
/retest

flakes

PR #302, opened 2020-01-17
Bug 1790629: use st.Manifest.OriginalFilename if object doesn't have a name

vrutkovs:
This is helpful when one of the manifests doesn't have a name (like `kind: List` in [image-registry PR](https://github.com/openshift/cluster-image-registry-operator/commit/d197841e7944138ee2dc0815f2749c4f67657889#diff-88bbb853bcd5a5fbc188c46767630fadR1))

This list still cannot be applied (`apply` requires a name to find existing resource) but it would be helpful to debug these kind of issues.

TODO:
* [x] Add unittest

### DIFF
--- a/pkg/payload/task_test.go
+++ b/pkg/payload/task_test.go
@@ -1,0 +1,169 @@
+package payload
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/openshift/cluster-version-operator/lib"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/diff"
+)
+
+func TestTaskString(t *testing.T) {
+	tests := []struct {
+		name string
+		task *Task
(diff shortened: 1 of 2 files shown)


wking:
/approve

Ping me when there's a test, or you decide you don't need one ;).

vrutkovs:
@wking added a few tests, PTAL

pkg/payload/task_test.go
```diff
@@ -0,0 +1,169 @@
...
+						"kind":       "List",
+						"apiVersion": "v1",
+						"items": []interface{}{
+							"foo",
+							"bar",
+						},
+					},
+				},
+			}},
+			want: ` "list_manifest.yaml" (0 of 0)`,
```
wking: Leading space when we have no GKV isn't great, but that's not new in this PR, and this is much better than the [current][1]:

```
Could not update list \"\" (188 of 529)"
```

for lists.

/lgtm

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1790629#c0

PR #303, opened 2020-01-20
Bug 1793051: cvo: pass limiter when creating a store

vrutkovs:
`NewStore` creates a limiter, but doesn't pass it to `Store`. Later Store 
attempts to use it in `DigestSignatures` - and explodes


vrutkovs:
/hold

That doesn't seem to fix it :/

smarterclayton:
/hold cancel

Force merging because we believe we are broken upgrading "from".

PR #304, opened 2020-01-20
Revert "Merge pull request #279 from smarterclayton/config_map"

deads2k:
This reverts commit 27c4671aa0cbfbbf20324cf06145b44ec2bca45e, reversing
changes made to b89aa75004e83f49002250917bcfac82eb17bd05.

reverts https://github.com/openshift/cluster-version-operator/pull/279 in case it's causing the CI issues.

PR #305, opened 2020-01-21
Revert #303 and #279

sdodson:
Reverts #303 then #279 
Supersedes #304


LalatenduMohanty:
There is also https://github.com/openshift/cluster-version-operator/pull/304

smarterclayton:
I would prefer #306

smarterclayton:
Going to give 306 a chance, otherwise will revert.

smarterclayton:
/close

in favor of #307 if we need it

PR #306, opened 2020-01-21
verify: Copy clientBuilder during cloning of a verifier

smarterclayton:
Otherwise clientBuilder is missing.


wking:
This is another fix for #279

PR #308, opened 2020-01-21
Bug 1793209: Don't use protobuf client for SecurityContextConstraints

tkashem:


sdodson:
/uncc @abhinavdahiya 
/cc @wking @smarterclayton 
PTAL

mfojtik:
/lgtm

SCC are CR resources, protobuf does not work for those, this is critical fix.

PR #309, opened 2020-01-22
Further refactoring of CVO in prep to move verify package

jottofar:
Further refactoring of CVO, building upon https://github.com/openshift/cluster-version-operator/pull/279, in preparation of moving signature verify code into ibrary-go for reuse by CVO and admin tooling.
As part of this pull the manifest package has been changed in preparation to move it to library-go as well. Manifest specific parts of payload were moved to a render package under manifest.
The top-level API method for verifying and caching signatures was moved to verify package.


pkg/cvo/cvo_scenarios_test.go
```diff
@@ -70,12 +70,12 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 	})
 
 	o := &Operator{
-		namespace:                   "test",
-		name:                        "version",
+		namespace: "test",
```
wking: I think this is backsliding on Go 1.13's formatting:

```console
$ git --no-pager log --oneline -1
33c473e (HEAD, origin/pr/309) Further refactoring of CVO in prep to move verify package
$ go1.13.5 fmt ./pkg/cvo/
pkg/cvo/cvo_scenarios_test.go
$ git diff
diff --git a/pkg/cvo/cvo_scenarios_test.go b/pkg/cvo/cvo_scenarios_test.go
index 7c978ba..712e97a 100644
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -70,12 +70,12 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 	})
 
 	o := &Operator{
-		namespace: "test",
-		name:      "version",
+		namespace:                   "test",
+		name:                        "version",
 		enableDefaultClusterVersion: true,
-		queue:    workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cvo-loop-test"),
-		client:   client,
-		cvLister: &clientCVLister{client: client},
+		queue:                       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cvo-loop-test"),
+		client:                      client,
+		cvLister:                    &clientCVLister{client: client},
 	}
 
 	dynamicScheme := runtime.NewScheme()
@@ -159,7 +159,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 		},
 		Status: configv1.ClusterVersionStatus{
 			History: []configv1.UpdateHistory{
-			// empty because the operator release image is not set, so we have no input
+				// empty because the operator release image is not set, so we have no input
 			},
 			Conditions: []configv1.ClusterOperatorStatusCondition{
 				{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse},
@@ -430,7 +430,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 		},
 		Status: configv1.ClusterVersionStatus{
 			History: []configv1.UpdateHistory{
-			// empty because the operator release image is not set, so we have no input
+				// empty because the operator release image is not set, so we have no input
 			},
 			Conditions: []configv1.ClusterOperatorStatusCondition{
 				{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse},
@@ -691,7 +691,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 		},
 		Status: configv1.ClusterVersionStatus{
 			History: []configv1.UpdateHistory{
-			// empty because the operator release image is not set, so we have no input
+				// empty because the operator release image is not set, so we have no input
 			},
 			Conditions: []configv1.ClusterOperatorStatusCondition{
 				{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse},
```

Can we leave those indents alone to avoid churning the code?

pkg/manifest/render/render.go
```diff
@@ -0,0 +1,268 @@
...
+
+	buf := new(bytes.Buffer)
+	if err := tmpl.Execute(buf, config); err != nil {
+		return nil, errors.Wrapf(err, "failed to execute template")
+	}
+
+	return buf.Bytes(), nil
+}
+
+func LoadManifests(dir, releaseImage string) ([]manifest.Manifest, string, error) {
```
wking: This is not new to this PR, but it would be nice if `LoadManifests` had a godoc comment.  Maybe:

```go
// LoadManifests reads manifests out of the given directory and returns a slice of manifests and a hash over all loaded manifests.  The releaseImage argument is... what is manifestRenderConfig about anyway?
```

Seems like the hashing should be pulled out into a separate function, but whatever, we can revisit that later.  And I'm ok revisiting the whole godoc later too.

wking:
Got through the rest of this.  Looks ok to me.  There's some shuffling going on with the loading stuff, so and the [two][1] [comments][2] I made above.  We might be able to land it faster if we focus on just the manifest rename here, since that's pretty trivial, and then spin the loading refactor out into a follow-up PR?

[1]: https://github.com/openshift/cluster-version-operator/pull/309#discussion_r369884441
[2]: https://github.com/openshift/cluster-version-operator/pull/309#discussion_r369886533

pkg/cvo/cvo_scenarios_test.go
jottofar: Fixed. Not purposeful but I believe due to new editor options which I have backed out.

pkg/manifest/render/render.go
jottofar: Split func and added comments. releaseImage ends up being used in a template but i didn't investigate further so didn't talk to it in comment.

pkg/cvo/cvo_scenarios_test.go
wking: Fixed with 33c473e -> be97bc5, thanks.

wking:
> > There's some shuffling going on with the loading stuff
>
> Assume there's not an issue though?

We want something in that space, it just takes me longer to decide if what you currently have is the right something ;).  I know the manifest shift is right already.

> Do you mean focus on manifests first when doing a PR to library-go?

No, I mean punt the loading/render stuff here into a follow-up CVO PR so I can land the manifest shift now.  But it's not a big deal to keep them in the same PR either.

pkg/manifest/render/render.go
```diff
@@ -0,0 +1,274 @@
...
+	buf := new(bytes.Buffer)
+	if err := tmpl.Execute(buf, config); err != nil {
+		return nil, errors.Wrapf(err, "failed to execute template")
+	}
+
+	return buf.Bytes(), nil
+}
+
+// LoadManifests reads manifests out of the given directory and
+// returns a slice of manifests over all loaded manifests.
```
wking: nit: Drop "over all loaded manifests"?  I don't think we need to distinguish from a hypothentical function that loaded a bunch of manifests and then returned some subset of them ;).

pkg/manifest/render/render.go
```diff
@@ -0,0 +1,274 @@
...
+
+	agg := utilerrors.NewAggregate(errs)
+	if agg != nil {
+		return nil, errors.New(fmt.Sprintf("Error loading manifests from %s: %v", dir, agg.Error()))
+	}
+	return manifests, nil
+}
+
+// LoadManifestsHash returns a hash over all given manifests.
+func LoadManifestsHash(manifests []manifest.Manifest) string {
```
wking: nit: This should just be `Hash`.  And maybe `LoadManifests` should just be `Load`?  The package already provides a good deal of qualification, and we don't need to distinguish between other loaders or hashers in the package API.

pkg/manifest/render/render.go
```diff
@@ -0,0 +1,274 @@
...
+
+	CVOManifestDir     = "manifests"
+	ReleaseManifestDir = "release-manifests"
+
+	cincinnatiJSONFile  = "release-metadata"
+	ImageReferencesFile = "image-references"
+)
+
+// Render renders all the manifests from /manifests to outputDir.
+func Render(outputDir, releaseImage string) error {
```
wking: `oc` will benefit from having the `Manifest` type and loading functions in the shared `library-go` space.  I'm not clear on why we want `Render` in library-go.  I guess `oc adm release new ...` will consume it?  Seems orthogonal to ConfigMap signatures though.

jottofar: It's original methods, Render and renderDir, are required to work with manifests and since it was already needed, it is where I stuck the manifest related methods pulled from payload (see https://github.com/openshift/cluster-version-operator/pull/309/files#diff-8fadc5d8bd084451d026d5bf128dad66)

pkg/manifest/render/render.go
jottofar: Agree. i originally just left the comments alone in an effort to get this out.

pkg/manifest/render/render.go
jottofar: Agree. Not sure how OO "we" are but that naming convention makes sense to me.

pkg/manifest/render/render.go
jottofar: You're right, I totally confused the use of manifests here with the signatures we want to import and optionally save to file.

PR #311, opened 2020-01-31
Bug 1797123: pkg/cvo: Fetch proxy CA certs from openshift-config-managed/trusted-ca-bundle

wking:
[The API docs][1] recommend avoiding `trustedCA` (unless you happen to be the "proxy validator") and instead pulling the trust bundle from this managed namespace.  The docs also explain that the `trusted-ca-bundle` ConfigMap has already been merged with the system certificates, so we don't need to inject those locally.  If `trusted-ca-bundle` doesn't exist, we'll fall back to our local system store.

Our logic was touched most recently in 5968cdf933 (#231), which resolved the "looking in the wrong place" issue by looking at the `trustedCA` source (which feeds the proxy validator).  With this commit we switch that around and look at the proxy validator's output.

[1]: https://github.com/openshift/api/blob/f2a771e1a90ceb4e65f1ca2c8b11fc1ac6a66da8/config/v1/types_proxy.go#L44-L52

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -115,11 +115,12 @@ type Operator struct {
 	// syncBackoff allows the tests to use a quicker backoff
 	syncBackoff wait.Backoff
 
-	cvLister       configlistersv1.ClusterVersionLister
-	coLister       configlistersv1.ClusterOperatorLister
-	cmConfigLister listerscorev1.ConfigMapNamespaceLister
-	proxyLister    configlistersv1.ProxyLister
-	cacheSynced    []cache.InformerSynced
+	cvLister              configlistersv1.ClusterVersionLister
+	coLister              configlistersv1.ClusterOperatorLister
+	cmConfigLister        listerscorev1.ConfigMapNamespaceLister
+	cmConfigManagedLister listerscorev1.ConfigMapNamespaceLister
+	proxyLister           configlistersv1.ProxyLister
+	cacheSynced           []cache.InformerSynced
 
 	// queue tracks applying updates to a cluster.
 	queue workqueue.RateLimitingInterface
@@ -212,6 +213,7 @@ func New(
 
 	optr.proxyLister = proxyInformer.Lister()
 	optr.cmConfigLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigNamespace)
+	optr.cmConfigManagedLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigManagedNamespace)
 
 	// make sure this is initialized after all the listers are initialized
 	optr.upgradeableChecks = optr.defaultUpgradeableChecks()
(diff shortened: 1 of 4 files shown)


abhinavdahiya:
If the operator that creates the config map in that managed namespace is created by CVO, there is a cycle of dependencies..

wking:
> If the operator that creates the config map in that managed namespace is created by CVO, there is a cycle of dependencies.

Is that a problem?  We shouldn't need either Cincy graph fetches or signature pulls until the network operator comes up.

abhinavdahiya:
> > If the operator that creates the config map in that managed namespace is created by CVO, there is a cycle of dependencies.
> 
> Is that a problem? We shouldn't need either Cincy graph fetches or signature pulls until the network operator comes up.

Yeah, i don't think we should add this cycle.
We only use the proxy for Cincinnati requests today but I don't think we should add this cycle that would make it difficult if it was necessary for anything important later on.

wking:
> We only use the proxy for Cincinnati requests today but I don't think we should add this cycle that would make it difficult if it was necessary for anything important later on.

I can't imagine a case where the CVO would need to make external HTTP(S) requests before the network operator comes up.  And I like having the CVO continue to be able to make HTTP(S) requests via a proxy even in the event that a cluster-admin borks the ConfigMap they control (because the network operator will refuse to copy the broken CAs into the openshift-config-managed ConfigMap that it controls).  But whatever, I don't care enough to push this through ;).  I'm happy to reopen and rebase if there's a consensus to pick this approach back up later, just let me know.  Until then.

/close

wking:
Rebased onto master (around a trivial context conflict with #319) with 02fd56f -> f2c34cf.

wking:
Rebased around #349's move to `egress.go` with f2c34cf -> 1b3ae21.

pkg/cvo/cvo.go
```diff
@@ -211,6 +212,7 @@ func New(
 
 	optr.proxyLister = proxyInformer.Lister()
 	optr.cmConfigLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigNamespace)
+	optr.cmConfigManagedLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigManagedNamespace)
```
LalatenduMohanty, line 215: :+1:

pkg/cvo/cvo.go
```diff
@@ -752,17 +754,15 @@ func (optr *Operator) HTTPClient() (*http.Client, error) {
 // getTransportOpts retrieves the URL of the cluster proxy and the CA
 // trust, if they exist.
 func (optr *Operator) getTransportOpts() (*url.URL, *tls.Config, error) {
-	proxyURL, cmNameRef, err := optr.getHTTPSProxyURL()
+	proxyURL, err := optr.getHTTPSProxyURL()
```
LalatenduMohanty, line 757: @wking Do you know why was `cmNameRef` here?

pkg/cvo/egress.go
```diff
@@ -11,39 +11,38 @@ import (
...
 	}
-	return nil, "", nil
+	return nil, nil
 }
 
-func (optr *Operator) getTLSConfig(cmNameRef string) (*tls.Config, error) {
-	cm, err := optr.cmConfigLister.Get(cmNameRef)
-
+func (optr *Operator) getTLSConfig() (*tls.Config, error) {
+	cm, err := optr.cmConfigManagedLister.Get("trusted-ca-bundle")
```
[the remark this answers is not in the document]
wking: It is part of [the adopted enhancement][1].  I only use it in this one place, so it doesn't seem useful to create a local `const` for it.  But I can if you want one.

[1]: https://github.com/openshift/enhancements/blame/19d5bc00db44cf2614634e7e09e28743b6d87a62/enhancements/proxy/global-cluster-egress-proxy.md#L166

wking:
Rebased around #411 with 1b3ae214 -> c9fab435.

wking:
[e2e-upgrade][1] died on `Application behind service load balancer with PDB is not disrupted`, which is not us.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/311/pull-ci-openshift-cluster-version-operator-master-e2e-upgrade/1287836375776759808

LalatenduMohanty:
/hold
Going through some conversation we had with @bparees on this approach to make sure we are doing what we agreed upon.

PR #312, opened 2020-02-03
Bug 1794823: lib/resourcemerge: Do not attempt to change Job's immutable spec.selector

jottofar:
Tracing code and unit testing to help understand

### DIFF
--- a/docs/user/reconciliation.md
+++ b/docs/user/reconciliation.md
@@ -164,3 +164,10 @@ For subsequent updates, the builder blocks until:
 ### Job
 
 After pushing the merged Job into the cluster, the builder blocks until the Job succeeds.
+
+The cluster-version operator will panic if spec.selector is set because there are no clear use-cases for setting it in release manifests.
+
+Subsequent updates:
+
+* The cluster-version operator is currently unable to delete and recreate a Job to track changes in release manifests. Please avoid making changes to Job manifests until the cluster-version operator supports Job delete/recreate.
+* A Job's spec.selector will never be updated because spec.selector is immutable.
--- a/lib/resourcemerge/batch.go
+++ b/lib/resourcemerge/batch.go
@@ -2,22 +2,16 @@ package resourcemerge
 
 import (
 	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 )
 
 // EnsureJob ensures that the existing matches the required.
 // modified is set to true when existing had to be updated with required.
 func EnsureJob(modified *bool, existing *batchv1.Job, required batchv1.Job) {
-	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if existing.Spec.Selector == nil {
-		*modified = true
-		existing.Spec.Selector = required.Spec.Selector
+	if required.Spec.Selector != nil {
+		panic("spec.selector is not supported in Job manifests")
 	}
-	if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
-		*modified = true
-		existing.Spec.Selector = required.Spec.Selector
-	}
+	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 	setInt32Ptr(modified, &existing.Spec.Parallelism, required.Spec.Parallelism)
 	setInt32Ptr(modified, &existing.Spec.Completions, required.Spec.Completions)
 	setInt64Ptr(modified, &existing.Spec.ActiveDeadlineSeconds, required.Spec.ActiveDeadlineSeconds)
--- a/lib/resourcemerge/batch_test.go
+++ b/lib/resourcemerge/batch_test.go
@@ -1,0 +1,91 @@
+package resourcemerge
+
+import (
+	"testing"
+
+	batchv1 "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pointer "k8s.io/utils/pointer"
+)
+
+func TestEnsureJob(t *testing.T) {
+	labelSelector := metav1.LabelSelector{}
+	tests := []struct {
+		name     string
+		existing batchv1.Job
+		required batchv1.Job
+
+		expectedModified bool
+		expectedPanic    bool
+		expected         batchv1.Job
+	}{
+		{
+			name: "different backofflimit count",
+			existing: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(2)}},
+			required: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(3)}},
+
+			expectedModified: true,
+			expected: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(3)}},
+		},
+		{
+			name: "same backofflimit count",
+			existing: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(2)}},
+			required: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(2)}},
+
+			expectedModified: false,
+			expected: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					BackoffLimit: pointer.Int32Ptr(2)}},
+		},
+		{
+			name: "required-Selector not nil",
+			existing: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					Selector:       &labelSelector,
+					ManualSelector: pointer.BoolPtr(false)}},
+			required: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					Selector:       &labelSelector,
+					ManualSelector: pointer.BoolPtr(true)}},
+
+			expectedPanic: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			defer func() {
+				switch r := recover(); r {
+				case nil:
+					if test.expectedPanic {
+						t.Errorf(test.name + " should have panicked!")
+					}
+				default:
+					if !test.expectedPanic {
+						panic(r)
+					}
+				}
+			}()
+			modified := pointer.BoolPtr(false)
+			EnsureJob(modified, &test.existing, test.required)
+			if *modified != test.expectedModified {
+				t.Errorf("mismatch modified got: %v want: %v", *modified, test.expectedModified)
+			}
+
+			if !equality.Semantic.DeepEqual(test.existing, test.expected) {
+				t.Errorf("mismatch Job got: %v want: %v", test.existing, test.expected)
+			}
+		})
+	}
+}


lib/resourcemerge/batch.go
```diff
@@ -3,26 +3,34 @@ package resourcemerge
...
+		// If ManualSelector is not True we don't care about Selector
+		if required.Spec.ManualSelector {
+			if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
+				return
+			}
+		}
+		setBoolPtr(modified, &existing.Spec.ManualSelector, required.Spec.ManualSelector)
 	}
+	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+
```
jottofar: @wking check my re-factoring logic here per my comments/code changes

lib/resourcemerge/batch.go
```diff
@@ -8,21 +8,26 @@ import (
...
+	// finish as Selector is immutable.
+	if required.Spec.ManualSelector != nil {
+
+		// If ManualSelector is not True we don't care about Selector
+		if *required.Spec.ManualSelector {
+			if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
+				return
+			}
+		}
+		setBoolPtr(modified, &existing.Spec.ManualSelector, required.Spec.ManualSelector)
```
wking: `setBoolPrt` [is a noop when the required pointer is `null`][1], so we can leave this `setBoolPtr` call alone in its previous location without changing the logic.  Can you reroll to do that?

[1]: https://github.com/openshift/cluster-version-operator/blob/abfb48da582c81937bb6a08abaef163199c68f6e/lib/resourcemerge/core.go#L511-L513

lib/resourcemerge/batch.go
```diff
@@ -8,21 +8,26 @@ import (
...
-		existing.Spec.Selector = required.Spec.Selector
+	// To set Selector you must set ManualSelector to True. If 'required'
+	// sets Selector to value not equal to existing, let current Job
+	// finish as Selector is immutable.
+	if required.Spec.ManualSelector != nil {
+
+		// If ManualSelector is not True we don't care about Selector
+		if *required.Spec.ManualSelector {
+			if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
+				return
```
wking: I don't think we want this early `return`, because it's going to skip `Parallelism` and all the subsequent bits getting set later in `EnsureJob`.  Maybe you want the previous:

```go
*modified = true
existing.Spec.Selector = required.Spec.Selector
```

here?

lib/resourcemerge/batch.go
```diff
@@ -8,21 +8,26 @@ import (
...
+			}
+		}
+		setBoolPtr(modified, &existing.Spec.ManualSelector, required.Spec.ManualSelector)
 	}
+	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
```
wking: Why shift this down?  Can we leave it where it was to make the diff smaller and easier to read?

lib/resourcemerge/batch.go
```diff
@@ -8,21 +8,26 @@ import (
...
-	if !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
-		*modified = true
-		existing.Spec.Selector = required.Spec.Selector
+	// To set Selector you must set ManualSelector to True. If 'required'
+	// sets Selector to value not equal to existing, let current Job
+	// finish as Selector is immutable.
+	if required.Spec.ManualSelector != nil {
+
+		// If ManualSelector is not True we don't care about Selector
+		if *required.Spec.ManualSelector {
```
wking: Shifting the `ManualSelector` `setBoolPtr` back to its original location will allow you to collapse these cases into a single:

```go
if required.Spec.ManualSelector != nil &&
   *required.Spec.ManualSelector &&
   !equality.Semantic.DeepEqual(existing.Spec.Selector, required.Spec.Selector) {
```

lib/resourcemerge/batch.go
wking: Assorted comments inline, no need for a separate thread on them all ;).

lib/resourcemerge/batch.go
wking: Talked to @jottofar out of band, and he linked to [these][1] [docs][2].  After reading them over, it seems like setting your own Job `selector` is useful for fancy shell games where you delete/replace Jobs without interrupting the underlying Pods.  It's not clear to me that CVO consumers would need anything that sophisticated, and certainly not clear to me how they would know what to feed into `selector` in their static manifest even if they did want something like that.  So how about:

```go
if required.Spec.Selector != nil {
  panic("spec.selector is not supported in Job manifests")
}
```

And ideally we'd reroll the resource managers to allow for returning errors instead of panicking, but a refactor like that is probably too invasive to be backportable, so we shouldn't do it in this PR.

[1]: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#specifying-your-own-pod-selector
[2]: https://cloud.google.com/kubernetes-engine/docs/how-to/jobs#specifying_a_pod_selector

lib/resourcemerge/batch.go
jottofar: We can't change selector or we'll get the immutable error. This basically says, let the current job finish and then take care of all changes in next cycle UNLESS the selector just happens to be the same. Then we go ahead and break out and pick up all other changes.

lib/resourcemerge/batch.go
jottofar: Because, if my logic stands, my first check is to see if I need to bail on doing any modifications.

lib/resourcemerge/batch.go
jottofar: True, if we move setBoolPtr. I find it more logical to have steBoolPtr where I moved it but could still do a collapse. My main goal here is to first ensure the logic is correct.

lib/resourcemerge/batch.go
wking: > ...my first check is to see if I need to bail on doing any modifications

These in-memory edits are not expensive.  And [we shouldn't be bailing][1] anyway.  So I don't think we need to shift this around in an attempt to save cycles.

[1]: https://github.com/openshift/cluster-version-operator/pull/312/#discussion_r374947620

lib/resourcemerge/batch.go
wking: > We can't change selector or we'll get the immutable error.

Wouldn't that apply to all changes that set `modified` `true`?  Or is `selector` especially immutable?  For example, if you change the pod template, I'd have naively expected that to also need a delete/create Job.

jmrodri:
@jottofar this was the PR that originated this bug: https://github.com/openshift/cluster-svcat-apiserver-operator/pull/74

jottofar:
> @jottofar this was the PR that originated this bug: [openshift/cluster-svcat-apiserver-operator#74](https://github.com/openshift/cluster-svcat-apiserver-operator/pull/74)

@jmrodri yep, this will become the PR to fix https://bugzilla.redhat.com/show_bug.cgi?id=1794823

docs/user/reconciliation.md
```diff
@@ -164,3 +164,10 @@ For subsequent updates, the builder blocks until:
 ### Job
 
 After pushing the merged Job into the cluster, the builder blocks until the Job succeeds.
+
+Since neither the use case requiring a Job's spec.selector to be explicitly set nor the handling thereof is well defined the code will panic if a user attempts to set spec.selector.
```
wking: Maybe rephrase to:

> The cluster-version operator will panic if `spec.selector` is set, because `selector` is immutable and there are no clear use-cases for setting it in release manifests.

or some such.  That leads with the behavior, so folks who don't care about the reasoning can skip over it more easily.

docs/user/reconciliation.md
```diff
@@ -164,3 +164,10 @@ For subsequent updates, the builder blocks until:
...
+
+Subsequent updates:
+
+* Job update handling is not well-defined and is therefore not recommended.
+* A Job's spec.selector is immutable so a Job's spec.selector will never be updated.
```
wking: I think this point is covered in your earlier `selector` sentence, so we can drop the reference here.

docs/user/reconciliation.md
```diff
@@ -164,3 +164,10 @@ For subsequent updates, the builder blocks until:
 ### Job
 
 After pushing the merged Job into the cluster, the builder blocks until the Job succeeds.
+
+Since neither the use case requiring a Job's spec.selector to be explicitly set nor the handling thereof is well defined the code will panic if a user attempts to set spec.selector.
+
+Subsequent updates:
+
+* Job update handling is not well-defined and is therefore not recommended.
```
wking: Is this getting at "we'll probably want to delete/recreate jobs on changes, but don't have code for that yet"?  If so, maybe rephrase to:

> Current limitations:
>
> * The cluster-version operator is currently unable to delete and recreate a job to track changes in release manifests.  Please avoid making changes to Job manifests until it grows that support.

or some such.

lib/resourcemerge/batch_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+				if !equality.Semantic.DeepEqual(test.existing, test.expected) {
+					t.Errorf("mismatch Job got: %v want: %v", test.existing, test.expected)
+				}
+			})
+		}
+	}
+}
+
+func int32Ptr(i int32) *int32 {
```
wking: You can use `pointer`, which you already import, instead of defining this and your later `boolPtr` yourself.  [Docs][1]. 

[1]: https://godoc.org/k8s.io/utils/pointer#Int32Ptr

lib/resourcemerge/batch_test.go
```diff
@@ -0,0 +1,103 @@
...
+			expectedModified: false,
+			expected: batchv1.Job{
+				Spec: batchv1.JobSpec{
+					Selector:       &labelSelector,
+					ManualSelector: boolPtr(false)}},
+		},
+	}
+
+	for i, test := range tests {
+		if i == len(tests)-1 {
```
wking: This is a brittle switch.  Can you add `expectedPanic bool` or some such to your test-case structure and switch on that instead?

docs/user/reconciliation.md
jottofar: Rephrased for clarity but left out reference to immutable which really has nothing to do with a user explicitly setting the spec.selector in the manifest. The fact it is immutable is relevant if CVO were to attempt an update so that's why I included it under 'Subsequent updates'.

docs/user/reconciliation.md
jottofar: Clarified a bit but left it here (see https://github.com/openshift/cluster-version-operator/pull/312#discussion_r375605029)

lib/resourcemerge/batch_test.go
jottofar: Fixed. I copied the original code from apps_test.go which I guess is OBE due to updates to pointer package perhaps.

lib/resourcemerge/batch_test.go
jottofar: Agree, not good!

wking: 6b58e7a has `expectedPanic`.

lib/resourcemerge/batch_test.go
wking: 6b58e7a uses `pointer` functions.

docs/user/reconciliation.md
wking: 6b58e7a wording looks good to me.

docs/user/reconciliation.md
wking: > Clarified a bit but left it here...

6b58e7a wording looks good to me.

docs/user/reconciliation.md
wking: 6b58e7a wording looks good to me.

jottofar:
> I think I liked this better when we didn't `recover` from unexpected panics. Can we do that with closures or something?

Check whether panic is expected and re-raise if not.

jottofar:
> no need to set `expected` or `expectedModified` in the panic case, is there? I don't think we need to make assertions about how panicking calls mutate their arguments.

Fixed. Making assertions was not intent.

wking:
/lgtm

[Previous update][1] failed with a CI-registry flake:

```
error: unable to extract layer sha256:... from registry.svc.ci.openshift.org/ci-...: received unexpected HTTP status: 504 Gateway Time-out
```

That's tracked in [rhbz#1792002][2], apparently @openshift/openshift-team-developer-productivity-test-platform needs to periodically kick the registry pods or something to keep them working.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/312/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/478
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1792002

PR #316, opened 2020-02-06
Bug 1768260: lib,pkg: provide detailed errors for workload failures

abhinavdahiya:
waitFor{Deployment,Daemonset} now return UpdateError's so that the sync cycle can include detailed errors.

deployment:
- when replicafailure condition is true, returns WorkloadNotProgressing update error.
- when available condition is false or progressing is true, returns WorkloadNotAvailable update error.

daemonset:
It returns WorkloadNotAvailbale update error whenever the daemonset has not rolled out.

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -6,16 +6,19 @@ import (
 	"strings"
 
 	configv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
-	"github.com/openshift/cluster-version-operator/lib"
-	"github.com/openshift/cluster-version-operator/lib/resourceapply"
-	"github.com/openshift/cluster-version-operator/lib/resourceread"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
+
+	"github.com/openshift/cluster-version-operator/lib"
+	"github.com/openshift/cluster-version-operator/lib/resourceapply"
+	"github.com/openshift/cluster-version-operator/lib/resourceread"
+	"github.com/openshift/cluster-version-operator/pkg/payload"
 )
 
 type deploymentBuilder struct {
@@ -77,7 +80,9 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 	return nil
 }
 func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.DeploymentsGetter, deployment *appsv1.Deployment) error {
-	return wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
+	iden := fmt.Sprintf("%s/%s", deployment.Namespace, deployment.Name)
+	var lastErr error
+	err := wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
 		d, err := client.Deployments(deployment.Namespace).Get(deployment.Name, metav1.GetOptions{})
 		if errors.IsNotFound(err) {
 			// exit early to recreate the deployment.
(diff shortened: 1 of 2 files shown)


abhinavdahiya:
ping @wking

lib/resourcebuilder/apps.go
```diff
@@ -86,41 +91,77 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
-			d.Status.UpdatedReplicas,
-			d.Status.ReadyReplicas,
-			d.Status.UnavailableReplicas,
-			strings.Join(deploymentConditions, ""))
+		klog.Errorf("deployment %s is in unknown state", iden)
 		return false, nil
 	}, ctx.Done())
+	if err != nil {
+		if err == wait.ErrWaitTimeout && lastErr != nil {
+			return lastErr
```
bparees: would it be worth dumping the deployment yaml, especially if we're in the "deployment is in unknown state" case?

It's a tradeoff between messy logs and making people jump through more hoops to dig up debug info.

lib/resourcebuilder/apps.go
```diff
@@ -203,7 +251,19 @@ func waitForDaemonsetRollout(ctx context.Context, client appsclientv1.DaemonSets
...
+			Nested:  fmt.Errorf("daemonset %s is progressing; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedNumberScheduled, d.Status.DesiredNumberScheduled, d.Status.NumberAvailable, d.Status.DesiredNumberScheduled),
+			Reason:  "WorkloadNotAvailable",
+			Message: fmt.Sprintf("daemonset %s is progressing", iden),
+			Name:    iden,
+		}
 		return false, nil
 	}, ctx.Done())
+	if err != nil {
+		if err == wait.ErrWaitTimeout && lastErr != nil {
+			return lastErr
```
bparees: same comment here about considering dumping the resource yaml.

lib/resourcebuilder/apps.go
abhinavdahiya: I don't think printing the entire YAML is the right thing to do, we shouldn't rely on this to provide the entire debugging, we have the `oc adm must-gather` to give us more details.

sdodson: I don't think we should dump the YAML unless in all situations we're confident that would immediately make root cause clear. Is there an example of other higher level tools doing such in the product already?

bparees: I can't think of one offhand, but is there any scenario in which an install fails with this error message and the user does not immediately have to go look up that yaml to figure out the next step?

abhinavdahiya: I think in all situations we are better served if we use bootstrap gather and must-gather to collect failure logs.

Also looking at the entire object wouldn't necessarily give user what's failing. If there is a specific section that provides more values that _might_ be more appropriate.

lib/resourcebuilder/apps.go
```diff
@@ -86,41 +91,77 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
+			lastErr = &payload.UpdateError{
+				Nested:  fmt.Errorf("deployment %s has some pods failing; unavailable replicas=%d", iden, d.Status.UnavailableReplicas),
+				Reason:  "WorkloadNotProgressing",
+				Message: fmt.Sprintf("deployment %s has a replica failure %s: %s", iden, replicafailureCondition.Reason, replicafailureCondition.Message),
+				Name:    iden,
+			}
+			return false, nil
+		}
+
+		if availableCondition == nil && availableCondition.Status == corev1.ConditionFalse {
```
jstuever: Should this be a `!= nil`?

lib/resourcebuilder/apps.go
```diff
@@ -86,41 +91,77 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
+			lastErr = &payload.UpdateError{
+				Nested:  fmt.Errorf("deployment %s is not available; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedReplicas, d.Status.Replicas, d.Status.AvailableReplicas, d.Status.Replicas),
+				Reason:  "WorkloadNotAvailable",
+				Message: fmt.Sprintf("deployment %s is not available %s: %s", iden, availableCondition.Reason, availableCondition.Message),
+				Name:    iden,
+			}
+			return false, nil
+		}
+
+		if progressingCondition == nil && progressingCondition.Status == corev1.ConditionTrue {
```
jstuever: `!= nil`?

lib/resourcebuilder/apps.go
sdodson: I think bootstrap gather and must-gather are really only useful to support and engineering. We don't provide any guidance on what to do with that data which is something that I think we need to consider as an organization going forward.

In my opinion this PR as is does a decent job making it clear where next to look. I worry that dumping YAML is both noisy and shifts more debugging into the CVO and Installer. Lets go with things as is and see how that helps.

sdodson:
/hold
e2e-aws-upgrade is broken after master went to 4.5, going to see if there's an ongoing effort to address that but will likely override this to enable 4.4 backport.

PR #317, opened 2020-02-06
bug 1794817: lib/resourcebuilder/batch.go: skip waiting for job to complete when in Init mode

abhinavdahiya:
When the cluster version operator is in Initialization mode, don't wait for jobs to complete to allow for maximum traversal of the requirement..


abhinavdahiya:
ping @wking

sdodson:
/assign @smarterclayton 
Or @wking PTAL

PR #318, opened 2020-02-07
cvo: When installing or upgrading, fast-fill cluster-operators

smarterclayton:
The must-gather and insights operator depend on cluster operators
and related objects in order to identify resources to create. Because
cluster operators are delegated to the operator install and upgrade
failures of new operators can fail to gather the requisite info if
the cluster degrades before those steps.

Add a new selective Precreating install mode and do a single pass
over all cluster operators in the payload without retries at the beginning
of an initializing or upgrading sync pass to attempt to create the
ClusterOperators if they don't exist.

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -8,6 +8,7 @@ import (
 	"time"
 	"unicode"
 
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -48,16 +49,16 @@ func readClusterOperatorV1OrDie(objBytes []byte) *configv1.ClusterOperator {
 }
 
 type clusterOperatorBuilder struct {
-	client   ClusterOperatorsGetter
-	raw      []byte
-	modifier resourcebuilder.MetaV1ObjectModifierFunc
-	mode     resourcebuilder.Mode
+	client       ClusterOperatorsGetter
+	createClient configclientv1.ClusterOperatorInterface
+	raw          []byte
+	modifier     resourcebuilder.MetaV1ObjectModifierFunc
+	mode         resourcebuilder.Mode
 }
 
 func newClusterOperatorBuilder(config *rest.Config, m lib.Manifest) resourcebuilder.Interface {
-	return NewClusterOperatorBuilder(clientClusterOperatorsGetter{
-		getter: configclientv1.NewForConfigOrDie(config).ClusterOperators(),
-	}, m)
+	client := configclientv1.NewForConfigOrDie(config).ClusterOperators()
+	return NewClusterOperatorBuilder(clientClusterOperatorsGetter{getter: client}, client, m)
 }
 
 // ClusterOperatorsGetter abstracts object access with a client or a cache lister.
@@ -75,10 +76,11 @@ func (g clientClusterOperatorsGetter) Get(name string) (*configv1.ClusterOperato
 
 // NewClusterOperatorBuilder accepts the ClusterOperatorsGetter interface which may be implemented by a
 // client or a lister cache.
-func NewClusterOperatorBuilder(client ClusterOperatorsGetter, m lib.Manifest) resourcebuilder.Interface {
+func NewClusterOperatorBuilder(client ClusterOperatorsGetter, createClient configclientv1.ClusterOperatorInterface, m lib.Manifest) resourcebuilder.Interface {
 	return &clusterOperatorBuilder{
-		client: client,
-		raw:    m.Raw,
+		client:       client,
+		createClient: createClient,
+		raw:          m.Raw,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -585,12 +585,14 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 	}
 	graph := payload.NewTaskGraph(tasks)
 	graph.Split(payload.SplitOnJobs)
+	var precreateObjects bool
 	switch work.State {
```
deads2k: for those of us less familiar, which other cases are there?  I don't see a default where the precreate would be false.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -97,6 +99,13 @@ func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
 	if b.modifier != nil {
 		b.modifier(os)
 	}
+	if b.mode == resourcebuilder.PrecreatingMode {
+		_, err := b.createClient.Create(os)
+		if err != nil && kerrors.IsAlreadyExists(err) {
```
deads2k: I think we can skip any error.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -97,6 +99,13 @@ func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
...
+			err = nil
+		}
+		return err
```
deads2k: after a successful create, need to `UpdateStatus` and ignore any errors.

pkg/cvo/internal/operatorstatus.go
smarterclayton: I just want to log unexpected errors in case we find out that early fill is hanging or something

pkg/cvo/sync_worker.go
smarterclayton: reconciling mode

wking:
[images][1] failed with a CI-registry flake:

```
could not copy stable imagestreamtag: Timeout: request did not complete within allowed duration
```

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/318/pull-ci-openshift-cluster-version-operator-master-images/837

smarterclayton:
Looks like machine-config ends up without any relatedObjects - not sure why.

kikisdeliveryservice:
> Looks like machine-config ends up without any relatedObjects - not sure why.

Where should we be looking to see that the relatedObjects are missing? Don't really know what I'm supposed to look at here?

deads2k:
> > Looks like machine-config ends up without any relatedObjects - not sure why.
> 
> Where should we be looking to see that the relatedObjects are missing? Don't really know what I'm supposed to look at here?

Based on the method name here, it looks like relatedObjects are only set by the MCO during the first updateStatus, not all updatestatus calls.  Meaning the MCO isn't reconciling.

deads2k:
opened https://github.com/openshift/machine-config-operator/pull/1566

sdodson:
@wking @abhinavdahiya Can we work towards getting this reviewed by end of week?

pkg/cvo/internal/operatorstatus.go
```diff
@@ -48,16 +49,16 @@ func readClusterOperatorV1OrDie(objBytes []byte) *configv1.ClusterOperator {
 }
 
 type clusterOperatorBuilder struct {
-	client   ClusterOperatorsGetter
-	raw      []byte
-	modifier resourcebuilder.MetaV1ObjectModifierFunc
-	mode     resourcebuilder.Mode
+	client       ClusterOperatorsGetter
+	createClient configclientv1.ClusterOperatorInterface
```
abhinavdahiya: qq: if this is supposed to be only the Create, is there a create only interface??

pkg/cvo/internal/operatorstatus.go
```diff
@@ -97,6 +99,26 @@ func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
...
+		if err != nil {
+			if kerrors.IsAlreadyExists(err) {
+				return nil
+			}
+			return err
+		}
+		clusterOperator.Status.RelatedObjects = os.Status.DeepCopy().RelatedObjects
+		if _, err := b.createClient.UpdateStatus(clusterOperator); err != nil {
+			if kerrors.IsConflict(err) {
+				return nil
```
abhinavdahiya: hmm.. why would we not error? this could cause the cvo to move out of precreating mode, without relatedObjects set and that seemed like was the entire goal of this..?

pkg/cvo/sync_worker.go
```diff
@@ -608,6 +610,28 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 		// perform an orderly roll out by payload order, using some parallelization
 		// but avoiding out of order creation so components have some base
 		graph.Parallelize(payload.ByNumberAndComponent)
+		precreateObjects = true
```
abhinavdahiya: the clusteroperator sync tries to sync the related objects in precreating mode.

for during initlization and upgrade? mode.. cvo will try to reset the relatedObjs to what's in release-image when cluster operators might be adding new ones in the same time.

pkg/cvo/internal/operatorstatus.go
deads2k: > qq: if this is supposed to be only the Create, is there a create only interface??

No and we need update status.

pkg/cvo/internal/operatorstatus.go
deads2k: > hmm.. why would we not error? this could cause the cvo to move out of precreating mode, without relatedObjects set and that seemed like was the entire goal of this..?

We want to have a shot at every other clusteroperator resource.  If we have a conflict, we're done with this one, but not all of them.

pkg/cvo/sync_worker.go
deads2k: > for during initlization and upgrade? mode.. cvo will try to reset the relatedObjs to what's in release-image when cluster operators might be adding new ones in the same time.

Yeah, this code defers all conflicts to avoid fighting.  The ones in the payload are a good starting point.


> for during initlization and upgrade?

Yes, in the end it should do this for both.

deads2k:
/hold

integration is failing consistently

deads2k:
It was green and went red.

deads2k:
/hold cancel

this is getting fixed.

sdodson:
/hold
Clear this once we've fixed the integration test. Or yell at me to override it on Monday if that doesn't end up happening.

deads2k:
waiting on https://github.com/openshift/release/pull/8224 I think

enxebre:
we need to update https://github.com/openshift/cluster-version-operator/blob/master/docs/user/reconciliation.md#clusteroperator

https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#how-should-i-include-clusteroperator-custom-resource-in-manifests to reflect this

PR #319, opened 2020-02-11
Bug 1797806: metrics: include upi installs in cluster_installer

crawford:
This is a follow-on to c1477328, which added the cluster_installer
series originally. Recently, the installer started injecting a new
ConfigMap (openshift-install-manifests), which includes information
about the manifest generation stage of the install. Because this phase
always happens, this ConfigMap is included in every cluster created by
this newer installer. We can now look at the presence of this new
ConfigMap and the absence of the openshift-install ConfigMap to
determine that the cluster was installed using UPI. We can additionally
look at its contents to determine information about the installer.

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


crawford:
https://infogw-proxy.api.openshift.com/graph?g0.range_input=1h&g0.max_source_resolution=0s&g0.expr=cluster_installer%7Btype%3D%22other%22%7D&g0.tab=1

crawford:
/retitle metrics: include upi installs in cluster_installer

crawford:
/retitle Bug 1797806: metrics: include upi installs in cluster_installer

PR #322, opened 2020-02-13
Bug 1801300: lib/resourcemerge: remove ports which are no longer required

vrutkovs:
This ensures container ports which are no longer specified are
being removed


wking:
Basically the same implementation as the existing `EnsureServicePorts`, so I'm fine with the more limited test suite here, because `EnsureServicePorts`' more extensive suite shows the approach is solid.

/lgtm

vrutkovs:
4.3 cherrypick - https://github.com/openshift/cluster-version-operator/pull/323

PR #327, opened 2020-02-18
Bug 1782982: pkg/verify/verifyconfigmap: Fix Store to use {algo}-{hash} prefix

jottofar:
…h} prefix


pkg/verify/verifyconfigmap/store.go
```diff
@@ -39,8 +39,8 @@ func NewStore(client corev1client.ConfigMapsGetter, limiter *rate.Limiter) *Stor
 		limiter = rate.NewLimiter(rate.Every(30*time.Second), 1)
 	}
 	return &Store{
-		client: client,
-		ns:     "openshift-config-managed",
+		client:  client,
+		ns:      "openshift-config-managed",
```
jottofar: This was an unintentional side affect of my editor but it did line things up

PR #331, opened 2020-02-25
OWNERS: add OTA team members to approvers list

vrutkovs:


sdodson:
I'm fine with adding more OTA team members but I'd like to see the list approvers somewhere around 6 people who are fully invested in the future of the project, aka maintainers. From the total list of proposed approvers who are the 6 most invested?

LalatenduMohanty:
@sdodson We need to add @vrutkovs  and @jottofar  for sure.  For others lets wait for some individual contributions and then we can add them.

sdodson:
/lgtm
/approve
We should probably prune some folks in the near future if they prove not to provide timely reviews/approvals :-)

PR #334, opened 2020-03-02
cvo: report only unavailable operators with cluster_operator_up

mfojtik:
The `ClusterOperatorDown`[1] alert is described as:

> Cluster Operator XYZ has not been available for 10 mins

However, the metric we fire this alert on is based on "available" AND NOT "degraded", which means we also report "ClusterOperatorDown" for operators that are degraded (but still available).

Alternative for this PR could be fixing description of this alert, but I don't see degraded operator for longer than 10 minutes as something we need to fire critical alert on?

[1] https://github.com/openshift/cluster-version-operator/blob/master/install/0000_90_cluster-version-operator_02_servicemonitor.yaml#L47

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -259,9 +259,8 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 			break
 		}
 		g := m.clusterOperatorUp.WithLabelValues(op.Name, firstVersion)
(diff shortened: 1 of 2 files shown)


mfojtik:
@eparis @smarterclayton why is LGTM restricted only to collaborators??

smarterclayton:
/hold

This is a fundamental definition.  Needs more discussion.

smarterclayton:
This ties into the product SLO.  Operators being degraded means things are bad.  Only go degraded when things are bad.

What inputs triggered this?

wking:
Ended up happening, more or less, via #550.

PR #335, opened 2020-03-13
docs/user/status: Document RetrievedUpdates reasons

wking:
The messages included in the condition are intended to be fairly clear, but add some additional docs here where we can talk about the impact of the failure and suggest possible mitigations.

### DIFF
--- a/docs/user/status.md
+++ b/docs/user/status.md
@@ -1,0 +1,111 @@
+# Conditions
+
+[The ClusterVersion object](../dev/clusterversion.md) sets `conditions` describing the state of the cluster-version operator (CVO).
+This document describes those conditions and, where appropriate, suggests possible mitigations.
+
+## RetrievedUpdates
+
+When `RetrievedUpdates` is `True`, the CVO is succesfully retrieving updates, which is good.
+When `RetrievedUpdates` is `False`, `reason` will be set to explain why, as discussed in the following subsections.
+In all cases, the impact is that the cluster will not be able to retrieve recommended updates, so cluster admins will need to monitor for available updates on their own or risk falling behind on security or other bugfixes.
+
+### NoUpstream
+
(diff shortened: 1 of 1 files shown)


docs/user/status.md
```diff
@@ -0,0 +1,108 @@
...
+
+### InvalidURI
+
+The configured `upstream` URI is not valid.
+
+Fix by setting `spec.upstream` in ClusterVersion to point to a valid [Cincinnati][] URI, for example https://api.openshift.com/api/upgrades_info/v1/graph .
+
+### InvalidID
+
+The configured `clusterID` is not a valid URI.
```
vrutkovs:
```suggestion
The configured `clusterID` is not a valid UUID.
```

wking: Thanks.  Fixed in 54b01d4 -> c79a2e4.

vrutkovs:
LGTM
/cc @jottofar

docs/user/status.md
```diff
@@ -0,0 +1,108 @@
...
+
+The configured `clusterID` is not a valid UUID.
+
+Fix by setting `spec.clusterID` to a valid UUID.
+The UUID should be unique to a given cluster, because it is the default value used for reporting Telemetry and Insights.
+It may also be used by the CVO when making Cincinnati requests, so that Cincinnati can return update recommentations tailored to the specific cluster.
+
+### NoArchitecture
+
+The set of architectures has not been configured.
```
jottofar: Why doesn't this "reason" have an explanation of how to fix? This is the only one that does not so it should have something here.

wking: > Why doesn't this "reason" have an explanation of how to fix?

I forgot ;).  Added the usual "CVO bug, need an update" boilerplate with c79a2e4 -> 69c81b9.

PR #336, opened 2020-03-16
Bug 1802553: Adding a precondition for upgrade in progress

LalatenduMohanty:
CVO must not start another upgrade process if an upgrade is already
 in progress. We have a similar check in oc adm command. But an user
 can start the upgrade through the web console too.

### DIFF
--- a/pkg/payload/precondition/clusterversion/upgradeable.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable.go
@@ -61,6 +61,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 		}
 	}
 
+	// When a cluster is already in progress with an upgrade we should not start an upgrade
+	if c := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil {
+		if c.Status == configv1.ConditionTrue {
+			return &precondition.Error{
+				Nested:  err,
+				Reason:  "UpgradeInProgress",
+				Message: "Update target can not be changed before the current update completes",
+				Name:    pf.Name(),
+			}
+		}
+	}
+
 	// if we are upgradeable==true we can always upgrade
 	up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable)
 	if up == nil {
--- a/pkg/payload/task.go
+++ b/pkg/payload/task.go
@@ -192,6 +192,9 @@ func SummaryForReason(reason, name string) string {
 	case "ImageVerificationFailed":
 		return "the image may not be safe to use"
 
+	case "UpgradeInProgress":
+		return "an upgrade is in progress"
+
 	case "UpgradePreconditionCheckFailed":
 		return "it may not be safe to apply this update"
 
--- a/pkg/payload/precondition/precondition.go
+++ b/pkg/payload/precondition/precondition.go
@@ -51,7 +51,7 @@ type Precondition interface {
 // List is a list of precondition checks.
 type List []Precondition
 
-// RunAll runs all the reflight checks in order, returning a list of errors if any.
+// RunAll runs all the preflight checks in order, returning a list of errors if any.
 // All checks are run, regardless if any one precondition fails.
 func (pfList List) RunAll(ctx context.Context, releaseContext ReleaseContext, cv *configv1.ClusterVersion) []error {
 	var errs []error


pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -50,6 +50,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 		return nil
 	}
 
+	// if ClusterStatusConditionType = "Progressing" we should not upgrade
```
vrutkovs: This needs a more detailed explanation

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -50,6 +50,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+	if c := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil {
+		if c.Status == configv1.ConditionTrue {
+			return &precondition.Error{
+				Nested:  err,
+				Reason:  "UpgradePreconditionCheckFailed",
```
wking: Reason should be `UpgradeInProgress`, right?  We want it to be specific without folks having to look at the human-oriented `Message`.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -50,6 +50,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+				Message: "An upgrade already in progress",
```
wking: nit: Maybe "Cannot change the update target before the current update completes." or some such?

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -50,6 +50,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 		return nil
```
wking: nit: the new check should live before this block that's reading out the `OperatorUpgradeable` condition.

LalatenduMohanty:
There might some catch in the CVO code, so I am working on to understand the code flow better.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -44,6 +44,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 		}
 	}
 
+	// When a cluster is already in progress with an upgrade we we should not start an upgrade
+	if c := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorProgressing); c != nil {
+		if c.Status == configv1.ConditionTrue {
+			return &precondition.Error{
+				Nested:  err,
+				Reason:  "UpgradeInProgress",
+				Message: "update target can not be changed before the current update completes",
```
wking: nit: "update" -> "Update".  From [the docs][1] for the very similar ClusterOperator conditions:

> All messages should start with a capital letter (like a sentence) and be written for an end user / admin to debug the problem.

[1]: https://github.com/openshift/cluster-version-operator/blame/94429fec12c7e3e709649b822cbfcf9cd55d176c/docs/dev/clusteroperator.md#L155
[1]:

pkg/payload/task.go
```diff
@@ -192,6 +192,9 @@ func SummaryForReason(reason, name string) string {
 	case "ImageVerificationFailed":
 		return "the image may not be safe to use"
 
+	case "UpgradeInProgress":
+		return "an upgrade in progress"
```
wking: nit: "in" -> "is in".

LalatenduMohanty:
@wking Pr is ready for another look. Thanks.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -44,6 +44,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 		}
 	}
 
+	// When a cluster is already in progress with an upgrade we we should not start an upgrade
```
vrutkovs:
```suggestion
	// When a cluster is already in progress with an upgrade we should not start an upgrade
```

vrutkovs:
I don't think this is correct. Users should be able to have Partial upgrade from 4.4.0 -> 4.4.1 -> .. -> 4.4.n. What we don't support is partial upgrade from 4.3.5 -> 4.3.8 -> 4.4.0.

This PR should ensure that all Partial upgrade in the list have the same minor version

wking:
> Users should be able to have Partial upgrade from 4.4.0 -> 4.4.1 -> .. -> 4.4.n.

Patch bumps are safer than minor bumps, but I don't see a need to allow any of them without an override.  For example, we sometimes block z-stream edges, and the fact that 4.9.1 -> 4.9.2 and 4.9.2 -> 4.9.3 are recommended doesn't say anything about the stability of 4.9.1 -> 4.9.3, which is what some components would be doing if you did not require an override for patch retargets.

LalatenduMohanty:
`ClusterStatusConditionType` can have values between Available, Progressing, Degraded, Upgradeable . However `UpdateState` can be either Completed or Partial [2]

Though `PartialUpdate` indicates an update was never completely applied or is currently being applied. I am wondering  if `ClusterStatusConditionType` would gets stuck in `Progressing` when update state is always `PartialUpdate` .

[1] https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L141-L175
[2] https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L141-L143

smarterclayton:
/hold

This changes the behavior of the API and would be a compatibility thing.  Probably cannot safely do this without it being opt in.

wking:
@smarterclayton has concerns about this possibly being an API change, although I'm not clear on the details.  Probably want him to sign off on this before it lands.

LalatenduMohanty:
> This changes the behavior of the API and would be a compatibility thing. Probably cannot safely do this without it being opt in.

Do you mean this does not work with `  --allow-upgrade-with-warnings=false` in its current form or something else?

LalatenduMohanty:
/close

As this does not seem to the right way to fix this.

PR #343, opened 2020-04-01
Do not report errors to status loop on progressing

smarterclayton:
Test whether the error on the progressing path is causing us to reset
log output during upgrades and flag spurious errors.

Exploring why we see an "unknown error" and all pod progressing status
printed.

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -142,13 +142,18 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
 			return false, nil
 		}
 
-		if progressingCondition != nil && progressingCondition.Status == corev1.ConditionTrue {
+		if progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse && progressingCondition.Reason == "ProgressDeadlineExceeded" {
 			lastErr = &payload.UpdateError{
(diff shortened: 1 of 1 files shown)


lib/resourcebuilder/apps.go
```diff
@@ -142,16 +142,21 @@ func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.Deploy
...
+				Nested:  fmt.Errorf("deployment %s is not progressing; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedReplicas, d.Status.Replicas, d.Status.AvailableReplicas, d.Status.Replicas),
 				Reason:  "WorkloadNotAvailable",
-				Message: fmt.Sprintf("deployment %s is progressing %s: %s", iden, progressingCondition.Reason, progressingCondition.Message),
+				Message: fmt.Sprintf("deployment %s is not progressing %s: %s", iden, progressingCondition.Reason, progressingCondition.Message),
 				Name:    iden,
 			}
 			return false, nil
 		}
 
+		if progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse {
```
smarterclayton, line 155: Wait.... this should be conditionTrue

smarterclayton:
Ok, this is ready for review (and more arguments for/against, if there are any)

/assign @abhinavdahiya

PR #353, opened 2020-04-20
Bug 1826115: pkg/cvo/status: Always set status.desired to match current CVO

wking:
As described in the bug, bumping `status.desired` when a new `spec.desiredUpdate` is set leads to trouble like:

```console
$ oc adm upgrade --to 4.3.13
Updating to 4.3.13
$ oc get -o json clusterversion version | jq -r '.status.conditions[] | .lastTransitionTime + " " + .type + " " + .status + " " + .message' | sort
2020-04-20T20:57:12Z RetrievedUpdates True
2020-04-20T21:23:40Z Available True Done applying 4.3.10
2020-04-20T22:01:55Z Upgradeable False Cluster operator kube-apiserver cannot be upgraded: DefaultSecurityContextConstraintsUpgradeable: Default SecurityContextConstraints object(s) have mutated [privileged]
2020-04-20T22:16:40Z Failing True Precondition "ClusterVersionUpgradeable" failed because of "DefaultSecurityContextConstraints_Mutated": Cluster operator kube-apiserver cannot be upgraded: DefaultSecurityContextConstraintsUpgradeable: Default SecurityContextConstraints object(s) have mutated [privileged]
2020-04-20T22:16:40Z Progressing True Unable to apply 4.3.13: it may not be safe to apply this update
$ oc get -o json clusterversion version | jq -r '.status.desired.version'
4.3.13
$ oc adm upgrade --to=4.3.13 --force
info: Cluster is already at version 4.3.13
$ oc adm upgrade --clear
Cleared the update field, still at 4.3.13
$ oc get -o json clusterversion version | jq -r '.status.desired.version'
4.3.10
```

Where the "already at..." and "still at ..." messages are relying on [the `status.desired` semantics][2]:

> desired is the version that the cluster is reconciling towards.

When the user sets `spec.desiredUpdate`, they are making their intention clear (and bumping history at this point is appropriate, because we need *somewhere* to store the `verified` history entry).  But to match the "reconciling towards" semantics, this commit shifts the actual `status.desired` bump so it happens right after the new CVO comes up (with the new `currentVersion`).

[2]: https://github.com/openshift/api/blob/0f159fee64dbf711d40dac3fa2ec8b563a2aaca8/config/v1/types_cluster_version.go#L82-L87

### DIFF
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -126,8 +126,6 @@ func mergeOperatorHistory(config *configv1.ClusterVersion, desired configv1.Upda
 
 	// TODO: prune Z versions over transitions to Y versions, keep initial installed version
 	pruneStatusHistory(config, 50)
-
-	config.Status.Desired = desired
 }
 
 func pruneStatusHistory(config *configv1.ClusterVersion, maxHistory int) {
@@ -181,6 +179,7 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
 	version := versionString(status.Actual)
 
 	mergeOperatorHistory(config, status.Actual, status.Verified, now, status.Completed > 0)
+	config.Status.Desired = optr.currentVersion()
 
 	// update validation errors
 	var reason string
@@ -400,7 +399,9 @@ func (optr *Operator) syncFailingStatus(original *configv1.ClusterVersion, ierr 
 		LastTransitionTime: now,
 	})
 
-	mergeOperatorHistory(config, optr.currentVersion(), false, now, false)
+	current := optr.currentVersion()
+	mergeOperatorHistory(config, current, false, now, false)
+	config.Status.Desired = current
 
 	updated, err := applyClusterVersionStatus(optr.client.ConfigV1(), config, original)
 	optr.rememberLastUpdate(updated)
@@ -410,7 +411,7 @@ func (optr *Operator) syncFailingStatus(original *configv1.ClusterVersion, ierr 
 	return ierr
 }
 
-// applyClusterVersionStatus attempts to overwrite the status subresource of required. If
+// applyClusterVersionStatus attempts to overwrite the status subresource, if required. If
 // original is provided it is compared to required and no update will be made if the
 // object does not change. The method will retry a conflict by retrieving the latest live
 // version and updating the metadata of required. required is modified if the object on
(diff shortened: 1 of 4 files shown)


wking:
/hold

I'm going to have to rework this a bit more, because blindly dropping `Actual` into the history (as the unit tests seem to expect) seems crazy.

wking:
/assign @smarterclayton 

I'm still working on updating the unit tests for this, but assigning to Clayton for review, because in 961873d66a (#82) he added a bunch of:

```go
// Prefers the payload version over the operator's version (although in general they will remain in sync
```

comments, and I'm reversing that here.

wking:
Ok, 3fd35dc -> e9c53b1 should get unit tests passing in this new world.  But it's a pretty large change to the tests which reverses the earlier choice of explicitly preferring the sync worker's status over the operator version.  So still would be good to have @smarterclayton sign off on the new direction.

wking:
Hmm, still need to update [the integration tests][1].

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/353/pull-ci-openshift-cluster-version-operator-master-integration/910

pkg/cvo/status.go
```diff
@@ -126,8 +126,6 @@ func mergeOperatorHistory(config *configv1.ClusterVersion, desired configv1.Upda
 
 	// TODO: prune Z versions over transitions to Y versions, keep initial installed version
 	pruneStatusHistory(config, 50)
-
-	config.Status.Desired = desired
```
LalatenduMohanty, line 130: I still not clear why this does not work.

wking: `syncStatus` is feeding `mergeOperatorHistory` `status.Actual`, which diverges from `optr.currentVersion()` when an update has been requested and we're working through preconditions.  With this commit, we're always setting `Desired` to `currentVersion()`.

wking: `syncStatus` is feeding `mergeOperatorHistory` `status.Actual`, which diverges from `optr.currentVersion()` when an update has been requested and we're working through preconditions.  With this commit, we're always setting `Desired` to `currentVersion()`.

LalatenduMohanty:
/lgtm cancel 
As I have not gone through the test changes.

PR #354, opened 2020-04-21
Bug 1826349: hack/build-go: Use OS_GIT_VERSION if available

vrutkovs:
This would update the build script, so that OCP could be built using
a custom semver string from OS_GIT_VERSION instead of `git describe` output


sosiouxme:
/lgtm
still :)

PR #356, opened 2020-04-22
Bug 1823306: pkg/cvo/upgradeable: Mention "between minor versions"

wking:
Catching the message strings up with the softening from bcd58d892e (#291).


wking:
Fixed at least some of [the unit failures][1] and rebased onto master with a9bb055 -> 9e76c64.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/356/pull-ci-openshift-cluster-version-operator-master-unit/913

PR #357, opened 2020-04-22
Extend CVO alerts to cover update retrieval

jottofar:


install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,6 +40,14 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: ClusterVersionOperatorNotRetrieving
```
wking, line 43: nit: The CVO retrieves a few things (including Cincy graph data, release signatures, and Kube API objects).  Can we call this `ClusterVersionOperatorUpdatesStale` or `ClusterVersionOperatorAvailableUpdatesStale` or some such?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,6 +40,14 @@ spec:
...
+      annotations:
+        message: Cluster version operator has not retrieved updates for 10 mins. Failure reason {{ "{{ $labels.reason }}" }}. {{ "{{ $labels.message }}" }}.
```
wking, line 45: nit: "mins" -> "minutes".  Less mental overhead from guessing at abbreviations.

pkg/cvo/metrics.go
```diff
@@ -297,6 +307,18 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 		g.Set(1.0)
 		ch <- g
 	}
+
+	if availableUpdates := m.optr.getAvailableUpdates(); availableUpdates != nil {
```
wking, line 311: when is this `nil`?  Is that just "we were just born and haven't completed a single graph-pull cycle yet"?

pkg/cvo/metrics.go
```diff
@@ -297,6 +307,18 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
...
+		klog.Infof("Status: %s", availableUpdates.Condition.Status)
+		if availableUpdates.Condition.Status == configv1.ConditionFalse {
```
wking, line 313: What about the "we're fine" case? Or the "unknown" case (which we may not set)?  How about something like:

```go
g := m.clusterVersionOperatorRetrieving
switch availableUpdates.Condition.Status {
case configv1.ConditionTrue:
  g.Set(1)
case configv1.ConditionFalse:
  g = g.WithLabelValues(availableUpdates.Condition.Reason, availableUpdates.Condition.Message)
  g.Set(0)
default:
  klog.Warnf("unknown update status: ...", ...)
}
ch <- g
```

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: From doc https://github.com/wking/cluster-version-operator/blob/master/docs/user/status.md:

"When RetrievedUpdates is False, ... In all cases, the impact is that the cluster will not be able to retrieve recommended updates"

So how about ClusterVersionOperatorCannotRetrieveUpdates?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: I will change them all in CVO. I thought that perhaps there was an OpenShift wide standard for Alert text and that mins was part of that but apparently not. Foolish me...

pkg/cvo/metrics.go
jottofar: Not sure when but here's the why I check: https://github.com/openshift/cluster-version-operator/blob/2c4931dc283c551938be1a00fee290de0b79d99c/pkg/cvo/availableupdates.go#L107

Do you think we should issue a warning when it's nil?  It will spit out every 30 seconds until it gets set. Could even be an Alert if we wanted.

pkg/cvo/metrics.go
jottofar: Will do. I actually wasn't done but that's okay because wasn't sure about the Set(1) since it doesn't appear to always be done in other places. However, it makes sense to set it. For example, it allows someone to simply add another dependency into the metrics yaml file based on it being true without updating the code.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
wking: sounds great

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
wking: hah, I dunno, maybe there is and I'm not aware of it.  If there's an existing pattern you're matching (sounds like there is), feel free to punt, and I'll take this on in follow-up work

pkg/cvo/metrics.go
wking: I think we should figure out why it can be `nil` and update the docstring there, and *then* figure out what to do about that case for alerts.

pkg/cvo/metrics.go
wking: if there is some established precedent you are following where we don't push a `Set(1)` version, then I'm fine following that precedent.  I just thought our other metrics were always set, not just set when there were problems.

pkg/cvo/metrics.go
jottofar: It appears it can only be nil if it does not get initialized. Here's the method that sets it, including initially: https://github.com/openshift/cluster-version-operator/blob/2c4931dc283c551938be1a00fee290de0b79d99c/pkg/cvo/cvo.go#L499

The first error where CVO can't find itself I would hope happens very rarely if ever and if it does perhaps the cluster is toast anyway. 

I would think the more likely scenario is ValidateClusterVersion https://github.com/openshift/cluster-version-operator/blob/2c4931dc283c551938be1a00fee290de0b79d99c/lib/validation/validation.go#L16 fails, but, don't know the circumstances or liklyhood behind that either.

Thoughts?

/cc @wking

wking: Both of those seem rare, and unrelated to getting updates, so probably ignore them if they happen.  Not sure if skipping the `ch <- g` means we continue to serve whatever the last value was or if it means we don't serve the metric at all.  Probably means we keep serving the last value, which is what we want for the `nil` case.

jottofar: I output a warning if availableUpdtates is null.
Set 'ch <- g' to status condition true or false. Unknown "should" never happen but if it does output a warning.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,27 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: ClusterVersionOperatorCannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates for 10 minutes. Failure reason {{ "{{ $labels.reason }}" }}. {{ "{{ $labels.message }}" }}.
+      expr: |
+        cluster_version_operator_retrieving{job="cluster-version-operator"} == 0
+      for: 10m
```
wking, line 48: 10 minutes seems overly strict given our loose Cincinnati service-level agreements and objectives.  Can we make this at least an hour?

pkg/cvo/metrics.go
```diff
@@ -83,6 +85,10 @@ version for 'cluster', or empty for 'initial'.
 			Name: "cluster_installer",
 			Help: "Reports info about the installation process and, if applicable, the install tool.",
 		}, []string{"type", "version", "invoker"}),
+		clusterVersionOperatorRetrieving: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_retrieving",
```
wking, line 89: Can this be `cluster_version_operator_update_retrieval` or some such?

pkg/cvo/metrics.go
```diff
@@ -34,6 +35,7 @@ type operatorMetrics struct {
 	clusterOperatorConditions           *prometheus.GaugeVec
 	clusterOperatorConditionTransitions *prometheus.GaugeVec
 	clusterInstaller                    *prometheus.GaugeVec
+	clusterVersionOperatorRetrieving    *prometheus.GaugeVec
```
wking, line 38: nit: this internal property does not need the `clusterVersionOperator` namespacing that the metric name needs.  Can we just use `cannotRetrieveUpdates` or some such?

pkg/cvo/metrics.go
wking: Or, I guess `cluster_version_operator_can_retrieve_updates` to match the 1 vs. 0 sense we're using for values.

jottofar: I chose it based on status.md, "...CVO is succesfully retrieving updates", but doesn't matter to me. It does appear that in other cases this name ends up being the camel-cased name used for the GaugeVec so this comment and https://github.com/openshift/cluster-version-operator/pull/357#discussion_r419700881 are related. Also see former comment for my reason for not wanting to using a negative named condition.

pkg/cvo/metrics.go
jottofar: I didn't use clusterVersionOperator due to namespace considerations but rather thought it best described what was being checked. Per status.md "...CVO is succesfully retrieving updates". I could make it CVORetrieving but think spelling it out better matches the pattern already being used. 
Do not want to use the negative named condition because it does not flow with the logic check of Status true.

wking: I think these are CVO/release metrics by default, unless they say otherwise.  So `version` would be CVO/release version.  `availableUpdates` is CVO/release updates.  This one would be `canRetrieveAvailableUpdates`.  Or, hmm, maybe it should be `availableUpdatesStaleSeconds` and record seconds since the last sync?

docs/user/status.md
```diff
@@ -8,6 +8,7 @@ This document describes those conditions and, where appropriate, suggests possib
 When `RetrievedUpdates` is `True`, the CVO is succesfully retrieving updates, which is good.
 When `RetrievedUpdates` is `False`, `reason` will be set to explain why, as discussed in the following subsections.
 In all cases, the impact is that the cluster will not be able to retrieve recommended updates, so cluster admins will need to monitor for available updates on their own or risk falling behind on security or other bugfixes.
+When CVO is unable to retrieve recommended updates the ClusterVersionOperatorCannotRetrieveUpdates alert will fire containing the reason.
```
LalatenduMohanty, line 11: IMO we should add that for `NoChannel` we will not alert or whatever the final conclusion on this.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,26 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: CannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates since {{ "{{ $labels.lastTime }}" }}. Failure reason {{ "{{ $labels.reason }}" }}. {{ "{{ $labels.message }}" }}.
+      expr: |
+        cannot_retrieve_updates{job="cluster-version-operator"} == 1
```
vrutkovs, line 47: We don't really need `cannot_retrieve_updates` metric for that as it can be derived from `cluster_version_operator_successful_update_retrieval`:

```
time() - cluster_version_operator_successful_update_retrieval > 3600
```

jottofar:
//retest

vrutkovs:
LGTM, checked that the new metrics appears in PromeCIeus and the alert is not triggered on CI clusters

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,26 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: CannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates since {{ "{{ $labels.lastTime }}" }}. Failure reason {{ "{{ $labels.reason }}" }}. {{ "{{ $labels.message }}" }}.
+      expr: |
+        time()-cluster_version_operator_successful_update_retrieval >= 3600 unless cluster_version_operator_successful_update_retrieval{reason="NoChannel"}
```
wking, line 47: My PromQL is not great, but I'd expect we could do this with:

```
time()-cluster_version_operator_successful_update_retrieval{reason!="NoChannel"} >= 3600
```

Does that not work for some reason?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,26 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: CannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates since {{ "{{ $labels.lastTime }}" }}. Failure reason {{ "{{ $labels.reason }}" }}. {{ "{{ $labels.message }}" }}.
```
wking, line 45: Do we want to... link [here][1]?  To docs for the specific reason?  Pin to `release-4.6`?  Float with master?  Not link to docs at all?  Suggest:

> Clear ClusterVersion.spec.channel if you have no upstream update recommendation service available.

I don't know if there's a cap on alert message text length or not, but I'd like to provide as much actionable context here as possible, so users are less likely to say "huh, another critical alert I know nothing about" and walk away without fixing it.

[1]: https://github.com/openshift/cluster-version-operator/blob/master/docs/user/status.md#retrievedupdates

pkg/cvo/metrics.go
```diff
@@ -9,12 +9,15 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	"github.com/openshift/cluster-version-operator/pkg/internal"
 )
 
+const noChannel string = "NoChannel"
```
wking, line 19: If we're going to define a constant for this string, probably use it in the places that are currently hard-coding the string:

```console
$ git --no-pager grep NoChannel origin/pr/357 pkg 
origin/pr/357:pkg/cvo/availableupdates.go:                      Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoChannel",
origin/pr/357:pkg/cvo/cvo_test.go:                                      Reason:  "NoChannel",
origin/pr/357:pkg/cvo/metrics.go:const noChannel string = "NoChannel"
origin/pr/357:pkg/cvo/metrics.go:                       // Reset time if reason has tranisitioned from NoChannel to avoid alerting too quickly
```

pkg/cvo/metrics.go
```diff
@@ -23,17 +26,27 @@ func (optr *Operator) registerMetrics(coInformer cache.SharedInformer) error {
 	return prometheus.Register(m)
 }
 
+type retrievingCondition struct {
+	condition                     configv1.ClusterOperatorStatusCondition
+	lastKnownRetrievingTime       time.Time
+	lastKnownRetrievingTimeString string
```
wking, line 32: I've filed #368 to store these in the existing `availableUpdates` struct.

pkg/cvo/metrics.go
```diff
@@ -23,17 +26,27 @@ func (optr *Operator) registerMetrics(coInformer cache.SharedInformer) error {
...
 
+	lastCheckedRetrievingCondition retrievingCondition
+
 	version                             *prometheus.GaugeVec
 	availableUpdates                    *prometheus.GaugeVec
 	clusterOperatorUp                   *prometheus.GaugeVec
 	clusterOperatorConditions           *prometheus.GaugeVec
 	clusterOperatorConditionTransitions *prometheus.GaugeVec
 	clusterInstaller                    *prometheus.GaugeVec
+	//cannotRetrieveUpdates                           *prometheus.GaugeVec
```
wking, line 48: probably drop this commented-out code.

pkg/cvo/metrics.go
```diff
@@ -42,6 +55,16 @@ func newOperatorMetrics(optr *Operator) *operatorMetrics {
 
 		conditionTransitions: make(map[conditionKey]int),
 
+		lastCheckedRetrievingCondition: retrievingCondition{
```
wking, line 58: No need for this here once we get #368.

pkg/cvo/metrics.go
```diff
@@ -83,6 +106,10 @@ version for 'cluster', or empty for 'initial'.
 			Name: "cluster_installer",
 			Help: "Reports info about the installation process and, if applicable, the install tool.",
 		}, []string{"type", "version", "invoker"}),
+		clusterVersionOperatorSuccessfulUpdateRetrieval: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_successful_update_retrieval",
+			Help: "Reports when updates were last succesfully retrieved and, if currently not able to retrieve updtaes, reports reason.",
```
wking, line 111: nit: "updtaes" -> "updates"

pkg/cvo/metrics.go
```diff
@@ -83,6 +106,10 @@ version for 'cluster', or empty for 'initial'.
...
+		}, []string{"lastTime", "reason", "message"}),
```
wking, line 112: `message` would be super-helpful, but... I'm concerned about label cardinality (apparently Prometheus doesn't like it if we have labels that have many different values, and `message` is basically unbounded).  We should probably stick to `lastTime` and `reason`, and have the alert message say "For more information..." and then... somehow point people at the condition.  By linking to a console view?  Or by saying "... see ClusterVersion's status.conditions"?  Or something...

pkg/cvo/metrics.go
```diff
@@ -133,6 +160,43 @@ func (m *operatorMetrics) Describe(ch chan<- *prometheus.Desc) {
 	ch <- m.clusterOperatorConditions.WithLabelValues("", "", "").Desc()
 	ch <- m.clusterOperatorConditionTransitions.WithLabelValues("", "").Desc()
 	ch <- m.clusterInstaller.WithLabelValues("", "", "").Desc()
+	/*
+		ch <- m.cannotRetrieveUpdates.WithLabelValues("", "", "").Desc()
+		ch <- m.clusterVersionOperatorSuccessfulUpdateRetrieval.WithLabelValues("").Desc()
+	*/
```
wking, line 166: No need to keep commented-out stuff.

pkg/cvo/metrics.go
```diff
@@ -133,6 +160,43 @@ func (m *operatorMetrics) Describe(ch chan<- *prometheus.Desc) {
...
+	ch <- m.clusterVersionOperatorSuccessfulUpdateRetrieval.WithLabelValues("", "", "").Desc()
+}
+
+func (m *operatorMetrics) setRetrievingCondition(current configv1.ClusterOperatorStatusCondition, atTime time.Time) {
+
```
wking, line 171: nit: empty line.

pkg/cvo/metrics.go
```diff
@@ -133,6 +160,43 @@ func (m *operatorMetrics) Describe(ch chan<- *prometheus.Desc) {
...
+	savedLastKnownRetrievingTime := m.lastCheckedRetrievingCondition.lastKnownRetrievingTime
+
+	if !m.lastCheckedRetrievingCondition.lastKnownRetrievingTime.IsZero() {
```
wking, line 174: No need for this block with #368.

pkg/cvo/metrics.go
jottofar: - How about if we just tell them to look in the cvo log and search for 'Reason'? It at least is an easy approach.
- Assume by "console view" you mean pop up a window of some kind with the message? Do you know if this approach used anywhere else OpenShift?
- I'll post a question to forum-monitoring for any recommendations/ideas as well.

wking: > Assume by "console view" you mean pop up a window of some kind with the message? Do you know if this approach used anywhere else OpenShift?

Yeah, I mean Administration > Cluster Settings > Overview, like [here][1].

[1]: https://docs.openshift.com/container-platform/4.2/updating/updating-cluster-between-minor.html#update-upgrading-web_updating-cluster-between-minor

jottofar: The base issue is that an unbounded label set, such as "labels.message"s strings, can cause an unbounded set of independent metrics rather than one continuous metric. I saw this behavior with the lastTime value when I wasn't careful about only setting it when it truly needed to be set. But do we really have this issue with 'message'? Our metric is only unique and alive for the period since LastSyncOrConfigChange and true, if the LastSyncOrConfigChange and Reason remain the same but the Message changes, which I think would be rare, we would get a new metric which would be incorrect. But as I said, that would be very rare.
Another approach is to go back to the way I had it with a separate "metric" just for the Alert.

pkg/cvo/metrics.go
```diff
@@ -83,6 +103,10 @@ version for 'cluster', or empty for 'initial'.
 			Name: "cluster_installer",
 			Help: "Reports info about the installation process and, if applicable, the install tool.",
 		}, []string{"type", "version", "invoker"}),
+		clusterVersionOperatorSuccessfulUpdateRetrieval: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_successful_update_retrieval",
```
simonpasquier, line 107: better to use `cluster_version_operator_update_retrieval_timestamp_seconds` to follow metric naming [conventions](https://prometheus.io/docs/practices/naming/#metric-names).

pkg/cvo/metrics.go
```diff
@@ -83,6 +103,10 @@ version for 'cluster', or empty for 'initial'.
...
+			Help: "Reports when updates were last succesfully retrieved and, if currently not able to retrieve updates, reports reason.",
```
simonpasquier, line 108: > Reports when updates were last succesfully retrieved and, if currently not able to retrieve updates, reports reason.

If you really want this, I'd recommend to have 2 distinct metrics: one reporting the last update timestamp and another reporting the reason.

jottofar: @simonpasquier 

If I have two metrics, how do I configure my alert to pull from the two metrics? With the one metric I use $value and $labels.reason:

- alert: CannotRetrieveUpdates
      annotations:
        message: Cluster version operator has not retrieved updates since {{ "{{ $value | humanizeDuration }}" }}. Failure reason {{ "{{ $labels.reason }}" }}.

pkg/cvo/metrics.go
wking: Following up on the console URI, the installer spits out:

https://console-openshift-console.apps.ci-ln-cy2dir2-d5d6b.origin-ci-int-aws.dev.rhcloud.com/

We want to link people to:

https://console-openshift-console.apps.ci-ln-cy2dir2-d5d6b.origin-ci-int-aws.dev.rhcloud.com/settings/cluster/

(so console URI + `/settings/cluster/`), where they will see a UI like:

Screenshot from 2020-05-20 09-40-02

jottofar: ACK. Is there a need for the get console URI method outside of the cvo package, e.g. should I create in lib perhaps?

pkg/cvo/metrics.go
wking: Obsolete as of 06c550ff9fed.

pkg/cvo/metrics.go
wking: Obsolete as of 06c550f.

pkg/cvo/metrics.go
wking: Obsolete as of 06c550f.

pkg/cvo/metrics.go
wking: Done with 06c550f.

pkg/cvo/metrics.go
wking: Done as of 06c550f.

pkg/cvo/metrics.go
wking: Obsolete as of 06c550f.

pkg/cvo/metrics.go
wking: Done as of 06c550f.

pkg/cvo/metrics.go
wking: Obsolete as of 06c550f.

pkg/cvo/metrics.go
wking: Obsolete as of 06c550f.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
wking: Done with 06c550ff9.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
wking: Done as of 06c550ff9f.

pkg/cvo/metrics.go
wking: 06c550ff9f's `cluster_version_operator_update_retrieval_timestamp_seconds` works for me.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,26 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: CannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates since {{ "{{ $value | humanizeDuration }}" }}. Failure reason {{ "{{ $labels.reason }}" }}. Console {{ "{{ $labels.console_url.url }}" }}.
+      expr: |
+        (time()-cluster_version_operator_update_retrieval_timestamp_seconds{reason!="NoChannel"} >= 300) and on(foobar) (console_url > 0)
```
wking, line 47: Is this 5m delay for testing?  Are we bumping it to [an hour][1] or more before landing this PR?

[1]: https://github.com/openshift/cluster-version-operator/pull/357#discussion_r419699248

wking: Also, I think we should alert regardless of `console_url > 0`.  Not sure what the PromQL is for "group in the `console_url` if you can find it, but leave the label unset if you can't".  And once we have that, some sort of conditional in `message` to skip the console link if we couldn't find the URI?

pkg/cvo/metrics.go
wking: Done as of 06c550ff9fed.

pkg/cvo/metrics.go
wking: Using a `reason` label and a timestamp value, as we have with 06c550ff9fed, makes sense to me.  If you still think it should be split into two metrics, can you explain your motivation?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -40,19 +40,26 @@ spec:
       for: 10m
       labels:
         severity: critical
+    - alert: CannotRetrieveUpdates
+      annotations:
+        message: Cluster version operator has not retrieved updates since {{ "{{ $value | humanizeDuration }}" }}. Failure reason {{ "{{ $labels.reason }}" }}. Console {{ "{{ $labels.console_url.url }}" }}.
```
wking, line 45: This console URI still needs [a `/settings/cluster/` suffix][1] to get you from the default console page to the ClusterVersion page.

[1]: https://github.com/openshift/cluster-version-operator/pull/357#discussion_r428156247

pkg/cvo/metrics.go
simonpasquier: > can you explain your motivation?

I feel that conflating 2 meanings into metric (when was the last successful updates + what is the last failure reason) is a source of confusion. Also if the update fails for reason A then for reason B without updating the value of `cluster_version_operator_successful_update_retrieval` then the alert will reset and go back to `pending` because the `reason` label would have changed.

You can inject the reason in the annotation using the `"cluster_version_operator_reason_failure > 0" | query | first | label "reason"` construct (see https://prometheus.io/docs/prometheus/latest/configuration/template_reference/#queries).

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: Yeah, all for testing. And just hacking around to try and get the console from that other metric.  Will re-do based on @simonpasquier suggestions.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: Ack. This was just trying to figure out how to pull the url from the metric.

pkg/cvo/metrics.go
jottofar: Added second metric as recommended.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: Changed delay back to an hour.
Inserted condition into annotation so only non-zero length console url message appears.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
jottofar: suffix added

jottofar:
This won't build until https://github.com/openshift/cluster-version-operator/pull/368 merges.

pkg/cvo/cvo_test.go
```diff
@@ -2375,7 +2375,7 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 				Condition: configv1.ClusterOperatorStatusCondition{
 					Type:    configv1.RetrievedUpdates,
 					Status:  configv1.ConditionFalse,
-					Reason:  "NoChannel",
+					Reason:  NoChannel,
```
wking, line 2378: `NoChannel` -> `noChannel`?

jottofar: Forgot about that one. Fixed now

pkg/cvo/metrics.go
```diff
@@ -83,6 +86,14 @@ version for 'cluster', or empty for 'initial'.
...
 			Help: "Reports info about the installation process and, if applicable, the install tool.",
 		}, []string{"type", "version", "invoker"}),
+		clusterVersionOperatorUpdateRetrievalTimestampSeconds: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_update_retrieval_timestamp_seconds",
+			Help: "Reports when updates were last succesfully retrieved.",
+		}, []string{"name"}),
+		clusterVersionOperatorReasonFailure: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_reason_failure",
+			Help: "Reports the reason when updates cannot be retrieved.",
+		}, []string{"reason"}),
```
wking, line 96: Surfacing [this previous thread][1] discussing the use of two metrics.  @simonpasquier in that comment:

> Also if the update fails for reason A then for reason B without updating the value of `cluster_version_operator_successful_update_retrieval` then the alert will reset and go back to pending because the reason label would have changed.

I'm not clear on why that would be the case, since with `cluster_version_operator_update_retrieval_timestamp_seconds` still over the threshold I'd expect the alert to continue to fire (but to switch to reporting the new reason).  But whatever, it seems to work the way we have it, and I can always float consolidation to a single metric in follow-up work if it bothers me and I can can get it to work that way ;).  This comment just gets a non-outdated discussion link where I can find it ;).

[1]: https://github.com/openshift/cluster-version-operator/pull/357#discussion_r429967805

pkg/cvo/metrics.go
```diff
@@ -297,6 +310,23 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 		g.Set(1.0)
 		ch <- g
 	}
+
+	// check ability to retrieve recommended updates
+	if availableUpdates := m.optr.getAvailableUpdates(); availableUpdates != nil {
+		g2 := m.clusterVersionOperatorUpdateRetrievalTimestampSeconds.WithLabelValues("")
```
wking, line 316: can we shift this `g2` bit down to just before we set it from `LastSyncOrConfigChange`?  No need to split the `g2` handling in half and drop the failure-reason stuff in the middle.

pkg/cvo/metrics.go
```diff
@@ -83,6 +86,14 @@ version for 'cluster', or empty for 'initial'.
 			Name: "cluster_installer",
 			Help: "Reports info about the installation process and, if applicable, the install tool.",
 		}, []string{"type", "version", "invoker"}),
+		clusterVersionOperatorUpdateRetrievalTimestampSeconds: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_update_retrieval_timestamp_seconds",
+			Help: "Reports when updates were last succesfully retrieved.",
+		}, []string{"name"}),
+		clusterVersionOperatorReasonFailure: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cluster_version_operator_reason_failure",
```
wking, line 94: Both `clusterVersionOperatorReasonFailure` and `cluster_version_operator_reason_failure` are too generic for failures to fetch *update* retrievals.  How about `cluster_version_operator_update_retrieval_failure` or some such?  And... do we want to record a non-empty reason on success too (like `AsExpected`?).  Maybe this should be `cluster_version_operator_update_retrieval_conditions` (like the existing `cluster_operator_conditions`)?

pkg/cvo/metrics.go
simonpasquier: because the `reason` label attached to the alert would change, it would resolve the current alert and create a new one (just like timeseries, alerts are identified by their labels set).

wking: Ahhh, ok, that makes sense.  So the question is, do we want folks to have to renew a silence or whatever if the reason changes?  I guess I could go either way on that.  If we say "no, they should be able to silence once regardless of changing `reason`", then we should be able to get `reason` from the existing `cluster_operator_conditions` instead of defining this new `cluster_version_operator_reason_failure` metric.

jottofar: @wking @simonpasquier 

Use of cluster_operator_conditions makes for the following rather ugly expr and alert message constructs. Re-use of the metric cluster_operator_conditions is awkward. For example, after setting the channel and getting the VersionNotFound reason there are 10 cluster_operator_conditions metrics with name="version".

        message: Cluster version operator has not retrieved updates in {{ "{{ $value | humanizeDuration }}" }}. Failure reason {{ "{{ with $cluster_operator_conditions := \"cluster_operator_conditions\" | query}}{{range $value := .}}{{if and (eq (label \"name\" $value) \"version\") (eq (label \"condition\" $value) \"RetrievedUpdates\") (eq (label \"endpoint\" $value) \"metrics\") (eq (value $value) 0.0)}}{{label \"reason\" $value}} {{end}}{{end}}{{end}}" }}. {{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} For more information refer to {{ label \"url\" (first $console_url ) }}/settings/cluster/.{{ end }}{{ end }}" }}
      expr: |
        (time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 120 and ignoring(condition, name, reason) cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"}

simonpasquier: You coud use `cluster_operator_conditions{name="version",condition="RetrievedUpdates"} == 0` as the initial query instead of `cluster_operator_conditions` and remove the inner `if` tests:

```
Failure reason {{ "{{ with $cluster_operator_conditions := \"cluster_operator_conditions{name=\\\"version\\\",condition=\\\"RetrievedUpdates\\\"} == 0\" | query}}{{label \"reason\" .}}{{end}}" }}. {{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} For more information refer to {{ label \"url\" (first $console_url ) }}/settings/cluster/.{{ end }}{{ end }}" }}
```

Not sure about the escaping of the query though.

wking: > ... there are 10 cluster_operator_conditions metrics with name="version".

And `condition="RetrievedUpdates"`?  We should only set one of those at a time.  But maybe the PromQL service is giving us the last few?  Have you tried:

```
(time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 120 and ignoring(condition, name, reason) max(cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"})
```

Or some such?

jottofar: I cannot get any form of `cluster_operator_conditions{name="version",condition="RetrievedUpdates"} == 0 ` to work so I'm going with my original using the `if`.

wking:
🎉 

/lgtm

PR #358, opened 2020-04-23
Bug 1809195: Send CVO metrics over https

jottofar:
CVO metrics are currently being sent insecurely over http. Change
metrics yaml file to use https and TLS encryption.


install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -9,9 +9,13 @@ metadata:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
wking, line 12: Are there API docs for ServiceMonitor?  [OpenShift docs][1] link [Prometheus docs][2], but I don't see anyone talking about `bearerTokenFile`.  Ah, [here we go][3].  But I don't see anything there about why we'd expect this path to work, who's putting what token in that path, or who's consuming it.  Would be good to get metrics input on this.

[1]: https://docs.openshift.com/container-platform/4.3/operators/operator_sdk/osdk-monitoring-prometheus.html#osdk-monitoring-prometheus-servicemonitor_osdk-monitoring-prometheus
[2]: https://github.com/coreos/prometheus-operator/blob/7a25bf6b6bb2347dacb235659b73bc210117acc7/Documentation/design.md#servicemonitor
[3]: https://github.com/coreos/prometheus-operator/blob/0c6d184c19c890af10d715310d1770424d4e370b/Documentation/api.md#apiserverconfig

install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -1,16 +1,19 @@
...
   type: ClusterIP
   selector:
     k8s-app: cluster-version-operator
   ports:
-  - name: metrics
-    port: 9099 # chosen to be in the internal open range
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 8443
```
wking, line 19: Isn't HTTPS sufficient?  I don't see why we'd need to expose TCP too.

install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -1,16 +1,19 @@
...
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   type: ClusterIP
   selector:
     k8s-app: cluster-version-operator
   ports:
-  - name: metrics
-    port: 9099 # chosen to be in the internal open range
+  - name: https
+    port: 443
```
wking, line 17: To back this, doesn't the CVO need to be serving HTTPS on 443?  How does it get the X.509 keypair and cert to do that?  Also, you'll probably need to adjust something in [here][1] to set [`--listen`][2] or change [the default][3].

[1]: https://github.com/openshift/cluster-version-operator/blob/fa452c2d270f1f989f3868ef97ae8cf825713583/install/0000_00_cluster-version-operator_03_deployment.yaml#L25-L29
[2]: https://github.com/openshift/cluster-version-operator/blob/fa452c2d270f1f989f3868ef97ae8cf825713583/install/0000_00_cluster-version-operator_03_deployment.yaml#L25-L29
[3]: https://github.com/openshift/cluster-version-operator/blob/ac394f1f73a6e1367d0af8622d4f9af545a6e39a/pkg/start/start.go#L86

install/0001_00_cluster-version-operator_03_service.yaml
wking: > Isn't HTTPS sufficient...

Scratch this, reading the Service docs, I see that `protocol: TCP` is [the default][1].  I'm fine with making that default explicit.  `targetPort` will need to match up with something that the CVO is actually listening on, though (currently [9099 by default][1]).

[1]: https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service
[2]: https://github.com/openshift/cluster-version-operator/blob/ac394f1f73a6e1367d0af8622d4f9af545a6e39a/pkg/start/start.go#L86

install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -1,16 +1,19 @@
...
+    service.alpha.openshift.io/serving-cert-secret-name: openshift-cluster-version-serving-cert
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   type: ClusterIP
   selector:
     k8s-app: cluster-version-operator
   ports:
-  - name: metrics
-    port: 9099 # chosen to be in the internal open range
+  - name: https
```
wking, line 16: I'd rather keep the port named `metrics`, in case the CVO somehow grows some other thing that it serves on a different port in the future.

install/0001_00_cluster-version-operator_03_service.yaml
wking: > To back this, doesn't the CVO need to be serving HTTPS on 443?

*This* portion of my comment is really about `targetPort`, and now there's an [appropriately positioned thread][1] for that.

About the external-facing `port`, we do have [existing docs][2] around host-port 9099 being used for the CVO, so we might want to leave that alone instead of moving it to 443.

[1]: https://github.com/openshift/cluster-version-operator/pull/358#discussion_r417057773
[2]: https://github.com/openshift/openshift-docs/blame/d646e0822ca61e1e5389ea80441d1389b5fd13bf/modules/installation-network-user-infra.adoc#L61

install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -1,16 +1,19 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: cluster-version-operator
+  name: metrics
   namespace: openshift-cluster-version
   labels:
     k8s-app: cluster-version-operator
   annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: openshift-cluster-version-serving-cert
```
wking, line 9: [Docs][1] say we can use `service.beta.openshift.io/serving-cert-secret-name` (not alpha) here.  They also have "For convenience, it is recommended that this be the same as `<service-name>`.", which suggests a value of `cluster-version-operator` to match the original Service name or `metrics` to match your rename.


[1]: https://github.com/openshift/openshift-docs/blob/enterprise-4.4/modules/customize-certificates-add-service-serving.adoc

wking:
After some wandering in the wilderness above, I think I'm starting to wrap my head around this.  Take a look at wking/cluster-version-operator@c8cb5cf12c29b72a107abe514f0d461c8d782258, pick up whatever seems useful, and then move on to figure out the Go behind the new X.509 options.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
simonpasquier: the token file is automatically made available by the [Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#service-account-admission-controller).

simonpasquier: Also there are 2 endpoint definitions. Maybe you wanted this rather:

```yaml
- bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
  interval: 30s
  port: https
  scheme: https
  tlsConfig:
    caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
    serverName: metrics.openshift-cluster-version.svc
```

jottofar:
/retitle Bug 1809195: Send CVO metrics over https

wking:
[Update issue][1] is that the Service is `0001_00_cluster-version-operator_03_service` with the `service.beta.openshift.io/serving-cert-secret-name` annotation is coming too late, so the monitoring tools aren't creating the secret, so the CVO container is failing to start.  From [the must-gather][2]:

```console
$ yaml2json <namespaces/openshift-cluster-version/core/events.yaml | event-log | tail -n1
2020-05-11T14:12:15Z - 2020-05-11T16:07:45Z (13) openshift-cluster-version Unable to attach or mount volumes: unmounted volumes=[serving-cert], unattached volumes=[default-token-4mrxv etc-ssl-certs etc-cvo-updatepayloads serving-cert]: timed out waiting for the condition
```

The deployment is getting bumped earlier in `0000_00_cluster-version-operator_03_deployment`, and the CVO hangs on that without pushing the new service.  Workaround would be to file one PR that just adds the `service.beta.openshift.io/serving-cert-secret-name` annotation to the Service, and we wait until that goes out with a release 4.y.z (with a new "No TLS certs available for HTTPS metrics" bug for backporting).  Once that lands, we land this PR with all the remaining pieces, and that goes out with 4.y.(z+n).  When 4.y.(z+n) enters Cincinnati, we ensure that it and all subsequent 4.y.(z+n) releases are never the update of a 4.y.(z-m) release.  So you have to go from the early z-stream (which has nothing about HTTPS) to a middle-ground z-stream (which has the annotation, but doesn't not consume the resulting secret) to a late z-stream (which has both the annotation and consuming code).  And we'd also need to backport the annotation to 4.(y-1), and block all 4.(y-1) -> 4.y from before the annotation landed in 4.(y-1) so you couldn't go straight to 4.y.(z+n) or later.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/679
[2]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/679/artifacts/e2e-aws-upgrade/must-gather.tar

wking:
In case it helps, here's a graph showing valid (black) and broken (red) edges, in a hypothetical world where the annotation landed in 1.1.2 and was backported to 1.0.3 (blue nodes), and the secret-consuming deployment changes landed in 1.1.4 (green nodes).  The rule is "you need to go black node -> blue node -> green node, and can't go straight from black to green".

jottofar:
Checked must-gather and confirmed service,yaml does not contain the annotation:

 service.alpha.openshift.io/serving-cert-secret-name: openshift-cluster-version-serving-cert

install/0001_00_cluster-version-operator_03_service.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   labels:
     k8s-app: cluster-version-operator
   annotations:
+    service.beta.openshift.io/serving-cert-secret-name: cluster-version-operator-serving-cert
```
wking, line 9: Oh, you should drop this and rebase onto master now that #366 has landed.

wking:
[e2e-aws][1] exploded on CSI tests with lots of errors like:

```
PersistentVolumeClaims [csi-hostpathz5p8w] not all in phase Bound within 5m0s
```

Seems unrelated to this change.

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1348

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
 }
 
+func (o *Options) handler() http.Handler {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	return mux
+}
+
+func (o *Options) makeTLSConfig() (*tls.Config, error) {
+	if !fileExists(o.ServingCertFile) {
```
wking, line 188: Do we need these guards?  I'd expect the TLS consumer would error out with a reasonable message if the cert or key file did not exist.

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
 	return nil
 }
 
+func (o *Options) handler() http.Handler {
```
wking, line 181: Can we drop this helper and just keep this inline?  I don't think saving a line in the caller is worth the indirection.

pkg/start/start.go
wking: Actually, right after these guards you load the bytes, and `ReadFile` returns great error messages [like][1]:

```
open /does/not/exist: no such file or directory
```

So we can definitely drop these guards.

[1]: https://play.golang.org/p/Yc0PxnjRvmH

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+	certificate, err := tls.X509KeyPair(certBytes, keyBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return crypto.SecureTLSConfig(&tls.Config{
+		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return &certificate, nil
+		},
+		ClientAuth: tls.RequestClientCert,
```
wking, line 214: Do we care about client certs?  [The default `NoClientCert`][1] seems reasonable to me.

[1]: https://golang.org/pkg/crypto/tls/#Config

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+
 func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourcelock.ConfigMapLock) {
 	// listen on metrics
 	if len(o.ListenAddr) > 0 {
-		mux := http.NewServeMux()
-		mux.Handle("/metrics", promhttp.Handler())
+		handler := o.handler()
+		tcpl, err := net.Listen("tcp", o.ListenAddr)
+		if err != nil {
+			klog.Errorf("Listen error: %v", err)
```
wking, line 224: I don't think we want to silently continue here.  Can we use `klog.Fatalf` like master does for the `ListenAndServe` today?

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+		// match HTTP first
+		httpl := m.Match(cmux.HTTP1())
 		go func() {
-			if err := http.ListenAndServe(o.ListenAddr, mux); err != nil {
-				klog.Fatalf("Unable to start metrics server: %v", err)
+			s := &http.Server{
+				Handler: handler,
+			}
+			if err := s.Serve(httpl); err != cmux.ErrListenerClosed {
+				klog.Errorf("Serve error (http): %v", err)
```
wking, line 240: I'd like `Fatalf` here too.

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
 			}
 		}()
+
+		var TLSConfig *tls.Config
+		if tlsConfig, err := o.makeTLSConfig(); err != nil {
+			klog.Errorf("Could not create tls Config: %v", err)
```
wking, line 246: nit: `tls Config` -> `TLS config` and `Errorf` -> `Fatalf`.  Maybe `Could not` -> `Failed to`, because I sometimes grep for "fail" in logs but rarely grep for "could not" ;).

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+		} else {
+			TLSConfig = tlsConfig
```
wking, line 248: I don't think it's worth defining a placeholder `var` just to keep `err` out of the scope.  Can we pivot to:

```go
tlsConfig, err := o.makeTLSConfig()
if err != nil {
  klog.Fatalf("Failed to create TLS config: %v", err)
}
```

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+				}
+			}()
+
+			go func() {
+				if err := m.Serve(); !strings.Contains(err.Error(), "use of closed network connection") {
+					klog.Errorf("%v", err)
+				}
+			}()
+		} else {
+			klog.V(0).Info("Metrics port listening on address", o.ListenAddr)
```
wking, line 270: `klog.Infof`?  Both for the formatting `f` and because I don't see a point to a `V(0)` vs. the package-scoped function.

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+
+		var TLSConfig *tls.Config
+		if tlsConfig, err := o.makeTLSConfig(); err != nil {
+			klog.Errorf("Could not create tls Config: %v", err)
+		} else {
+			TLSConfig = tlsConfig
+		}
+
+		if TLSConfig != nil {
+			klog.V(0).Info("Metrics port listening on HTTP and HTTPS on address ", o.ListenAddr)
```
wking, line 252: `klog.Infof`? Both for the formatting f and because I don't see a point to a `V(0)` vs. the package-scoped function.

pkg/start/start.go
```diff
@@ -153,16 +178,97 @@ func (o *Options) Run() error {
...
+			tlsl := m.Match(cmux.Any())
+			tlsl = tls.NewListener(tlsl, TLSConfig)
```
wking, line 254: nit:

```go
tlsListener = tls.NewListener(m.Match(cmux.Any()), tlsConfig)
```

to avoid defining and them immediately clobbering a single variable, and to spell out `l` -> `Listener` (because I like chatty variable names so they're easier to understand at a glance and grep for).

pkg/start/start.go
```diff
@@ -153,16 +177,81 @@ func (o *Options) Run() error {
...
+				Handler: handler,
+			}
+			if err := s.Serve(httpl); err != cmux.ErrListenerClosed {
+				klog.Fatalf("HTTP serve error (http): %v", err)
 			}
 		}()
+
+		tlsConfig, err := o.makeTLSConfig()
+		if err != nil {
+			klog.Fatalf("Failed to create tls Config: %v", err)
```
wking, line 232: nit: [still][1] "tls Config" -> "TLS config".

[1]: https://github.com/openshift/cluster-version-operator/pull/358#discussion_r431280989

wking:
[Died on install][1].

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1401

pkg/start/start.go
jottofar: @wking Changing to Fatalf causes failures. I had hoped that it would go away with https://github.com/openshift/cluster-version-operator/pull/366 but apparently not. It seemed similar but what I originally believe was happening was the pod was coming up before the mount was there causes it to be restarted afterwhich the file is there. With the Fatalf it just get killed off and not restarted. That was the best theory I had at the time in any case.
So when it has Fatalf the fails are at [1405](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1405) and [772](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/358/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/772) . Unfortunately the must-gathers fail as well.

install/0001_00_cluster-version-operator_03_service.yaml
wking: This was fixed and spun out to #366.

install/0001_00_cluster-version-operator_03_service.yaml
wking: fixed as of 16da75689

install/0001_00_cluster-version-operator_03_service.yaml
wking: Fixed as of 16da75689

install/0001_00_cluster-version-operator_03_service.yaml
wking: Fixed as of 16da75689.

jottofar:
e2e-aws failing with ClusterAutoscalerOperatorDown alert firing

pkg/start/start.go
jottofar: fixed in [abf62a5](https://github.com/openshift/cluster-version-operator/pull/358/commits/abf62a532d457b0cb0943ddc98763daec2a6eb2c)

pkg/start/start.go
jottofar: fixed in abf62a5

pkg/start/start.go
jottofar: changed in abf62a5 to use default NoClientCert

pkg/start/start.go
```diff
@@ -79,6 +88,15 @@ func defaultEnv(name, defaultValue string) string {
 	return env
 }
 
+// fileExists checks if a file exists and is not a directory.
+func fileExists(filename string) bool {
```
wking, line 92: You can drop this function now that you've dropped its consumers.

pkg/start/start.go
```diff
@@ -103,6 +121,15 @@ func (o *Options) Run() error {
 	if o.ReleaseImage == "" {
 		return fmt.Errorf("missing --release-image flag, it is required")
 	}
+	if o.ServingCertFile == "" && o.ServingKeyFile != "" {
+		klog.Errorf("--serving-key-file was set, so --serving-cert-file must also be set")
```
wking, line 125: These should be `return fmt.Errorf(...)`, right?

pkg/start/start.go
```diff
@@ -103,6 +121,15 @@ func (o *Options) Run() error {
...
+	}
+	if o.ServingKeyFile == "" && o.ServingCertFile != "" {
+		klog.Errorf("--serving-cert-file was set, so --serving-key-file must also be set")
+	}
+	if o.ServingKeyFile == "" {
+		klog.Errorf("missing --serving-key-file flag, it is required")
```
wking, line 131: It is not required.  Leaving both `--serving-...` options unset just means "don't server HTTPS".  That's valid, and is what we want to do during CVO bootstrapping.  I think we should drop this `if` block.

pkg/start/start.go
```diff
@@ -153,16 +180,80 @@ func (o *Options) Run() error {
...
+				s := &http.Server{
+					Handler: handler,
+				}
+				if err := s.Serve(tlsListener); err != cmux.ErrListenerClosed {
+					klog.Fatalf("HTTPS serve error: %v", err)
+				}
+			}()
+
+			go func() {
+				if err := m.Serve(); !strings.Contains(err.Error(), "use of closed network connection") {
```
wking, line 250: What's the story for not logging `use of closed network connection`?

PR #359, opened 2020-04-24
Do not build with CGO_ENABLED=0

eparis:
Back in ddeb1408 this made sense, since we were based on SCRATCH. But
now we use origin:base and should not try to be fancy.
    
Statically linking the binary and basing on SCRATCH allows us to minimize
the layer size. If we're using a full base image, it no longer makes
sense to statically link, since we have the libraries we need in the image.
    
Also, when using the RHEL golang compiler fork, instead of upstream,
building wth CGO_ENABLED=0 disables the ability to use nss and to get
FIPS compliance. This will cause the CVO image to follow the same pattern
as everything else.


russellb:
s/This well call/This will cause/ ?

PR #360, opened 2020-04-24
Bug 1827166: pkg/cvo/sync_worker: Log precondition handling

wking:
To make it easier to debug precondition failures like [rhbz#1827166][1].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1827166

PR #346, opened 2020-04-13
pkg/cvo/sync_worker: Fix "successfull" -> "successful" typo

wking:
Typo from aceb5bc66f (#243).

PR #361, opened 2020-04-24
pkg/cvo/sync_worker: Log precondition handling

wking:
To make it easier to debug precondition failures like [rhbz#1827166][1].

This backports #360 to 4.4 for testing.  It also backports #346 to avoid a rebase conflict.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1827166

### DIFF
--- a/pkg/payload/precondition/clusterversion/upgradeable.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable.go
@@ -8,6 +8,7 @@ import (
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/klog"
 
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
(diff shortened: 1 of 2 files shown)


pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -46,12 +47,18 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 
 	// if we are upgradeable==true we can always upgrade
```
jottofar, line 48: I always found this comment confusing, or at least overly simplistic, even before your changes. Can it be removed or enhanced.

wking: This comment is from #291, so orthogonal to my change.  Since this PR is a already a backport, I'd rather not pick it up here.  Feel free to file a separate master PR adjusting the wording.

sdodson:
/hold to stop unintentional ci tests, though maybe the invalid-bug flags are enough
Are we leaving this for testing purposes or should we close this?

wking:
/close

We can re-open if/when we fix the master bug

PR #362, opened 2020-04-24
pkg/cvo/sync_worker: Log precondition handling

wking:
To make it easier to debug precondition failures like [rhbz#1827166][1].

This cherry-picks #361 back to 4.3 for testing.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1827166


sdodson:
/hold to stop unintentional ci tests, though maybe the invalid-bug flags are enough
Are we leaving this for testing purposes or should we close this?

wking:
/close

We can re-open if/when we fix the master bug.

PR #363, opened 2020-04-27
Update readme to include fast fill logic

enxebre:
This https://github.com/openshift/cluster-version-operator/pull/318 introduced a PrecreatingMode which actually let the CVO to push the clusterOperator in manifests.
This changes previous behaviour and enables a new pattern for individual operators to set related objects into the clusterOperator defined in manifests/ https://github.com/openshift/cluster-kube-apiserver-operator/pull/833
https://github.com/openshift/cluster-etcd-operator/pull/312
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/398


enxebre:
cc @smarterclayton @abhinavdahiya @deads2k

PR #364, opened 2020-04-30
Bug 1822844: Block z level upgrades when ClusterVersionOverridesSet is set

jottofar:
CVO Upgradeable=False should block all upgrades including z level.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/time/rate"
 	"k8s.io/klog"
@@ -29,7 +30,7 @@ import (
 
 // ConfigSyncWorker abstracts how the image is synchronized to the server. Introduced for testing.
 type ConfigSyncWorker interface {
-	Start(ctx context.Context, maxWorkers int)
+	Start(ctx context.Context, maxWorkers int, cvoOptrName string, lister configlistersv1.ClusterVersionLister)
 	Update(generation int64, desired configv1.Update, overrides []configv1.ComponentOverride, state payload.State) *SyncWorkerStatus
 	StatusCh() <-chan SyncWorkerStatus
 }
@@ -241,7 +242,7 @@ func (w *SyncWorker) Update(generation int64, desired configv1.Update, overrides
 // Start periodically invokes run, detecting whether content has changed.
 // It is edge-triggered when Update() is invoked and level-driven after the
 // syncOnce() has succeeded for a given input (we are said to be "reconciling").
-func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
+func (w *SyncWorker) Start(ctx context.Context, maxWorkers int, cvoOptrName string, lister configlistersv1.ClusterVersionLister) {
 	klog.V(5).Infof("Starting sync worker")
 
 	work := &SyncWork{}
@@ -306,11 +307,15 @@ func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
 				w.lock.Unlock()
 				defer cancelFn()
 
+				config, err := lister.Get(cvoOptrName)
+				if err != nil {
+					return err
+				}
 				// reporter hides status updates that occur earlier than the previous failure,
 				// so that we don't fail, then immediately start reporting an earlier status
 				reporter := &statusWrapper{w: w, previousStatus: w.Status()}
 				klog.V(5).Infof("Previous sync status: %#v", reporter.previousStatus)
-				return w.syncOnce(ctx, work, maxWorkers, reporter)
+				return w.syncOnce(ctx, work, maxWorkers, reporter, config)
 			}()
 			if err != nil {
 				// backoff wait
@@ -466,7 +471,7 @@ func (w *SyncWorker) Status() *SyncWorkerStatus {
 // sync retrieves the image and applies it to the server, returning an error if
 // the update could not be completely applied. The status is updated as we progress.
 // Cancelling the context will abort the execution of the sync.
-func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers int, reporter StatusReporter) error {
+func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers int, reporter StatusReporter, clusterVersion *configv1.ClusterVersion) error {
 	klog.V(4).Infof("Running sync %s (force=%t) on generation %d in state %s at attempt %d", versionString(work.Desired), work.Desired.Force, work.Generation, work.State, work.Attempt)
 	update := work.Desired
 
@@ -524,7 +529,7 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 				Actual:      update,
 				Verified:    info.Verified,
 			})
-			if err := precondition.Summarize(w.preconditions.RunAll(ctx, precondition.ReleaseContext{DesiredVersion: payloadUpdate.ReleaseVersion})); err != nil {
+			if err := precondition.Summarize(w.preconditions.RunAll(ctx, precondition.ReleaseContext{DesiredVersion: payloadUpdate.ReleaseVersion}, clusterVersion)); err != nil {
 				if update.Force {
 					klog.V(4).Infof("Forcing past precondition failures: %s", err)
 				} else {
--- a/pkg/payload/precondition/clusterversion/upgradeable.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable.go
@@ -28,10 +28,26 @@ func NewUpgradeable(lister configv1listers.ClusterVersionLister) *Upgradeable {
 	}
 }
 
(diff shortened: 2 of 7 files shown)


pkg/cvo/updatepayload.go
```diff
@@ -84,24 +84,27 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
...
+			info.VerificationError = vErr
+		} else {
+			info.Verified = true
 		}
-		klog.Warningf("An image was retrieved from %q that failed verification: %v", update.Image, vErr)
-		info.VerificationError = vErr
-	} else {
-		info.Verified = true
-	}
+	*/
```
wking, line 106: This is just commented out for testing right?  We should drop the comments and remove the indent injection now that you have things working locally.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -66,8 +66,22 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 	desiredMinor := getEffectiveMinor(releaseContext.DesiredVersion)
 
 	// if there is no difference in the minor version (4.y.z where 4.y is the same for current and desired), then we can still upgrade
+	// if no cluster overrides have been set
 	if currentMinor == desiredMinor {
 		klog.V(4).Infof("Precondition %q passed: minor from the current %s matches minor from the target %s (both %s).", pf.Name(), cv.Status.History[0].Version, releaseContext.DesiredVersion, currentMinor)
+
+		for _, override := range clusterVersion.Spec.Overrides {
+			if override.Unmanaged {
+				klog.V(4).Infof("ClusterVersionOverridesSet. Z-level upgrade will not start.")
```
wking, line 75: How about:

```go
klog.V(4).Infof("Update from %s to %s blocked by %s: %s", cv.Status.History[0].Version, releaseContext.DesiredVersion, condition.Reason, condition.Message)
```

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -66,8 +66,22 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+
+				return &precondition.Error{
+					Nested:  err,
+					Reason:  "ClusterVersionOverridesSet",
+					Message: "Disabling ownership via cluster version overrides prevents upgrades. Please remove overrides before continuing.",
+					Name:    pf.Name(),
```
wking, line 81: Can we factor [this logic][1] out into a public:

```go
// ClusterVersionOverridesCondition returns an UpgradeableClusterVersionOverrides condition when overrides are set, and nil when no overrides are set.
func ClusterVersionOverridesCondition(cv *configv1.ClusterVersion) *configv1.ClusterOperatorStatusCondition
```

or some such?  And then use that helper here too?  Then future improvements like listing the overridden objects or otherwise copy-editing the reason or message can be made in a single place.

[1]: https://github.com/openshift/cluster-version-operator/blob/cb8241a33846546ab63ec8c01112f633b6980182/pkg/cvo/upgradeable.go#L203-L215

LalatenduMohanty:
@jottofar Can we change the commit message to 
```
Block z level upgrades when ClusterVersionOverridesSet is set

ClusterVersionOverridesSet should prevent upgrades as it disabling ownership via cluster version operator
```

jottofar:
> @jottofar Can we change the commit message to below as the current message does not reflect the actual intent.
Done

sdodson:
/retitle Bug 1822844: Block z level upgrades when ClusterVersionOverridesSet is set

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -66,9 +82,21 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 	desiredMinor := getEffectiveMinor(releaseContext.DesiredVersion)
 
 	// if there is no difference in the minor version (4.y.z where 4.y is the same for current and desired), then we can still upgrade
+	// if no cluster overrides have been set
 	if currentMinor == desiredMinor {
 		klog.V(4).Infof("Precondition %q passed: minor from the current %s matches minor from the target %s (both %s).", pf.Name(), cv.Status.History[0].Version, releaseContext.DesiredVersion, currentMinor)
-		return nil
+		if condition := ClusterVersionOverridesCondition(clusterVersion); condition != nil {
+			klog.V(4).Infof("Update from %s to %s blocked by %s: %s", cv.Status.History[0].Version, releaseContext.DesiredVersion, condition.Reason, condition.Message)
```
wking, line 89: Can we test this with `--to-image`?  I'd expect `releaseContext.DesiredVersion` to be an empty string in that case until we fix [rhbz#1822513][1].  I'm not clear on what we'd do here before that bug gets fixed, but we should know what we're getting ourselves into in case we want to hold this on a fix for the `--to-image` bug.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1822513

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -66,9 +82,21 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+
+			return &precondition.Error{
+				Nested:  err,
```
wking, line 92: we should drop this line right?  Because `err` was last-touched way up at `pf.lister.Get`, and we already handled the `err != nil` case there.  This condition error is born here, and not wrapping an underlying error.

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: Yes, this is dependent on where we end up with [rhbz#1822513](https://bugzilla.redhat.com/show_bug.cgi?id=1822513), i.e. `releaseContext.DesiredVersion` will be an empty string.

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: removed in [9336d69](https://github.com/openshift/cluster-version-operator/pull/364/commits/9336d69bb60bf43d4a5a60e403a3c28a7016c2f9)

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: Further analysis of [rhbz#1822513](https://bugzilla.redhat.com/show_bug.cgi?id=1822513) shows it to be an issue with pulling incorrect history entry as current version (see https://bugzilla.redhat.com/show_bug.cgi?id=1822513#c14). That fix is independent of this fix although both of course are needed.

pkg/cvo/sync_worker.go
```diff
@@ -307,11 +308,16 @@ func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
 				w.lock.Unlock()
 				defer cancelFn()
 
+				config, err := lister.Get(cvoOptrName)
+				if err != nil {
+					klog.Errorf("Failed to get cluster version data from ClusterVersionLister: %v", err)
```
wking, line 313: do we need to log this here?  I'd rather log it up the stack wherever we decide to ignore the returned `err` instead of exiting.

pkg/payload/precondition/clusterversion/upgradeable.go
wking: Going to hold this until #394 fixes the `cv.Status.History[0].Version` approach to use completed updates instead.

/hold

pkg/cvo/sync_worker.go
jottofar: Not clear to me what you're after here wrt "instead of exiting".

jottofar: Fixed in [64e0aba](https://github.com/openshift/cluster-version-operator/pull/364/commits/64e0aba9f9748579edc3988458e0b060ef1736c9)

wking:
[images][1] hit CI-cluster saturation (`Pod pending timeout`).

/retest

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/364/pull-ci-openshift-cluster-version-operator-master-images/1286030598544035840/prowjob.json

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: Fixed in [b72e843](https://github.com/openshift/cluster-version-operator/pull/364/commits/b72e843974ad165f33417da7addeaf4a76a2b690)

wking: /hold cancel

wking:
Looks good to me, but [unit failed][1], including:

```
 --- FAIL: TestCVO_UpgradePreconditionFailing (0.00s)
    cvo_scenarios_test.go:1429: ([]testing.Action) (len=3 cap=3) {
         (testing.GetActionImpl) {
...
```

Possibly needs some CI updates for the `Completed` pivot like [this one][2]?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/364/pull-ci-openshift-cluster-version-operator-master-unit/1286379886650331136
[2]: https://github.com/openshift/cluster-version-operator/pull/394/files#diff-69bb206c424083262fe76e51fed63ba3L119

PR #368, opened 2020-05-16
pkg/cvo/availableupdates: Split 'At' into LastAttempt and LastSyncOrConfigChange

wking:
The previous `At` had been tracking the most recent attempt, regardless of success.  With this PR, we rename it to the more explicit `LastAttempt` and add godocs explaining the intended semantics.

Also grow a `LastSyncOrConfigChange` sibling, tracking the time since the last success (or since the configured upstream had been changed).  This will make it more convenient to alert when the slice of available updates is stale, although that alert itself and the metric that feeds it will come from #357.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -65,14 +65,27 @@ type availableUpdates struct {
 	Upstream string
 	Channel  string
 
-	At time.Time
+	// LastAttempt records the time of the most recent attempt at update
+	// retrieval, regardless of whether it was successful.
+	LastAttempt time.Time
+
+	// LastSyncOrConfigChange records the most recent time when any of
+	// the following events occurred:
+	//
+	// * Upstream changed, reflecting a new authority, and obsoleting
+	//   any information retrieved from (or failures // retrieving from) the
+	//   previous authority.
+	// * Channel changes.  Same reasoning as for Upstream.
+	// * A slice of Updates was successfully retrieved, even if that
+	//   slice was empty.
+	LastSyncOrConfigChange time.Time
 
 	Updates   []configv1.Update
 	Condition configv1.ClusterOperatorStatusCondition
 }
 
 func (u *availableUpdates) RecentlyChanged(interval time.Duration) bool {
-	return u.At.After(time.Now().Add(-interval))
+	return u.LastAttempt.After(time.Now().Add(-interval))
 }
 
 func (u *availableUpdates) NeedsUpdate(original *configv1.ClusterVersion) *configv1.ClusterVersion {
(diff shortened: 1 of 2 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -95,12 +108,24 @@ func (u *availableUpdates) NeedsUpdate(original *configv1.ClusterVersion) *confi
...
 // setAvailableUpdates updates the currently calculated version of updates.
 func (optr *Operator) setAvailableUpdates(u *availableUpdates) {
+	success := false
 	if u != nil {
-		u.At = time.Now()
+		u.LastAttempt = time.Now()
+		if u.Condition.Type == configv1.RetrievedUpdates {
+			success = u.Condition.Status == configv1.ConditionTrue
+		} else {
+			klog.V(4).Warnf("Unrecognized condition %s=%s (%s: %s): cannot judge update retrieval success", condition.Type, condition.Status, condition.Reason, condition.Message)
```
wking, line 117: Apparently `Warnf` [is not a thing][1] and I should drop `V` and use [`Warningf`][2].

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/368/pull-ci-openshift-cluster-version-operator-master-images/1032#1:build-log.txt%3A65
[2]: https://godoc.org/github.com/kubernetes/klog#Warningf

wking: Fixed with 8e0f5c5 -> d62aba1.

pkg/cvo/availableupdates.go
```diff
@@ -95,12 +108,24 @@ func (u *availableUpdates) NeedsUpdate(original *configv1.ClusterVersion) *confi
...
 	}
 
 	optr.statusLock.Lock()
 	defer optr.statusLock.Unlock()
+	if u != nil && (optr.availableUpdates == nil ||
+		optr.availableUpdates.Upstream != u.Upstream ||
+		optr.availableUpdates.Channel != u.Channel ||
+		success) {
+		u.LastSyncOrConfigChange = u.LastAttempt
+	}
```
jottofar, line 128: Is it okay to set LastSyncOrConfigChange even when we had "Unrecognized condition"? Depends on the semantics of LastSyncOrConfigChange but it can't be "the CVO is succesfully retrieving updates" unless Status is ConditionTrue.

wking: > Is it okay to set `LastSyncOrConfigChange` even when we had "Unrecognized condition"?

When you change `Upstream` or `Channel`, yes, because that's the `OrConfigChange` portion of this.  The `LastSync` portion is switching on `success`, which I have `true` on:

```go
if u.Condition.Type == configv1.RetrievedUpdates {
  success = u.Condition.Status == configv1.ConditionTrue
} ...
```

so it will be `false` in all "Unrecognized condition" cases, right?

jottofar: But 'if u != nil && optr.availableUpdates == nil' LastSyncOrConfigChange gets regardless of Condition.Status. Should it just be based on 'success'?

wking: > `if u != nil && optr.availableUpdates == nil`.

This is "we're setting this for the first time".  Might mean we're resetting this value whenever we restart a CVO, which isn't great (and would be fixed by having the CVO populate `optr.availableUpdates` from an in-cluster ClusterVersion on start), but othewise I don't see a problem with this.

jottofar: I was thinking of it failing on the first time, init, and it setting Status to False but if it fails on the first time it leaves it nil so, yeah, should be fine.

jottofar, line 128: LastSyncOrConfigChange is getting clobbered after its initially set. Need below change or similar:

`        if u != nil {
                if optr.availableUpdates == nil ||
                        optr.availableUpdates.Upstream != u.Upstream ||
                        optr.availableUpdates.Channel != u.Channel ||
                        success {
                        u.LastSyncOrConfigChange = u.LastAttempt
                } else if optr.availableUpdates != nil {
                        u.LastSyncOrConfigChange = optr.availableUpdates.LastSyncOrConfigChange
                }
        }`

wking: Good catch; should be fixed with 326ccf5 -> 3812e5f.

wking:
[upgrade][1]:

```
1 pass, 0 skip (1h0m37s)
2020/05/28 20:38:36 Container test in pod e2e-aws-upgrade failed, exit code 1, reason Error 
```

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/368/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/764

PR #369, opened 2020-05-20
install/0000_90_cluster-version-operator_02_servicemonitor: Drop "unstable" from ClusterOperatorDegraded

wking:
If we considered `Degraded=True` something that impacted update stability, I expect we'd want a guard against it that blocked new updates until admins coaxed the operator back into happiness.  This commit softens wording from af62f04483 (#232).

/assign @abhinavdahiya

### DIFF
--- a/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
+++ b/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
@@ -52,7 +52,7 @@ spec:
         severity: critical
     - alert: ClusterOperatorDegraded
       annotations:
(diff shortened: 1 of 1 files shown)


wking:
And apparently I was wrong.  While we don't consider Degraded=True sufficient to block update initialization, we do consider it [sufficient][1] to [block][2] manifest-graph application (except for installation).

/close

[1]: https://github.com/openshift/cluster-version-operator/blob/e318b86d674ccc850266b74169a0a4403d3b633b/docs/user/reconciliation.md#clusteroperator
[2]: https://github.com/openshift/cluster-version-operator/blob/e318b86d674ccc850266b74169a0a4403d3b633b/docs/user/reconciliation.md#clusteroperator

PR #370, opened 2020-05-22
Bug 1838872: Avoid pre-creating clusteroperators that should be excluded

csrwng:
In 4.5 we added code to precreate cluster-operator resources. However, in the precreate loop we're not checking whether resources should be excluded based on annotation. This adds a check to the loop so we can preserve pre-4.5 behavior.

### DIFF
--- a/hack/cluster-version-util/task_graph.go
+++ b/hack/cluster-version-util/task_graph.go
@@ -30,7 +30,7 @@ func newTaskGraphCmd() *cobra.Command {
 
 func runTaskGraphCmd(cmd *cobra.Command, args []string) error {
 	manifestDir := args[0]
-	release, err := payload.LoadUpdate(manifestDir, "")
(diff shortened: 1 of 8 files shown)


csrwng:
@abhinavdahiya ptal

wking:
Can we handle exclusions [when we load manifests][1]?  That way we can't forget to exclude things later on.

[1]: https://github.com/openshift/cluster-version-operator/blob/e318b86d674ccc850266b74169a0a4403d3b633b/pkg/payload/payload.go#L116-L162

csrwng:
> Can we handle exclusions [when we load manifests](https://github.com/openshift/cluster-version-operator/blob/e318b86d674ccc850266b74169a0a4403d3b633b/pkg/payload/payload.go#L116-L162)? That way we can't forget to exclude things later on.

Makes sense, will make the change

csrwng:
@wking updated

PR #371, opened 2020-05-28
pkg/cvo: Set NoDesiredImage reason when desired.Image is empty

wking:
Because every `Failing=True` condition should have a reason.

Also wordsmith the user-facing docs to replace "synchronize" with "reconcile", because our merge logic is more nuanced than the complete match "synchronize" implies for me.

### DIFF
--- a/docs/user/status.md
+++ b/docs/user/status.md
@@ -3,6 +3,27 @@
 [The ClusterVersion object](../dev/clusterversion.md) sets `conditions` describing the state of the cluster-version operator (CVO).
 This document describes those conditions and, where appropriate, suggests possible mitigations.
 
+## Failing
+
+When `Failing` is True, the CVO is failing to reconcile the cluster with the desired release image.
+In all cases, the impact on the cluster will be that dependent nodes in [the manifest graph](reconciliation.md#manifest-graph) may not be [reconciled](reconciliation.md#reconciling-the-graph).
+Note that the graph [may be flattened](reconciliation.md#manifest-graph), in which case there are no dependent nodes.
+
+Most reconciliation errors will result in `Failing=True`, although [`ClusterOperatorNotAvailable`](#clusteroperatornotavailable) has special handling.
+
+### NoDesiredImage
+
+The CVO has not been given a release image to reconcile.
+
+If this happens it is a CVO coding error, because clearing [`desiredUpdate`][api-desired-update] should return you to the current CVO's release image.
+
+### ClusterOperatorNotAvailable
+
+`ClusterOperatorNotAvailable` (or the consolidated `ClusterOperatorsNotAvailable`) is set when the CVO fails to retrieve the ClusterOperator from the cluster or when the retrieved ClusterOperator does not satisfy [the reconciliation conditions](reconciliation.md#clusteroperator).
+
+Unlike most manifest-reconciliation failures, this error does not immediately result in `Failing=True`.
+Under some conditions during installs and updates, the CVO will treat this condition as a `Progressing=True` condition and give the operator up to ten minutes to level before reporting `Failing=True`.
+
 ## RetrievedUpdates
 
 When `RetrievedUpdates` is `True`, the CVO is succesfully retrieving updates, which is good.
@@ -107,5 +128,6 @@ If this error occurs because you forced an update to a release that is not in an
 If this happens it is a CVO coding error.
 There is no mitigation short of updating to a new release image with a fixed CVO.
 
+[api-desired-update]: https://github.com/openshift/api/blob/34f54f12813aaed8822bb5bc56e97cbbfa92171d/config/v1/types_cluster_version.go#L40-L54
 [channels]: https://docs.openshift.com/container-platform/4.3/updating/updating-cluster-between-minor.html#understanding-upgrade-channels_updating-cluster-between-minor
 [Cincinnati]: https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md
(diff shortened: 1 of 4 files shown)


wking:
[e2e-aws][1]:

```
level=fatal msg="failed to initialize the cluster: Cluster operator console is reporting a failure: RouteHealthDegraded: route not yet available, https://console-openshift-console.apps.ci-op-y4iighzl-928f7.origin-ci-int-aws.dev.rhcloud.com/health returns '503 Service Unavailable'"
```

Dunno what that's about.

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/371/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1527#1:build-log.txt%3A110

wking:
Prow must have hiccuped, because [e2e-aws][1] failed (on `RouteHealthDegraded` on install), while [e2e-aws-upgrade][2] passed, despite Prow currently claiming `Job triggered`.

/refresh

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/371/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1527
[2]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/371/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/840

docs/user/status.md
```diff
@@ -3,6 +3,27 @@
 [The ClusterVersion object](../dev/clusterversion.md) sets `conditions` describing the state of the cluster-version operator (CVO).
 This document describes those conditions and, where appropriate, suggests possible mitigations.
 
+## Failing
+
+When `Failing` is True, the CVO is failing to reconcile the cluster with the desired release image.
+In all cases, the impact on the cluster will be that later nodes in [the manifest graph](reconciliation.md#manifest-graph) may not be [reconciled](reconciliation.md#reconciling-the-graph).
```
LalatenduMohanty, line 9: ` that later nodes` does not communicate clearly.  Can we just say some nodes?

docs/user/status.md
```diff
@@ -3,6 +3,27 @@
...
+Note that the graph [may be flattened](reconciliation.md#manifest-graph), in which case there are no "later nodes".
```
LalatenduMohanty, line 10: Another mention of `later nodes`.

docs/user/status.md
```diff
@@ -3,6 +3,27 @@
...
+
+Most reconciliation errors will result in `Failing=True`, although [`ClusterOperatorNotAvailable`](#clusteroperatornotavailable) has special handling.
+
+### NoDesiredImage
+
+The CVO has not been given a release image to reconcile.
+
+This should not be possible, because clearing [`desiredUpdate`][api-desired-update] should return you to the current CVO's release image.
```
LalatenduMohanty, line 18: When you say this should  not be possible, do you mean this is a bug?

LalatenduMohanty: Also I will suggest pointing to code i.e. `[api-desired-update]: https://github.com/openshift/api/blob/34f54f12813aaed8822bb5bc56e97cbbfa92171d/config/v1/types_cluster_version.go#L40-L54` is not a good idea. As code can change and it will result in to a broken link.

docs/user/status.md
wking: How about "dependent nodes"?

docs/user/status.md
wking: > When you say this should not be possible, do you mean this is a bug?

I mean that if it happens, it is a CVO bug.

> As code can change and it will result in to a broken link.

My link pins a specific API commit by hash, so it does not float with new API code changes.

LalatenduMohanty: > I mean that if it happens, it is a CVO bug.

In that case, lets mention that if that happens it is a CVO bug. 

> My link pins a specific API commit by hash, so it does not float with new API code changes.

Still this does not sound like a good idea because what happens the API changes and the code snippet does not reflect the current code correctly.

docs/user/status.md
LalatenduMohanty: Works for me

docs/user/status.md
wking: > Still this does not sound like a good idea because what happens the API changes and the code snippet does not reflect the current code correctly.

Then we yell at the API folks for the breaking change ;).  But there's no way to float with master and also auto-guard against breaking changes (or even line renumbers).  This commit-pinned link uses the same approach we use for vendored code, and I think that makes sense, and we periodically review and bump our pinned commit as needed.

docs/user/status.md
wking: Fixed "later nodes" -> "dependent nodes" with 89b9af1 -> d459a57.

docs/user/status.md
wking: Fixed https://github.com/openshift/cluster-version-operator/pull/371#discussion_r441246725

docs/user/status.md
wking: > In that case, lets mention that if that happens it is a CVO bug.

Echoed the existing "If this happens it is a CVO coding error" wording with d459a57 -> 1249588.

Still outstanding for this thread is getting approval or a motivated alternative for the commit-pinned API repo link.

PR #372, opened 2020-05-28
Bug 1838497: pkg/cvo/sync_worker: Do not treat "All errors were context errors..." as success

wking:
With this commit, I drop `contextIsCancelled` in favor of `Context.Err()`. From [the docs][1]:

> If Done is not yet closed, Err returns nil.  If Done is closed, Err returns a non-nil error explaining why: Canceled if the context was canceled or DeadlineExceeded if the context's deadline passed.  After Err returns a non-nil error, successive calls to Err return the same error.

I dunno why we'd been checking `Done()` instead, but `contextIsCancelled` dates back to 961873d66a (#82).

I've also generalized a number of `*Cancel*` helpers to be `*Context*` to remind folks that `Context.Err()` can be `DeadlineExceeded` as well as `Canceled`, and the CVO uses both `WithCancel` and `WithTimeout`.  The new error messages will be either:

    update context deadline exceeded at 1 of 2

or:

    update context canceled at 1 of 2

Instead of always claiming:

    update was cancelled at 1 of 2

[1]: https://golang.org/pkg/context/#Context

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -626,8 +626,8 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 	if precreateObjects {
 		payload.RunGraph(ctx, graph, 8, func(ctx context.Context, tasks []*payload.Task) error {
 			for _, task := range tasks {
-				if contextIsCancelled(ctx) {
-					return cr.CancelError()
+				if err := ctx.Err(); err != nil {
+					return cr.ContextError(err)
 				}
 				if task.Manifest.GVK != configv1.SchemeGroupVersion.WithKind("ClusterOperator") {
 					continue
@@ -645,8 +645,8 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 	// update each object
 	errs := payload.RunGraph(ctx, graph, maxWorkers, func(ctx context.Context, tasks []*payload.Task) error {
 		for _, task := range tasks {
-			if contextIsCancelled(ctx) {
-				return cr.CancelError()
+			if err := ctx.Err(); err != nil {
+				return cr.ContextError(err)
 			}
 			cr.Update()
 
@@ -690,11 +690,11 @@ func init() {
 	)
 }
 
-type errCanceled struct {
+type errContext struct {
 	err error
 }
 
-func (e errCanceled) Error() string { return e.err.Error() }
+func (e errContext) Error() string { return e.err.Error() }
 
 // consistentReporter hides the details of calculating the status based on the progress
 // of the graph runner.
@@ -731,7 +731,7 @@ func (r *consistentReporter) Error(err error) {
 	copied := r.status
 	copied.Step = "ApplyResources"
 	copied.Fraction = float32(r.done) / float32(r.total)
-	if !isCancelledError(err) {
+	if !isContextError(err) {
 		copied.Failure = err
 	}
 	r.reporter.Report(copied)
@@ -752,10 +752,10 @@ func (r *consistentReporter) Errors(errs []error) error {
 	return err
 }
 
-func (r *consistentReporter) CancelError() error {
+func (r *consistentReporter) ContextError(err error) error {
 	r.lock.Lock()
 	defer r.lock.Unlock()
-	return errCanceled{fmt.Errorf("update was cancelled at %d of %d", r.done, r.total)}
+	return errContext{fmt.Errorf("update %s at %d of %d", err, r.done, r.total)}
 }
 
 func (r *consistentReporter) Complete() {
@@ -771,11 +771,11 @@ func (r *consistentReporter) Complete() {
 	r.reporter.Report(copied)
 }
 
-func isCancelledError(err error) bool {
+func isContextError(err error) bool {
 	if err == nil {
 		return false
 	}
-	_, ok := err.(errCanceled)
+	_, ok := err.(errContext)
 	return ok
 }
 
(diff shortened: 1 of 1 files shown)


wking:
Added on the fix for [rhbz#1838497][1], since it is also in the "are these context errors?" space.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1838497#c23

pkg/cvo/sync_worker.go
```diff
@@ -668,8 +668,10 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 		return nil
 	})
 	if len(errs) > 0 {
-		err := cr.Errors(errs)
-		return err
+		if err := cr.Errors(errs); err != nil {
+			return err
+		}
+		return errs[0]
```
jottofar, line 674: So with this change we will always return an error. either summarized or err[0], even if all errors were context errors. Whereas before we could possibly be ignoring DeadlineExceeded errors, which is incorrect, but this also will not ignore Canceled either. Am I interpreting this correctly and is the latter intended?

LalatenduMohanty, line 674: We are only sending the first context error to syncOnce as you mentioned in the commit message. However in the code it is not clear to me that sending all the errors will cause any damage. 
```
		return w.syncOnce(ctx, work, maxWorkers, reporter)
			}()
			if err != nil {
				// backoff wait
				// TODO: replace with wait.Backoff when 1.13 client-go is available
				consecutiveErrors++
```

wking: yup.  Neither `DeadlineExceeded` nor `Canceled` should result in us claiming the update completed.

wking: > However in the code it is not clear to me that sending all the errors will cause any damage.

We could probably return all of them if we wanted, but `errs` is an `[]error` and `apply()` returns an `error`, so we'd need to do something to collapse this down to a single `error`.  I'm fine trying with `errors.NewAggregate(errs)` instead if you prefer, and don't see a lot of space between the two approaches.  We will already be logging the unfiltered list under `All errors were cancellation errors...`, so I don't see a problem with excluding some of the additional context errors in other logging fed by the `syncOnce` return value.

wking:
e2e:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/372/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1395/build-log.txt | grep -A6 'Failing tests:' | head -n7
Failing tests:

[sig-builds][Feature:Builds] custom build with buildah  being created from new-build should complete build with custom builder image [Suite:openshift/conformance/parallel]
[sig-imageregistry][Feature:ImageInfo] Image info should display information about images [Suite:openshift/conformance/parallel]
[sig-instrumentation][Late] Alerts shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured [Suite:openshift/conformance/parallel]
[sig-operator] an end user use OLM can subscribe to the cockroachdb operator [Suite:openshift/conformance/parallel]

```

One of the alerts is `CertifiedOperatorConnectionErrors`, so possibly related to the Quay outage.

/retest

wking:
openshift/ci-tools#860 should help update CI.

/retest

pkg/cvo/sync_worker.go
```diff
@@ -731,7 +733,7 @@ func (r *consistentReporter) Error(err error) {
 	copied := r.status
 	copied.Step = "ApplyResources"
 	copied.Fraction = float32(r.done) / float32(r.total)
-	if !isCancelledError(err) {
+	if !isContextError(err) {
```
abhinavdahiya, line 736: Replace this with errors.Is ?

pkg/cvo/sync_worker.go
LalatenduMohanty: Thanks for the explanation. I missed that part when I was reviewing the PR. `func NewAggregate()` sounds better because we are not just returning the first error.

LalatenduMohanty: So the goal is to return context errors if they exist and not ignore them. So either we just use `NewAggregate` and not use `cr.Errors` as we do not want to filter context errors or create a function for this specific purpose. Is there benefit of current code VS just using `NewAggregate`?

wking: > Is there benefit of current code VS just using NewAggregate?

The context errors are from the individual graph workers, so depending on how many graph workers are aborted this way you can get a number of stuttering values.  For example, looking at one of the logs attached to the bug:

```console
$ grep 'Result of work.*update was cancelled' | head -n2
Logs_for_both_environments/unsuccessful_oc_logs_cluster_version_operator.log | head
I0524 16:08:36.953283       1 task_graph.go:611] Result of work: [update was cancelled at 0 of 573]
I0524 16:17:14.532134       1 task_graph.go:611] Result of work: [Cluster operator machine-config is still updating update was cancelled at 196 of 573 Cluster operator openshift-apiserver is reporting a failure: APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable update was cancelled at 196 of 573]
```

Formatting there is not super clear, but the 16:17 entry had four workers fail:

* Cluster operator machine-config is still updating
* update was cancelled at 196 of 573
* Cluster operator openshift-apiserver is reporting a failure: APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable
* update was cancelled at 196 of 573

I don't see a need to include a bunch of stutters in cases where the only worker issues were context failures.  But if you feel a strong preference for `NewAggregate` I can use that instead ;).

LalatenduMohanty: You are right as per https://github.com/openshift/cluster-version-operator/blob/master/pkg/payload/task_graph.go#L592-L599 I was expecting below
```
I0524 16:17:14.532134       1 task_graph.go:611] Result of work: [Cluster operator machine-config is still updating update was cancelled at 196 of 573 Cluster operator openshift-apiserver is reporting a failure: APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable update was cancelled at 196 of 573
```
The only question remaining is whether we will miss out any useful information if we do not report all the errors but `return errs[0]` ?

pkg/cvo/sync_worker.go
LalatenduMohanty: Which `errors.Is` we are talking about here. Is it from https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go ?

pkg/cvo/sync_worker.go
wking: We won't, because the `All errors were cancellation errors...` log line will contain the full slice.

pkg/cvo/sync_worker.go
abhinavdahiya: > Which errors.Is we are talking about here. 

https://blog.golang.org/go1.13-errors

wking:
[upgrade job][1] seems to have gotten lost.  Kick it.

/test e2e-aws-upgrade

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/372/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/810

wking:
CI registry/cluster flake.

/retest

PR #373, opened 2020-05-28
pkg/cvo/sync_worker: Promote "Sync succeeded" to v4

wking:
To match the existing v4 logs for:

    Running sync %s (force=%t) on generation %d in state %s at attempt %d

To avoid excessive noise, also restrict it to edge transitions between states (the `Running sync ...` line exposes the level) and include the source state in the logged message.  Because the * -> `Reconciling` transition is important, and deserves at least as much weight in the logs as "we're about to start a new sync cycle" and a log entry that comes before the next `Until` sleep.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -329,7 +329,9 @@ func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
 				utilruntime.HandleError(fmt.Errorf("unable to synchronize image (waiting %s): %v", interval, err))
 				continue
 			}
-			klog.V(5).Infof("Sync succeeded, reconciling")
+			if work.State != payload.ReconcilingPayload {
+				klog.V(4).Infof("Sync succeeded, transitioning from %s to %s", work.State, payload.ReconcilingPayload)
+			}
 
 			work.Completed++
 			work.State = payload.ReconcilingPayload


pkg/cvo/sync_worker.go
```diff
@@ -329,7 +329,9 @@ func (w *SyncWorker) Start(ctx context.Context, maxWorkers int) {
 				utilruntime.HandleError(fmt.Errorf("unable to synchronize image (waiting %s): %v", interval, err))
 				continue
 			}
-			klog.V(5).Infof("Sync succeeded, reconciling")
+			if work.State != payload.ReconcilingPayload {
+				klog.V(4).Infof("Sync succeeded, transitioning from %s to %s", work.State, payload.ReconcilingPayload)
```
LalatenduMohanty, line 333: Doesv(4) gets printed in the CVO log by default?

wking: yup.  See our existing CVO logs for lovely `Running sync...in state...` entries which I refer to all the time, because we currently don't bubble the failing manifests up into the ClusterVersion.

wking:
[`Some ingresscontrollers are degraded: default`][1].

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/373/pull-ci-openshift-cluster-version-operator-master-e2e-aws-upgrade/752

PR #375, opened 2020-05-28
pkg/cvo/sync_worker: Drop unused 'reconciling' from SyncWorker

wking:
The property landed with SyncWorker in 961873d66a (#82), but has never been used.


wking:
`CommunityOperatorConnectionErrors` from the Quay outage.

/retest

wking:
[e2e-aws][1] seems to have lost a node and blown up, but not sure which test actually failed.

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/375/pull-ci-openshift-cluster-version-operator-master-e2e-aws/1399

PR #386, opened 2020-06-16
WIP: *: Unify terminology around "upgrades" instead of "updates"

wking:
We were not very consistent before, which can be confusing for folks who wonder if we are attempting to make a distinction between the two terms.  While some folks try to distinguish between bugfix updates and new-feature upgrades, the CVO does not make that distinction.  This commit moves us to the "upgrades" where we can do so without breaking APIs.  For example, it does not change:

* The `cluster_version_available_updates` metric.
* The `RetrievedUpdates` condition type and its reasons like `UpdatePayloadIntegrity`.
* External API's like `ClusterVersion`'s `desiredUpdate`, `availableUpdates`, and `updateHistory`.

To avoid slippery slopes, all internal handlers for those external APIs are being changed to use "upgrades", to minimize confusion about where boundaries should be drawn between the external API and internal naming.

I am not changing `update-codegen.sh` or `update-vendor.sh`, since those are internal tools that are distinct from cluster upgrades.

I am changing `--enable-auto-update` -> `--enable-auto-upgrade`, because while that is technically an external API, the only consumers who could possibly touch it today are folks who are overriding the CVO deployment, and those folks are unlikely to be upgrading anyway.

WIP while I work any remaining kinds out of overly-broad `sed`s ;).

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -58,7 +58,7 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 		if err != nil && !errors.IsNotFound(err) {
 			return err
 		}
-		err = updatePodSpecWithProxy(
+		err = upgradePodSpecWithProxy(
 			&deployment.Spec.Template.Spec,
 			strings.Split(containerNamesString, ","),
 			proxyConfig.Status.HTTPProxy,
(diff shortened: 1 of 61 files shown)


lib/resourcebuilder/apps.go
```diff
@@ -58,7 +58,7 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 		if err != nil && !errors.IsNotFound(err) {
 			return err
 		}
-		err = updatePodSpecWithProxy(
+		err = upgradePodSpecWithProxy(
```
LalatenduMohanty, line 61: We do not need to change this or similar code. We should only focus on documentation changes and if anyone complains we can change the CLI options.

PR #387, opened 2020-06-17
CONTRIBUTING: Add a contributing file

wking:
Based on [the installer's][1], but with a few local changes like:

* No section about mock tests, because we don't have any.
* No section about the Developer Certificate of Origin, because we don't use `Signed-off-by`, and I'm skeptical about how binding a `CONTRIBUTING.md` -> `DCO` link is vs. a contributor claiming "I just wrote a commit; I didn't read all the docs in the repo first".  If we decide that a `CONTRIBUTING.md` -> `DCO` link is binding, we can always add it later.
* Trimmed down section about linting, because we don't do as much of that as the installer.
* One sentence per line, so we don't have to worry about rewrapping (and complicating diff/blame) as we edit this content going forward.
* Whatever else seemed reasonable when I went through ;).

Also wash the repository through `go fmt ./...`.

[1]: https://github.com/openshift/installer/blob/4e46d0a347533263903beb3349a33f53eee7a6c2/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# How to Contribute
+
+The cluster-version operator is [Apache 2.0 licensed](LICENSE) and accepts contributions via GitHub pull requests.
+This document outlines some of the conventions on development workflow, commit message formatting, contact points and other resources to make it easier to get your contribution accepted.
+
+## Security Response
+
+If you've found a security issue that you'd like to disclose confidentially, please contact Red Hat's Product Security team.
+Details [here][security].
+
+## Getting Started
+
+- Fork the repository on GitHub.
+- Read the [README](README.md) and [developer documentation](docs/dev) for build and test instructions.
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,70 @@
...
+<why this change was made>
+<BLANK LINE>
+<footer>
+```
+
+The first line is the subject and should be no longer than 70 characters, the second line is always blank, and other lines should be wrapped at 80 characters.
+This allows the message to be easier to read on GitHub as well as in various Git tools.
+
+[golang-style]: https://github.com/golang/go/wiki/CodeReviewComments
+[new-issue]: https://github.com/openshift/cluster-verison-operator/issues/new
```
jottofar, line 68: broken link, typo

wking: Fixed with aff7628 -> dd14258.

PR #388, opened 2020-06-17
doc/dev/upgrades: Add a blurb about restarting upgrades

cgwalters:
I was going to make this a blog entry, but it got shot down
as sounding like a bug and we didn't want customers to worry.

Let's stash it here in upstream git where few customers
will find it, and the ones that do are hopefully more technically
inclined and will understand the logic.

(Originally stored at https://hackmd.io/Yph-TnRmR9ekzEJht0IzUA )

### DIFF
--- a/docs/dev/upgrades.md
+++ b/docs/dev/upgrades.md
@@ -57,3 +57,27 @@ Which in practice can be described in runlevels:
 0000_80_*: machine operators
 0000_90_*: reserved for any post-machine updates
 ```
+
+## Why does the OpenShift 4 upgrade process "restart" in the middle?
+
+Since the release of OpenShift 4, a somewhat frequently asked question is: Why sometimes during an `oc adm upgrade` (cluster upgrade) does the process appear to re-start partway through?  [This bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1690816) for example has a number of duplicates, and I've seen the question appear in chat and email forums.
+
+The answer to this question is worth explaining in detail, because it illustrates some fundamentals of the [self-driving, operator-focused OpenShift 4](https://blog.openshift.com/openshift-4-a-noops-platform/).  During the initial development of OpenShift 4, the toplevel [cluster-version-operator](https://github.com/openshift/cluster-version-operator/) (CVO) and the [machine-config-operator](https://github.com/openshift/machine-config-operator/) (MCO) were developed concurrently (and still are).
+
+The MCO is just one of a number of "second level" operators that the CVO manages.  However, the relationship between the CVO and MCO is somewhat special because the MCO [updates the operating system itself](https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md) for the control plane.
+
+If the new release image has an updated operating system (`machine-os-content`), the CVO pulling down an update ends up causing it to (indirectly) restart itself.
+
+This is because in order to apply the OS update (or any config changes) MCO will drain each node it is working on updating, then reboot.  The CVO is just a regular pod (driven by a `deployment`) running in the cluster (`oc -n openshift-cluster-version get pods`); it gets drained and rescheduled just like the rest of the platform it manages, as well as user applications.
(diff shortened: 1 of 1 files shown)


docs/dev/upgrades.md
```diff
@@ -56,4 +56,28 @@ Which in practice can be described in runlevels:
...
+
+The answer to this question is worth explaining in detail, because it illustrates some fundamentals of the [self-driving, operator-focused OpenShift 4](https://blog.openshift.com/openshift-4-a-noops-platform/).  During the initial development of OpenShift 4, the toplevel [cluster-version-operator](https://github.com/openshift/cluster-version-operator/) (CVO) and the [machine-config-operator](https://github.com/openshift/machine-config-operator/) (MCO) were developed concurrently (and still are).
+
+The MCO is just one of a number of "second level" operators that the CVO manages.  However, the relationship between the CVO and MCO is somewhat special because the MCO [updates the operating system itself](https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md) for the control plane.
+
+If the new release image has an updated operating system (`machine-os-content`), the CVO pulling down an update ends up causing it to (indirectly) restart itself.
```
sdodson, line 69: Idle curiosity, what percentage of OCP releases bump `machine-os-content`?

wking:
I'd rather cache state in ClusterVersion so we can pick up where we left off (#264 was a step in that direction).  I am agnostic about landing docs in the meantime.

sdodson:
> I'd rather cache state in ClusterVersion so we can pick up where we left off (#264 was a step in that direction). I am agnostic about landing docs in the meantime.

I'd go ahead and take this until such a time as we persist state somehow. Our track record for fixing non critical issues in CVO doesn't suggest we'll get to this anytime soon.

docs/dev/upgrades.md
```diff
@@ -56,4 +56,28 @@ Which in practice can be described in runlevels:
...
+
+Also, besides operating system updates, there's the case where an updated payload changes the CVO image itself.
+
+Today, there's no special support in the CVO for passing "progress" between the previous and new pod; the new pod just looks at the current cluster state and attemps to reconcile between the observed and desired state.  This is generally true of the "second level" operators as well, from the MCO to the network operator, the router, etc.
+
+Hence, the fact that the CVO is terminated and restarted is visible to components watching the `clusterversion` object as the status is recalculated.
+
+I could imagine at some point adding clarification for this; perhaps a basic boolean flag state in e.g. a `ConfigMap` or so that denoted that the pod was drained due to an upgrade, and the new CVO pod would "consume" that flag and include "Resuming upgrade..." text in its status. But I think that's probably all we should do.
```
vrutkovs, line 79: @jottofar @wking correct me if I'm wrong, but a boolean flag for this is not necessary. New CVO can detect in-progress update when started (since current version doesn't yet matches desiredVersion). 

Not sure if it can autodetect the percentage though

jottofar: It knows it's in an update but I don't know how much state it saves, e.g. RetrievedUpdates, to know it has already loaded the update.

wking: I am in favor of storing information about synced/failing manifests in ClusterVersion, which would allow us to pick back up where the previous CVO left off.  But there's no such stored state today.

LalatenduMohanty: This should be tracked in a Jira.

LalatenduMohanty (approved):
/lgtm
/hold in case others have any different opinion. If we do not receive any comment to change the PR after a day we can remove the hold.

docs/dev/upgrades.md
wking: no need to track in Jira, this is already tracked in [the bug][1] that these docs link.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1690816#c3

sdodson:
/override ci/prow/e2e-upgrade
docs only changes

PR #392, opened 2020-06-24
pkg/cvo/status: Change "downloading update" to "downloading release image"

wking:
Folks are seeing `downloading update` messages during install, [like][1]:

```
time="2020-06-05T15:01:43Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 4.5.0-0.nightly-2020-06-04-025914: 57% complete"
time="2020-06-05T15:02:53Z" level=debug msg="Still waiting for the cluster to initialize: Multiple errors are preventing progress:\n* Could not update oauthclient \"console\" (351 of 584): the server does not recognize this resource, check extension API servers\n* Could not update prometheusrule \"openshift-cloud-credential-operator/cloud-credential-operator-alerts\" (186 of 584): the server does not recognize this resource, check extension API servers"
time="2020-06-05T15:03:44Z" level=debug msg="Still waiting for the cluster to initialize: Multiple errors are preventing progress:\n* Could not update oauthclient \"console\" (351 of 584): the server does not recognize this resource, check extension API servers\n* Could not update prometheusrule \"openshift-cloud-credential-operator/cloud-credential-operator-alerts\" (186 of 584): the server does not recognize this resource, check extension API servers"
time="2020-06-05T15:10:44Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 4.5.0-0.nightly-2020-06-04-025914"
time="2020-06-05T15:10:44Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 4.5.0-0.nightly-2020-06-04-025914: downloading update"
time="2020-06-05T15:10:44Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 4.5.0-0.nightly-2020-06-04-025914"
time="2020-06-05T15:10:44Z" level=debug msg="Still waiting for the cluster to initialize: Working towards 4.5.0-0.nightly-2020-06-04-025914: 0% complete"
```

when the CVO pod is rescheduled and the new pod is initializing itself.  We might want to adjust things so we can skip that step when the target release image matches the CVO pod's release image, but regardless of whether we do that, using the more geneneric `release image` works for both updates and initial release images.  I'm leaving the `DownloadingUpdate` reason unchanged, because reasons are part of our committed API, and this seems like a minor enough change that I don't think we should break that API.  Messages, on the other hand, are for human consumption and [explicitly][2] not intended for machine matching.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1844529#c0
[2]: https://github.com/openshift/api/blob/eb651a5bb0ad41d771003d277a6d33b8ee307085/config/v1/types_cluster_operator.go#L133

### DIFF
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -1140,7 +1140,7 @@ func TestOperator_sync(t *testing.T) {
 							{Type: configv1.OperatorAvailable, Status: configv1.ConditionFalse},
 							{Type: ClusterStatusFailing, Status: configv1.ConditionFalse},
 							// we correct the message that was incorrect from the previous state
-							{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Reason: "DownloadingUpdate", Message: "Working towards image/image:v4.0.1: downloading update"},
+							{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Reason: "DownloadingUpdate", Message: "Working towards image/image:v4.0.1: downloading release image"},
 							{Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse},
 						},
 					},
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -295,7 +295,7 @@ func (optr *Operator) syncStatus(original, config *configv1.ClusterVersion, stat
 				if len(reason) == 0 {
 					reason = "DownloadingUpdate"
 				}
-				message = fmt.Sprintf("Working towards %s: downloading update", version)
+				message = fmt.Sprintf("Working towards %s: downloading release image", version)
 			case skipFailure:
 				reason = progressReason
 				message = fmt.Sprintf("Working towards %s: %s", version, progressShortMessage)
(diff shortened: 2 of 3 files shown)


sdodson:
/hold
What does this improve? This feels like change for the sake of change.

jottofar:
> /hold
> What does this improve? This feels like change for the sake of change.

Some recent history: https://coreos.slack.com/archives/CEGKQ43CP/p1593013365496100

sdodson:
It's still inaccurate right? Is is actually downloading the release image again?
@abhinavdahiya Have you any thoughts on whether we should change this?

abhinavdahiya:
During installation we should not see that message at all. the downloading should only should up when we are actually trying to download. During installation, even if CVO is restarted it shouldn't need to download any thing new since it itself is the correct version.

AFAIK this was not a problem in previous versions like 4.4 etc.. so I think there is a regression here that needs to be fixed instead of the renaming.

jottofar:
@abhinavdahiya I was looking at CVO log in this must-gather http://file.rdu.redhat.com/lbednar/must-gather.local.8978213012992196160.tar.gz with the issue trying to understand the flow.
I only see an initial "Loading payload". Is that correct, should it even call it at startup?

It only spits out that log if the % <=0 and Step = RetrievePayload:
https://github.com/openshift/cluster-version-operator/blob/2b849292e49f1365ce1c84541194b736b76d7b18/pkg/cvo/status.go#L294-L298

sdodson:
@wking Do you wish to move forward with this or should we close it?

PR #294, opened 2020-01-07
pkg/verify: Refactor signature fetching to only use a Store interface

wking:
This abstracts signature retrieval, setting the stage for future pivots such as fetching signatures from config maps in the local cluster.

Spun off from [these suggestions][1] for #279.  I've left the file implementation in place to minimize the diff (in 0cc3e41c1d04 I'd replaced it with an in-memory sig store).  I've also left both the file and HTTP stores in the `verify.go` file to minimize the diff (in 0cc3e41c1d04 I'd moved the HTTP store into a subpackage to simplify `verify.go`).

[1]: https://github.com/openshift/cluster-version-operator/pull/279#discussion_r359239295

PR #393, opened 2020-06-24
Bug 1840343: pkg/verify: Parallelize HTTP(S) signature retrieval

wking:
Builds on top of #294.  You can review that one first if you like, or review both commits in this PR (and GitHub will mark #294 merged if this one lands first, because it will pull #294's tip into the target `master` branch).  With #294, the change we need to make to parallelize HTTP(S) retrieval is pretty small.  I'm just adding a parallel store implementation and using that instead of the serial wrapper for HTTP(S) retrieval.

### DIFF
--- a/pkg/verify/store/configmap/configmap.go
+++ b/pkg/verify/store/configmap/configmap.go
@@ -1,4 +1,4 @@
-package verifyconfigmap
+package configmap
 
 import (
 	"context"
@@ -16,6 +16,8 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog"
+
+	"github.com/openshift/cluster-version-operator/pkg/verify/store"
 )
 
 // ReleaseLabelConfigMap is a label applied to a configmap inside the
@@ -87,10 +89,10 @@ func digestToKeyPrefix(digest string) (string, error) {
 	return fmt.Sprintf("%s-%s", algo, hash), nil
 }
 
-// DigestSignatures returns a list of signatures that match the request
+// Signatures returns a list of signatures that match the request
 // digest out of config maps labelled with ReleaseLabelConfigMap in the
 // openshift-config-managed namespace.
-func (s *Store) DigestSignatures(ctx context.Context, digest string) ([][]byte, error) {
+func (s *Store) Signatures(ctx context.Context, name string, digest string, fn store.Callback) error {
 	// avoid repeatedly reloading config maps
 	items := s.mostRecentConfigMaps()
 	r := s.limiter.Reserve()
@@ -100,7 +102,7 @@ func (s *Store) DigestSignatures(ctx context.Context, digest string) ([][]byte, 
 		})
 		if err != nil {
 			s.rememberMostRecentConfigMaps([]corev1.ConfigMap{})
-			return nil, err
+			return err
 		}
 		items = configMaps.Items
 		s.rememberMostRecentConfigMaps(configMaps.Items)
@@ -108,23 +110,28 @@ func (s *Store) DigestSignatures(ctx context.Context, digest string) ([][]byte, 
 
 	prefix, err := digestToKeyPrefix(digest)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	var signatures [][]byte
 	for _, cm := range items {
 		klog.V(4).Infof("searching for %s in signature config map %s", prefix, cm.ObjectMeta.Name)
 		for k, v := range cm.BinaryData {
 			if strings.HasPrefix(k, prefix) {
 				klog.V(4).Infof("key %s from signature config map %s matches %s", k, cm.ObjectMeta.Name, digest)
-				signatures = append(signatures, v)
+				done, err := fn(ctx, v, nil)
+				if err != nil || done {
+					return err
+				}
+				if err := ctx.Err(); err != nil {
+					return err
+				}
 			}
 		}
 	}
-	return signatures, nil
+	return nil
 }
 
-// Store attempts to persist the provided signatures into a form DigestSignatures will
+// Store attempts to persist the provided signatures into a form Signatures will
 // retrieve.
 func (s *Store) Store(ctx context.Context, signaturesByDigest map[string][][]byte) error {
 	cm := &corev1.ConfigMap{
--- a/pkg/verify/store/memory/memory.go
+++ b/pkg/verify/store/memory/memory.go
@@ -1,0 +1,36 @@
+// Package memory implements an in-memory signature store.  This is
+// mostly useful for testing.
+package memory
+
+import (
+	"context"
+
+	"github.com/openshift/cluster-version-operator/pkg/verify/store"
+)
+
+// Store provides access to signatures stored in memory.
+type Store struct {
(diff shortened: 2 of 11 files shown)


pkg/verify/store/configmap/configmap.go
```diff
@@ -87,10 +89,10 @@ func digestToKeyPrefix(digest string) (string, error) {
 	return fmt.Sprintf("%s-%s", algo, hash), nil
 }
 
-// DigestSignatures returns a list of signatures that match the request
+// Signatures returns a list of signatures that match the request
 // digest out of config maps labelled with ReleaseLabelConfigMap in the
 // openshift-config-managed namespace.
-func (s *Store) DigestSignatures(ctx context.Context, digest string) ([][]byte, error) {
+func (s *Store) Signatures(ctx context.Context, name string, digest string, fn store.Callback) error {
```
jottofar, line 95: Is `name` used anywhere?

pkg/verify/verify.go
```diff
@@ -348,11 +255,20 @@ func (v *ReleaseVerifier) cacheVerification(releaseDigest string, signedWith [][
...
-// checkFileSignatures reads signatures as "signature-1", "signature-2", etc out of a directory until
-// either the provided fn returns an error, false, or no such file exists. No more than maxSignaturesToCheck
-// will be read.
-func checkFileSignatures(ctx context.Context, dir string, maxSignaturesToCheck int, fn func(path string, signature []byte) (bool, error)) error {
-	base := filepath.Join(dir, "signature-")
+type fileStore struct {
+	directory string
+}
+
+// Signatures reads signatures as "signature-1", "signature-2", etc. out of a directory.
```
jottofar, line 262: `// Signatures reads signatures as "signature-1", "signature-2", etc. out of digest named sub-directory of a directory.`
or some such

pkg/verify/verify.go
```diff
@@ -348,11 +255,20 @@ func (v *ReleaseVerifier) cacheVerification(releaseDigest string, signedWith [][
...
+func (s *fileStore) Signatures(ctx context.Context, name string, digest string, fn store.Callback) error {
+	parts := strings.SplitN(digest, ":", 3)
+	if len(parts) != 2 || len(parts[0]) == 0 || len(parts[1]) == 0 {
+		return fmt.Errorf("the provided release image digest must be of the form ALGO:HASH")
+	}
+	algo, hash := parts[0], parts[1]
+	digestPathSegment := fmt.Sprintf("%s=%s", algo, hash)
```
jottofar, line 269: Make helper. Could be used by  configmap.digestToKeyPrefix too if separator passed in.

pkg/verify/store/memory/memory.go
```diff
@@ -0,0 +1,36 @@
+// Package memory implements an in-memory signature store.  This is
```
jottofar, line 1: This feels more like another Store implementation - fileStore, httpStore, memoryStore. It could then be used to test parallel and serial too.

pkg/verify/store/parallel/parallel.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Signatures fetches signatures for the provided digest.
+func (s *Store) Signatures(ctx context.Context, name string, digest string, fn store.Callback) error {
+	nestedCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	responses := make(chan signatureResponse, len(s.Stores))
+	errorChannelCount := 0
+	errorChannel := make(chan error, 1)
+
+	for i := range s.Stores {
```
jottofar, line 32: The parallelism is indirectly limited by current implementations of Store via maxSignatureSearch but do you think it needs an explicit limit here? If not perhaps just a comment.

pkg/verify/store/parallel/parallel.go
```diff
@@ -0,0 +1,92 @@
...
+			err := <-errorChannel
+			errorChannelCount--
+			if loopError == nil && err != nil && err != context.Canceled && err != context.DeadlineExceeded {
+				loopError = err
+			}
+		} else {
+			select {
+			case response := <-responses:
+				done, err := fn(ctx, response.signature, response.errIn)
+				if done || err != nil {
```
jottofar, line 59: The current implementation of fn always returns err nil but looking into the future why would we not want to continue trying to verify even if it did return an error? Hard to say without knowing the nature of the error whther it would be fruitless but I would think all we care about is that one of the signatures satisfies verification.

pkg/verify/store/memory/memory.go
wking: It is.  But this one is already its own sub-package, so `Store` is sufficiently scoped.  I'll move HTTP/file stores into sub-packages in follow-up work.

pkg/verify/store/parallel/parallel.go
wking: `maxSignatureSearch` limits the number of sig's a single HTTP or file store will return.  Parallelism here is just limited by the number of stores we feed the parallel store.  For OCP builds, that's currently hard-coded to two (GCS and the mirrors).

pkg/verify/store/parallel/parallel.go
wking: `fn` is a passed-in callback.  Are you taking issue with the parallel handling, or with the wrapped HTTP implementation?  This particular line is "if the caller is satisfied with this most-recent sig, or they considered `errIn` or something else a fatal error, we're done, and can wrap up the parallel handling".

pkg/verify/verify.go
wking: Reworded to:

```go
// Signatures reads signatures as "signature-1", "signature-2", etc. out of a digest-based subdirectory.
```

with 487d860 -> 6044c9c.

pkg/verify/verify.go
wking: > Make helper.

This is just five lines.  A helper might make sense now, but seems overcomplicated for [the future][1] where the HTTP and file stores will live in separate sub-packages.

[1]: https://github.com/openshift/cluster-version-operator/pull/393#discussion_r445916218

jottofar: I'm fine with that but thought I had gotten a similar comment from someone to use a helper in a similar situation (not you necessarily).

pkg/verify/store/parallel/parallel.go
jottofar: Oh, right, I see that now. I looked right pass the local Store stuct.

pkg/verify/store/parallel/parallel.go
jottofar: Ok, this is me again at the wrong "level", i.e. not at the local Store struct level, and thinking that if someone came along and implemented a different callback fn this if check would short-circuit checking the remaining signatures in the given store, e.g. httpStore, but that's not the case. 

So yeah, we're good here.

pkg/verify/store/memory/memory.go
jottofar: Yeah, I know it is a `Store` but `fileStore` and `httpStore` seem different than parallel and serial. The former and `memoryStore` being `Store`s and the later being "store managers" whereby the managers don't need a callback as they just invoke the underlying `Store`'s callback. Even though I know we currently and will probably continue to have only one callback implementation. Not advocating a major redo here, or even any changes, just explaining my line of thought.

PR #396, opened 2020-06-29
pkg/verify/store/sigstore: Factor HTTP store into its own package

wking:
Decluttering the root verify package and making it more store-agnostic.

I've also replaced the `ClientBuilder` interface with a single `HTTPClient` function type, which removes the need for `verifyClientBuilder` and other wrappers.  I'm also adding a rate-limiter around the client builder to avoid pounding away at that logic, even if it's mostly fed through informers (so should be no API-call reduction, but will save some local CPU cycles).

Spun off from my [earlier][1] 0cc3e41c1d04 as another nibbly bit (hopefully easy to review), now that the bulk of the earlier change has landed via #294.  Still to go after this change will be a similar sub-package for the file store, or deprecating the file store in favor of #294's memory store for unit testing.

[1]: https://github.com/openshift/cluster-version-operator/pull/279#discussion_r359239295

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -49,6 +49,7 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/verify/store"
 	"github.com/openshift/cluster-version-operator/pkg/verify/store/configmap"
 	"github.com/openshift/cluster-version-operator/pkg/verify/store/serial"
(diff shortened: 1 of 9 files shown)


wking:
/assign @jottofar 

Since he'd reviewed #393 and will be familiar with all of this.

PR #397, opened 2020-07-01
WIP: *: Discuss impact and suggested next steps in condition messages

wking:
Make these more actionable, so users don't have to track down our Git docs.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -20,7 +20,11 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/cincinnati"
 )
 
-const noChannel string = "NoChannel"
+const (
+	noChannel string = "NoChannel"
+
+	failedToRetrieveUpdatesImpact string = `Because the cluster cannot retrieve recommended updates, cluster administrators will need to monitor for available updates on their own or risk falling behind on security or other bugfixes (clear the channel if you want to monitor for updates yourself).`
+)
 
 // syncAvailableUpdates attempts to retrieve the latest updates and update the status of the ClusterVersion
 // object. It will set the RetrievedUpdates condition. Updates are only checked if it has been more than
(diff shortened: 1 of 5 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -143,9 +147,10 @@ func (optr *Operator) getAvailableUpdates() *availableUpdates {
 
 func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsConfig *tls.Config, upstream, arch, channel, version string) ([]configv1.Update, configv1.ClusterOperatorStatusCondition) {
 	if len(upstream) == 0 {
+		baseConsoleURI := FIXME()
 		return nil, configv1.ClusterOperatorStatusCondition{
 			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoUpstream",
-			Message: "No upstream server has been set to retrieve updates.",
+			Message: fmt.Sprintf("No upstream server has been set to retrieve updates.%s  Fix by setting spec.upstream in ClusterVersion to point to an update recommendation service, for example https://api.openshift.com/api/upgrades_info/v1/graph , %s/settings/cluster/", failedToRetrieveUpdatesImpact, baseConsoleURI),
```
vrutkovs, line 153: I don't think its worth pointing to a console, there are multiple ways how to update clusterversion - via `oc` or ArgoCD, so some might think the console is the only way to do that

wking: How do message-readers discover those ways?

vrutkovs: Probably a link to the docs would be best. Currently its split between [CLI](https://docs.openshift.com/container-platform/4.4/updating/updating-cluster-cli.html#update-upgrading-cli_updating-cluster-cli) and [web console](https://docs.openshift.com/container-platform/4.4/updating/updating-cluster.html#update-upgrading-web_updating-cluster)

wking: But also not available for some restricted-network clusters.  Still, docs links are easy, and we can pivot to console links later if there's interest.  Will reroll...

PR #398, opened 2020-07-01
Bug 1825396: upgrades with force should set verified to false

jottofar:


jottofar:
/hold
See discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1825396#c6 as to final bug resolution.

wking:
/retest

I'm curious about what happens to no-longer defined jobs like e2e-aws.

jottofar:
/close

Bug closed with a doc update.

PR #400, opened 2020-07-07
lib/resourcebuilder: Replace wait-for with single-shot "is it alive now?"

wking:
We've had `if updated` guards around `waitFor*Completion` since the library landed in 2d334c2190 (#10).  But, only waiting when `updated` is true is a weak block, because if/when we fail to complete, [`Task.Run` will back-off and call `builder.Apply` again][0].  That new `Apply` will see the already-updated object, set `updated` false, and not wait.  So whether we block or not is orthogonal to `updated`; nobody cares about whether the most recent update happened in this `builder.Apply`, this sync cycle, or a previous cycle.

We don't even care all that much about whether the Deployment, DaemonSet, CustomResourceDefinition, or Job succeeded.  Most feedback is going to come from the ClusterOperator, so with this commit we continue past the resource wait-for unless the resource is really hurting, in which case we fail immediately (inside `builder.Apply`, `Task.Run` will still hit us a few times) to bubble that up.  In situations where we don't see anything too terrible going on, we'll continue on past and later block on ClusterOperator not being ready.

Other changes in this commit:

* I've pulled `b.modifier(crd)` out of the switch, because that should happen regardless of the CRD version.
* I've added an error for unrecognized CRD version, because that will be easier to debug than trying to figure out why a CRD manifest is being silently ignored by the CVO.
* There's no object status for CRDs or DaemonSets that marks "we are really hurting".  The v1.18.0 Kubernetes CRD and DaemonSet controllers do not set any conditions in their operand status (although the API for those conditions exists [here][1] and [here][2]).  With this commit, we have very minimal wait logic for either.  Sufficiently unhealthy DaemonSet should be reported on via their associated ClusterOperator, and sufficiently unhealthy CRD should be reported on when we fail to push any custom resources consuming them (`Task.Run` retries will give the API server time to ready itself after accepting a CRD update before the CVO fails its sync cycle).
* `deploymentBuilder` and `daemonsetBuilder` grow `mode` properties.  They had been using `actual.Generation > 1` as a proxy for "post-install" since 14fab0b2c8 (#26), but generation 1 is just "we haven't changed the object since it was created", not "we're installing a fresh cluster".  For example, a new Deployment or DaemonSet could be added as part of a cluster update, and we don't want special install-time "we don't care about specific manifest failures" then.

[0]: https://github.com/openshift/cluster-version-operator/blob/e70a259d712975513cd365a6dea39b096ca4ed08/pkg/payload/task.go#L71-L75
[1]: https://github.com/kubernetes/api/blob/v0.18.0/apps/v1/types.go#L586-L590
[2]: https://github.com/kubernetes/apiextensions-apiserver/blob/v0.18.0/pkg/apis/apiextensions/types.go#L319-L320

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -12,7 +12,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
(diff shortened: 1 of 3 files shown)


smarterclayton:
Can you split the unrelated changes out to smaller separate PRs?  Rather review those individually

wking:
> Can you split the unrelated changes out to smaller separate PRs?

Split into #401 (`Generation > 1` -> `mode`), #402 (centralize `b.modifier(crd)`), and #403 (erroring on unrecognized CRD version).  Once #401 and #402 land, I'll rebase this one on top.  #403 may or may not count as a conflict; depends on how ornery GitHub is feeling ;).

abhinavdahiya:
> Can you split the unrelated changes out to smaller separate PRs? Rather review those individually

/uncc @abhinavdahiya

wking:
> Can you split the unrelated changes out to smaller separate PRs?

Rebased onto master around #401 (and dropping the rejected #402) with cebb958 -> c6afa88.  #403 is still waiting on review, but it's maybe far enough away to avoid getting marked as a conflict due to overlapping context.  And it will be easy enough to rebase this PR or #403 depending on which lands first if GitHub decides there is a conflict.

wking:
[Failures][1] [look][2] unrelated to this change.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/400/pull-ci-openshift-cluster-version-operator-master-e2e/1285620892390395905
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/400/pull-ci-openshift-cluster-version-operator-master-e2e-upgrade/1285620892390395904

wking:
Rebased around #412 with 507d91b -> 100231e.  No conflicts; I guess GitHub's conflict detector is just jumpier than Git's diff-context detector.

lib/resourcebuilder/apps.go
```diff
@@ -113,102 +111,68 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+	if progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse {
+		return &payload.UpdateError{
+			Nested:  fmt.Errorf("deployment %s is not progressing; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedReplicas, d.Status.Replicas, d.Status.AvailableReplicas, d.Status.Replicas),
+			Reason:  "WorkloadNotAvailable",
+			Message: fmt.Sprintf("deployment %s is not progressing %s: %s", iden, progressingCondition.Reason, progressingCondition.Message),
+			Name:    iden,
 		}
-		return err
 	}
+
```
jottofar, line 175: So with this change we don't care why we're not progressing, only that we're not, at this point in the process. I understand that to be in line with the overall goal of this change but would it not still be useful to post a log if we were in an `unknown state`? Would it be at all useful in debugging a deployment failure that bubbles up?

lib/resourcebuilder/batch.go
```diff
@@ -45,40 +44,51 @@ func (b *jobBuilder) Do(ctx context.Context) error {
...
-		return WaitForJobCompletion(ctx, b.client, job)
+	if b.mode != InitializingMode {
+		_, err := checkJobHealth(ctx, b.client, job)
+		return err
 	}
 	return nil
 }
 
 // WaitForJobCompletion waits for job to complete.
 func WaitForJobCompletion(ctx context.Context, client batchclientv1.JobsGetter, job *batchv1.Job) error {
```
jottofar, line 58: No longer used or needed?

lib/resourcebuilder/batch.go
```diff
@@ -45,40 +44,51 @@ func (b *jobBuilder) Do(ctx context.Context) error {
...
-		}
+// checkJobHealth returns an error if the job status is bad enough to block further manifest application.
+func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *batchv1.Job) (bool, error) {
+	j, err := client.Jobs(job.Namespace).Get(job.Name, metav1.GetOptions{})
+	if err != nil {
+		return false, fmt.Errorf("error getting Job %s: %v", job.Name, err)
+	}
 
-		// Since we have filled in "activeDeadlineSeconds",
-		// the Job will 'Active == 0' iff it exceeds the deadline.
```
jottofar, line 72: might as well fix `iff`

lib/resourcebuilder/batch.go
wking: > No longer used or needed?

We still need it; discussed in the last paragraph of 100231e7.

lib/resourcebuilder/batch.go
wking: Expanded to "if and only if" with 100231e -> c2af13b, which also rebases onto master to resolve conflicts with #403.

lib/resourcebuilder/apps.go
wking: > So with this change we don't care why we're not progressing, only that we're not...

Did we ever?

```console
$ git show --oneline -w c2af13b -- lib/resourcebuilder/apps.go
...
-func waitForDeploymentCompletion(ctx context.Context, client appsclientv1.DeploymentsGetter, deployment *appsv1.Deployment) error {
+
+func checkDeploymentHealth(ctx context.Context, client appsclientv1.DeploymentsGetter, deployment *appsv1.Deployment) error {
        iden := fmt.Sprintf("%s/%s", deployment.Namespace, deployment.Name)
-       var lastErr error
-       err := wait.PollImmediateUntil(defaultObjectPollInterval, func() (bool, error) {
...
-               if d.Generation <= d.Status.ObservedGeneration && d.Status.UpdatedReplicas == d.Status.Replicas && d.Status.UnavailableReplicas == 0 {
-                       return true, nil
...
```

The main changes I'm making here are removing the `PollImmediateUntil` (because we are no longer waiting, per the single-shot motivation in the PR topic) and removing this `UpdatedReplicas` check (because we no longer care if the Deployment has updated or not.  It's up to ClusterOperator to tell us if/when the component has leveled, as discussed in the PR topic).

Or are you asking specifically about the `progressingCondition == nil` case or the unrecognized `progressingCondition.Status` cases?  I think the new logic is about "if we see something we know is bad; block" and that not seeing things we think are good is no longer sufficient to block.  I'm fine logging them if we are concerned that will introduce mysterious bugs.

jottofar: My comment is a long-winded roundabout to whether the `unknown state` logging should be left as something that could be helpful since `unknown state` sounds rather scary to me. Since you're more familiar with the history of this beast I'm fine with leaving it to your judgement.

wking:
Rebased around #406's `Context` additions with c2af13b2 -> 9be62c45.

lib/resourcebuilder/apps.go
wking: Ok, I've pushed 9be62c45 -> cc9292a0 adding something sort of like the old `unknown state` logging.  Discussion in a new paragraph in the commit message of cc9292a0.

wking:
[`e2e`][1] failure is just in `an end user can use OLM can subscribe to the operator`, which is blowing up all over the place.

/override ci/prow/e2e

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/400/pull-ci-openshift-cluster-version-operator-master-e2e/1288951856990523392

PR #401, opened 2020-07-08
lib/resourcebuilder/apps: Explicit mode instead of 'Generation > 1'

wking:
`deploymentBuilder` and `daemonsetBuilder` grow `mode` properties.  They had been using `actual.Generation > 1` as a proxy for "post-install" since 14fab0b2c8 (#26), but generation 1 is just "we haven't changed the object since it was created", not "we're installing a fresh cluster".  For example, a new Deployment or DaemonSet could be added as part of a cluster update, and we don't want special install-time "we don't care about specific manifest failures" then.

Spun out from #400 at @smarterclayton's [request](https://github.com/openshift/cluster-version-operator/pull/400/#issuecomment-655152378).

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -26,6 +26,7 @@ type deploymentBuilder struct {
 	proxyGetter configv1.ProxiesGetter
 	raw         []byte
 	modifier    MetaV1ObjectModifierFunc
+	mode        Mode
 }
 
 func newDeploymentBuilder(config *rest.Config, m lib.Manifest) Interface {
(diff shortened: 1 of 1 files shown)


wking:
Service disruption

/retest

wking:
/override e2e-gcp-upgrade

We don't need that anymore, it's now the genrically-named e2e-upgrade.

sdodson:
> /override e2e-gcp-upgrade
> 
> We don't need that anymore, it's now the genrically-named e2e-upgrade.

I think `/skip` is what's meant to be used when tests are removed or made non blocking but I guess that worked too.

wking:
I tried `/skip` up [here][1], but maybe that was before the `e2e-gcp` -> `e2e` change.

[1]: https://github.com/openshift/cluster-version-operator/pull/401#issuecomment-656456038

PR #402, opened 2020-07-08
lib/resourcebuilder/apiext: Centralize b.modifier(crd) call

wking:
No need to repeat this for every version, which is how we've had things since 4ee7b07e99 (#259).  Spun off from #400 at @smarterclayton's [request](https://github.com/openshift/cluster-version-operator/pull/400/#issuecomment-655152378).

### DIFF
--- a/lib/resourcebuilder/apiext.go
+++ b/lib/resourcebuilder/apiext.go
@@ -45,24 +45,22 @@ func (b *crdBuilder) WithModifier(f MetaV1ObjectModifierFunc) Interface {
 func (b *crdBuilder) Do(ctx context.Context) error {
 	crd := resourceread.ReadCustomResourceDefinitionOrDie(b.raw)
 
+	if b.modifier != nil {
+		b.modifier(crd)
+	}
+
 	var updated bool
 	var err error
 	var name string
 
 	switch crd := crd.(type) {
 	case *apiextv1beta1.CustomResourceDefinition:
-		if b.modifier != nil {
-			b.modifier(crd)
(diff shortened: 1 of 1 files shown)


wking:
/close

Nevermind.  [Unit tests][1] fail with:

```console
# github.com/openshift/cluster-version-operator/lib/resourcebuilder [github.com/openshift/cluster-version-operator/lib/resourcebuilder.test]
lib/resourcebuilder/apiext.go:49:13: cannot use crd (type "k8s.io/apimachinery/pkg/runtime".Object) as type "k8s.io/apimachinery/pkg/apis/meta/v1".Object in argument to b.modifier:
	"k8s.io/apimachinery/pkg/runtime".Object does not implement "k8s.io/apimachinery/pkg/apis/meta/v1".Object (missing GetAnnotations method)
```

Because 4ee7b07 also replaced the structured:

```go
func ReadCustomResourceDefinitionV1Beta1OrDie(objBytes []byte) *apiextv1beta1.CustomResourceDefinition {
```

with the less-structured:

```go
func ReadCustomResourceDefinitionOrDie(objBytes []byte) runtime.Object {
```

We can maybe pivot to support for just the v1 CRD types, but that would be a different PR.

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/402/pull-ci-openshift-cluster-version-operator-master-unit/1280668446677274624#1:build-log.txt%3A20

PR #404, opened 2020-07-09
Bug 1907329: Add cluster profile support

guillaumerose:
Implementation of https://github.com/openshift/enhancements/pull/200

This adds a CLUSTER_PROFILE env variable that is used to determine which manifests to apply based on the `include.release.openshift.io/` annotation.

It is exactly what is written in the enhancement, not more.

This env. variable is not used by any components and, as it is empty, it defaults to the value of the default profile `self-managed-high-availability`.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -148,6 +148,8 @@ type Operator struct {
 	// exclude is an optional identifier used to exclude certain manifests
 	// via annotation
 	exclude string
+
+	clusterProfile string
 }
 
 // New returns a new cluster version operator.
@@ -166,6 +168,7 @@ func New(
 	client clientset.Interface,
 	kubeClient kubernetes.Interface,
 	exclude string,
+	clusterProfile string,
 ) *Operator {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(klog.Infof)
@@ -194,7 +197,8 @@ func New(
 		availableUpdatesQueue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "availableupdates"),
 		upgradeableQueue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "upgradeable"),
 
-		exclude: exclude,
+		exclude:        exclude,
+		clusterProfile: clusterProfile,
 	}
 
 	cvInformer.Informer().AddEventHandler(optr.eventHandler())
@@ -219,7 +223,7 @@ func New(
 // controller that loads and applies content to the cluster. It returns an error if the payload appears to
 // be in error rather than continuing.
 func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestConfig *rest.Config) error {
-	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.release.Image, optr.exclude)
+	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.release.Image, optr.exclude, optr.clusterProfile)
 	if err != nil {
 		return fmt.Errorf("the local release contents are invalid - no current version can be determined from disk: %v", err)
 	}
@@ -261,6 +265,7 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 		},
 		optr.exclude,
 		optr.eventRecorder,
+		optr.clusterProfile,
 	)
 
 	return nil
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -78,6 +78,7 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 		cvLister:                    &clientCVLister{client: client},
 		exclude:                     "exclude-test",
 		eventRecorder:               record.NewFakeRecorder(100),
+		clusterProfile:              payload.DefaultClusterProfile,
 	}
 
 	dynamicScheme := runtime.NewScheme()
@@ -94,6 +95,7 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 		},
 		"exclude-test",
 		record.NewFakeRecorder(100),
+		o.clusterProfile,
 	)
 	o.configSync = worker
 
@@ -228,7 +230,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Generation:  1,
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -241,7 +243,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Fraction:    float32(1) / 3,
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -255,7 +257,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Fraction:    float32(2) / 3,
 			Initial:     true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -269,7 +271,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Reconciling: true,
 			Completed:   1,
 			Fraction:    1,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -312,7 +314,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 				URL:      configv1.URL("https://example.com/v1.0.0-abc"),
 				Channels: []string{"channel-a", "channel-b", "channel-c"},
 			},
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			History: []configv1.UpdateHistory{
 				// Because image and operator had mismatched versions, we get two entries (which shouldn't happen unless there is a bug in the CVO)
 				{State: configv1.CompletedUpdate, Image: "image/image:1", Version: "1.0.0-abc", StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
@@ -334,7 +336,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Generation:  1,
 			Reconciling: true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -347,7 +349,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Reconciling: true,
 			Fraction:    float32(1) / 3,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -360,7 +362,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Reconciling: true,
 			Fraction:    float32(2) / 3,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -373,7 +375,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Reconciling: true,
 			Completed:   2,
 			Fraction:    1,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -542,7 +544,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 		SyncWorkerStatus{
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -555,7 +557,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Fraction:    float32(1) / 3,
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -569,7 +571,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Fraction:    float32(2) / 3,
 			Initial:     true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -583,7 +585,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Reconciling: true,
 			Completed:   1,
 			Fraction:    1,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -627,7 +629,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 				URL:      configv1.URL("https://example.com/v1.0.0-abc"),
 				Channels: []string{"channel-a", "channel-b", "channel-c"},
 			},
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			History: []configv1.UpdateHistory{
 				// Because image and operator had mismatched versions, we get two entries (which shouldn't happen unless there is a bug in the CVO)
 				{State: configv1.CompletedUpdate, Image: "image/image:1", Version: "1.0.0-abc", StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
@@ -648,7 +650,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 		SyncWorkerStatus{
 			Reconciling: true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -661,7 +663,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Reconciling: true,
 			Fraction:    float32(1) / 3,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -674,7 +676,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Reconciling: true,
 			Fraction:    float32(2) / 3,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -687,7 +689,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Reconciling: true,
 			Completed:   2,
 			Fraction:    1,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -847,7 +849,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 		SyncWorkerStatus{
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -860,7 +862,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 			Fraction:    float32(1) / 3,
 			Step:        "ApplyResources",
 			Initial:     true,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -874,7 +876,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 			Fraction:    float32(2) / 3,
 			Initial:     true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -888,7 +890,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 			Reconciling: true,
 			Completed:   1,
 			Fraction:    1,
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
@@ -932,7 +934,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 				URL:      configv1.URL("https://example.com/v1.0.0-abc"),
 				Channels: []string{"channel-a", "channel-b", "channel-c"},
 			},
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			History: []configv1.UpdateHistory{
 				// Because image and operator had mismatched versions, we get two entries (which shouldn't happen unless there is a bug in the CVO)
 				{State: configv1.CompletedUpdate, Image: "image/image:1", Version: "1.0.0-abc", StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
@@ -953,7 +955,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 		SyncWorkerStatus{
 			Reconciling: true,
 			Step:        "ApplyResources",
-			VersionHash: "6GC9TkkG9PA=",
+			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
 				Image:    "image/image:1",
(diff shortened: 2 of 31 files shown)


vrutkovs:
AWS failed to bootstrap

/retest

guillaumerose:
@vrutkovs yes it is because I didn't solve the issue of "where we should store the profile". Currently, the CVO expects to find it in a ConfigMap and it is still created by hand. Do you think it's a good idea to read it from a ConfigMap ? There is a chicken and egg issue with this: if the ConfigMap is not present, the CVO cannot run but the ns openshift-config is created by the CVO.

vrutkovs:
CVO should fall back to default profile (so that upgrade would succeed) if the configmap is not present

guillaumerose:
OK, so I will change this PR to:
* load the ConfigMap as soon as the CVO starts.
* if the ConfigMap doesn't exist, use the default profile.

It will be the responsibility of the installer to have the ConfigMap in place before the CVO runs for the first time. Does it sounds good ?

guillaumerose:
I experimented a bit more and found an interesting problem by making optional the ConfigMap.

During the boostrap, the CVO is created via a bootstrap-manifest and it runs before manifests are created (located in /manifests - even before the first one 0000_00_cluster-version-operator_00_namespace.yaml). It could not find the ConfigMap and therefore load the default profile.
Couple of minutes later, bootstrap ends, the ConfigMap is in place, the CVO reboot in the master and it's fine. But there was a period of time where the profile was not the good one. 

Any thoughts on this ? How can we make the CVO waiting for the ConfigMap but still having the choice to not have it ?
Isn't it a problem to have the CVO in the bootstrap node not running the right profile ?

vrutkovs:
>But there was a period of time where the profile was not the good one.

I'm not sure I'm following - why installed profile is incorrect? I assumed that no profile specified == "default" profile. If its not set then all operators (unless explicitly excluded) are enabled

guillaumerose:
I wanted to change the behavior to load once the profile at startup but it doesn't work. CVO is launched before manifests are created. At start the ClusterVersion object doesn't exist and locks the CVO. 

To work correctly with a profile, the user have to ensure the ConfigMap is created before the ClusterVersion. For snc, I did something like:

```
openshift-install create manifests

cp myns.yaml ${INSTALL_DIR}/openshift/0000_00_cluster-version-operator_00_myns.yaml
cp myconfigmap.yaml ${INSTALL_DIR}/openshift/0000_00_cluster-version-operator_00_myconfigmap.yaml

openshift-install create cluster
```

pkg/cvo/sync_worker.go
```diff
@@ -58,6 +58,10 @@ type PayloadRetriever interface {
 	RetrievePayload(ctx context.Context, desired configv1.Update) (PayloadInfo, error)
 }
 
+type ClusterProfileRetriever interface {
```
csrwng, line 61: add godoc

pkg/cvo/sync_worker.go
```diff
@@ -58,6 +58,10 @@ type PayloadRetriever interface {
...
+	RetrieveClusterProfile(ctx context.Context) string
```
csrwng, line 62: You should return an error in this function. It's probably safer to fail in the case you can't read the profile than to assume "default"

pkg/cvo/updatepayload.go
```diff
@@ -315,3 +317,22 @@ func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string
...
+}
+
+type clusterProfileRetriever struct {
+	cmConfigLister listerscorev1.ConfigMapNamespaceLister
+}
+
+func (r *clusterProfileRetriever) RetrieveClusterProfile(_ context.Context) string {
+	configMap, err := r.cmConfigLister.Get(internal.ClusterProfileConfigMap)
+	if err != nil {
+		klog.Infof("cannot get configmap openshift-config/cluster-profile, using default profile")
```
csrwng, line 334: see comment for the interface

pkg/cvo/sync_worker.go
csrwng: Just to clarify, you can check the error for IsNotFound and return default... however other errors probably indicate an api server error and you should probably report.

pkg/cvo/sync_worker.go
guillaumerose: done :+1:

pkg/cvo/sync_worker.go
guillaumerose: yes you're right, it's better now.

pkg/cvo/updatepayload.go
guillaumerose: done :+1:

pkg/cvo/sync_worker.go
```diff
@@ -58,6 +58,12 @@ type PayloadRetriever interface {
 	RetrievePayload(ctx context.Context, desired configv1.Update) (PayloadInfo, error)
 }
 
+// ClusterProfileRetriever abstracts how the cluster profile is retrieved.
+type ClusterProfileRetriever interface {
+	// RetrieveClusterProfile attempts to retrieve the profile
+	RetrieveClusterProfile(ctx context.Context) (string, error)
```
wking, line 64: I'm not a huge fan of `interface` that wrap a single function.  And I think the risk of needing to grow a new function in the future is small.  Can we make this:

```go
// RetrieveClusterProfile retrieves the cluster profile.
type RetrieveClusterProfile func(ctx context.Context) (string, error)
```

That would allow us to drop `fakeClusterProfileRetriever` and just inline `func (ctx context.Context) {return "default", nil}` where we needed a dummy.

pkg/cvo/sync_worker.go
wking: a84ec7a1e1 has a godoc; marking this thread resolved.

pkg/cvo/sync_worker.go
```diff
@@ -505,8 +515,14 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 			return err
 		}
 
+		profile, err := w.clusterProfileRetriever.RetrieveClusterProfile(ctx)
+		if err != nil {
+			klog.Warningf("failed to retrieve profile: %v", err)
```
wking, line 520: I don't think we should `klog` here and then return the `err`.  The `syncOnce` caller is in charge of logging and otherwise announcing the error, as it does [here][1].  Can we drop this line?

[1]: https://github.com/openshift/cluster-version-operator/blob/a49fef5c66c6b0707c54fd93f84d2f51d3d28aca/pkg/cvo/sync_worker.go#L336

pkg/cvo/updatepayload.go
```diff
@@ -315,3 +318,25 @@ func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string
...
+
+type clusterProfileRetriever struct {
+	cmConfigLister listerscorev1.ConfigMapNamespaceLister
+}
+
+func (r *clusterProfileRetriever) RetrieveClusterProfile(_ context.Context) (string, error) {
+	configMap, err := r.cmConfigLister.Get(internal.ClusterProfileConfigMap)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Cannot get ConfigMap openshift-config/cluster-profile, using default profile: %v", err)
```
wking, line 336: Instead of hard-coding `openshift-config/cluster-profile` here, can we use `%s/%s`, `ConfigNamespace`, and `ClusterProfileConfigMap`?

pkg/internal/constants.go
```diff
@@ -1,7 +1,8 @@
 package internal
 
 const (
-	ConfigNamespace    = "openshift-config"
-	InstallerConfigMap = "openshift-install"
-	ManifestsConfigMap = "openshift-install-manifests"
+	ConfigNamespace         = "openshift-config"
+	InstallerConfigMap      = "openshift-install"
+	ManifestsConfigMap      = "openshift-install-manifests"
+	ClusterProfileConfigMap = "cluster-profile"
```
wking, line 7: There are a few comments in the PR about if/how to use a ConfigMap to store this config (which was [out-of-scope][1] in the enhancement), starting [here][2].  Can we move them to this thread so they don't get interleaved with other PR-scoped comments?  I feel like managing this config may deserve its own enhancement :p.  Is this intended to be a install-time config?  What happens if admins twiddle the value post-install?  What happens if they set an invalid value (seems like the CVO blindly accepting an invalid profile would effectively cripple manifest application)?  Should we be reporting the configured profile through metrics/telemetry?  My initial knee-jerk is that this should be a ClusterVersion `status` property, to dodge some of these mutable-location wrinkles, but we're unlikely to be able to alter ClusterVersion without a formal enhancement.  Thoughts?

[1]: https://github.com/openshift/enhancements/pull/200#discussion_r394654607
[2]: https://github.com/openshift/cluster-version-operator/pull/404#issuecomment-662516200

pkg/payload/payload_test.go
```diff
@@ -123,3 +123,73 @@ func mustRead(path string) []byte {
...
+			annotations: map[string]interface{}{"exclude.release.openshift.io/identifier": "true"},
+		},
+		{
+			exclude:     "identifier",
+			annotations: map[string]interface{}{"exclude.release.openshift.io/identifier": "true"},
+			isExcluded:  true,
+		},
+		{
+			profile:     "",
+			annotations: map[string]interface{}{"include.release.openshift.io/edge": "true"},
```
wking, line 149: this is a weird test-case.  I'd expect an empty-string profile to cause `shouldExclude` to return an error (which would require us to update its call signature) or panic or some-such.  If we happen to get into that situation, loading a bunch of competing manifests from different profiles does not seem like something that would work well.

pkg/payload/payload_test.go
```diff
@@ -123,3 +123,73 @@ func mustRead(path string) []byte {
...
+						"annotations": tt.annotations,
+					},
+				},
+			},
+		})
+		if ret != tt.isExcluded {
+			t.Errorf("(exclude: %v, profile: %v, annotations: %v) %v != %v", tt.exclude, tt.profile, tt.annotations, tt.isExcluded, ret)
+		}
+	}
+
```
wking, line 194: nit: useless empty space.  Can we drop this line?

pkg/payload/payload_test.go
guillaumerose: Yes it's a problem.
At the initialization of the CVO, it loads the payload to get the verifier ConfigMap (https://github.com/openshift/cluster-version-operator/pull/404/files#diff-85c31beb4341b4c52f892c0581bbb5d6R226) but don't have access to the profile. At this point of the install, the CVO container is mostly alone and no resources are present. Either we wait for the ConfigMap (but it's optional, so how can we do that ?) or we load all or default manifests and take the first verifier.
In the sync loop, it works correctly because the ClusterVersion object acts as a lock and the ConfigMap cluster-profile is created before.

pkg/internal/constants.go
guillaumerose: For CRC, it is definitely an install-time config and we don't want the admin to modify it.
In this PR, the ClusterVersion object is key and the ConfigMap with the profile must be created before (as the CV is a lock for the sync loop). The admin have to insert the correct manifest between the `openshift-install create manifests` and `openshift-install create cluster` phases.  
I must admit it would cleaner if we can have a profile property in the CV like CVO overrides.
Is this first version acceptable still? WDYT?

pkg/cvo/sync_worker.go
guillaumerose: done :+1:

pkg/cvo/sync_worker.go
guillaumerose: done :+1:

pkg/cvo/updatepayload.go
guillaumerose: done :+1:

pkg/payload/payload_test.go
guillaumerose: done :+1:

pkg/payload/payload_test.go
wking: >  Either we wait for the ConfigMap (but it's optional, so how can we do that ?) or we load all or default manifests and take the first verifier.

"verifier"?  The enhancement line I comment on [here][1] suggests the no-configured-profile behavior should be "exclude manifests with any profile annotations".  That also mitigates the config-loading race, by pushing no profile-specific manifests until a profile config has been found.

[1]: https://github.com/openshift/enhancements/pull/200#discussion_r457408715

pkg/payload/payload.go
```diff
@@ -188,10 +189,38 @@ func LoadUpdate(dir, releaseImage, excludeIdentifier string) (*Update, error) {
...
+		return false
+	}
+
+	excludeAnnotation := fmt.Sprintf("exclude.release.openshift.io/%s", excludeIdentifier)
+	if annotations[excludeAnnotation] == "true" {
+		return true
+	}
+
+	if profile == "" {
+		return false
```
wking, line 204: Per [this][1], we should move the empty-profile check down and have:

```go
if !shouldCheckProfile{
  return false // include manifests which declare no profiles
}

if profile == "" {
  return true // exclude manifests which declare profiles when no profile is configured
}

profileAnnotation := fmt.Sprintf("include.release.openshift.io/%s", profile)
...
```


[1]: https://github.com/openshift/cluster-version-operator/pull/404#discussion_r461551007

pkg/cvo/updatepayload.go
```diff
@@ -315,3 +317,15 @@ func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string
 	}
 	return configv1.Update{}, false
 }
+
+func (optr *Operator) clusterProfileRetriever() (string, error) {
+	configMap, err := optr.cmConfigLister.Get(internal.ClusterProfileConfigMap)
```
wking, line 322: This needs to check for [the `CLUSTER_PROFILE` environment variable][1] before looking at ConfigMaps or other locations.

[1]: https://github.com/openshift/enhancements/pull/200/files#diff-b0d3a9e089e5cf5a125a408f84f934abR67

pkg/cvo/updatepayload.go
```diff
@@ -315,3 +317,15 @@ func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Cannot get ConfigMap %s/%s, using default profile: %v", internal.ConfigNamespace, internal.ClusterProfileConfigMap, err)
+			return "default", nil
```
wking, line 326: Per [this][1], I think we shoulf `return "", nil` here.

[1]: https://github.com/openshift/cluster-version-operator/pull/404#discussion_r461555262

pkg/cvo/updatepayload.go
```diff
@@ -315,3 +317,15 @@ func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string
 	}
 	return configv1.Update{}, false
 }
+
+func (optr *Operator) clusterProfileRetriever() (string, error) {
+	configMap, err := optr.cmConfigLister.Get(internal.ClusterProfileConfigMap)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Cannot get ConfigMap %s/%s, using default profile: %v", internal.ConfigNamespace, internal.ClusterProfileConfigMap, err)
```
wking, line 325: No need to format `err` in here.  We already know it's not-found and we can express that in English like `ConfigMap %s/%s not found, excluding all profiles`.

pkg/payload/payload_test.go
guillaumerose: The verifier is created at this point using an update: https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/cvo.go#L245.
This is done once at the initialization of the CVO, so we can't wait for the profile to appear I suppose.

wking: Why would you need profile-specific verification?

wking: > This is done once at the initialization of the CVO...

We probably need to cache the profile and reload from disk if it's changed since the last sync cycle.

guillaumerose: > Why would you need profile-specific verification? 

You're right. I don't want it. 
At initialization, we can load default profile manifests and take the verifier. It removes this empty string profile and simplify `shouldExclude`. I will give it a try a do that.

> We probably need to cache the profile and reload from disk if it's changed since the last sync cycle.

Cache the profile is I think a good thing. Can the informer notify when a change appear?

wking: > Can the informer notify when a change appear?

The ConfigMap informer could, but it would make a generic interface/getter more awkward to expose that.  I'd rather just poll once at the start of each sync cycle.

pkg/cvo/updatepayload.go
guillaumerose: done

pkg/cvo/updatepayload.go
guillaumerose: done! Now there is no more mention of the default profile, it's simpler.

pkg/cvo/updatepayload.go
guillaumerose: added!

pkg/payload/payload.go
guillaumerose: done

guillaumerose: I just realized that with this change we can only add more manifests but not replace or deactivate default manifests. Let's say I want a different image for the ingress operator. One manifest will have the include annotation, the other will not have it. If I run the CVO with the profile, the 2 manifests will be applied.
How can we avoid this ?

pkg/cvo/updatepayload.go
LalatenduMohanty: @guillaumerose Where is the env variable?

guillaumerose: Here https://github.com/openshift/cluster-version-operator/pull/404/files#diff-4f43e557624e0fc47bc4e31f4f236ab5R99

LalatenduMohanty: I do not see the env viariable check before looking at ConfigMaps

pkg/internal/constants.go
LalatenduMohanty: @guillaumerose Why do you think CRC need it as install-time config? To me it should work for CRC even if you provide the ConfigMap post the first installation.

pkg/cvo/updatepayload.go
guillaumerose: It's weird it doesn't appear here. If I browse the `Files changed` tab I can see it: https://github.com/openshift/cluster-version-operator/pull/404/files#diff-d540a41404f678a1c438f4c1e5b92a87R322-R325

pkg/internal/constants.go
guillaumerose: It should work post-install but it will take extra time (10-20min?) to install unnecessary components. It can also have weird side effects. It would be great to guaranty that a profile is not dependent of the default profile.

LalatenduMohanty: Any change to API also impact to multiple components (Installer, CVO) should be an enhancement https://github.com/openshift/enhancements#is-my-thing-an-enhancement

guillaumerose: To move this PR forward, I propose to:
* remove the use of the ConfigMap,
* only use the env. variable (as stated in the enhancement).
* open a PR in the installer to add the env variable.

WDYT?

pkg/payload/payload.go
wking: > Let's say I want a different image for the ingress operator...

Open a new PR to adjust the enhancement and say:

> Manifests with no explicit inclusion annotations implicitly belong to the `default` profile.  If no profile is configured, the cluster-version operator should use the `default` profile.

pkg/internal/constants.go
wking: You'd need to adjust the `render` logic here and also pur default CVO Deployment templating to propagate the env var between updates.  But sounds plausible to me.

pkg/payload/payload.go
guillaumerose: PR opened https://github.com/openshift/enhancements/pull/414

pkg/internal/constants.go
guillaumerose: I updated the PR with only the env variable and the flag. It's now clearer although the flag `--cluster-profile` is a little bit invasive. Let me know what you think.

cmd/render.go
```diff
@@ -39,7 +41,10 @@ func runRenderCmd(cmd *cobra.Command, args []string) {
 	if renderOpts.releaseImage == "" {
 		klog.Fatalf("missing --release-image flag, it is required")
 	}
-	if err := payload.Render(renderOpts.outputDir, renderOpts.releaseImage); err != nil {
+	if renderOpts.clusterProfile == "" {
+		klog.Fatalf("empty --cluster-profile flag, it is required")
```
wking, line 45: It should default to `default`, not be required.  And I see you set `default` above when declaring the `StringVar`.  So the error message should probably be `"invalid empty --cluster-profile value"` or some such.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -94,6 +94,7 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 		},
 		"exclude-test",
 		record.NewFakeRecorder(100),
+		"default",
```
wking, line 97: This should get set in `&Operator` above and then passed through here.

cmd/start.go
```diff
@@ -32,5 +32,6 @@ func init() {
 	cmd.PersistentFlags().StringVar(&opts.ReleaseImage, "release-image", opts.ReleaseImage, "The Openshift release image url.")
 	cmd.PersistentFlags().StringVar(&opts.ServingCertFile, "serving-cert-file", opts.ServingCertFile, "The X.509 certificate file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file, or neither.")
 	cmd.PersistentFlags().StringVar(&opts.ServingKeyFile, "serving-key-file", opts.ServingKeyFile, "The X.509 key file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file, or neither.")
+	cmd.PersistentFlags().StringVar(&opts.ClusterProfile, "cluster-profile", opts.ClusterProfile, "Select a different deployment model")
```
wking, line 35: `start.NewOptions()` should be defaulting this to `default`.  But I'm not seeing that code anywhere.  Am I just missing it?

pkg/payload/payload.go
```diff
@@ -188,10 +189,34 @@ func LoadUpdate(dir, releaseImage, excludeIdentifier string) (*Update, error) {
...
+
+	excludeAnnotation := fmt.Sprintf("exclude.release.openshift.io/%s", excludeIdentifier)
+	if annotations[excludeAnnotation] == "true" {
+		return true
+	}
+
+	shouldCheckProfile := false
+	for k := range annotations {
+		if strings.HasPrefix(k, "include.release.openshift.io/") {
+			shouldCheckProfile = true
```
wking, line 206: simpler implementation might be:

```go
prefix := "include.release.openshift.io/"
exists := struct{}{}
profiles := map[string]struct{}{}
for key, value := range annotations {
  if strings.HasPrefix(key, prefix) && value == "true" {
    profiles[key[len(prefix):]] = exists
  }
}

if len(profiles) == 0 {
  profiles["default"] = exists
}

_, belongsInProfile := profiles[profile]
return !belongsInProfile
```

cmd/start.go
wking: Never mind, found it with the `defaultClusterProfile` stuff.

pkg/payload/payload_test.go
```diff
@@ -123,3 +123,52 @@ func mustRead(path string) []byte {
...
+			profile:     "crc",
+			annotations: map[string]interface{}{"include.release.openshift.io/edge": "true"},
+			isExcluded:  true,
+		},
+		{
+			profile:     "edge",
+			annotations: map[string]interface{}{"include.release.openshift.io/edge": "true"},
+		},
+		{
+			profile: "default",
```
wking, line 154: add a test case where `profile: "default"` shows us excluding `annotations: map[string]interface{}{"include.release.openshift.io/edge": "true"}`, i.e. that the manifest only belongs implicitly to `default` if it declares no explicit profile inclusion.

pkg/payload/payload_test.go
```diff
@@ -123,3 +123,52 @@ func mustRead(path string) []byte {
...
+		},
+		{
+			profile: "edge",
```
wking, line 157: this should exclude the manifest, because the manifest is implicitly a member of the `default` profile, and not a member of the `edge` profile.

guillaumerose: This case ensures we don't have to duplicate all manifests for all profiles. We can probably change the enhancemen to say 
`Manifests with no explicit inclusion annotations implicitly belong to all profiles. `.

pkg/cvo/cvo_scenarios_test.go
guillaumerose: done :+1:

pkg/payload/payload_test.go
guillaumerose: done :+1:

cmd/render.go
guillaumerose: done :+1:

bootstrap/bootstrap-pod.yaml
```diff
@@ -17,6 +17,7 @@ spec:
       - "--enable-default-cluster-version=false"
       - "--v=4"
       - "--kubeconfig=/etc/kubernetes/kubeconfig"
+      - "--cluster-profile={{.ClusterProfile}}"
```
guillaumerose, line 20: We might want to add a condition like this. It will prevent to add the default value in the manifest and spread this flag too much.
``` 
{{if ne .ClusterProfile "default" }}
      - "--cluster-profile={{.ClusterProfile}}"
{{end}}
```
WDYT @wking ?

wking: I'm fine as this stands.  One less place to hard-code `default`

pkg/payload/payload_test.go
wking: > This case ensures we don't have to duplicate all manifests for all profiles.

David [wants][1] that legwork to be a new-profile entry barrier.

[1]: https://github.com/openshift/enhancements/pull/200#discussion_r375837903

bbrowning:
This looks like a real failure in that e2e-upgrade test:
```
fail [github.com/openshift/origin@/test/e2e/upgrade/upgrade.go:139]: during upgrade to registry.svc.ci.openshift.org/ci-op-jstn9hsr/release@sha256:64f525746fcc9663be6c4b70bb4e4d98d9f8d57fa5db6c90b0de8072869ce761
Unexpected error:
    <*errors.errorString | 0xc00208be10>: {
        s: "Cluster did not complete upgrade: timed out waiting for the condition: Error loading manifests from /etc/cvo/updatepayloads/QGnAQsjGa670WqA9XFnWRg: error running preprocess on 0000_00_cluster-version-operator_03_deployment.yaml: failed to execute template: template: manifest:32:33: executing \"manifest\" at <.ClusterProfile>: can't evaluate field ClusterProfile in type payload.manifestRenderConfig",
    }
```

It's specifically hitting issues around the `ClusterProfile` field.

bbrowning:
My reading of the code and upgrade error here is that the previous version of the CVO needs to have the `ClusterProfile` field added to the manifestRenderConfig struct BEFORE that field is ever used in an actual manifest. During an upgrade, it looks like the existing CVO reads the upgrade payload to validate it and it's failing because the existing CVO doesn't know how to handle this new `ClusterProfile` field.

I think that means there are a couple of options:
1. Release this change over two releases, where the first release adds the new `ClusterProfile` field but does not reference it in any manifest. Then, in a later release, it can be referenced from upgrade manifests.
2. Release this change in one release, but explicitly do NOT reference this field in any manifest. That means the logic of what profile to default to when none is specified may have to get spread around a bit more.

I'm not super familiar with the mechanics of how CVO operates here, and the above is just based on a quick reading of the failing test and code. Perhaps someone else in this thread can verify the above is correct.

guillaumerose:
Thanks @bbrowning, I didn't see the failure.

I guess we have 3 options:
* Continue with the env. variable/new parameter. First release, we define the new variable in the templating, second release we use it. It still requires some changes in the installer.
* Use a ConfigMap. No changes are needed in the installer but an enhancement PR should define it.
* Add a new field in ClusterVersion. We would need also an enhancement PR.

WDYT @abhinavdahiya @wking ?

derekwaynecarr:
Is there an enhancement that references how CRC will use cluster profiles?  The IBM Cloud managed service case is clear since its already a production service and the alternative/excluded manifests in that setting is clearly understood.

LalatenduMohanty:
@guillaumerose I guess we are still planning for this PR in 4.7, right?

guillaumerose:
@LalatenduMohanty I guess, yes. The major issue is that I/we don't know how we should implement it. We need to make a choice between the 3 solutions here https://github.com/openshift/cluster-version-operator/pull/404#issuecomment-668557003 of if someone has an other idea.

rkukura:
> Is there an enhancement that references how CRC will use cluster profiles? The IBM Cloud managed service case is clear since its already a production service and the alternative/excluded manifests in that setting is clearly understood.
> 
> see: [openshift/enhancements#418](https://github.com/openshift/enhancements/pull/418)

The enhancement PR for a single-node-developer cluster profile for CRC is: [openshift/enhancements#482](https://github.com/openshift/enhancements/pull/482)

guillaumerose:
I reduced the scope of this PR to the minimal possible set of changes. It is linked to this enhancement PR: https://github.com/openshift/enhancements/pull/414

Before merging it, we would need all subcomponents to have the proper include annotation (see bug 1871890). 

This changes only addresses what is required for IBM Cloud. For single node profile, we still need to bring this option to the installer.

pkg/cvo/cvo.go
```diff
@@ -221,7 +225,7 @@ func New(
 // controller that loads and applies content to the cluster. It returns an error if the payload appears to
 // be in error rather than continuing.
 func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestConfig *rest.Config) error {
-	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.release.Image, optr.exclude)
+	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.release.Image, optr.exclude, "self-managed-highly-available")
```
cfergeau, line 228: The  default name is `self-managed-high-availability`

guillaumerose: fixed, thanks

guillaumerose:
Rebased.

@LalatenduMohanty PTAL

eparis:
The unit test failure looks like a flake. But the lack of manifest getting applied, thus breaking cluster reader for insights operator does seem to be specific and repeatable to this PR.

wking:
/hold

Blocked by operator-framework/operator-lifecycle-manager#1887 to label the InstallPlans CRD manifest (among other things) to avoid [`cannot list resource "installplans" in API group "operators.coreos.com" in the namespace "openshift"`][1].

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/404/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1338647777357336576#1:build-log.txt%3A73

LalatenduMohanty (approved):
/lgtm

wking:
/hold cancel

I'd [held this][1] for operator-framework/operator-lifecycle-manager#1887, which has since landed.

[1]: https://github.com/openshift/cluster-version-operator/pull/404#issuecomment-746662144

LalatenduMohanty (approved):
/lgtm

PR #406, opened 2020-07-10
Bump client-go and library-go to current 4.6 tips

wking:
Builds on #405; you may want to review that first.

Pulling in openshift/client-go@83993cebb5aee53 and openshift/library-go@cc498c355c9998.  Generated with:

```context
$ go get -u github.com/openshift/client-go@83993cebb5aee533bf2f2dded2d87b9e32835f4a
go: github.com/openshift/client-go 83993cebb5aee533bf2f2dded2d87b9e32835f4a => v0.0.0-20200623090625-83993cebb5ae
$ go get -u github.com/openshift/library-go@cc498c355c99983057e9e01f3abcceb40ca5c298
go: github.com/openshift/library-go cc498c355c99983057e9e01f3abcceb40ca5c298 => v0.0.0-20200709151624-cc498c355c99
$ go mod vendor
go: downloading k8s.io/apimachinery v0.18.4
go: downloading github.com/openshift/api v0.0.0-20200623075207-eb651a5bb0ad
go: downloading k8s.io/api v0.18.4
go: downloading golang.org/x/net v0.0.0-20200602114024-627f9648deb9
go: downloading gopkg.in/yaml.v2 v2.3.0
go: downloading golang.org/x/text v0.3.3
$ go mod tidy
$ git add -A go.* vendor
```

using:

```context
$ go version
go version go1.14.4 linux/amd64
```

The client-go bump in turn addresses the golang.org/x/text v0.3.3 to address [CVE-2020-14040][3].

```context
$ grep '# golang.org/x/text' vendor/modules.txt
```

WIP is because I still need things like `Context` arguments to keep up with the Kube-client API, which has been pulled from v0.17.1 to v0.18.3.  Will get to that in a follow-up commit, and then pull the WIP from this PR.

[1]: https://github.com/openshift/client-go/commit/83993cebb5aee533bf2f2dded2d87b9e32835f4a
[2]: https://github.com/openshift/library-go/commit/cc498c355c99983057e9e01f3abcceb40ca5c298
[3]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-14040

PR #408, opened 2020-07-15
Bug 1855577: Updating the golang.org/x/text version to v0.3.3

LalatenduMohanty:


LalatenduMohanty:
With this change

LalatenduMohanty:
There are indirect dependency on `golang.org/x/text` , so I am not sure how to resolve this elegantly.  
for example 
```
$ go list -m all | grep 'golang.org/x/text'
golang.org/x/text v0.3.3
```

LalatenduMohanty:
Another approach https://github.com/openshift/cluster-version-operator/pull/406/. However we need to merge the fix asap.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	golang.org/x/sys v0.0.0-20191003212358-c178f38b412c // indirect
+	golang.org/x/text v0.3.3 // indirect
```
vrutkovs, line 28: `replace` would ensure other packages won't overwrite that

PR #410, opened 2020-07-17
Use context to add timeout to cincinnati HTTP request

jottofar:
Change CVO wait.Until instances to wait.UntilWithContext making top-level context available to the lower levels of the code. Use the added context to add a timeout to the remote cincinnati HTTP call which has the potential to hang the caller at the mercy of the target server.

The timeout has been set to an hour which was determined to be long enough to allow all expected well-behaved requests to complete but short enough to not be catastrophic if the calling thread were to hang.

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -1,12 +1,14 @@
 package cincinnati
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/blang/semver/v4"
 	"github.com/google/uuid"
@@ -16,6 +18,9 @@ const (
 	// GraphMediaType is the media-type specified in the HTTP Accept header
 	// of requests sent to the Cincinnati-v1 Graph API.
 	GraphMediaType = "application/json"
+
+	// Timeout when calling upstream Cincinnati stack.
+	getUpdatesTimeout = time.Minute * 60
 )
 
 // Client is a Cincinnati client which can be used to fetch update graphs from
@@ -58,7 +63,7 @@ func (err *Error) Error() string {
 // finding all of the children. These children are the available updates for
 // the current version and their payloads indicate from where the actual update
 // image can be downloaded.
-func (c Client) GetUpdates(uri *url.URL, arch string, channel string, version semver.Version) ([]Update, error) {
+func (c Client) GetUpdates(ctx context.Context, uri *url.URL, arch string, channel string, version semver.Version) ([]Update, error) {
 	transport := http.Transport{}
 	// Prepare parametrized cincinnati query.
 	queryParams := uri.Query()
@@ -83,7 +88,9 @@ func (c Client) GetUpdates(uri *url.URL, arch string, channel string, version se
 	}
 
 	client := http.Client{Transport: &transport}
-	resp, err := client.Do(req)
+	timeoutCtx, cancel := context.WithTimeout(ctx, getUpdatesTimeout)
+	defer cancel()
+	resp, err := client.Do(req.WithContext(timeoutCtx))
 	if err != nil {
 		return nil, &Error{Reason: "RemoteFailed", Message: err.Error(), cause: err}
 	}
--- a/pkg/cincinnati/cincinnati_test.go
+++ b/pkg/cincinnati/cincinnati_test.go
@@ -1,6 +1,7 @@
 package cincinnati
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
@@ -132,7 +133,7 @@ func TestGetUpdates(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			updates, err := c.GetUpdates(uri, arch, channelName, semver.MustParse(test.version))
+			updates, err := c.GetUpdates(context.Background(), uri, arch, channelName, semver.MustParse(test.version))
 			if test.err == "" {
 				if err != nil {
 					t.Fatalf("expected nil error, got: %v", err)
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -326,19 +326,19 @@ func (optr *Operator) Run(ctx context.Context, workers int) {
 	optr.queue.Add(optr.queueKey())
 
 	// start the config sync loop, and have it notify the queue when new status is detected
-	go runThrottledStatusNotifier(stopCh, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
+	go runThrottledStatusNotifier(ctx, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
 	go optr.configSync.Start(ctx, 16, optr.name, optr.cvLister)
-	go wait.Until(func() { optr.worker(ctx, optr.availableUpdatesQueue, optr.availableUpdatesSync) }, time.Second, stopCh)
-	go wait.Until(func() { optr.worker(ctx, optr.upgradeableQueue, optr.upgradeableSync) }, time.Second, stopCh)
-	go wait.Until(func() {
+	go wait.UntilWithContext(ctx, func(ctx context.Context) { optr.worker(ctx, optr.availableUpdatesQueue, optr.availableUpdatesSync) }, time.Second)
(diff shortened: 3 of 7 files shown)


wking:
[unit][1] includes:

```
# github.com/openshift/cluster-version-operator/pkg/cvo [github.com/openshift/cluster-version-operator/pkg/cvo.test]
pkg/cvo/cvo_scenarios_test.go:114:15: not enough arguments in call to o.sync
	have (string)
	want (context.Context, string)
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/410/pull-ci-openshift-cluster-version-operator-master-unit/1284249488935882752

pkg/cvo/sync_worker.go
```diff
@@ -971,21 +971,21 @@ func ownerRefModifier(config *configv1.ClusterVersion) resourcebuilder.MetaV1Obj
...
 		for {
 			select {
-			case <-stopCh:
+			case <-ctx.Done():
 				return
 			case next := <-ch:
 				// only throttle if we aren't on an edge
 				if next.Generation == last.Generation && next.Actual == last.Actual && next.Reconciling == last.Reconciling && (next.Failure != nil) == (last.Failure != nil) {
-					if err := throttle.Wait(ctx); err != nil {
+					if err := throttle.Wait(localCtx); err != nil {
```
wking, line 988: Why do we need `localCtx` here?  Can we pass `ctx` through and... I guess catch and ignore [the context values][1] with:

```go
if err := throttle.Wait(ctx); err != nil && err != context.Canceled && err != context.DeadlineExceeded {
  utilruntime.HandleError(fmt.Errorf("unable to throttle status notification: %v", err))
}
```

[1]: https://golang.org/pkg/context/#pkg-variables

pkg/cvo/cvo_test.go
```diff
@@ -3130,7 +3135,9 @@ func TestOperator_upgradeableSync(t *testing.T) {
 			optr.cvLister = &clientCVLister{client: optr.client}
 			optr.upgradeableChecks = optr.defaultUpgradeableChecks()
 
-			err := optr.upgradeableSync(optr.queueKey())
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
```
wking, line 3139: If we aren't setting a timeout, I don't see why we can't just use:

```go
ctx := context.Background()
```

in these test functions.

pkg/cvo/cvo.go
```diff
@@ -325,19 +325,19 @@ func (optr *Operator) Run(ctx context.Context, workers int) {
...
 	// start the config sync loop, and have it notify the queue when new status is detected
-	go runThrottledStatusNotifier(stopCh, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
+	go runThrottledStatusNotifier(ctx, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
 	go optr.configSync.Start(ctx, 16)
-	go wait.Until(func() { optr.worker(optr.availableUpdatesQueue, optr.availableUpdatesSync) }, time.Second, stopCh)
-	go wait.Until(func() { optr.worker(optr.upgradeableQueue, optr.upgradeableSync) }, time.Second, stopCh)
-	go wait.Until(func() {
+	go wait.UntilWithContext(ctx, func(context.Context) { optr.worker(ctx, optr.availableUpdatesQueue, optr.availableUpdatesSync) }, time.Second)
+	go wait.UntilWithContext(ctx, func(context.Context) { optr.worker(ctx, optr.upgradeableQueue, optr.upgradeableSync) }, time.Second)
+	go wait.UntilWithContext(ctx, func(context.Context) {
```
wking, line 332: These look like you are ignoring the passed-in `Context` and instead relying on a closure to get `ctx` into the functions.  Can we use:

```go
go wait.UntilWithContext(ctx, func(ctx context.Context) { optr.worker(ctx, ....
```

and similar instead?

pkg/cincinnati/cincinnati_test.go
```diff
@@ -132,7 +133,7 @@ func TestGetUpdates(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			updates, err := c.GetUpdates(uri, arch, channelName, semver.MustParse(test.version))
+			updates, err := c.GetUpdates(context.TODO(), uri, arch, channelName, semver.MustParse(test.version))
```
wking, line 136: tests can use [`context.Background()`][1].  `TODO` is for when you intend to go back later and provide a real `Context`, and for this case we are fine with no `TestGetUpdates`-side timeouts.

[1]: https://golang.org/pkg/context/#Background

jottofar:
/retitle Use context to add timeout to cincinnati HTTP request

pkg/cvo/cvo_test.go
jottofar: fixed

pkg/cvo/cvo.go
jottofar: fixed

pkg/cincinnati/cincinnati_test.go
jottofar: fixed

pkg/cvo/sync_worker.go
jottofar: My thinking was to make as little change as possible to the existing code and context handling and therefore maintained the local declaration of the local context there and just renamed it. Keeping it separate from the context now passed into the func. Couldn't find a "best-practice" on topic of when/if one should create a separate context but it does seem more clear to me to keep the two separate contexts.

pkg/cincinnati/cincinnati.go
```diff
@@ -51,14 +56,29 @@ func (err *Error) Error() string {
...
 }
 
+// Run HTTP request in a goroutine and pass the response to f.
+func clientDo(ctx context.Context, req *http.Request, client http.Client, f func(*http.Response, error) error) error {
+	c := make(chan error, 1)
+	getUpdatesCtx, cancel := context.WithTimeout(ctx, getUpdatesTimeout)
+	defer cancel()
+	go func() { c <- f(client.Do(req.WithContext(getUpdatesCtx))) }()
+	select {
+	case <-getUpdatesCtx.Done():
```
wking, line 66: What's the story around this?  It seems to be "we don't trust `f` or `client.Do` or `req.WithContext` to respect `ctx`, so we'll wrap it in `clientDo` and also watch `getUpdatesCtx` ourselves" or some such.

pkg/cincinnati/cincinnati.go
```diff
@@ -51,14 +56,29 @@ func (err *Error) Error() string {
...
+		<-c // Wait for f to return.
```
wking, line 67: But then you wait for `f` anyway?

pkg/cincinnati/cincinnati.go
jottofar: Not any of that. It seemed like a clean way to encapsulate waiting for either a timeout/cancel, an error or the response. But I'm not partial to it if you have something cleaner in mind.

pkg/cincinnati/cincinnati.go
jottofar: Correct. See https://github.com/openshift/cluster-version-operator/pull/410#discussion_r461198315 for my reasoning - right or wrong :)

pkg/cincinnati/cincinnati.go
wking: [This][1] seems sufficient to me.

[1]: https://github.com/openshift/cluster-version-operator/pull/83/files#diff-3a796940f9df2373810e52aae0d83795R62

pkg/cvo/sync_worker.go
jottofar: Removed localCtx in commit [8ee0d52](https://github.com/openshift/cluster-version-operator/pull/410/commits/8ee0d52e67593922f965e8c48bc22275acc62bab)

pkg/cincinnati/cincinnati.go
jottofar: Removed clientDo method in commit [8ee0d52](https://github.com/openshift/cluster-version-operator/pull/410/commits/8ee0d52e67593922f965e8c48bc22275acc62bab)

wking:
```
pkg/cincinnati/cincinnati.go:89:36: undefined: transport
```

wking:
[e2e][1] passed, except for `an end user can use OLM can subscribe to the operator`, which is [rhbz#1862322][2].

/override ci/prow/e2e

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/410/pull-ci-openshift-cluster-version-operator-master-e2e/1289016291406909440

PR #412, opened 2020-07-22
update CVO to inject internal loadbalancer for use by the CVO pod

deads2k:
During upgrades and kube-apiserver rollouts, the localhost connection can become
unavailable for multiple minutes. Using the internal loadbalancer prevents the
outage and lease loss.

/hold  

unit tests are owed, so I'm holding the PR, but I'd like to demonstrate the idea and see how CI reacts.

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -45,9 +45,9 @@ spec:
             name: serving-cert
             readOnly: true
         env:
-          - name: KUBERNETES_SERVICE_PORT # allows CVO to communicate with apiserver directly on same host.
+          - name: KUBERNETES_SERVICE_PORT # allows CVO to communicate with apiserver directly on same host.  Is substituted with port from infrastructures.status.apiServerInternalURL if available.
             value: "6443"
-          - name: KUBERNETES_SERVICE_HOST # allows CVO to communicate with apiserver directly on same host.
+          - name: KUBERNETES_SERVICE_HOST # allows CVO to communicate with apiserver directly on same host.  Is substituted with hostname from infrastructures.status.apiServerInternalURL if available.
             value: "127.0.0.1"
           - name: NODE_NAME
             valueFrom:
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -3,6 +3,8 @@ package resourcebuilder
 import (
 	"context"
 	"fmt"
+	"net"
+	"net/url"
 	"strings"
 
 	configv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -22,18 +24,20 @@ import (
 )
 
 type deploymentBuilder struct {
-	client      *appsclientv1.AppsV1Client
-	proxyGetter configv1.ProxiesGetter
-	raw         []byte
-	modifier    MetaV1ObjectModifierFunc
-	mode        Mode
+	client               *appsclientv1.AppsV1Client
+	proxyGetter          configv1.ProxiesGetter
+	infrastructureGetter configv1.InfrastructuresGetter
+	raw                  []byte
+	modifier             MetaV1ObjectModifierFunc
+	mode                 Mode
 }
 
 func newDeploymentBuilder(config *rest.Config, m lib.Manifest) Interface {
 	return &deploymentBuilder{
-		client:      appsclientv1.NewForConfigOrDie(withProtobuf(config)),
-		proxyGetter: configv1.NewForConfigOrDie(config),
-		raw:         m.Raw,
+		client:               appsclientv1.NewForConfigOrDie(withProtobuf(config)),
+		proxyGetter:          configv1.NewForConfigOrDie(config),
+		infrastructureGetter: configv1.NewForConfigOrDie(config),
+		raw:                  m.Raw,
 	}
 }
 
@@ -69,6 +73,43 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 		)
 		if err != nil {
 			return err
+		}
+	}
+
+	// if we detect the CVO deployment we need to replace the KUBERNETES_SERVICE_HOST env var with the internal load
+	// balancer to be resilient to kube-apiserver rollouts that cause the localhost server to become non-responsive for
+	// multiple minutes.
+	if deployment.Namespace == "openshift-cluster-version" && deployment.Name == "cluster-version-operator" {
+		infrastructureConfig, err := b.infrastructureGetter.Infrastructures().Get("cluster", metav1.GetOptions{})
+		// not found just means that we don't have infrastructure configuration yet, so we should tolerate not found and avoid substitution
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		if !errors.IsNotFound(err) {
+			lbURL, err := url.Parse(infrastructureConfig.Status.APIServerInternalURL)
+			if err != nil {
+				return err
+			}
+			// if we have any error and have empty strings, substitution below will do nothing and leave the manifest specified value
+			// errors can happen when the port is not specified, in which case we have a host and we write that into the env vars
+			lbHost, lbPort, err := net.SplitHostPort(lbURL.Host)
+			if err != nil {
+				if strings.Contains(err.Error(), "missing port in address") {
+					lbHost = lbURL.Host
+					lbPort = ""
+				} else {
+					return err
+				}
+			}
+			err = updatePodSpecWithInternalLoadBalancerKubeService(
+				&deployment.Spec.Template.Spec,
+				[]string{"cluster-version-operator"},
+				lbHost,
+				lbPort,
+			)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
--- a/lib/resourcebuilder/podspec.go
+++ b/lib/resourcebuilder/podspec.go
@@ -45,3 +45,78 @@ func updatePodSpecWithProxy(podSpec *corev1.PodSpec, containerNames []string, ht
 	return nil
 
 }
+
+// updatePodSpecWithInternalLoadBalancerKubeService mutates the input podspec by setting the KUBERNETES_SERVICE_HOST to the internal
+// loadbalancer endpoint and the KUBERNETES_SERVICE_PORT to the specified port
+func updatePodSpecWithInternalLoadBalancerKubeService(podSpec *corev1.PodSpec, containerNames []string, internalLoadBalancerHost, internalLoadBalancerPort string) error {
+	hasInternalLoadBalancer := len(internalLoadBalancerHost) > 0
+	if !hasInternalLoadBalancer {
+		return nil
+	}
+
+	for _, containerName := range containerNames {
+		found := false
+		for i := range podSpec.Containers {
+			if podSpec.Containers[i].Name != containerName {
+				continue
+			}
+			found = true
+
+			podSpec.Containers[i].Env = setKubeServiceValue(podSpec.Containers[i].Env, internalLoadBalancerHost, internalLoadBalancerPort)
+		}
+		for i := range podSpec.InitContainers {
+			if podSpec.InitContainers[i].Name != containerName {
+				continue
+			}
+			found = true
+
+			podSpec.InitContainers[i].Env = setKubeServiceValue(podSpec.Containers[i].Env, internalLoadBalancerHost, internalLoadBalancerPort)
+		}
+
+		if !found {
+			return fmt.Errorf("requested injection for non-existent container: %q", containerName)
+		}
+	}
+
+	return nil
+}
+
+// setKubeServiceValue replaces values if they are present and adds them if they are not
+func setKubeServiceValue(in []corev1.EnvVar, internalLoadBalancerHost, internalLoadBalancerPort string) []corev1.EnvVar {
+	ret := []corev1.EnvVar{}
+
+	portVal := "443"
+	if len(internalLoadBalancerPort) != 0 {
+		portVal = internalLoadBalancerPort
+	}
+
+	foundPort := false
+	foundHost := false
+	for j := range in {
+		ret = append(ret, *in[j].DeepCopy())
+		if ret[j].Name == "KUBERNETES_SERVICE_PORT" {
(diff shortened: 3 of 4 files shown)


deads2k:
from the CI run
```
    env:
    - name: KUBERNETES_SERVICE_PORT
      value: "443"
    - name: KUBERNETES_SERVICE_HOST
      value: api-int.ci-op-z0zhpvh0-11cd2.origin-ci-int-gce.dev.openshift.com
```

so it substituted,  but got 

2020-07-22T15:06:17.196524747Z E0722 15:06:17.196365       1 leaderelection.go:331] error retrieving resource lock openshift-cluster-version/version: Get https://api-int.ci-op-z0zhpvh0-11cd2.origin-ci-int-gce.dev.openshift.com:443/api/v1/namespaces/openshift-cluster-version/configmaps/version: dial tcp 10.0.0.2:443: connect: connection refused

because the port is 6443.  I'll update and write unit tests.

deads2k:
/hold cancel

now with tests and port detection.

lib/resourcebuilder/apps.go
```diff
@@ -72,6 +76,34 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
 		}
 	}
 
+	// the CVO is the only operator which doesn't have knowledge of its own operand (itself).  If we detect the CVO deployment
```
wking, line 79: Can we drop this opening sentence?  It doesn't seem relevant to me, and the CVO has multiple operands (all the manifests supplied in the release image), of which the CVO's own deployment is one.

lib/resourcebuilder/apps.go
```diff
@@ -72,6 +76,34 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+	// we need to replace the KUBERNETES_SERVICE_HOST env var with the internal load balancer to be resilient to kube-apiserver
+	// rollouts that cause the localhost server to become non-responsive for multiple minutes.
+	if deployment.Namespace == "openshift-cluster-version" && deployment.Name == "cluster-version-operator" {
+		infrastructureConfig, err := b.infrastructureGetter.Infrastructures().Get("cluster", metav1.GetOptions{})
+		// not found just means that we don't have proxy configuration, so we should tolerate and fill in empty
```
wking, line 84: how does "proxy configuration" come into this?  I'd expect that to be `Proxies().Get("cluster", ...)`

lib/resourcebuilder/apps.go
```diff
@@ -72,6 +76,34 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		lbURL, err := url.Parse(infrastructureConfig.Status.APIServerInternalURL)
```
wking, line 88: Sadly, API does not declare whether `APIServerInternalURL` is [optional or required][1].  But it _does_ declare [`Status` to be `+optional`][2].  So I think you need an `infrastructureConfig.Status == nil`, and possibly also an `infrastructureConfig.Status.APIServerInternalURL == ""`, guard which... falls back to `127.0.0.1` or `localhost` or some such.

[1]: https://github.com/openshift/api/blob/59994290fe0e13f63ae65f4c5a5f734b0d9f1eb4/config/v1/types_infrastructure.go#L74-L78
[2]: https://github.com/openshift/api/blob/59994290fe0e13f63ae65f4c5a5f734b0d9f1eb4/config/v1/types_infrastructure.go#L19-L21

lib/resourcebuilder/podspec.go
```diff
@@ -45,3 +45,78 @@ func updatePodSpecWithProxy(podSpec *corev1.PodSpec, containerNames []string, ht
 	return nil
 
 }
+
+// updatePodSpecWithInternalLoadBalancerKubeService mutates the input podspec by setting the KUBERNETES_SERVICE_HOST to the internal
+// loadbalancer endpoint and the KUBERNETES_SERVICE_PORT to the specified port
+func updatePodSpecWithInternalLoadBalancerKubeService(podSpec *corev1.PodSpec, containerNames []string, internalLoadBalancerHost, internalLoadBalancerPort string) error {
```
wking, line 51: can we add a comment to [the deployment itself][1], so folks reading that know that we may be fiddling with some fields dynamically?

[1]: https://github.com/openshift/cluster-version-operator/blob/6132bc33a76762dd1071eb049e308a4df75673ee/install/0000_00_cluster-version-operator_03_deployment.yaml

lib/resourcebuilder/apps.go
```diff
@@ -72,6 +76,34 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		lbHost, lbPort, err := net.SplitHostPort(lbURL.Host)
+		if err != nil {
```
wking, line 93:
```go
package main

import (
	"fmt"
	"net"
	"net/url"
)

func main() {
	config := ""
	lbURL, err := url.Parse(config)
	fmt.Printf("%v %s\n", lbURL, err)
	lbHost, lbPort, err := net.SplitHostPort(lbURL.Host)
	fmt.Printf("%q %s %s\n", lbHost, lbPort, err)
}
```

[gives][1] `missing port in address`.  You probably need to catch `missing port in address` and use `lbPort := ""` for that case.

[1]: https://play.golang.org/p/BgvPzFudICv

lib/resourcebuilder/apps.go
deads2k: > how does "proxy configuration" come into this? I'd expect that to be `Proxies().Get("cluster", ...)`

will fix comment

lib/resourcebuilder/apps.go
deads2k: sure

lib/resourcebuilder/apps.go
deads2k: > [gives](https://play.golang.org/p/BgvPzFudICv) `missing port in address`. You probably need to catch `missing port in address` and use `lbPort := ""` for that case.

since missing values result in no substitution, I'll add a comment explaining and skip the error

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -45,9 +45,9 @@ spec:
             name: serving-cert
             readOnly: true
         env:
-          - name: KUBERNETES_SERVICE_PORT # allows CVO to communicate with apiserver directly on same host.
+          - name: KUBERNETES_SERVICE_PORT # allows CVO to communicate with apiserver directly on same host.  Is substituted with port from infrastructures.status.apiServerInternalURL if available.
```
wking, line 48: `... Is substituted with port from infrastructures.status.apiServerInternalURL if available, defaulting to 443 if apiServerInternalURL is set but defines no port`?  Or somehow working the 443 default in here...

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -45,9 +45,9 @@ spec:
...
             value: "6443"
-          - name: KUBERNETES_SERVICE_HOST # allows CVO to communicate with apiserver directly on same host.
+          - name: KUBERNETES_SERVICE_HOST # allows CVO to communicate with apiserver directly on same host.  Is substituted with port from infrastructures.status.apiServerInternalURL if available.
```
wking, line 50: s/port/hostname/

install/0000_00_cluster-version-operator_03_deployment.yaml
wking: Never mind, I'll handle this where `apiServerInternalURL` is defined in openshift/api.

wking: > Never mind, I'll handle this where apiServerInternalURL is defined in openshift/api.

openshift/api#704.  Marking this thread resolved.

LalatenduMohanty:
Apart from https://github.com/openshift/cluster-version-operator/pull/412/#discussion_r458972629 and  https://github.com/openshift/cluster-version-operator/pull/412/#discussion_r458972824 the PR looks good to me.

lib/resourcebuilder/apps.go
```diff
@@ -72,6 +76,35 @@ func (b *deploymentBuilder) Do(ctx context.Context) error {
...
+			return err
+		}
+		if !errors.IsNotFound(err) {
+			lbURL, err := url.Parse(infrastructureConfig.Status.APIServerInternalURL)
+			if err != nil {
+				return err
+			}
+			// if we have any error and have empty strings, substitution below will do nothing and leave the manifest specified value
+			// errors can happen when the port is not specified, in which case we have a host and we write that into the env vars
+			lbHost, lbPort, _ := net.SplitHostPort(lbURL.Host)
```
wking, line 95: I don't think a blanket-ignore works.

```go
package main

import (
	"fmt"
	"net"
	"net/url"
)

func main() {
	config := "https://example.com"
	lbURL, err := url.Parse(config)
	fmt.Printf("%v %s\n", lbURL, err)
	lbHost, lbPort, err := net.SplitHostPort(lbURL.Host)
	fmt.Printf("%q %q %s\n", lbHost, lbPort, err)
}
```

[gives][1]:

```
https://example.com %!s(<nil>)
"" "" address example.com: missing port in address
```

```
lbHost, lbPort, err := net.SplitHostPort(lbURL.Host)
if err != nil {
  lbHost = lbURL.Host
}
```

Would be nice if there was an error detector for `missing port in address`, but I don't see one.  I'd be ok with:

```go
if err != nil {
  if strings.Contains(err.Error(), "missing port in address") {
    lbHost = lbURL.Host
  } else {
    return err
  }
}
```

too, if we just want to treat the substring as a committed API.


[1]: https://play.golang.org/p/stQUuNkqp3m

lib/resourcebuilder/podspec.go
wking: > can we add a comment to the deployment itself...

95c6c57 has this; marking the thread resolved.

lib/resourcebuilder/podspec.go
```diff
@@ -45,3 +45,78 @@ func updatePodSpecWithProxy(podSpec *corev1.PodSpec, containerNames []string, ht
...
+	}
+
+	return nil
+}
+
+// setKubeServiceValue replaces values if they are present and adds them if they are not
+func setKubeServiceValue(in []corev1.EnvVar, internalLoadBalancerHost, internalLoadBalancerPort string) []corev1.EnvVar {
+	ret := []corev1.EnvVar{}
+
+	portVal := "443"
```
wking, line 88: We should switch on scheme, so we default to 80 for the crazy folks that configure an HTTP LB.  Which is [techincally allowed][1] today.

[1]: https://github.com/openshift/api/blob/0ba2c3658da6cade27a5066575cd8446eb03914c/config/v1/types_infrastructure.go#L74

wking: or ensure that `KUBERNETES_SERVICE_PORT` is unset when `internalLoadBalancerPort == ""`, and leave the defaulting to the library code, which hopefully does not require an explicit `KUBERNETES_SERVICE_PORT`.  But I can check...

wking: [The docs][1] suggest that separate variables exist for `KUBERNETES_SERVICE_PORT` and `KUBERNETES_SERVICE_PORT_HTTPS`.  Might be recent, because I don't see any `KUBERNETES_SERVICE_PORT_HTTPS` matches in the CVO codebase.

[1]: https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#environment-variables

wking: Dunno what those docs are about, maybe they're aspirational.  But current CVO has:

```console
$ git --no-pager grep 'KUBERNETES_SERVICE_PORT must' origin/master
origin/master:vendor/k8s.io/client-go/rest/config.go:var ErrNotInCluster = errors.New("unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined")
```

So back to my "we should switch the default depending on the configured scheme" suggestion.

deads2k: it is not allowed. The kube-apiserver cannot run in http. Those people will die out.

lib/resourcebuilder/apps.go
wking: f0f00ff9b9b no longer contains the line.  Marking this thread resolved.

lib/resourcebuilder/apps.go
wking: f0f00ff9b9b has "we don't have infrastructure configuration yet".  Marking this thread resolved.

lib/resourcebuilder/apps.go
wking: Ah, `Status` is `+optional`, but not a pointer (for mysterious, historical reasons).  No need for a `nil` guard.  Marking this thread resolved.

lib/resourcebuilder/apps.go
wking: f0f00ff9b9b1 has the explicit `missing port in address` catch.  Marking this thread resolved.

install/0000_00_cluster-version-operator_03_deployment.yaml
wking: f0f00ff9b9b has `hostname` here.  Marking the thread resolved.

lib/resourcebuilder/apps.go
wking: f0f00ff9b9b18 has the substring check.  Marking this thread resolved.

lib/resourcebuilder/podspec.go
sdodson: We're on 6443, right?

wking: most folks will explicitly set `6443`, yes.  An `https://{host}` URI without an explicit port means "443" though, so this default makes sense to me there.

wking: openshift/api#704 now forbids `http`.  Once that's approved I will be satisfied with the hard-coded 443 here.

wking: [in the pipe][1].  Marking this thread resolved.

[1]: https://github.com/openshift/api/pull/704#issuecomment-662678578

PR #413, opened 2020-07-22
Changes to verify package in preparation for move to library-go

jottofar:
Changes to verify package in preparation for move to library-go
    
The CVO verify package will be moved to library-go for reuse by 'oc adm release mirror' and CVO. Changes are being made here in CVO before the move to provide a cleaner move to library-go.

Logic to create verifier was broken out into a new configmap method NewFromManifests. This method uses k8s encoding and as a result existing test cases had to be changed from creating manifests from from declarations within the test cases itself to creating them from config map files. In this manner the lib/Manifest object is properly created with its Raw bytes field populated.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -12,11 +12,9 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/google/uuid"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	informerscorev1 "k8s.io/client-go/informers/core/v1"
@@ -281,32 +279,20 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 // package for more details on the algorithm for verification. If the annotation is set, a verifier or error
 // is always returned.
 func loadConfigMapVerifierDataFromUpdate(update *payload.Update, clientBuilder sigstore.HTTPClient, configMapClient coreclientsetv1.ConfigMapsGetter) (verify.Interface, *verify.StorePersister, error) {
-	configMapGVK := corev1.SchemeGroupVersion.WithKind("ConfigMap")
-	for _, manifest := range update.Manifests {
-		if manifest.GVK != configMapGVK {
-			continue
-		}
-		if _, ok := manifest.Obj.GetAnnotations()[verify.ReleaseAnnotationConfigMapVerifier]; !ok {
-			continue
-		}
-		src := fmt.Sprintf("the config map %s/%s", manifest.Obj.GetNamespace(), manifest.Obj.GetName())
-		data, _, err := unstructured.NestedStringMap(manifest.Obj.Object, "data")
-		if err != nil {
-			return nil, nil, errors.Wrapf(err, "%s is not valid: %v", src, err)
-		}
-		verifier, err := verify.NewFromConfigMapData(src, data, clientBuilder)
-		if err != nil {
-			return nil, nil, err
-		}
-
-		// allow the verifier to consult the cluster for signature data, and also configure
-		// a process that writes signatures back to that store
-		signatureStore := configmap.NewStore(configMapClient, nil)
-		verifier.Store = &serial.Store{Stores: []store.Store{signatureStore, verifier.Store}}
-		persister := verify.NewSignatureStorePersister(signatureStore, verifier)
-		return verifier, persister, nil
+	verifier, err := verify.NewFromManifests(update.Manifests, clientBuilder)
+	if err != nil {
+		return nil, nil, err
 	}
-	return nil, nil, nil
+	if verifier == nil {
+		return nil, nil, nil
+	}
+
+	// allow the verifier to consult the cluster for signature data, and also configure
+	// a process that writes signatures back to that store
+	signatureStore := configmap.NewStore(configMapClient, nil)
+	verifier.Store = &serial.Store{Stores: []store.Store{signatureStore, verifier.Store}}
+	persister := verify.NewSignatureStorePersister(signatureStore, verifier)
+	return verifier, persister, nil
 }
 
 // Run runs the cluster version operator until stopCh is completed. Workers is ignored for now.
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -1,6 +1,7 @@
 package cvo
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -20,7 +21,6 @@ import (
 	apiextclientv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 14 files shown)


pkg/verify/verifyutil/util.go
```diff
@@ -0,0 +1,18 @@
+package verifyutil
```
wking, line 1: Can this just be `util` under `pkg/verify/util`?  Folks who need to distinguish from another `util` can always:

```go
import (
  verifyutil ".../pkg/verify/util"
)
```

pkg/verify/configmap.go
```diff
@@ -8,18 +8,69 @@ import (
...
+const (
+	// ReleaseAnnotationConfigMapVerifier is an annotation set on a config map in the
+	// release payload to indicate that this config map controls signing for the payload.
+	// Only the first config map within the payload should be used, regardless of whether
+	// it has data. See NewFromConfigMapData for more.
+	ReleaseAnnotationConfigMapVerifier = "release.openshift.io/verification-config-map"
+
+	// NamespaceLabelConfigMap is the Namespace label applied to a configmap
+	// containing signatures.
+	NamespaceLabelConfigMap = "openshift-config-managed"
```
wking, line 30:
```console
$ git --no-pager log --oneline -1 origin/pr/413
e792735 (origin/pr/413) Changes to verify package in preparation for move to library-go
$ git --no-pager grep openshift-config-managed origin/pr/413
...some *_test.go hits, where I'm fine sticking with the local string...
origin/pr/413:pkg/verify/store/configmap/configmap.go:		ns:      "openshift-config-managed",
...more places where I'm fine leaving things as they stand...
```

I think we want `ns: importName.NamespaceLabelConfigMap` there.  Or maybe because of import direction we need to turn it around and define the constant in `pkg/verify/store/configmap/configmap.go`?

pkg/verify/configmap.go
```diff
@@ -8,18 +8,69 @@ import (
...
+
+	// ReleaseLabelConfigMap is a label applied to a configmap inside the
+	// NamespaceLabelConfigMap namespace that indicates it contains signatures
+	// for release image digests. Any binaryData key that starts with the digest
+	// is added to the list of signatures checked.
+	ReleaseLabelConfigMap = "release.openshift.io/verification-signatures"
```
wking, line 36: Like [the namespace][1], I don't like this being a public API with two separate definitions:

```console
$ git --no-pager grep 'verification-signatures' origin/pr/413
origin/pr/413:pkg/verify/configmap.go:	ReleaseLabelConfigMap = "release.openshift.io/verification-signatures"
origin/pr/413:pkg/verify/store/configmap/configmap.go:const ReleaseLabelConfigMap = "release.openshift.io/verification-signatures"
```

Can consumers just use the existing `pkg/verify/store/configmap` source?

[1]: https://github.com/openshift/cluster-version-operator/pull/413/files#r458951378

pkg/verify/configmap.go
```diff
@@ -8,18 +8,69 @@ import (
...
+
+	// Unique portion of key used within a config map identifying data field containing
+	// one or more GPG public keys in ASCII form that must have signed the
+	// release image by digest.
+	VerifierPublicKeyPrefix = "verifier-public-key-"
+
+	// Unique portion of key used within a config map identifying data field containing a URL
+	// (scheme http://, or https://) location that contains signatures.
+	StorePrefix = "store-"
```
wking, line 45: Convention [is][1]:

> Notice this comment is a complete sentence that begins with the name of the element it describes.

So these should be `VerifierPublicKeyPrefix is the unique portion...` and `StorePrefix is the unique portion...`.  We should grow some `golint` validation or some such to auto-warn about this...

[1]: https://blog.golang.org/godoc

pkg/verify/configmap.go
```diff
@@ -8,18 +8,69 @@ import (
...
+)
+
+// GetSignaturesAsConfigmap returns the given signatures in a configmap. Uses
+// digestToKeyPrefix to replace colon with dash when saving digest to configmap.
```
wking, line 49: nit: "digestToKeyPrefix" -> "util.DigestToKeyPrefix" or whatever, now that that is a public API.

pkg/verify/verifyutil/util.go
jottofar: Fixed in [2bde02f](https://github.com/openshift/cluster-version-operator/pull/413/commits/2bde02f1bc058d9b91c9b5cce3afb3259523b74c)

pkg/verify/configmap.go
jottofar: Fixed in [2bde02f](https://github.com/openshift/cluster-version-operator/pull/413/commits/2bde02f1bc058d9b91c9b5cce3afb3259523b74c)

pkg/verify/configmap.go
jottofar: Fixed in [2bde02f](https://github.com/openshift/cluster-version-operator/pull/413/commits/2bde02f1bc058d9b91c9b5cce3afb3259523b74c)

pkg/verify/configmap.go
jottofar: Fixed in [2bde02f](https://github.com/openshift/cluster-version-operator/pull/413/commits/2bde02f1bc058d9b91c9b5cce3afb3259523b74c)

pkg/verify/configmap.go
jottofar: Fixed in [2bde02f](https://github.com/openshift/cluster-version-operator/pull/413/commits/2bde02f1bc058d9b91c9b5cce3afb3259523b74c)

pkg/verify/verify.go
```diff
@@ -29,6 +30,8 @@ type Interface interface {
 	// Verify should return nil if the provided release digest has suffient signatures to be considered
 	// valid. It should return an error in all other cases.
 	Verify(ctx context.Context, releaseDigest string) error
+
+	Signatures() map[string][][]byte
```
wking, line 34: Does this need to be part of the verification interface?  Can't `oc` use our `ReleaseVerifier` type directly instead of going through the opaque `verify.Interface`?

jottofar: Good catch. Misstep when I wrapped it in a method in oc. Fixed in [f16cb2d](https://github.com/openshift/cluster-version-operator/pull/413/commits/f16cb2d3f4f0f68ba83532d9090ce64f51dd5b94).

jottofar:
@wking I went back to my original approach which adds a method LoadConfigMapVerifierDataFromManifest, to be used by oc, and allows the prefix constants to remain internal. Note that I added k8s decode/encode to util which is used by new method.

pkg/verify/util/encode.go
```diff
@@ -0,0 +1,40 @@
...
+	if err := corev1.AddToScheme(coreScheme); err != nil {
+		panic(err)
+	}
+	coreEncoderCodecFactory := serializer.NewCodecFactory(coreScheme)
+	coreEncoder = coreEncoderCodecFactory.LegacyCodec(corev1.SchemeGroupVersion)
+}
+
+// ReadConfigMap reads config map object from bytes. nil is returned if
+// the object cannot be decoded as a config map.
+func ReadConfigMap(objBytes []byte) (*corev1.ConfigMap, error) {
```
wking, line 25: Seems like this function should go into a separate helper, outside of `verify`, and it can replace [our existing `ReadConfigMapV1OrDie`][1] (we'd need to `panic` locally if the library function returned an error).  And I guess move `ConfigMapAsBytes` over too?  But whatever, doesn't need to block the move to library-go.

[1]: https://github.com/openshift/cluster-version-operator/blob/a49fef5c66c6b0707c54fd93f84d2f51d3d28aca/lib/resourceread/core.go#L21-L26

pkg/verify/configmap.go
```diff
@@ -8,18 +8,81 @@ import (
...
+
+	// storePrefix is the unique portion of the key used within a config map identifying
+	// data field containing a URL (scheme http://, or https://) location that contains
+	// signatures.
+	storePrefix = "store-"
+)
+
+// LoadConfigMapVerifierDataFromManifest decodes the manifest as a config map and checks it for the correct
+// verifier annotations. If found returns data as a map otherwise returns an empty map.
+func LoadConfigMapVerifierDataFromManifest(manifestBytes []byte) map[string]string {
```
wking, line 42: How will `oc` use this?  Today it [iterates over manifests][1], [like we do in the CVO][2].  Can we use `oc`'s `LoadConfigMapVerifierDataFromUpdate` here instead?  Although maybe call it `NewFromManifests` or some such?

[1]: https://github.com/openshift/oc/blob/49e8811ef52d58451e317cd616bcbca1865505f6/pkg/helpers/release/verify.go#L341-L343
[2]: https://github.com/openshift/cluster-version-operator/blob/a49fef5c66c6b0707c54fd93f84d2f51d3d28aca/pkg/cvo/cvo.go#L277-L308

pkg/verify/util/encode.go
jottofar: I thought of same but simply wanted to concentrate on what was immediately required for this reuse task.

pkg/verify/configmap.go
jottofar: I removed this method as oc will send in manifests like CVO does now. As a result, I broke out logic to create the verifier from CVO method loadConfigMapVerifierDataFromUpdate into a new configmap method NewFromManifests since oc only needs a verifier without stores or a persister.

pkg/verify/testdata/manifests/only-first-used.yaml
```diff
@@ -0,0 +1,112 @@
...
+    C/JJ7/ufz3UZmyViH/WDl+AYdgk3JqCIO5w5ryrC9IyBzYv2m0HqYbWfphY3uHw5
+    un3ndLJcu8+BGP5F+ONQEGl+DRH58Il9Jp3HwbRa7dvkPgEhfFR+1hI+Btta2C7E
+    0/2NKzCxZw7Lx3PBRcU92YKyaEihfy/aQKZCAuyfKiMvsmzs+4poIX7I9NQCJpyE
+    IGfINoZ7VxqHwRn/d5mw2MZTJjbzSf+Um9YJyA0iEEyD6qjriWQRbuxpQXmlAJbh
+    8okZ4gbVFv1F8MzK+4R8VvWJ0XxgtikSo72fHjwha7MAjqFnOq6eo6fEC/75g3NL
+    Ght5VdpGuHk0vbdENHMC8wS99e5qXGNDued3hlTavDMlEAHl34q2H9nakTGRF5Ki
+    JUfNh3DVRGhg8cMIti21njiRh7gyFI2OccATY7bBSr79JhuNwelHuxLrCFpY7V25
+    OFktl15jZJaMxuQBqYdBgSay2G0U6D1+7VsWufpzd/Abx1/c3oi9ZaJvW22kAggq
+    dzdA27UUYjWvx42w9menJwh/0jeQcTecIUd0d0rFcw/c1pvgMMl/Q73yzKgKYw==
+    =zbHE
```
wking, line 71: nit:

```console
$ git --no-pager show --check --oneline origin/pr/413
b39b35fe (origin/pr/413, jottofar/ota-190-verify-reuse-prep) Changes to verify package in preparation for move to library-go
pkg/verify/testdata/manifests/only-first-used.yaml:71: trailing whitespace.
+    =zbHE     
```

pkg/cvo/cvo_test.go
```diff
@@ -3272,169 +3272,64 @@ func fakeClientsetWithUpdates(obj *configv1.ClusterVersion) *fake.Clientset {
...
 			wantVerifiers: 1,
 		},
 	}
 	for _, tt := range tests {
+		raw, err := ioutil.ReadFile(filepath.Join("..", "verify", "testdata", "manifests", tt.fileName))
+		if err != nil {
+			t.Fatal(err)
+		}
+		if tt.fileName != "no-op.yaml" {
+			ms, err := lib.ParseManifests(bytes.NewReader(raw))
```
wking, line 3327: Why are we not parsing `no-op.yaml`?  And if not parsing it makes sense, why are we reading it from disk at all?

pkg/verify/testdata/manifests/only-first-used.yaml
jottofar: Fixed. Thanks!

pkg/cvo/cvo_test.go
jottofar: Good catch. Left over from testing. It contains no manifests, which is the test, so, right, no need for an actual file.

pkg/verify/testdata/manifests/loads-valid.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: release-verification
+  namespace: configmap.NamespaceLabelConfigMap
```
wking, line 5: Shouldn't these be literal `openshift-config-managed`, a not the Go variable?  Although the docs are pretty clear that only the annotation matters, so maybe we want `namespace: only-the-annotation-matters` ;).

pkg/verify/testdata/manifests/loads-valid.yaml
```diff
@@ -0,0 +1,39 @@
...
+    0/2NKzCxZw7Lx3PBRcU92YKyaEihfy/aQKZCAuyfKiMvsmzs+4poIX7I9NQCJpyE
+    IGfINoZ7VxqHwRn/d5mw2MZTJjbzSf+Um9YJyA0iEEyD6qjriWQRbuxpQXmlAJbh
+    8okZ4gbVFv1F8MzK+4R8VvWJ0XxgtikSo72fHjwha7MAjqFnOq6eo6fEC/75g3NL
+    Ght5VdpGuHk0vbdENHMC8wS99e5qXGNDued3hlTavDMlEAHl34q2H9nakTGRF5Ki
+    JUfNh3DVRGhg8cMIti21njiRh7gyFI2OccATY7bBSr79JhuNwelHuxLrCFpY7V25
+    OFktl15jZJaMxuQBqYdBgSay2G0U6D1+7VsWufpzd/Abx1/c3oi9ZaJvW22kAggq
+    dzdA27UUYjWvx42w9menJwh/0jeQcTecIUd0d0rFcw/c1pvgMMl/Q73yzKgKYw==
+    =zbHE
+    -----END PGP PUBLIC KEY BLOCK-----"
+  store-local: file://../verify/testdata/signatures
```
wking, line 39: I feel like this path is a hint that these unit tests belong under `pkg/verify`.

pkg/verify/testdata/manifests/only-first-used.yaml
```diff
@@ -0,0 +1,112 @@
+# NOTE: This file is parsed by CVO's lib/Manifest ParseManifests method which
```
wking, line 1: The fact that ParseManifests will load two manifests with identical kind/namespace/name without complaining is probably a bug.  Can we pick different names for these two ConfigMaps, and file a follow-up ticket about failing noisly on that sort of conflict?

pkg/cvo/cvo_test.go
```diff
@@ -3272,169 +3272,64 @@ func fakeClientsetWithUpdates(obj *configv1.ClusterVersion) *fake.Clientset {
...
-			},
+			name:          "only the first valid configuration is used",
+			fileName:      "only-first-used.yaml",
+			update:        &payload.Update{},
 			want:          true,
 			wantVerifiers: 1,
 		},
 	}
 	for _, tt := range tests {
+		if tt.fileName != "no-op.yaml" {
```
wking, line 3322: possibly `s/no-op.yaml//`, because an empty string is a clearer marker of "don't load any filenames" to me than `no-op.yaml`.  But obviously works either way, so feel free to leave it as it stands if you feel `no-op.yaml` is more intuitive.

pkg/verify/testdata/manifests/only-first-used.yaml
```diff
@@ -0,0 +1,112 @@
...
+# reverses the order of the below config map. The valid one is therefore the
+# last one in the below list.
```
wking, line 3: I don't think it reverses the order, see [this test-case][1].  One issue is that `wantVerifiers` is currently an integer.  I'd rather see `expectedVerifier string` and `expectedError string` (or `*regexp.Regexp` for either, if we need that flexibility), to be compared with the output of [`got.String()`][2] and `err.Error()` like [this][3].

[1]: https://github.com/openshift/cluster-version-operator/blob/16d3d8465b8e3f3c557becab28763d6cc18829a6/lib/manifest_test.go#L66-L99
[2]: https://github.com/openshift/cluster-version-operator/blob/16d3d8465b8e3f3c557becab28763d6cc18829a6/pkg/verify/verify.go#L82-L83
[3]: https://github.com/openshift/cluster-version-operator/blob/692a8a37b52c2bc00c45fb84622ba2d5ed3125bb/pkg/verify/store/serial/serial_test.go#L86-L98

pkg/verify/testdata/manifests/loads-valid.yaml
jottofar: Hmm, it's testing method loadConfigMapVerifierDataFromUpdate which is in package cvo. But yeah, let's wait until it's in library-go if we do want to make a move.

pkg/verify/testdata/manifests/loads-valid.yaml
jottofar: They should. Fixed.

pkg/cvo/cvo_test.go
jottofar: Changed.

pkg/verify/testdata/manifests/only-first-used.yaml
jottofar: Yeah, it wasn't reversing order but instead my failure to separate the manifests with `---` causing my issue. Works as expected now.
I also made the changes to the test case per your recommendation.

pkg/verify/testdata/manifests/only-first-used.yaml
jottofar: I changed the manifests names to match the original test data - which was my original intent - which do differ.

I post a PR to fix ParseManifests shortly.

pkg/cvo/cvo.go
```diff
@@ -281,32 +279,20 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
...
-		// allow the verifier to consult the cluster for signature data, and also configure
-		// a process that writes signatures back to that store
-		signatureStore := configmap.NewStore(configMapClient, nil)
-		verifier.Store = &serial.Store{Stores: []store.Store{signatureStore, verifier.Store}}
-		persister := verify.NewSignatureStorePersister(signatureStore, verifier)
-		return verifier, persister, nil
+	verifier, err := verify.NewFromManifests(update.Manifests, clientBuilder)
+	if err != nil {
+		return nil, nil, err
 	}
```
vrutkovs, line 285:
```suggestion
	verifier, err := verify.NewFromManifests(update.Manifests, clientBuilder)
	if err != nil || verifier == nil {
		return nil, nil, err
	}
```

pkg/verify/configmap.go
```diff
@@ -8,20 +8,94 @@ import (
...
+		BinaryData: make(map[string][]byte),
+	}
+	prefix, err := util.DigestToKeyPrefix(digest, "-")
+	if err != nil {
+		return nil, err
+	}
+	cm.Name = prefix
+	for i := 0; i < len(signatures); i++ {
+		cm.BinaryData[fmt.Sprintf("%s-%d", prefix, i+1)] = signatures[i]
+	}
```
vrutkovs, line 65: nitpick:
```suggestion
	for i, v := range signatures {
		cm.BinaryData[fmt.Sprintf("%s-%d", prefix, i+1)] = v
	}
```

pkg/cvo/cvo.go
wking: I don't think we need to be *that* defensive, but I'm ok adding language to `NewFromManifests`' docstring like "... returns a verifier or an error" to show that `err == nil` means `verifier` is populated.

jottofar: Since this is a code port to common I've tried to minimize changes. Per the original code and comment, it is not an error to return a nil verifier: "It returns an error if the data is not valid, or no verifier if no config map is found."

vrutkovs: Okay, lets leave it as is

pkg/verify/configmap.go
jottofar: Fixed

pkg/cvo/cvo_test.go
```diff
@@ -3272,190 +3272,97 @@ func fakeClientsetWithUpdates(obj *configv1.ClusterVersion) *fake.Clientset {
...
-							},
-						},
-					},
-				},
-			},
-			wantErr: true,
+			name:          "requires verifiers",
+			fileName:      "requires-verifiers.yaml",
+			update:        &payload.Update{},
+			expectedError: ExpectedError,
```
wking, line 3309: Hmm, `requires-verifiers.yaml` declares `store-local`.  Why are we saying `...did not provide any signature stores...` for it?

wking:
One unit test [I don't understand][1], otherwise looks good to me.  Pull the hold if I'm just confused, or if you want to punt a fix to follow-up work.

/lgtm
/hold

[1]: https://github.com/openshift/cluster-version-operator/pull/413#discussion_r471917264

pkg/cvo/cvo_test.go
jottofar: `expectStore` indicates whether we expect `loadConfigMapVerifierDataFromUpdate` to return non nil 'signatureStore' which it only does for the last two test cases where `expectStore` is set to `true`. This is per the original test cases as well but they used `want: true`.

PR #415, opened 2020-07-23
install/0000_90_cluster-version-operator_02_servicemonitor: Info-level alert for available updates

wking:
The purpose of this alert is to give cluster-admins something they can subscribe to (via Alertmanager notifications) to let them know when they can initiate an alert for a particular cluster.

It is info severity because we want the quick push notification, but letting your cluster sit on the old version for a few days or a week or so is unlikely to cause much trouble.  For things like CVE fixes, a more timely update might be advisable, but the cluster-version operator currently has no way to make the "critical CVE" vs. "no fixes, but adds a feature" distinction.  So I'm punting on the warn-level or so "CVO is getting grumpy about your failure to take the long-recommended update" alert for now, and just addressing the info-level "you could update, no pressure".


vrutkovs:
I like the idea of adding this to alertmanager, but wouldn't it be too annoying for admin console users?

We show "upgrade available" on the top navbar, on Overview page and on Cluster Settings page. This alert would show on Notifications page - isn't that enough?

Lets have this approved by UI/UX team?

jottofar:
> they can initiate an alert for a particular cluster.
In description, shouldn't this be:
"they can initiate an `update` for a particular cluster."

sdodson:
@ncameronbritt please lend us your eyes

ncameronbritt:
I think this is a good approach. And I believe we discussed that if we had alerts for available upgrades (as proposed here), then we probably don't need the indication we have in the console today, because customers could still see the alert in the console. This design shows that: https://marvelapp.com/prototype/64j205c/screen/70755417
@cshinn @megan-hall @beanh66, what do you all think?

wking:
> "they can initiate an update for a particular cluster."

Fixed with c53d639 -> b397d8f, and edited the initial PR comment too.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -51,6 +51,13 @@ spec:
         (time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 3600 and ignoring(condition, name, reason) cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"}
       labels:
         severity: critical
+    - alert: UpdateAvailable
+      annotations:
+        message: $value {{ if eq $value 1 }}update is{{ else }}updates are{{ end }} available.  For more information refer to 'oc adm upgrade'{{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} or {{ label \"url\" (first $console_url ) }}/settings/cluster/{{ end }}{{ end }}" }}.
```
vrutkovs, line 56: `cluster_version_available_updates > 1` would return true, not the number of alerts, right? 
In any case I don't see a benefit of mentioning number of alerts - even if its 9999 updates available admins are usually guided by maintenance windows

wking: Rebased onto master and reworded to be less dynamic with b397d8f3 -> ccf08821.

wking:
[e2e][1] failed on `an end user can use OLM can subscribe to the operator`, which is the unrelated [rhbz#1862322][2].

/override ci/prow/e2e

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/415/pull-ci-openshift-cluster-version-operator-master-e2e/1289039093132431360

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -51,6 +51,13 @@ spec:
         (time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 3600 and ignoring(condition, name, reason) cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"}
       labels:
         severity: critical
+    - alert: UpdateAvailable
+      annotations:
+        message: Your upstream update recommendation service recommends you update your cluster.  For more information refer to 'oc adm upgrade'{{ "{{ with $console_url := \"console_url\" | query }}{{ if ne (len (label \"url\" (first $console_url ) ) ) 0}} or {{ label \"url\" (first $console_url ) }}/settings/cluster/{{ end }}{{ end }}" }}.
+      expr: |
+        cluster_version_available_updates > 1
```
wking, line 58: Sigh, this should be `>0`, right?

vrutkovs: I think so - I'll try it out and file bugs

wking: Confirmed; filed #432 to fix.

PR #418, opened 2020-07-28
pkg/cvo/cvo_scenarios_test: Fix "Retrive" -> "Retrieve" typo

wking:
Typo is from d06f4be481 (#244).


wking:
This is local unit testing only.

/override ci/prow/e2e
/override ci/prow/e2e-upgrade

PR #419, opened 2020-07-28
*: Port from Update to Release for ClusterVersion status

wking:
Bump dependencies to pull in openshift/api#521, and then port our status code to use the new `Release` type.

This is mostly mechanical, except for the next/last `Actual` comparison, where I had to avoid:

```
invalid operation: next.Actual == last.Actual (struct containing []string cannot be compared)
```

now that `Release` has a slice of channels.  Now I'm comparing only `Image`, which is sufficient for "is this a different target?".  The test suite's expected `LastProgress` changes are because the shift from:

```go
{Version: "4.0.1", Image: "image/image:1"}
```

to:

```go
{Version: "1.0.0-abc", Image: "image/image:1"}
```

no longer counts as a `LastProgress`-bumping change, because the `Image` remains unchanged.

WIP because I still have some FIXMEs in there about populating the new `Release` properties.


pkg/autoupdate/autoupdate.go
```diff
@@ -175,7 +175,10 @@ func (ctrl *Controller) sync(ctx context.Context, key string) error {
 		return nil
 	}
 	up := nextUpdate(clusterversion.Status.AvailableUpdates)
-	clusterversion.Spec.DesiredUpdate = &up
+	clusterversion.Spec.DesiredUpdate = &v1.Update{
+		Version: up.Version,
+		Image:   up.Image,
+	}
```
LalatenduMohanty, line 181: Why can not we just use `up` here as `nextUpdate` returning `v1.Release` now?

pkg/cvo/availableupdates.go
```diff
@@ -45,14 +47,15 @@ func (optr *Operator) syncAvailableUpdates(config *configv1.ClusterVersion) erro
...
-	updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), proxyURL, tlsConfig, upstream, arch, channel, optr.releaseVersion)
+	current, updates, condition := calculateAvailableUpdatesStatus(string(config.Spec.ClusterID), proxyURL, tlsConfig, upstream, arch, channel, optr.releaseVersion)
 
 	if usedDefaultUpstream {
 		upstream = ""
 	}
 	optr.setAvailableUpdates(&availableUpdates{
 		Upstream:  upstream,
 		Channel:   config.Spec.Channel,
+		Current:   current,
```
LalatenduMohanty, line 58: `setAvailableUpdates()` function does not use `Current` . Also I am not sure why we need `Current`?

pkg/autoupdate/autoupdate.go
wking: `Spec.DesiredUpdate` is an `Update`, but `nextUpdate` is a `Release`, because it pulls from `AvailableUpdates`.  We need to convert from `Release` to `Update` somewhere.

pkg/cvo/availableupdates.go
wking: `Current` will allow us to set `ClusterVersion.Status.Desired.URL`.  Will get that FIXME addressed today.

wking:
Pulling the WIP, because 3a84314f -> f666975b addresses the outstanding issues with populating the new Release properties.  Review is probably most convenient via `git show --word-diff=color ...`, because that helps isolate the bits I touched from the bits I didn't (event when an untouched string shifts lines).

pkg/cvo/cvo.go
```diff
@@ -223,17 +227,32 @@ func New(
...
 func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestConfig *rest.Config) error {
-	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.releaseImage, optr.exclude)
+	update, err := payload.LoadUpdate(optr.defaultPayloadDir(), optr.release.Image, optr.exclude)
 	if err != nil {
 		return fmt.Errorf("the local release contents are invalid - no current version can be determined from disk: %v", err)
 	}
-	// XXX: set this to the cincinnati version in preference
-	if _, err := semver.Parse(update.ImageRef.Name); err != nil {
-		return fmt.Errorf("the local release contents name %q is not a valid semantic version - no current version will be reported: %v", update.ImageRef.Name, err)
+	optr.release.Version = update.Metadata.Version
```
jottofar, line 234: Does `update.Metadata.Version` = `update.ReleaseVersion`? Wondering why they are two and diff if any.

pkg/cvo/cvo.go
```diff
@@ -223,17 +227,32 @@ func New(
...
+		} else {
+			klog.Warningf("the local release contents define a non-string URL: %v", urlInterface)
+		}
+	}
+	if channelsInterface, ok := update.Metadata.Metadata["io.openshift.upgrades.graph.release.channels"]; ok {
+		if channelsString, ok := channelsInterface.(string); ok {
+			optr.release.Channels = strings.Split(channelsString, ",")
+			sort.Strings(optr.release.Channels)
+		} else {
+			klog.Warningf("the local release contents define a non-string URL: %v", channelsInterface)
```
jottofar, line 247: "the local release contents define a non-string **_list of channels_**" or some such

pkg/cvo/cvo.go
```diff
@@ -223,17 +227,32 @@ func New(
...
+		}
+	}
+
```
jottofar, line 250: Shouldn't `optr.release.Version = update.Metadata.Version` be moved here

pkg/cvo/cvo.go
```diff
@@ -613,11 +636,26 @@ func versionString(update configv1.Update) string {
 }
 
 // currentVersion returns an update object describing the current known cluster version.
```
jottofar, line 638: it now returns a release but also has the new update logic

pkg/cvo/status.go
```diff
@@ -180,8 +181,16 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 
 	now := metav1.Now()
 	version := versionString(status.Actual)
+	desired := configv1.Release{
+		Version: status.Actual.Version,
+		Image:   status.Actual.Image,
+	}
+	if availableUpdates != nil && desired.Image == availableUpdates.Current.Image {
```
jottofar, line 188: Should check `availableUpdates` above where first set.

pkg/cvo/status.go
```diff
@@ -180,8 +181,16 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
+		desired.URL = availableUpdates.Current.URL
+		desired.Channels = availableUpdates.Current.Channels
```
jottofar, line 190: Are you just setting these for completeness, in case they are ever used in `mergeOperatorHistory` for example?

pkg/cvo/updatepayload.go
```diff
@@ -304,13 +304,21 @@ func findUpdateFromConfig(config *configv1.ClusterVersion) (configv1.Update, boo
 
 func findUpdateFromConfigVersion(config *configv1.ClusterVersion, version string, force bool) (configv1.Update, bool) {
 	for _, update := range config.Status.AvailableUpdates {
-		if update.Version == version {
-			return update, len(update.Image) > 0
+		if update.Version == version && len(update.Image) > 0 {
```
jottofar, line 307: So the semantics have changed -> if `Image` is not set do not return `Version`? And same applies for history check below.

pkg/cvo/cvo.go
wking: Yes, [they match][1], or we fail to load the release.

[1]: https://github.com/openshift/cluster-version-operator/commit/f666975b299ba42b9d2194ecbcac2e6cf62ce2f8#diff-8fadc5d8bd084451d026d5bf128dad66R263-R269

pkg/cvo/cvo.go
wking: > Shouldn't optr.release.Version = update.Metadata.Version be moved here

Why?  It's a few lines up, and `InitializeFromPayload` shouldn't be racing serious activity in other goroutines (because what would those routines be doing before we'd initialized ourselves?).

pkg/cvo/status.go
wking: Yup, because `mergeOperatorHistory` uses them today [for `status.desired`][1].

[1]: https://github.com/openshift/cluster-version-operator/blob/e49df43aaa578f0f420337bc28b7fc0f8cf35c09/pkg/cvo/status.go#L131

pkg/cvo/status.go
wking: There's no such guard [in master today][1], because `NeedsUpdate` [has a `nil` guard][2].  So I think we're ok.  This call is just "hey, let's look inside that `availableUpdates` we were getting before and, if the images match, flesh out `desired` with any available data.  Although... I guess that will miss information from the release image itself.  We probably want to use `optr.currentVersion()` if that's an image-match instead.

[1]: https://github.com/openshift/cluster-version-operator/blob/e49df43aaa578f0f420337bc28b7fc0f8cf35c09/pkg/cvo/status.go#L163
[2]: https://github.com/openshift/cluster-version-operator/blob/e49df43aaa578f0f420337bc28b7fc0f8cf35c09/pkg/cvo/availableupdates.go#L91-L93

pkg/cvo/cvo.go
wking: Fixed with f666975b -> 1316b58c

pkg/cvo/cvo.go
wking: Documented with f666975b -> 1316b58c.

pkg/cvo/updatepayload.go
wking: yeah.  Before if some crazy upstream service had declared a node with a matching version but no image, we'd have said "sorry, no match" without looking through history.  Now we say "skipping this useless entry" and move on to consider later entries and history.  Documented in the commit message with 1316b58c -> 33333ecb.

pkg/cvo/cvo.go
jottofar: Nothing to do with a race, just keeping it close to were it's being checked/used for first time which is typically a best-practice. But not anything critical so I'm okay with doing nothing if you prefer.

pkg/cvo/status.go
jottofar: Yeah, we went through this availableUpdates being nil thing before. I checked it and as I recall I determined that it can only be nil at startup/initialization which is probably why it wasn't originally being checked in here. I just saw that you had checked it down below which stood out so I'd pull that check.

wking: >  Although... I guess that will miss information from the release image itself...

I've pushed 33333ecb -> c1de00c3, pulling `mergeReleaseMetadata` out into a separate helper (with tests), pushing `Release`-from-payload calculations down into `pkg/payload` itself (so both of our `LoadUpdate` consumers can share it), and adding a `mergeReleaseMetadata` call in `syncStatus` to catch cases like "I'm updating to a new release, and still want to layer in any metadata from `availableUpdates`".

pkg/cvo/availableupdates.go
wking: Addressed in c1de00c39d.

pkg/cvo/availableupdates.go
```diff
@@ -187,40 +192,71 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
...
 			Message: fmt.Sprintf("Unable to retrieve available updates: %s", err),
 		}
 	}
 
-	var cvoUpdates []configv1.Update
+	cvoCurrent, err = convertCincinnatiUpdateToRelease(current)
+	if err != nil {
+		return cvoCurrent, nil, configv1.ClusterOperatorStatusCondition{
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "ResponseInvalid",
+			Message: fmt.Sprintf("Invalid Cinncinnati node: %s", err),
```
vrutkovs, line 221:
```suggestion
			Message: fmt.Sprintf("Invalid retrieved update node: %s", err),
```

We should avoid mentioning Cincinnati (:cry:) due to upcoming rename

pkg/cvo/availableupdates.go
```diff
@@ -187,40 +192,71 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
...
 		// this should never happen
-		return nil, configv1.ClusterOperatorStatusCondition{
+		return cvoCurrent, nil, configv1.ClusterOperatorStatusCondition{
 			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "Unknown",
 			Message: fmt.Sprintf("Unable to retrieve available updates: %s", err),
 		}
 	}
 
-	var cvoUpdates []configv1.Update
+	cvoCurrent, err = convertCincinnatiUpdateToRelease(current)
```
vrutkovs, line 217:
```suggestion
	cvoCurrent, err = convertRetreivedUpdateToRelease(current)
```

pkg/cvo/availableupdates.go
```diff
@@ -187,40 +192,71 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
...
+		}
+	}
+
+	var cvoUpdates []configv1.Release
 	for _, update := range updates {
-		cvoUpdates = append(cvoUpdates, configv1.Update{
-			Version: update.Version.String(),
-			Image:   update.Image,
-		})
+		cvoUpdate, err := convertCincinnatiUpdateToRelease(update)
```
vrutkovs, line 227:
```suggestion
		cvoUpdate, err := convertRetreivedUpdateToRelease(update)
```

pkg/cvo/availableupdates.go
```diff
@@ -187,40 +192,71 @@ func calculateAvailableUpdatesStatus(clusterID string, proxyURL *url.URL, tlsCon
...
-	return cvoUpdates, configv1.ClusterOperatorStatusCondition{
+	return cvoCurrent, cvoUpdates, configv1.ClusterOperatorStatusCondition{
 		Type:   configv1.RetrievedUpdates,
 		Status: configv1.ConditionTrue,
 
 		LastTransitionTime: metav1.Now(),
 	}
 }
+
+func convertCincinnatiUpdateToRelease(update cincinnati.Update) (configv1.Release, error) {
```
vrutkovs, line 245:
```suggestion
func convertRetreivedUpdateToRelease(update cincinnati.Update) (configv1.Release, error) {
```

pkg/cvo/status.go
```diff
@@ -159,8 +159,9 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 	klog.V(5).Infof("Synchronizing errs=%#v status=%#v", validationErrs, status)
 
 	cvUpdated := false
+	availableUpdates := optr.getAvailableUpdates()
 	// update the config with the latest available updates
-	if updated := optr.getAvailableUpdates().NeedsUpdate(config); updated != nil {
+	if updated := availableUpdates.NeedsUpdate(config); updated != nil {
```
vrutkovs, line 164: Does `availableUpdates` need to kept around?

pkg/payload/payload.go
```diff
@@ -139,19 +162,19 @@ func LoadUpdate(dir, releaseImage, excludeIdentifier string) (*Update, error) {
 
 			raw, err := ioutil.ReadFile(p)
 			if err != nil {
-				errs = append(errs, errors.Wrapf(err, "error reading file %s", file.Name()))
+				errs = append(errs, err)
```
vrutkovs, line 165: Why would we want a less specific error here?

pkg/cvo/availableupdates.go
wking: We are renaming our implementation.  The protocol remains "Cincinnati", and the CVO only cares about the protocol, not its backing implementation.  I'd be fine with `recommended update node` though, since that seems more intuitive than mentioning Cincinnati.  I'd rather avoid `retrieved` since that might give a false `image retrieved` impression.  Thoughts?

pkg/cvo/availableupdates.go
wking: This is internal, s)not user-facing, so I think it should stay per [here][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/419#discussion_r462978410

pkg/cvo/status.go
wking: oops, need to drop this hunk with the most-recent reroll.

pkg/payload/payload.go
wking: `ReadFile` errors already include this information.

pkg/cvo/availableupdates.go
vrutkovs: Oh, right, "recommended update" sounds good to me

pkg/cvo/availableupdates.go
LalatenduMohanty: `convertRetreivedUpdateToRelease` is a better name than convertCincinnatiUpdateToRelease IMO.

wking:
> integration test fails with CVO refusing to pull available updates and throwing `NoCurrentVersion`

I was missing an `optr.release = update.Release` in `InitializeFromPayload`.  Fixed as part of 0a0483ae -> cd2d7013, which also fixes a few other nits and rebases on top of the current #420 tip.

pkg/cvo/availableupdates.go
wking: Done as part of 0a0483ae -> cd2d7013.

pkg/cvo/availableupdates.go
wking: 2 votes for renaming vs. my one, so renamed with 0a0483ae -> cd2d7013.

pkg/cvo/availableupdates.go
wking: renamed with 0a0483a -> cd2d701.

pkg/cvo/availableupdates.go
wking: renamed with 0a0483a -> cd2d701.

pkg/cvo/status.go
wking: dropped the hunk with 0a0483a -> cd2d701.

wking:
[e2e][1] failed on `an end user can use OLM can subscribe to the operator`, which is the unrelated [rhbz#1862322][2].

/override ci/prow/e2e

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/420/pull-ci-openshift-cluster-version-operator-master-e2e/1289021479937118208

LalatenduMohanty:
@wking  The commit message in https://github.com/openshift/cluster-version-operator/pull/419/commits/58356deee675c900b7cd4742ceefc0e6a84d7efb is too verbose IMO and hard to read/understand. If you can reduce the text and make it more readable, it would be good.

pkg/cvo/cvo.go
```diff
@@ -611,12 +612,50 @@ func versionString(update configv1.Update) string {
...
+// are used as fallbacks for any properties not defined in the release
+// image itself.
+func (optr *Operator) currentVersion() configv1.Release {
+	return optr.mergeReleaseMetadata(optr.release)
+}
+
+// mergeReleaseMetadata returns a deep copy of the input release.
+// Values from any matching availableUpdates release are used as
+// fallbacks for any properties not defined in the input release.
+func (optr *Operator) mergeReleaseMetadata(release configv1.Release) configv1.Release {
```
LalatenduMohanty, line 626: I would have named this as `updateReleaseMetadata()` but we can handle this in another PR.

wking:
> ...too verbose IMO and hard to read/understand...

The commit diff is also verbose and hard to understand.  I'd like to explain my motivation behind the fiddly bits.  Did you have specific sections which need rewording?

pkg/cvo/cvo.go
wking: It doesn't update in place, it returns a new object merging data from two sources

LalatenduMohanty:
> The commit diff is also verbose and hard to understand. I'd like to explain my motivation behind the fiddly bits. Did you have specific sections which need rewording?

```
The next/last Actual comparison avoids:

  invalid operation: next.Actual == last.Actual (struct containing []string cannot be compared)

The image pullspec is sufficient for that matching.  The test suite's
expected LastProgress changes are because the shift from:

  {Version: "4.0.1", Image: "image/image:1"}

to:

  {Version: "1.0.0-abc", Image: "image/image:1"}

no longer counts as a LastProgress-bumping change (because we're now
only comparing the Image pullspecs).
```

pkg/cvo/cvo.go
```diff
@@ -87,13 +86,13 @@ type Operator struct {
...
-	// releaseImage is the image the current operator points to and allows
-	// templating of the CVO deployment manifest.
-	releaseImage string
-	// releaseVersion is a string identifier for the current version, read
-	// from the image of the operator. It may be empty if no version exists, in
-	// which case no available updates will be returned.
-	releaseVersion string
+	// release is the release the current operator points to and
+	// metadata read from the release image.  It allows templating of
+	// the CVO deployment manifest.
```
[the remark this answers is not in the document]
wking: This is internal state not directly related to external APIs.  Rewording suggestions?

wking:
> The next/last Actual comparison avoids:...

Is that pivot obvious without an explanation?  Or is the explanation unclear?

LalatenduMohanty:
> Is that pivot obvious without an explanation? Or is the explanation unclear?

It is just hard to understand the context and why the text is there.

wking:
Context is changes like [this][1], which you should be able to find in the diff by searching for the mentioned `next.Actual == last.Actual`, and the remaining text explains (or attempts to explain) why the move to `Release` forced me to make a change, and why I chose the change I did.

[1]: https://github.com/openshift/cluster-version-operator/commit/58356deee675c900b7cd4742ceefc0e6a84d7efb#diff-b2099007e7494e37fc4f7a4320f3622bL1003

LalatenduMohanty:
> Context is changes like [this](https://github.com/openshift/cluster-version-operator/commit/58356deee675c900b7cd4742ceefc0e6a84d7efb#diff-b2099007e7494e37fc4f7a4320f3622bL1003), which you should be able to find in the diff by searching for the mentioned `next.Actual == last.Actual`, and the remaining text explains (or attempts to explain) why the move to `Release` forced me to make a change, and why I chose the change I did.

I still cannot understand why this needs to go to the commit message.

wking:
> I still cannot understand why this needs to go to the commit message.

So in a year when we reroll this, we remember the motivation ;)

wking:
[e2e][1] had som Prom issues.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/419/pull-ci-openshift-cluster-version-operator-master-e2e/1289299837585461248

pkg/cvo/sync_worker.go
```diff
@@ -475,54 +478,81 @@ func (w *SyncWorker) Status() *SyncWorkerStatus {
...
 		payloadUpdate.VerifiedImage = info.Verified
 		payloadUpdate.LoadedAt = time.Now()
 
+		if work.Desired.Version == "" {
+			work.Desired.Version = payloadUpdate.Release.Version
+			desired.Version = payloadUpdate.Release.Version
+		} else if payloadUpdate.Release.Version != work.Desired.Version {
+			err = fmt.Errorf("release image version %s does not match the expected upstream version %s", payloadUpdate.Release.Version, work.Desired.Version)
+			w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeWarning, "VerifyPayloadVersionFailed", "verifying payload failed version=%q image=%q failure=%v", work.Desired.Version, work.Desired.Image, err)
+			/* FIXME: Ignore for now.  I will make this fatal in a follow-up pivot
```
wking, line 542: I've opened #431 to address this.

PR #421, opened 2020-07-29
pkg/cvo/cvo_test: Distinguish names for the "single clusteroperator" cases

wking:
The three cases landed with the same name in 04528144fe (#243).  The new names make the differences, which all involve the presence and state of `default-operator-2`, more obvious.

### DIFF
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -2741,7 +2741,7 @@ func TestOperator_upgradeableSync(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


wking:
[Update][1] died on the unrelated `Application behind service load balancer with PDB is not disrupted`.

/override ci/prow/e2e-upgrade

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/421/pull-ci-openshift-cluster-version-operator-master-e2e-upgrade/1288766724526575616

PR #422, opened 2020-07-31
Bug 1862524: pkg/cvo/status: Raise Operator leveling grace-period to 20 minutes

wking:
Reduce false-positives when operators take a while to level (like the machine-config operator, which has to roll the control plane machines).  We may want to raise this further in the future, but baby steps ;).


sdodson:
/retitle Bug 1862524: pkg/cvo/status: Raise Operator leveling grace-period to 20 minutes

sdodson:
/hold
We'll wait in the queue behind other 4.6 feature work since CI is slammed right now.

wking:
Manually backported in #427.

PR #423, opened 2020-08-02
pkg/cincinnati/cincinnati: Change "installed" -> "reconciling" for VersionNotFound

wking:
When you initiate an update from A -> B, if B is not found in the current channel, the cluster would start reporting VersionNotFound after it's first attempt to retrieve recommended updates for B.  That usually happens before the update completes, and the old "currently installed version" wasn't a very accurate description of B mid-update.  The new wording relaxes our claim about the cluster state (maybe it has been reconciled, maybe not) and just talks about the version we're attempting to reconcile.

### DIFF
--- a/docs/user/status.md
+++ b/docs/user/status.md
@@ -117,9 +117,9 @@ It could also be caused by response corruption, e.g. if the configured `upstream
 
 ### VersionNotFound
 
-The currently installed version was not found in the configured `channel`.
+The currently reconciling cluster version was not found in the configured `channel`.
 
-This usually means that the configured `channel` is known to Cincinnati, but the cluster's current version is not found in that channel's graph.
+This usually means that the configured `channel` is known to Cincinnati, but the version the cluster is currently applying is not found in that channel's graph.
 Fix by setting `channel` to [a valid value][channels], e.g. `stable-4.3`.
 
 If this error occurs because you forced an update to a release that is not in any channel, fix by updating back to a release that occurs in a channel, although you are on your own to determine a safe update path.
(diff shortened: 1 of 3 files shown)


docs/user/status.md
```diff
@@ -117,9 +117,9 @@ It could also be caused by response corruption, e.g. if the configured `upstream
 
 ### VersionNotFound
 
-The currently installed version was not found in the configured `channel`.
+The currently reconciling version was not found in the configured `channel`.
```
LalatenduMohanty, line 120: Suggestion: `The version CVO is reconciling not found in the configured channel`.

vrutkovs: I'm fine with both (the first one sounds less complex though)

wking: I would rather not use the CVO acronym in user-facing condition messages (which this doc entry just echos).  If you feel like my wording here is ambiguous, how about:

> The currently reconciling cluster version was not found in the configured `channel`.

LalatenduMohanty: `The currently reconciling cluster version was not found in the configured channel` sounds good to me.

LalatenduMohanty: Also intent was to be more explicit. Not using `CVO` as an acronym is a good idea. However the doc already have `CVO` in the text.

wking: This line appears in both the doc and the condition message.  The CVO references should appear only in the doc.  Will update In a bit.

wking: Rerolled to use "currently reconciling cluster version" with 9f15457f -> ea6899ae.

PR #11, opened 2018-08-22
Add: autoupdate controller

abhinavdahiya:
autoupdate controller sets the DesiredUpdate for CVOConfig to the greatest available version from OperatorStatus AvailabeUpdates.

Requires #10 
/cc @crawford @smarterclayton 
/assign @crawford @smarterclayton

PR #425, opened 2020-08-05
pkg/cvo: Fix maxRetries comment (not about machineconfig pools)

wking:
This seems to have been copy-pasted from the machine-config operator in 6c8f725f5d (#7) and then from there over to the auto-update package with ad4bd6c428 (#11).  Update to generic wording that is less likely to go stale.

### DIFF
--- a/pkg/autoupdate/autoupdate.go
+++ b/pkg/autoupdate/autoupdate.go
@@ -27,11 +27,7 @@ import (
 )
 
 const (
-	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
-	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
-	// a machineconfig pool is going to be requeued:
-	//
-	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
+	// maxRetries is the number of times a work-item will be retried before it is dropped out of the queue.
 	maxRetries = 15
 )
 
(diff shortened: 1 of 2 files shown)


pkg/autoupdate/autoupdate.go
```diff
@@ -27,11 +27,7 @@ import (
 )
 
 const (
-	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
-	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
-	// a machineconfig pool is going to be requeued:
-	//
-	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
+	// maxRetries is the number of times a work-item be retried before it is dropped out of the queue.
```
jottofar, line 30: // maxRetries is the number of times a work-item `will` be retried before it is dropped out of the queue.

pkg/cvo/cvo.go
```diff
@@ -52,11 +52,7 @@ import (
 )
 
 const (
-	// maxRetries is the number of times a machineconfig pool will be retried before it is dropped out of the queue.
-	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
-	// a machineconfig pool is going to be requeued:
-	//
-	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
+	// maxRetries is the number of times a work-item be retried before it is dropped out of the queue.
```
jottofar, line 55: // maxRetries is the number of times a work-item `will` be retried before it is dropped out of the queue.

pkg/autoupdate/autoupdate.go
wking: Thanks; fixed with ae0b69b3 -> 69b459ef.

pkg/cvo/cvo.go
wking: https://github.com/openshift/cluster-version-operator/pull/425/files#r465981007

PR #426, opened 2020-08-05
Bug 1843505: Have the leader election use faster defaults

deads2k:
This resolves a problem where status doesn't reflect the payload actually being applied.
wking is chasing an alternative using different graceful leader election rules.
It would also be resolved by synchronously writing the status out when we start applying a new payload.

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -43,9 +43,13 @@ const (
 
 	minResyncPeriod = 2 * time.Minute
 
-	leaseDuration = 90 * time.Second
-	renewDeadline = 45 * time.Second
-	retryPeriod   = 30 * time.Second
+	// TODO set the slower lease values after we're able to reliably shut down gracefully.  This fixes the slow status update for now.
+	//leaseDuration = 90 * time.Second
+	//renewDeadline = 45 * time.Second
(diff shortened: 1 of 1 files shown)


pkg/start/start.go
```diff
@@ -211,10 +211,11 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 	// moved back inside RunOrDie
 	// TODO: properly wire ctx here
 	go leaderelection.RunOrDie(context.TODO(), leaderelection.LeaderElectionConfig{
-		Lock:          lock,
-		LeaseDuration: leaseDuration,
-		RenewDeadline: renewDeadline,
-		RetryPeriod:   retryPeriod,
+		Lock: lock,
+		// TODO set the slower lease values after we're able to reliably shut down gracefully.  This fixes the slow status update for now.
```
smarterclayton, line 215: Set constants instead of using defaults.

I would prefer halving / thirding our current vs this more dramatic change.

wking:
/retitle Bug 1843505: Have the leader election use faster defaults

Not a fix for the bug, but it will be related to the bug and something we end up backporting if we backport the bugfix to previous versions.

PR #427, opened 2020-08-05
Bug 1866480: pkg/cvo/status: Raise Operator leveling grace-period to 20 minutes

wking:
Backporting #422 and, to avoid rebase conflicts, the doc-only #371 to 4.5.

### DIFF
--- a/docs/user/reconciliation.md
+++ b/docs/user/reconciliation.md
@@ -93,22 +93,22 @@ So the graph nodes are all parallelized with the by-number ordering flattened ou
 
 For the usual reconciliation loop (neither an upgrade between releases nor a fresh install), the flattened graph is also randomly permuted to avoid hanging on ordering bugs.
 
-## Synchronizing the graph
+## Reconciling the graph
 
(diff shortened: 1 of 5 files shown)


docs/user/reconciliation.md
```diff
@@ -93,22 +93,22 @@ So the graph nodes are all parallelized with the by-number ordering flattened ou
 
 For the usual reconciliation loop (neither an upgrade between releases nor a fresh install), the flattened graph is also randomly permuted to avoid hanging on ordering bugs.
 
-## Synchronizing the graph
+## Reconciling the graph
```
vrutkovs, line 96:
```suggestion
## Reconciling the cluster version
```
nit: "graph" sounds a bit ambiguous to me  - I've immediately though of "updates graph" instead "manifests graph"

wking: This is a backport PR.  Drop an `/lgtm`, and then PR master with your change to "manifest graph", which sounds good to me.

PR #430, opened 2020-08-06
lib/resourcebuilder/apps: Only error on Deployment Available=False *and* Progressing=False

wking:
`Available=True`, `Progressing=False` is the happy, steady state.
`Available=True`, `Progressing=True` is a happy update.
`Available=False`, `Progressing=True` is acceptable outage, e.g. [during an update][1] with the `Recreate` strategy:

```console
$ curl -s https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-upgrade-4.6/1291426211527921664/artifacts/e2e-gcp-upgrade/container-logs/test.log | grep MinimumReplicasUnavailable | head -n1
Aug  6 17:56:00.674: INFO: deployment status: v1.DeploymentStatus{ObservedGeneration:1, Replicas:1, UpdatedReplicas:1, ReadyReplicas:0, AvailableReplicas:0, UnavailableReplicas:1, Conditions:[]v1.DeploymentCondition{v1.DeploymentCondition{Type:"Available", Status:"False", LastUpdateTime:v1.Time{Time:time.Time{wall:0x0, ext:63732333358, loc:(*time.Location)(0x9e74040)}}, LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63732333358, loc:(*time.Location)(0x9e74040)}}, Reason:"MinimumReplicasUnavailable", Message:"Deployment does not have minimum availability."}, v1.DeploymentCondition{Type:"Progressing", Status:"True", LastUpdateTime:v1.Time{Time:time.Time{wall:0x0, ext:63732333358, loc:(*time.Location)(0x9e74040)}}, LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63732333358, loc:(*time.Location)(0x9e74040)}}, Reason:"ReplicaSetUpdated", Message:"ReplicaSet \"dp-7f9df745ff\" is progressing."}}, CollisionCount:(*int32)(nil)}
```

`Available=False`, `Progressing=False` is the Deployment controller saying "I cannot deliver my expected service level for this Deployment", so that's when we should be complaining.  Fixes noise like:

```
  Aug  6 18:03:00.500: INFO: cluster upgrade is Failing: Multiple errors are preventing progress:
  * Could not update namespace "openshift-service-ca-operator" (467 of 608)
  * deployment openshift-cluster-machine-approver/machine-approver is not available MinimumReplicasUnavailable: Deployment does not have minimum availability.
  * deployment openshift-ingress-operator/ingress-operator is not available MinimumReplicasUnavailable: Deployment does not have minimum availability.
```

(the namespace part of that message is a separate issue).

[1]: https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-upgrade-4.6/1291426211527921664

### DIFF
--- a/lib/resourcebuilder/apps.go
+++ b/lib/resourcebuilder/apps.go
@@ -114,20 +114,11 @@ func (b *builder) checkDeploymentHealth(ctx context.Context, deployment *appsv1.
 		}
 	}
 
-	if availableCondition != nil && availableCondition.Status == corev1.ConditionFalse {
+	if availableCondition != nil && availableCondition.Status == corev1.ConditionFalse && progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse {
 		return &payload.UpdateError{
-			Nested:  fmt.Errorf("deployment %s is not available; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedReplicas, d.Status.Replicas, d.Status.AvailableReplicas, d.Status.Replicas),
+			Nested:  fmt.Errorf("deployment %s is not available and not progressing; updated replicas=%d of %d, available replicas=%d of %d", iden, d.Status.UpdatedReplicas, d.Status.Replicas, d.Status.AvailableReplicas, d.Status.Replicas),
 			Reason:  "WorkloadNotAvailable",
-			Message: fmt.Sprintf("deployment %s is not available %s: %s", iden, availableCondition.Reason, availableCondition.Message),
-			Name:    iden,
-		}
-	}
-
-	if progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse {
-		return &payload.UpdateError{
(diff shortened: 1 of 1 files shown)


smarterclayton:
I think this looks right, I'm going to trigger an upgrade via clutser bot between 4.5 and this.

smarterclayton:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1291465594092130304 should not display the error

smarterclayton:
Logs in that test run look correct

wking:
In this space is the upstream kubernetes/kubernetes#93933 fix.  In the meantime, the CVO will continue to assume that Deployments which fall into that hole are happy enough, even if they should have been set `ProgressDeadlineExceeded` and blocked the CVO's manifest graph application.  It's hard to work around that in the CVO though, without implementing our own deployment controller that monitors backing ReplicaSets directly.  I think we should ignore the issue for now, and carry the upstream patch in our local deployment controllers instead of carrying a workaround in the CVO.

I think this PR is orthogonal enough that it should not block on any of the above paragraph getting addressed.

smarterclayton:
Agree, we should carry the fix to the workload controller for error detection, not CVO

PR #434, opened 2020-08-13
Removing EnsureObjectMeta as it is called in ensurePodTemplateSpec()

LalatenduMohanty:
Removing EnsureObjectMeta as it is called in ensurePodTemplateSpec() again.


wking:
The subsequent call is:

```go
ensurePodTemplateSpec(modified, &existing.Spec.Template, required.Spec.Template)
```

So I don't see a way it could be merging the deployment's metadata.

PR #438, opened 2020-08-17
OTA-222: Add a manifest annotation to be used for object deletion

jottofar:
The new manifest annotation will trigger new CVO delete logic, added by this PR, to delete the in-cluster object instead of creating/updating it. This will provide a more straightforward way for engineering to remove content.

See [Enhancement for adding a manifest annotation for object removal
#419](https://github.com/openshift/enhancements/pull/419) for implementation details.

Per the enhancement whenever object deletions are pending, one or more deletions have been requested but object removal has not completed, the cluster is blocked from upgrading to the next minor release (`Upgradeable==false`).

if an object previously removed reappears a warning is logged which meets what is specified by the enhancement. This does not result in `Upgradeable==false` and no attempt is made to remove the object again.

Regarding implementation of enhancement section [Upgrade / Downgrade Strategy](https://github.com/jottofar/enhancements/blob/ab61123d74aeb90ebef5a8964a6d1686946f084b/enhancements/update/object-removal-manifest-annotation.md#upgrade--downgrade-strategy), since an upgrade can result in >1 CVO restarts there is no straightforward way to distinguish the cause of an IsNotFound error. That is, discerning whether a given object being requested for deletion was ever present in the release or had been removed by a previous CVO pod. A warning is logged whenever an object specified for deletion is not found, and no indication exists in the current CVO pod that an object a delete request had been issued.

Since k8s.io v1beta1 APIs are no longer served In kuberentes v1.22 and are being removed via [#545](https://github.com/openshift/cluster-version-operator/pull/545) delete is not implemented for these types. Users are expected to use v1 apiVersion in these delete manifests.

### DIFF
--- a/docs/dev/object-deletion.md
+++ b/docs/dev/object-deletion.md
@@ -1,0 +1,110 @@
+# Manifest Annotation For Object Deletion
+
+Developers can remove any of the currently managed CVO objects by modifying an existing manifest and adding the delete annotation `.metadata.annotations["release.openshift.io/delete"]="true"`. This manifest annotation is a request for the CVO to delete the in-cluster object instead of creating/updating it.
+
+Actual object deletion and subsequent deletion monitoring and status checking will only occur during an upgrade. During initial installation the delete annotation prevents further processing of the manifest since the given object should not be created.
+
+## Implementation Details
+
+When the following annotation appears in a CVO supported manifest and is set to `true` the associated object will be removed from the cluster by the CVO.
+Values other than `true` will result in a CVO failure and should therefore result in CVO CI failure.
+```yaml
+apiVersion: apps/v1
+...
+metadata:
+...
+  annotations:
+    release.openshift.io/delete: "true"
+```
+The existing CVO ordering scheme defined [here](operators.md) is also used for object removal. This provides a simple and familiar method of deleting multiple objects by reversing the order in which they were created. It is the developer's responsibility to ensure proper deletion ordering and to ensure that all items originally created by an object are deleted when that object is deleted. For example, an operator may have to be modified, or a new operator created, to take explicit actions to remove external resources. The modified or new operator would then be removed in a subsequent update.
+
+Similar to how CVO handles create/update requests, deletion requests are implemented in a non-blocking manner whereby CVO issues the initial request to delete an object kicking off resource finalization and after which resource removal. CVO does not wait for actual resource removal but instead continues. CVO logs when a delete is initiated, that the delete is ongoing when a manifest is processed again and found to have a deletion time stamp, and delete completion upon resource finalization.
+
+If an object cannot be successfully removed CVO will set `Upgradeable=False` which in turn blocks cluster update to the next minor release.
+
+## Examples
+
+The following examples provide guidance to OpenShift developers on how resources may be removed but this will vary depending on the component.
+Ultimately it is the developer's responsibility to ensure the removal works by thoroughly testing.
+In all cases, and as general guidance, an operator should never allow itself to be removed if the operator's operand has not been removed.
+
+### The autoscaler operator
+
+Remove the cluster-autoscaler-operator deployment.
+The existing cluster-autoscaler-operator deployment manifest 0000_50_cluster-autoscaler-operator_07_deployment.yaml is modified to contain the delete annotation:
+```yaml
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-autoscaler-operator
+  namespace: openshift-machine-api
+  annotations:
+    release.openshift.io/delete: "true"
+...
+```
+Additional manifest properties such as `spec` may be set if convenient (e.g. because you are looking to make a minimal change vs. a previous version of the manifest), but those properties have no affect on manifests with the delete annotation.
+
+### The service-catalog operators
+
+In release 4.5 two jobs were created to remove the Service Catalog - openshift-service-catalog-controller-manager-remover and openshift-service-catalog-apiserver-remover.
+In release 4.6, these jobs and all their supporting cluster objects also needed to be removed.
+The following example shows how to do Service Catalog removal using the object deletion manifest annotation.
+
+The Service Catalog is composed of two components, the cluster-svcat-apiserver-operator and the cluster-svcat-controller-manager-operator.
+Each of these components use manifests for creation/update of the component's required resources: namespace, roles, operator deployment, etc.
+The cluster-svcat-apiserver-operator had [these associated manifests][svcat-apiserver-4.4-manifests].
+
+The deletion annotation would be added to these manifests:
+
+* `0000_50_cluster-svcat-apiserver-operator_00_namespace.yaml` containing the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_50_cluster-svcat-apiserver-operator_02_config.crd.yaml` containing a cluster-scoped CRD.
+* `0000_50_cluster-svcat-apiserver-operator_03_config.cr.yaml` containing a cluster-scoped, create-only ServiceCatalogAPIServer.
+* `0000_50_cluster-svcat-apiserver-operator_04_roles.yaml` containing a cluster-scoped ClusterRoleBinding.
+* `0000_50_cluster-svcat-apiserver-operator_08_cluster-operator.yaml` containing a cluster-scoped ClusterOperator.
+
+These manifests would be dropped because their removal would occur as part of one of the above resource deletions:
+
+* `0000_50_cluster-svcat-apiserver-operator_03_configmap.yaml` containing a ConfigMap in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_50_cluster-svcat-apiserver-operator_03_version-configmap.yaml` containing another ConfigMap in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_50_cluster-svcat-apiserver-operator_05_serviceaccount.yaml` containing a ServiceAccount in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_50_cluster-svcat-apiserver-operator_06_service.yaml` containing a Service in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_50_cluster-svcat-apiserver-operator_07_deployment.yaml` containing a Deployment in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_90_cluster-svcat-apiserver-operator_00_prometheusrole.yaml` containing a Role in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_90_cluster-svcat-apiserver-operator_01_prometheusrolebinding.yaml` containing a RoleBinding in the `openshift-service-catalog-apiserver-operator` namespace.
+* `0000_90_cluster-svcat-apiserver-operator_02-operator-servicemonitor.yaml` containing a ServiceMonitor in the `openshift-service-catalog-apiserver-operator` namespace.
+
+So the remaining manifests with deletion annotations would be the namespace and the cluster-scoped CRD, ServiceCatalogAPIServer, ClusterRoleBinding, and ClusterOperator.
+The ordering of the surviving manifests would not be particularly important, although keeping the namespace first would avoid removing the ClusterRoleBinding while the consuming Deployment was still running.
+If multiple deletions are required, it is up to the developer to name the manifests such that deletions occur in the correct order.
+
+Similar handling would be required for the svcat-controller-manager operator.
+
+If resources external to kubernetes must be removed the developer must provide the means to do so.
+This is expected to be done through modification of an operator to do the removals during it's finalization.
+If operator modification for object removal is necessary that operator would be deleted in a subsequent update.
+
+The deletion manifests described above would have been preserved through 4.5.z release and removed in 4.6.
+See the [Subsequent Releases Strategy](#subsequent-releases-strategy) section for details.
+
+## Removing functionality that users might notice
+
+Below is the flow for removing functionality that users might notice, like the web console.
+
+* The first step is deprecating the functionality.
+    During the deprecation release 4.y, the functionality should remain available, with the operator setting Upgradeable=False and linking release notes like [these][deprecated-marketplace-apis].
+* Cluster Administrators must follow the linked release notes to opt in to the removal before updating to the next minor release in 4.(y+1).
+    When the administrator opts in to the removal, the operator should stop setting Upgradeable=False.
+* Depending on how the engineering team that owns the functionality implements its removal, the operand components may be removed when Cluster Administrators opt-in to the removal, or they may be left running and be removed during the transition to the next minor release 4.(y+1).
+* During the update to the next minor release 4.(y+1) the manifest delete annotation would be used to remove the operator, and, if they have not already been removed as part of the opt-in in release 4.y, any remaining operand components.
+
+## Subsequent Releases Strategy
+
+Special consideration must be given to subsequent updates which may still contain manifests with the delete annotation.
+These manifests will result in `object no longer exists` errors assuming the current release had properly and fully removed the given objects.
+It is acceptable for subsequent z-level updates to still contain the delete manifests but minor level updates should not and therefore the handling of the delete error will differ between these update levels.
+A z-level update will be allowed to proceed while a minor level update will be blocked.
+This will be accomplished through the existing CVO precondition mechanism which already behaves in this manner with regard to z-level and minor updates.
+
+[svcat-apiserver-4.4-manifests]: https://github.com/openshift/cluster-svcat-apiserver-operator/tree/aa7927fbfe8bf165c5b84167b7c3f5d9cb394e14/manifests
+[deprecated-marketplace-apis]: https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-marketplace-apis-deprecated
--- a/lib/resourcedelete/apiext.go
+++ b/lib/resourcedelete/apiext.go
@@ -1,0 +1,41 @@
+package resourcedelete
+
+import (
+	"context"
+	"fmt"
+
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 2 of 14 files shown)


jottofar:
/retitle Add a manifest annotation to be used for object deletion

LalatenduMohanty:
> if an object previously removed reappears a warning is logged which meets what is specified by the enhancement. This does not result in `Upgradeable==false` and no attempt is made to remove the object again.

Do we need to update the option 1 in the enhancement PR about this. ? https://github.com/jottofar/enhancements/blob/ab61123d74aeb90ebef5a8964a6d1686946f084b/enhancements/update/object-removal-manifest-annotation.md#finalization-option-1 

Not setting `Upgradeable==false` is fine with me in this case.  

Also as per the enhancment `If an object cannot be successfully removed CVO will set Upgradeable=false which in turn blocks cluster update to the next minor release.` I do not think we should do that. I do not see a reason why we will set `Upgradeable=false ` .

wking:
> Also as per the enhancment `If an object cannot be successfully removed CVO will set Upgradeable=false which in turn blocks cluster update to the next minor release`. I do not think we should do that. I do not see a reason why we will set `Upgradeable=false`.

Because potentially the existence of the resource would break something in the next minor release.  If nothing else, that next minor might remove the deletion manifest (because we don't want to have to carry these forever), and we don't want to leak a resource that we meant to delete.

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
+# Manifest Annotation For Object Deletion
+
+Developers can remove any of the currently managed CVO objects by modifying an existing manifest and adding the delete annotation .metadata.annotations["release.openshift.io/delete"]="true". This manifest annotation requests the CVO delete the in-cluster object instead of creating/updating it.
```
wking, line 3: nit: backticks around the code literal: `.metadata.annotations["release.openshift.io/delete"]="true"`

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+
+Actual object deletion and subsequent deletion monitoring and status checking only occur during an upgrade. During initial installation the delete annotation prevents further processing of the manifest since the given object should not be created.
+
+## Implementation Details
+
+When the following annotation appears in a CVO supported manifest and is set to "true" the associated object will be removed from the cluster by the CVO.
```
wking, line 9: nit: "true" -> `true` to match the subsequent line.

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+
+Similar to how CVO handles create/update requests, deletion requests are implemented in a non-blocking manner whereby CVO issues the initial request to delete an object kicking off resource finalization and after which resource removal. CVO does not wait for actual resource removal but instead continues. CVO logs when a delete is initiated, that the delete is ongoing when a manifest is processed again and found to have a deletion time stamp, and delete completion upon resource finalization.
+
+If an object cannot be successfully removed CVO will set `Upgradeable=false` which in turn blocks cluster update to the next minor release.
```
wking, line 23: nit: `Upgradeable=false` -> [`Upgradeable=False`][1]

[1]: https://github.com/openshift/api/blob/25175d9d392d6d858ff2da6a0117d751d5c9927c/config/v1/types_cluster_operator.go#L105

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+* Cluster Administrators must follow the linked release notes to opt in to the removal before updating to the next minor release in 4.(y+1).
+    When the administrator opts in to the removal, the operator should stop setting Upgradeable=False.
+* The operand may be removed when Cluster Administrators opt-in to the removal, or it may be left running and be removed during the transition to the next minor release 4.(y+1).
+* The update to the next minor release 4.(y+1) may use the manifest delete annotation to remove the operator, and, if they have not already been removed, any remaining operand components.
+
+## Subsequent Releases Strategy
+
+Special consideration must be given to subsequent updates which may still contain manifests with the delete annotation.
+These manifests will result in `object no longer exists` errors assuming the current release had properly and fully removed the given objects.
+This enhancement proposes that it is acceptable for subsequent z-level updates to still contain the delete manifests but minor level updates should not and therefore the handling of the delete error will differ between these update levels.
```
wking, line 113: nit: these local docs are no longer an enhancement proposal.  Maybe just drop the beginning and start with "It is acceptable..."?

pkg/cvo/internal/generic.go
```diff
@@ -29,6 +30,32 @@ func readUnstructuredV1OrDie(objBytes []byte) *unstructured.Unstructured {
...
 
+func deleteUnstructured(ctx context.Context, client dynamic.ResourceInterface, required *unstructured.Unstructured) (bool, error) {
+	if required.GetName() == "" {
+		return false, fmt.Errorf("Error running delete, invalid object: name cannot be empty")
+	}
+	if delAnnoFound, err := resourcedelete.ValidDeleteAnnotation(required.GetAnnotations()); !delAnnoFound || err != nil {
+		return delAnnoFound, err
+	}
+	resource := resourcedelete.Resource{
+		Kind:      "unstructured",
```
wking, line 41: Shouldn't we pull this out of `required`?

lib/resourcedelete/apiext.go
```diff
@@ -0,0 +1,41 @@
+package resourcedelete
```
wking, line 1: Did we want to auto-generate any of this in [`generate-lib-resource`][1]?  Or do these need enough hand-touching to make them annoying to automate?

[1]: https://github.com/openshift/cluster-version-operator/blob/7e7544097149708190903b1f4d088a4529704886/hack/generate-lib-resources.py#L7-L8

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -61,15 +63,25 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
-	up := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, configv1.OperatorUpgradeable)
-	if up == nil {
-		klog.V(4).Infof("Precondition %s passed: no Upgradeable condition on ClusterVersion.", pf.Name())
-		return nil
-	}
-	if up.Status != configv1.ConditionFalse {
-		klog.V(4).Infof("Precondition %s passed: Upgradeable %s since %v: %s: %s", pf.Name(), up.Status, up.LastTransitionTime, up.Reason, up.Message)
-		return nil
+	var reason, msg string
+	if deletes := resourcedelete.DeletesInProgress(); len(deletes) > 0 {
```
wking, line 67: I'd expect this `Run` code to continue to look at the `Upgradeable` condition on the ClusterVersion, and the new `DeletesInProgress` entries to feed that ClusterVersion condition around [here][1].

[1]: https://github.com/openshift/cluster-version-operator/blob/7e7544097149708190903b1f4d088a4529704886/pkg/cvo/upgradeable.go#L221

docs/dev/object-deletion.md
jottofar: fixed

docs/dev/object-deletion.md
jottofar: fixed

docs/dev/object-deletion.md
jottofar: fixed

docs/dev/object-deletion.md
jottofar: fixed

pkg/cvo/internal/generic.go
jottofar: Yes. Otherwise I could get collisions at the Namespace level I believe. Corrected.

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: I added a new `Check` into `pkg/cvo/upgradeable.go`.  Please review. 

Assuming that I can back my changes out of `pkg/payload/precondition/clusterversion/upgradeable.go` and no changes are required to that file?

jottofar:
>> if an object previously removed reappears a warning is logged which meets what is specified by the enhancement. This does not result in Upgradeable==false and no attempt is made to remove the object again.

> Do we need to update the option 1 in the enhancement PR about this. ?

Implementation is currently compliant with the enhancement. The applicable enhancement statement is:

`Applicable to both options, CVO will produce a warning if a resource previously removed reappears.`

The question that came up during the demo, and that would affect the enhancement, is whether we want to also continue attempting to delete the resource?

vrutkovs:
/retitle OTA-222: Add a manifest annotation to be used for object deletion
/retest

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
+# Manifest Annotation For Object Deletion
+
+Developers can remove any of the currently managed CVO objects by modifying an existing manifest and adding the delete annotation `.metadata.annotations["release.openshift.io/delete"]="true"`. This manifest annotation requests the CVO delete the in-cluster object instead of creating/updating it.
```
LalatenduMohanty, line 3: s/delete/to delete/

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+
+Actual object deletion and subsequent deletion monitoring and status checking only occur during an upgrade. During initial installation the delete annotation prevents further processing of the manifest since the given object should not be created.
```
LalatenduMohanty, line 5: s/only occur/will only occur/

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+  namespace: openshift-machine-api
+  annotations:
+    release.openshift.io/delete: "true"
+...
+```
+Additional manifest properties such as `spec` may be set if convenient (e.g. because you are looking to make a minimal change vs. a previous version of the manifest), but those properties have no affect on manifests with the delete annotation.
+
+### The service-catalog operators
+
+In release 4.5 two jobs, openshift-service-catalog-controller-manager-remover and openshift-service-catalog-apiserver-remover, were created to remove the Service Catalog.
```
LalatenduMohanty, line 50: In 4.5 release two jobs were created to remove the service catalog i.e. openshift-service-catalog-controller-manager-remover and openshift-service-catalog-apiserver-remover

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+In release 4.6, these jobs and all their supporting cluster objects also needed to be removed.
+This example shows how Service Catalog removal would have been executed had manifest based object deletion been in place.
```
LalatenduMohanty, line 52: Suggestion. Substitute `This example shows how Service Catalog removal would have been executed had manifest based object deletion been in place. ` 
with 
`Here is an example of how service catalog could have been removed using manifest annotation to delete an object `. 

Also I do not this we need to give an example what was the process to remove the object without this feature.

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+Each of these components use manifests for creation/update of the component's required resources: namespace, roles, operator deployment, etc.
+The cluster-svcat-apiserver-operator had [the following associated manifests][svcat-apiserver-4.4-manifests]:
+
+* `0000_50_cluster-svcat-apiserver-operator_00_namespace.yaml` containing the `openshift-service-catalog-apiserver-operator` namespace.
+    The deletion annotation would be added to this manifest.
+* `0000_50_cluster-svcat-apiserver-operator_02_config.crd.yaml` containing a cluster-scoped CRD.
+    The deletion annotation would be added to this manifest.
+* `0000_50_cluster-svcat-apiserver-operator_03_config.cr.yaml` containing a cluster-scoped, create-only ServiceCatalogAPIServer.
+    The deletion annotation would be added to this manifest.
+* `0000_50_cluster-svcat-apiserver-operator_03_configmap.yaml` containing a ConfigMap in the `openshift-service-catalog-apiserver-operator` namespace.
```
LalatenduMohanty, line 64: We should say ` The deletion annotation needs be added to these manifests below` and then list all the manifests that need the annotation.

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+    This manifest would be dropped, because the Deployment would be removed as part of the namespace deletion.
+* `0000_50_cluster-svcat-apiserver-operator_08_cluster-operator.yaml` containing a cluster-scoped ClusterOperator.
+    The deletion annotation would be added to this manifest.
+* `0000_90_cluster-svcat-apiserver-operator_00_prometheusrole.yaml` containing a Role in the `openshift-service-catalog-apiserver-operator` namespace.
+    This manifest would be dropped, because the Role would be removed as part of the namespace deletion.
+* `0000_90_cluster-svcat-apiserver-operator_01_prometheusrolebinding.yaml` containing a RoleBinding in the `openshift-service-catalog-apiserver-operator` namespace.
+    This manifest would be dropped, because the RoleBinding would be removed as part of the namespace deletion.
+* `0000_90_cluster-svcat-apiserver-operator_02-operator-servicemonitor.yaml` containing a ServiceMonitor in the `openshift-service-catalog-apiserver-operator` namespace.
+    This manifest would be dropped, because the ServiceMonitor would be removed as part of the namespace deletion.
+
```
LalatenduMohanty, line 84: We need to restructure the above as it is hard to read as there are so many manifest names. We can create two broad categories
* The deletion annotation should be added to these manifests below
* These manifests should be dropped

docs/dev/object-deletion.md
```diff
@@ -0,0 +1,118 @@
...
+## Removing functionality that users might notice
+
+Below is the flow for removing functionality that users might notice, like the web console.
+
+* The first step is deprecating the functionality.
+    During the deprecation release 4.y, the functionality should remain available, with the operator setting Upgradeable=False and linking release notes like [these][deprecated-marketplace-apis].
+* Cluster Administrators must follow the linked release notes to opt in to the removal before updating to the next minor release in 4.(y+1).
+    When the administrator opts in to the removal, the operator should stop setting Upgradeable=False.
+* The operand may be removed when Cluster Administrators opt-in to the removal, or it may be left running and be removed during the transition to the next minor release 4.(y+1).
+* The update to the next minor release 4.(y+1) may use the manifest delete annotation to remove the operator, and, if they have not already been removed, any remaining operand components.
```
LalatenduMohanty, line 107: Can yo clarify the above sentence? It is not clear to me.

LalatenduMohanty (changes requested):

docs/dev/object-deletion.md
jottofar: done

docs/dev/object-deletion.md
jottofar: done

docs/dev/object-deletion.md
jottofar: reworded

docs/dev/object-deletion.md
jottofar: reworded

docs/dev/object-deletion.md
jottofar: done

docs/dev/object-deletion.md
jottofar: done

docs/dev/object-deletion.md
jottofar: done

lib/resourcedelete/apiext.go
```diff
@@ -0,0 +1,41 @@
...
+// If not in UpdatingMode it simply returns an indication that the delete annotation was found. An error is
+// returned if an invalid annotation is found or an error occurs during delete processing.
+func DeleteCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter,
+	required *apiextv1.CustomResourceDefinition, updateMode bool) (bool, error) {
+
+	if delAnnoFound, err := ValidDeleteAnnotation(required.Annotations); !delAnnoFound || err != nil {
+		return delAnnoFound, err
+	} else if !updateMode {
+		return true, nil
+	}
```
LalatenduMohanty, line 23: In both the conditions, you are returning from the function. It is not clear not clear from code which situation the code will not meet these conditions and continue executing the rest of the function.

pkg/cvo/upgradeable.go
```diff
@@ -215,9 +216,28 @@ func (check *clusterVersionOverridesUpgradeable) Check() *configv1.ClusterOperat
 	return cond
 }
 
+type clusterDeletesInProgressUpgradeable struct {
```
LalatenduMohanty, line 219: I think `manifestDeleteInProgressUpgradeable` or `clusterManifestDeleteInProgressUpgradeable` would be a better name for this type.

lib/resourcedelete/apiext.go
jottofar: In the first case, we either didn't find the delete annotation or we got an error. The only error we can get is that the delete annotation is found but doe not have value `true`. It must have value `true`. Otherwise, if a valid delete annotation is found we check that the cluster is actually in `UpdateMode` since this is the only time we should be performing the delete (see `lib/resourcebuilder/resourcebuilder.go`.

This is explained in the method comment.

pkg/cvo/upgradeable.go
jottofar: I followed the pattern [here](https://github.com/openshift/cluster-version-operator/blob/2ea3d9e1fe2a0c2719c3b8fbda30610ce637df28/pkg/cvo/upgradeable.go#L239-L240). `cluster` is just meant to indicate that it is a cluster wide condition.

LalatenduMohanty: Yeah. I saw that but did not know `cluster` was for clusterwide. I suggested `clusterManifestDeleteInProgressUpgradeable` because `ManifestDelete` is better to understand as compared to `Deletes`

LalatenduMohanty (changes requested):

lib/resourcedelete/apiext.go
LalatenduMohanty: Makes sense. Thanks

LalatenduMohanty (approved):
/lgtm

jottofar:
/cc wking

Can I get an override on `/test e2e-agnostic-upgrade`. Failures are unrelated to my PR and it's unchanged since passing this test before.

PR #439, opened 2020-08-17
docs/dev/operators: Replace stale manifest filename regexp with generic discussion

wking:
The regular expression that the cluster-version operator actually uses is unchanged since it landed in cb4e0375e6 (#88).  But the regular expression given in the docs was over-simplified (e.g. it allowed the component to begin with a dash, while the actual expression forbids component-leading dashes).  Instead of talking about the specifics of the regular expression, just explain the different portions in the docs.


vrutkovs (approved):
/lgtm

PR #441, opened 2020-08-22
Bug 1797123: pkg/cvo: Separate ConfigMap informer for openshift-config-managed

wking:
With c9fab435c1 (#311), I attempted to pivot to loading the TLS configration from the `openshift-config-managed` namespace.  But I hadn't realized that `cmConfigInformer` was scoped to the `openshift-config` namespace, so attempts to retrieve `trusted-ca-bundle` failed via the "not found" no-op path regardless of whether the ConfigMap actually existed.  With this commit, I create a new informer for the `openshift-config-managed` namespace, so we can successfully retrieve the ConfigMap when it exists.

An alternative approach would be to drop the `informers.WithNamespace` filter.  My impression is that having two namespace-filtered informers will be more efficient than a single unfiltered informer, but I'm not really sure.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -163,6 +163,7 @@ func New(
 	cvInformer configinformersv1.ClusterVersionInformer,
 	coInformer configinformersv1.ClusterOperatorInformer,
 	cmConfigInformer informerscorev1.ConfigMapInformer,
+	cmConfigManagedInformer informerscorev1.ConfigMapInformer,
 	proxyInformer configinformersv1.ProxyInformer,
 	client clientset.Interface,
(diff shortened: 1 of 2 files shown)


jottofar:
/hold
In case @LalatenduMohanty still wants to look.
/lgtm

LalatenduMohanty:
> An alternative approach would be to drop the `informers.WithNamespace` filter. My impression is that having two namespace-filtered informers will be more efficient than a single unfiltered informer, but I'm not really sure.

I am not sure which is the better way to move ahead. So going ahead with the current implementation.

LalatenduMohanty (approved):
/lgtm

PR #443, opened 2020-08-25
Modify to use library-go manifest and verify packages

jottofar:
library-go manifest and verify packages were created since both CVO and the oc client use them.


pkg/cvo/cvo.go
```diff
@@ -691,14 +691,14 @@ func (b *resourceBuilder) builderFor(m *lib.Manifest, state payload.State) (reso
 	if resourcebuilder.Mapper.Exists(m.GVK) {
 		return resourcebuilder.New(resourcebuilder.Mapper, config, *m)
 	}
-	client, err := dynamicclient.New(config, m.GVK, m.Object().GetNamespace())
+	client, err := dynamicclient.New(config, m.GVK, m.Obj.GetNamespace())
```
wking, line 694: I don't understand this change.  Did we loose the getter somewhere?  Although not clear to me why we'd have a getter in front of a public property...

jottofar: Comment from when I merged `manifest` into library-go. The method isn't needed: https://github.com/openshift/library-go/pull/680#discussion_r385556651

pkg/cvo/cvo_test.go
```diff
@@ -39,10 +39,10 @@ import (
 	clientset "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/client-go/config/clientset/versioned/fake"
 
-	"github.com/openshift/cluster-version-operator/lib"
 	"github.com/openshift/cluster-version-operator/pkg/payload"
-	"github.com/openshift/cluster-version-operator/pkg/verify"
-	"github.com/openshift/cluster-version-operator/pkg/verify/store/sigstore"
+	"github.com/openshift/library-go/pkg/manifest"
+	//"github.com/openshift/library-go/pkg/verify"
```
wking, line 44: nit: drop this commented-out import?

jottofar: Done

jottofar:
retest

wking:
Integration is still blocked on openshift/release#13745 or similar.

/override ci/prow/integration

/lgtm

PR #444, opened 2020-08-25
Bug 1872080: Dockerfile.rhel: Use a RHEL-version-agnostic filename

wking:
So we don't have a nominally-rhel7 Dockerfile building RHEL 8 images and also so we don't need to manually bump this filename when we bump RHEL versions.


jottofar:
makes sense to but others can have a look as well
/hold
/lgtm

wking:
Rebased around #442 with 1100906d -> 32204c00.

LalatenduMohanty (approved):
/lgtm

wking:
/hold cancel

Both Lala and Jack have chimed in.

PR #424, opened 2020-08-04
Bug 1843505: pkg/start: Release leader lease on graceful shutdown

wking:
So the incoming cluster-version operator doesn't need to wait for the outgoing operator's lease to expire, which [can take a while][1]:

```
I0802 10:06:01.056591       1 leaderelection.go:243] attempting to acquire leader lease  openshift-cluster-version/version...
...
I0802 10:07:42.632719       1 leaderelection.go:253] successfully acquired lease openshift-cluster-version/version
```

and [time out the `Cluster did not acknowledge request to upgrade in a reasonable time` testcase][2].  Using `ReleaseOnCancel` has been the plan since 2b81f47626 (#87).  I'm not clear on why it (sometimes?) doesn't work today.

The discrepancy between the "exit after 2s no matter what" comment and the 5s `After` dates back to dbedb7accb (#179), which bumped the `After` from 2s to 5s, but forgot to bump the comment.  I'm removing that code here in favor of the two-minute timeout from b30aa0ecdb (#349).  We still exit immediately on a second `TERM`, for folks who get impatient waiting for the graceful timeout.

I'm also pushing the `Background()` initialization all the way up to the command-line handler, to make it more obvious that the context is scoped to the whole `start` invocation.

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/25365/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1289853267223777280/artifacts/e2e-gcp-upgrade/pods/openshift-cluster-version_cluster-version-operator-5b6ff896c6-57ppb_cluster-version-operator.log
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1843505#c7

PR #445, opened 2020-08-27
Bug 1872906:

wking:
Picking #349 and #424 back to 4.5 so we release the leader lock when we exit, which in turn allows the incoming CVO pod to pick up the leader lock and update the ClusterVersion status in a timely manner.

A number of mostly-minor manual tweaks to adjust to 4.5, which is missing #358's TLS metrics and #406 and #410's `Context` changes.


wking:
Oops, wrong target branch.

/close

wking:
Filed #446 against release-4.5.

PR #447, opened 2020-08-28
docs: update cluster_installer description

crawford:
This should have been updated in e57f170c. Shame on me.

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -7,7 +7,7 @@ The cluster version is reported as seconds since the epoch with labels for `vers
 
 * `current` - the version the operator is applying right now (the running CVO version) and the age of the payload
 * `initial` - the initial version of the cluster, and value is the creation timestamp of the cluster version (cluster age)
-* `cluster` - the current version of the cluster with from_version set to the initial version, and value is the creation timestamp of the cluster version (cluster age)
+* `cluster` - the current version of the cluster with `from_version` set to the initial version, and value is the creation timestamp of the cluster version (cluster age)
 * `failure` - if the failure condition is set, reports the last transition time for the condition.
 * `desired` - reported if different from current as the most recent timestamp on the cluster version
 * `completed` - the time the most recent version was completely applied, or absent if not reached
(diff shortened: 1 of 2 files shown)


crawford:
Documentation fixes don't need end-to-end tests.

/override ci/prow/e2e
/override ci/prow/e2e-upgrade

docs/dev/metrics.md
```diff
@@ -59,10 +59,10 @@ cluster_operator_payload_errors{version="4.0.3"} 10
...
+`cluster_installer` records information about the installation process. The type is either "openshift-install", indicating that `openshift-install` was used to install the cluster (IPI) or "other", indicating that an unknown process installed the cluster (UPI). When `openshift-install` creates a cluster, it will also report its version and invoker. When an unknown process installed the cluster, the version and invoker reported will be that of the `openshift-install` invocation which created the manifests. The version is helpful for determining exactly which builds are being used to install (e.g. were they official builds or had they been modified). The invoker is "user" by default, but it may be overridden by a consuming tool (e.g. Hive, CI).
```
wking, line 62: Can we update [this][1] too?

/hold

[1]: https://github.com/openshift/cluster-version-operator/blob/05f3d334cec0f3e187c9a64a55934a1c5011122b/pkg/cvo/metrics.go#L94

crawford: I'm open to it. What do you think it should say? It's very terse right now, but it _is_ accurate.

wking: Seems like we should mention the `openshift-install` and `other` values for `type`.  And point out that `version` is the invoker version, not the cluster version, just to be extra clear?  Maybe wording like:

> Reports info about the installation process and, if applicable, the install tool.  The type is either "openshift-install", indicating that openshift-install was used to install the cluster, or "other", indicating that an unknown process installed the cluster.  The invoker is "user" by default, but it may be overridden by a consuming tool.  FIXME: something about version...

Based on your example, looks like we set an installer version even when `type=other`?  What if some crazy person goes and installs a cluster from scratch, without even using `openshift-install` to create Ignition configs?  Do we want to say something explicit about `version` always being the version of `openshift-install` that was involved in installation, even if the user did something else to actually spin up the initial cluster resources?

crawford: Oops, this fell off of my radar. I've updated the description. Let me know what you think.

crawford:
It's times like these that I really miss bazel...

/override ci/prow/images
/override ci/prow/e2e
/override ci/prow/e2e-upgrade
/override ci/prow/integration
/override ci/prow/unit

PR #450, opened 2020-09-01
Update clusterversion.md

theodor2311:
Update oc get clusterversion command.


wking:
Manually labeled valid-bug.  No need to hang this small doc fix on a bug or wait until we open for 4.7.

PR #455, opened 2020-09-15
Bug 1873900: pkg/payload/task_graph: Avoid deadlocking on cancel with workCh queue

wking:
Before 55ef3d3027 (#264, new in 4.6), `RunGraph` had a separate goroutine that managed the work queue, with results fed into `errCh` to be collected by the main `RunGraph` goroutine.  It didn't matter if that work queue goroutine hung; as long as all the worker goroutines exited, `RunGraph` would collect their errors from `errCh` and return.  In 55ef3d3027, I removed the queue goroutine and moved queue management into the main `RunGraph` goroutine.  With that change, we became exposed to the following race:

1. Main goroutine pushes work into `workCh`.
2. Context canceled.
3. Workers exit via the `Canceled worker...` case, so they don't pick the work out of `workCh`.
4. Main goroutine deadlocks because there is work in flight, but nothing in `resultCh`, and no longer any workers to feed `resultCh`.

In logs, this looks like "[sync-worker goroutine has gone to sleep, and is no longer synchronizing manifests][1]".

With this commit, we drain results when they are available, but we also respect the context to allow the `resultCh` read to be canceled.  When we have been canceled with work in flight, we also attempt a non-blocking read from `workCh` to drain out anything there that has not yet been picked up by a worker.  Because `done` will not be set true, we'll call `getNextNode` again and come in with a fresh pass through the for loop.  `ctx.Err()` will no longer be `nil`, but if the `workCh` drain worked, we may now have `inflight == 0`, and we'll end up in the case that sets `done` true, and break out of the for loop on that round.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1873900

### DIFF
--- a/pkg/payload/task_graph.go
+++ b/pkg/payload/task_graph.go
@@ -507,9 +507,18 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
 			case <-ctx.Done():
 			}
 		case inflight > 0: // no work available to push; collect results
-			result := <-resultCh
-			results[result.index] = &result
-			inflight--
+			select {
+			case result := <-resultCh:
+				results[result.index] = &result
+				inflight--
+			case <-ctx.Done():
+				select {
+				case runTask := <-workCh: // workers canceled, so remove any work from the queue ourselves
+					inflight--
+					submitted[runTask.index] = false
+				default:
+				}
+			}
 		default: // no work to push and nothing in flight.  We're done
 			done = true
 		}
--- a/pkg/payload/task_graph_test.go
+++ b/pkg/payload/task_graph_test.go
@@ -832,6 +832,26 @@ func TestRunGraph(t *testing.T) {
 			},
 		},
 		{
+			name: "mid-task cancellation with work in queue does not deadlock",
+			nodes: []*TaskNode{
+				{Tasks: tasks("a1", "a2", "a3")},
+				{Tasks: tasks("b")},
+			},
+			sleep:    time.Millisecond,
+			parallel: 1,
+			errorOn: func(t *testing.T, name string, ctx context.Context, cancelFn func()) error {
+				if err := ctx.Err(); err != nil {
+					return err
+				}
+				if name == "a2" {
+					cancelFn()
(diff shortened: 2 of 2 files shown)


wking:
[Upgrade job][1] timed out some pods while creating the release.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/455/pull-ci-openshift-cluster-version-operator-master-e2e-upgrade/1305929088954798080

vrutkovs:
Makes sense to me, I wonder if it can be unit-tested?

wking:
> Makes sense to me, I wonder if it can be unit-tested?

We scale `workCh` by worker parallelism, so... set parallelism to one, create two slow, parallel tasks, and cancel halfway through the first?  Or something...

wking:
I've added a unit test with cae62913c7 -> 59824985f7.  59824985f7's commit message has a paragraph talking through the test.

wking:
Fixed [gofmt errors][1] with 59824985f7 -> 24e9e91b2d.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/455/pull-ci-openshift-cluster-version-operator-master-gofmt/1306279385069785088

wking:
And moved from two workers to one worker in the unit test with 24e9e91b2d -> 622e04fe74, which means we don't have to care about keeping the tasks synchronized-enough between two workers to figure out what `want` should be.

wking:
I've pushed 622e04fe74 -> 632e763fa2 with some discussion of `workCh` capacity and `select` randomness to show that, even with the old code, there were some cases where in-flight work would still not have deadlocked.  That reasoning is why the new unit test will sometimes fail with two cancels instead of deadlocking before the code-fix from this PR.

jottofar:
I had the same comment as Vadim - "make sense to me but would feel better with a test". I see that's done so...LGTM.

vrutkovs (approved):
/lgtm

unit-test seems sane

PR #457, opened 2020-09-17
Bug 1878163: Dockerfile.rhel: Bump to Go 1.15

wking:
ART is [waffling][1] around the 1.14 <-> 1.15 transition, but we don't need to wait.  Let the future start today!

Replaces #456, but without the commit message `to mach` typo or the bogus link to an ocp-build-data branch where the CVO is currently reverted back to Go 1.14.

[1]: https://github.com/openshift/ocp-build-data/pull/665


LalatenduMohanty (approved):
/lgtm

PR #458, opened 2020-09-18
Fixing long lines as they are hard to see in vscode or in github

LalatenduMohanty:
Used  https://github.com/segmentio/golines to create this PR

The command used to generate this patch is
$ golines -w -m 120 pkg/ cmd/ lib/

### DIFF
--- a/cmd/render.go
+++ b/cmd/render.go
@@ -25,8 +25,18 @@ var (
 
 func init() {
 	rootCmd.AddCommand(renderCmd)
-	renderCmd.PersistentFlags().StringVar(&renderOpts.outputDir, "output-dir", "", "The output directory where the manifests will be rendered.")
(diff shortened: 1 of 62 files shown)


wking:
I would like a presubmit job to avoid backsliding.  The more we can automate our style preferences, the easier it is to stick to them, and the less PR submitters and approvers have to think about them.  See [here][1] and [here][2] for an example presubmit job that runs a tool and errors out with a diff if that tool made changes.

[1]: https://github.com/openshift/release/pull/11752/commits/5cf5ba3f8707699a8345cd2aaf8d2c7b6615bc47#diff-5761978dde4343debc233a6f56af0a7eR1
[2]: https://github.com/openshift/release/pull/11752/commits/5cf5ba3f8707699a8345cd2aaf8d2c7b6615bc47#diff-011e89dbbc47830cc280f67b223c64e7R41

PR #460, opened 2020-09-23
Bug 1802553: Adding upgradeable = false when an upgrade in progress

LalatenduMohanty:
CVO must not start another upgrade process if an upgrade is already
 in progress for a minor version (Y stream). We have a similar check in oc adm command. But an user
 can start the upgrade through the web console too.

### DIFF
--- a/pkg/cvo/upgradeable.go
+++ b/pkg/cvo/upgradeable.go
@@ -135,6 +135,7 @@ type upgradeableCheck interface {
 
 type clusterOperatorsUpgradeable struct {
 	coLister configlistersv1.ClusterOperatorLister
+	operator *Operator
 }
 
 func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatusCondition {
@@ -142,7 +143,7 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
 		Type:   configv1.ClusterStatusConditionType("UpgradeableClusterOperators"),
 		Status: configv1.ConditionFalse,
 	}
-	ops, err := check.coLister.List(labels.Everything())
+	ops, err := check.operator.coLister.List(labels.Everything())
 	if meta.IsNoMatchError(err) {
 		return nil
 	}
@@ -150,6 +151,31 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
 		cond.Reason = "FailedToListClusterOperators"
 		cond.Message = errors.Wrap(err, "failed to list cluster operators").Error()
 		return cond
+	}
+
+	// Set upgradeable = false to stop the y stream upgrades to happen when an upgrade is in progress.
+	// This will only impact the Y stream updates as CVO ignores this for z stream updates
+	var isUpgradeInProgress bool
+	for _, op := range ops {
+		if sc := resourcemerge.FindOperatorStatusCondition(op.Status.Conditions, configv1.OperatorProgressing); sc != nil {
+			isUpgradeInProgress = true
+		}
+	}
+	now := metav1.Now()
+	if isUpgradeInProgress {
+		var statusConds []configv1.ClusterOperatorStatusCondition
+		statusConds = append(statusConds, configv1.ClusterOperatorStatusCondition{
+			Type:               configv1.OperatorUpgradeable,
+			Status:             configv1.ConditionFalse,
+			Reason:             "UpgradeInProgress",
+			Message:            fmt.Sprintf("Cluster cannot be upgraded between minor versions"),
+			LastTransitionTime: now,
+		})
+		check.operator.setUpgradeable(&upgradeable{
+			Conditions: statusConds,
+		})
+		// requeue
+		//optr.queue.Add(optr.queueKey())
 	}
 
 	type notUpgradeableCondition struct {
@@ -217,7 +243,7 @@ func (check *clusterVersionOverridesUpgradebale) Check() *configv1.ClusterOperat
 
 func (optr *Operator) defaultUpgradeableChecks() []upgradeableCheck {
 	return []upgradeableCheck{
-		&clusterOperatorsUpgradeable{coLister: optr.coLister},
+		&clusterOperatorsUpgradeable{operator: optr},
 		&clusterVersionOverridesUpgradebale{name: optr.name, cvLister: optr.cvLister},
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/cvo/upgradeable.go
```diff
@@ -152,6 +153,31 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
 		return cond
 	}
 
+	// Set upgradeable = false to stop the y stream upgrades to happen when an upgrade is in progress.
```
jottofar, line 156: Seems like this logic should be moved up into syncUpgradeable and this Check method should simply continue to return cond(s) to be handled there. syncUpgradeable already has access to `optr`, calls `setUpgradeable` and does the requeue.

pkg/cvo/upgradeable.go
```diff
@@ -152,6 +153,31 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
...
+	// This will only impact the Y stream updates as CVO ignores this for z stream updates
+	var isUpgradeInProgress bool
+	for _, op := range ops {
+		if sc := resourcemerge.FindOperatorStatusCondition(op.Status.Conditions, configv1.OperatorProgressing); sc != nil {
```
jottofar, line 160: Will this only be true during an upgrade? I don't know but wondering if some thing else affecting a resource could cause it to enter this state. Perhaps that's okay and valid - i.e. if a resource is `OperatorProgressing` for any reason we want to block upgrades?

pkg/cvo/upgradeable.go
LalatenduMohanty: syncUpgradeable is called from a different go routine and it is only supposed to be synchronizing the upgradeable false conditions. But in this case we want to set the upgradeable = false and then syncUpgradeable will do it jobs.

pkg/cvo/upgradeable.go
LalatenduMohanty: > Will this only be true during an upgrade? I don't know but wondering if some thing else affecting a resource could cause it to enter this state. Perhaps that's okay and valid

Mostly yes , as per the API docs 
```
// Progressing indicates that the operator is actively rolling out new code,
// propagating config changes, or otherwise moving from one steady state to
// another.  Operators should not report progressing when they are reconciling
// a previously known state.
```

pkg/payload/task.go
```diff
@@ -192,6 +192,9 @@ func SummaryForReason(reason, name string) string {
 	case "ImageVerificationFailed":
 		return "the image may not be safe to use"
 
+	case "UpgradeInProgress":
+		return "an upgrade is in progress"
```
smarterclayton, line 196: We use “update” over “upgrade”.

Might suggest “an upgrade is already in progress”.

pkg/cvo/upgradeable.go
```diff
@@ -152,6 +153,31 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
...
+			Status:             configv1.ConditionFalse,
+			Reason:             "UpgradeInProgress",
+			Message:            fmt.Sprintf("Cluster cannot be upgraded between minor versions"),
+			LastTransitionTime: now,
+		})
+		check.operator.setUpgradeable(&upgradeable{
+			Conditions: statusConds,
+		})
+		// requeue
+		//optr.queue.Add(optr.queueKey())
```
smarterclayton, line 178: Why is this commented out?

pkg/cvo/upgradeable.go
```diff
@@ -135,14 +135,15 @@ type upgradeableCheck interface {
 
 type clusterOperatorsUpgradeable struct {
 	coLister configlistersv1.ClusterOperatorLister
+	operator *Operator
```
smarterclayton, line 138: I would prefer an interface or pass two objects vs passing a static type.  Keep this testable.

pkg/cvo/upgradeable.go
```diff
@@ -152,6 +153,31 @@ func (check *clusterOperatorsUpgradeable) Check() *configv1.ClusterOperatorStatu
...
+	if isUpgradeInProgress {
+		var statusConds []configv1.ClusterOperatorStatusCondition
+		statusConds = append(statusConds, configv1.ClusterOperatorStatusCondition{
+			Type:               configv1.OperatorUpgradeable,
+			Status:             configv1.ConditionFalse,
+			Reason:             "UpgradeInProgress",
+			Message:            fmt.Sprintf("Cluster cannot be upgraded between minor versions"),
+			LastTransitionTime: now,
+		})
+		check.operator.setUpgradeable(&upgradeable{
```
smarterclayton, line 174: This really isn’t safe.  Methods called check are supposed to be side effect free.  I think you’re going to have to figure out how to keep the information flow strictly one way here (the sync loop needs to be the one setting this).  Remember, setting spec is intent and so there is no guarantee of an atomic transition from the upgrading step.  And I’m a bit worried that the logic really needs to be part of the state reconciler (from spec to the sync loop) and that is a much bigger change.

smarterclayton (changes requested):
I think this is a much more complicated change to be correct - the spec -> desired upgrade sync is what has to control this, and that means you need to look at both user desired intent and the current status and that’s also a concerning change (removes the ability to bypass the current state).

The core design principle is that spec.desired is the thing we have to converge to, and the CVO does not maintain a stack of incremental steps between current and desired.  Any change like that is super impactful too and needs an enhancement.

smarterclayton:
A more approachable option might be to prune available updates to the current Y level while desired != current and desired and current are different Y levels.  Same outcome, doesn’t require a redesign of cvo or break the convergence properties (I still believe the ability to force convergence is the correct top priority for Cvo)

pkg/cvo/upgradeable.go
wking: your current code is checking "is there a Progressing condition?", which should pretty much always be true, even outside of updates when we are Progressing=False.  What you want is "is there a Progressing=True condition?", which would be:

```go
if sc := resourcemerge.FindOperatorStatusCondition(op.Status.Conditions, configv1.OperatorProgressing); sc != nil && sc.Status == configv1.ConditionTrue {
```

pkg/cvo/upgradeable.go
jottofar: 'Check` is called by `syncUpgradeable` here https://github.com/openshift/cluster-version-operator/blob/478e60d7efed233d9979c70fcaf8c097b2d5d6f5/pkg/cvo/upgradeable.go#L36. Is there another call to `Check` that you plan to add on one you have added that I missed?

PR #469, opened 2020-10-09
Bug 1886900: pkg/cvo/sync_worker: Drop "Manifest: ..." logging

wking:
This line is descended from logging we grew way back in 847f71bcef (#14).  But we triggered it recently with 88c222c954 (#448).  Drop the line, because it's noisy spew in the log files, and we can get the manifest content via:

```console
$ oc adm release extract --to=manifests $PULLSPEC
```


LalatenduMohanty (approved):
/lgtm

wking:
I've added some follow-up reductions in #472.

PR #473, opened 2020-10-13
Set FieldManager in managedFields to track CVO activity

vrutkovs:
This should visualize which fields are managed by CVO, showing when was the object last updated and which fields were set by CVO. 
Hopefully this would help us find out which fields are extra and relieve some work from controller.

@sttts any other advice how to get more info on this?

### DIFF
--- a/lib/manifest.go
+++ b/lib/manifest.go
@@ -116,3 +116,17 @@ func ParseManifests(r io.Reader) ([]Manifest, error) {
 		manifests = append(manifests, m)
 	}
 }
+
+// Metav1CreateOptions returns default create options for new objects
+func Metav1CreateOptions() metav1.CreateOptions {
+	return metav1.CreateOptions{
+		FieldManager: "cluster-version-operator",
+	}
+}
+
+// Metav1UpdateOptions returns default update objects for existing objects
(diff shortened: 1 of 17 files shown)


lib/manifest.go
```diff
@@ -116,3 +116,17 @@ func ParseManifests(r io.Reader) ([]Manifest, error) {
 		manifests = append(manifests, m)
 	}
 }
+
+// Metav1CreateOptions returns default create options for new objects
+func Metav1CreateOptions() metav1.CreateOptions {
+	return metav1.CreateOptions{
+		FieldManager: "cluster-version-operator",
```
sttts, line 123: this default to the user-agent. Is the user-agent wrong in CVO ?

sttts:
/hold

I think this the wrong approach. The field manager default to the user-agent. Is the user-agent not set correctly in CVO? If it isn't, let's fix it.

lib/manifest.go
vrutkovs: Ah, right, its being set, sorry

vrutkovs:
This field is already being set, no need to rewrite it

PR #474, opened 2020-10-13
Fixing go doc for equalUpdate()

LalatenduMohanty:

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -429,12 +429,10 @@ func (w *SyncWorker) calculateNext(work *SyncWork) bool {
 	return changed
 }
 
-// equalUpdate returns true if two updates have the same image.
+// equalUpdate returns true if two updates are semantically equivalent.
+// It checks if the updates have have the same force and image values
 func equalUpdate(a, b configv1.Update) bool {
-	if a.Force != b.Force {
-		return false
-	}
-	return a.Image == b.Image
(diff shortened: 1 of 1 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -429,7 +429,8 @@ func (w *SyncWorker) calculateNext(work *SyncWork) bool {
 	return changed
 }
 
-// equalUpdate returns true if two updates have the same image.
+// equalUpdate returns true if two updates are semantically equivalent.
+// It checks if the updates have same force flag values or have the same images
```
wking, line 433: I think this sentence is an internal detail that we don't need to document, but if you want to document the implementation I'd document it as "... have the same force and image values.".  In which case, we might want to adjust the implementation to be:

```go
return a.Force == b.Force && a.Image == b.Image
```

which reads a little more clearly to me than the current `if`-block double-negative.

wking: Also, might be worth mentioning f8eff25191 (#170) in the commit message, because that's the commit that added the `Force` check but neglected to update the docstring ;).

LalatenduMohanty: Fixed in https://github.com/openshift/cluster-version-operator/commit/b060ad156db352ba34abf0a09e56f7b8b43243d8

wking (approved):
/lgtm

PR #475, opened 2020-10-14
hack/log-explainer: Don't crash if run on truncated logs

wking:
Avoid crashing on undefined `sync` if someone pushes in logs from a pod that has been running for so long that we no longer have the lines that start the opening sync.  Instead, just throw that partially-logged sync out.


vrutkovs (approved):
/lgtm

PR #476, opened 2020-10-14
pkg/cvo: Add 'State' property to SyncWorkerStatus

wking:
Instead of tracking reconciling/initializing via independent booleans, to avoid the possiblility of setting both at once.  This also allows consumers to use `State.String()` to render the state without having to write a special renderer for a unique-to-`SyncWorkerStatus` representation.

### DIFF
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -220,14 +220,14 @@ func TestCVO_StartupAndSync(t *testing.T) {
 		SyncWorkerStatus{
 			Generation: 1,
 			Step:       "RetrievePayload",
-			Initial:    true,
+			State:      payload.InitializingPayload,
 			// the desired version is briefly incorrect (user provided) until we retrieve the image
 			Actual: configv1.Release{Version: "4.0.1", Image: "image/image:1"},
 		},
 		SyncWorkerStatus{
 			Generation:  1,
 			Step:        "ApplyResources",
-			Initial:     true,
+			State:       payload.InitializingPayload,
 			VersionHash: "6GC9TkkG9PA=",
 			Actual: configv1.Release{
 				Version:  "1.0.0-abc",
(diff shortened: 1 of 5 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -1031,7 +1022,7 @@ func runThrottledStatusNotifier(ctx context.Context, interval time.Duration, buc
 				return
 			case next := <-ch:
 				// only throttle if we aren't on an edge
-				if next.Generation == last.Generation && next.Actual.Image == last.Actual.Image && next.Reconciling == last.Reconciling && (next.Failure != nil) == (last.Failure != nil) {
+				if next.Generation == last.Generation && next.Actual.Image == last.Actual.Image && next.State == last.State && (next.Failure != nil) == (last.Failure != nil) {
```
jottofar, line 1025: Since the semantics have changed are you sure that this will never result in a change in behavior? I.e. whereas before it was limited to whether the reconciling bool differed and now it's been widened to multiple possible State values, e.g. next.State == ReconcilingPayload, last.State == InitializingPayload or some such.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -2554,12 +2554,12 @@ func TestCVO_ParallelError(t *testing.T) {
...
-	// check the worker status is initially set to reconciling
-	if status := worker.Status(); status.Reconciling || status.Completed != 0 {
-		t.Fatalf("The worker should be reconciling from the beginning: %#v", status)
+	// check the worker status is initially set to initializing
+	if status := worker.Status(); status.State != payload.InitializingPayload || status.Completed != 0 {
+		t.Fatalf("The worker should be initializing from the beginning: %#v", status)
 	}
 	if worker.work.State != payload.InitializingPayload {
-		t.Fatalf("The worker should be reconciling: %v", worker.work)
+		t.Fatalf("The worker should be initializing: %v", worker.work)
```
jottofar, line 2562: This makes sense that it should initially be initializing but how did the test case pass before? Maybe I didn't see where the test case functionality was changed?

PR #477, opened 2020-10-23
Bug 1891143: pkg/cvo/metrics: Abandon child goroutines after shutdownContext expires

wking:
Apparently there's something in the HTTPS server goroutine that can hang up even if we've called `Shutdown()` on the server.  Defend against that with a safety valve to abandon stuck goroutines if `shutdownContext` expires.

Also pivot to `resultChannel` and `asyncResult`, so we can get names for the collected channels (and more easily identify the stuck channels by elimination), following the pattern set by cc1921d186 (#424).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1891143#c1

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -100,6 +100,11 @@ version for 'cluster', or empty for 'initial'.
 	}
 }
 
+type asyncResult struct {
+	name  string
+	error error
+}
+
 // RunMetrics launches an server bound to listenAddress serving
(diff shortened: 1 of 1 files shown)


wking:
John pruned some leaked GCP service accounts.

/retest

wking:
We have doubled our GCP service account quota.

/retest

wking:
Bug VERIFIED in 4.7.

/cherrypick release-4.6

PR #478, opened 2020-10-30
Doc fix for cluster operator creation step

andfasano:
It looks like that a small portion of the documentation was not yet updated due the changes introduced in https://github.com/openshift/cluster-version-operator/pull/318. Thanks @sdodson for pointing it out


sdodson:
docs only, don't waste CI cycles on this

PR #481, opened 2020-11-18
Bug 1950430: pkg/cvo/metrics: Drop HTTP, require HTTPS for metrics access

wking:
We began serving metrics over HTTPS with 6132bc33a7 (#358), which also requested monitoring to scrape us over HTTPS.  Now that that is all in place in 4.6, we no longer need to serve over HTTP in 4.7 and later.  This commit pivots us to always serving over HTTPS.

Because we are no longer serving HTTP, move to requiring `--serving-cert-file` and `--serving-key-file` when `--listen` is non-empty. I'd like to drop the `--listen` default, to make it an explicit opt-in, but I don't want to lose metrics when folks update from 4.6 -> 4.7. With this commit we start setting `--listen` explicitly when we launch child CVOs, and in 4.8 we can drop:

    ListenAddr: "0.0.0.0:9099",

from `pkg/start`.  It's possible that the manifest for the incoming CVO is constructed from the incoming release image, in which case we may be able to drop the `--listen` default now.

I'm setting `--listen` empty in the bootstrap manifest, because we don't need to serve metrics then (it's long before we have Prometheus around to scrape us).


wking:
Rebased onto master with 5a1e4aea -> 8fd5ce4e.

wking:
[Update/rollback][1] was too slow:

```
INFO[2021-04-19T19:13:18Z] Running step e2e-agnostic-upgrade-openshift-e2e-test. 
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2021-04-19T21:25:55Z"}
INFO[2021-04-19T21:25:55Z] Received signal.                              signal=interrupt
```

But the update/rollback [did complete][2] by the time we'd gathered, so:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/481/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1384196544659460096
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/481/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1384196544659460096/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/clusterversion.json

PR #136, opened 2019-03-11
sync: Completely parallelize the initial payload

smarterclayton:
Increase parallism to maximum tasks, increase backoff, wait longer for cluster operator, and generally try to load everything at the same time.

This results in a significant reduction in boot time and allows us to keep run-levels in the payload in the desired update order, not the desired install order.

Builds on top of #133

PR #482, opened 2020-11-24
pkg/cvo/internal: Do not block on Degraded=True ClusterOperator

wking:
We have blocking on this condition since 545c3424aa (#31) when it was `Failing`.  We'd softened our install-time handling to act this way back in b0b4902fce (#136), [motivated by install speed][1].  And a degraded operator may slow dependent components in their own transitions.  But as long as the operator/operand are available at all, it should not block depndent components from transitioning, so this commit removes the `Degraded=True` block from the remaining modes.

We still have the critical `ClusterOperatorDegraded` waking admins up when an operator goes `Degraded=True` for a while, we will just no longer block updates at that point.  We won't block `ReconcilingMode` manifest application either, but since that's already flattened and permuted, and `ClusterOperator` tend to be towards the end of their `TaskNode`, the impact on `ReconcilingMode` is minimal (except that we will no longer go `Failing=True` in `ClusterVersion` when the only issue is some `Degraded=True` ClusterOperator).

CC @abhinavdahiya, @deads2k, @smarterclayton as folks who were involved in the logic I'm removing here.

[1]: https://github.com/openshift/cluster-version-operator/pull/136#issue-260160975

### DIFF
--- a/docs/user/reconciliation.md
+++ b/docs/user/reconciliation.md
@@ -135,7 +135,6 @@ The ClusterOperator builder only monitors the in-cluster object and blocks until
     would block until the in-cluster ClusterOperator reported `operator` at version 4.1.0.
 
     The progressing check is deprecated and will be removed once all operators are reporting versions.
-* Not degraded (except during initialization, where we ignore the degraded status)
 
 ### CustomResourceDefinition
 
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -169,63 +169,23 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration, 
 		}
 
 		available := false
+		degraded := true
 		progressing := true
-		failing := true
-		var failingCondition *configv1.ClusterOperatorStatusCondition
-		degradedValue := true
-		var degradedCondition *configv1.ClusterOperatorStatusCondition
 		for i := range actual.Status.Conditions {
 			condition := &actual.Status.Conditions[i]
 			switch {
 			case condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue:
 				available = true
+			case condition.Type == configv1.OperatorDegraded && condition.Status == configv1.ConditionFalse:
+				degraded = false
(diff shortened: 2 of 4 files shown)


wking:
/hold

Big change; we want to give folks time to weigh in.

sdodson:
I would personally prefer that we tighten up the definition of what degraded means such that it's only used when things are critical enough that they should block the forward progress of upgrades.

wking:
> I would personally prefer that we tighten up the definition of what degraded means...

[Current definition][1] hinges on quality-of-service.  [Available definition][2] requires the operand to be "functional and available".  Can you float an example of component behavior that would be Available=True Degraded=False but still sufficiently severe to need to block later-manifest reconciliation?

[1]: https://github.com/openshift/api/blob/d8862a308f98bba6ce3614bd18b4ffe61c236908/config/v1/types_cluster_operator.go#L153-L168
[2]: https://github.com/openshift/api/blob/d8862a308f98bba6ce3614bd18b4ffe61c236908/config/v1/types_cluster_operator.go#L143-L144

sdodson:
Thanks for referencing that, I'd say my view was mostly inline with the statement `A service should not report Degraded during the course of a normal upgrade.` rather than a deep consideration of available versus degraded. I'm concerned that we're being too eager to move into a degraded state during normal upgrade behavior. I'd prefer those operators tune themselves to align more closely with real world cluster behavior, I fear their timeouts are tuned based on behaviors observed from unloaded or only CI clusters.

smarterclayton:
I agree with Scott (I think) - I don’t believe degraded is an acceptable state for operators ever, especially during upgrade, as we have historically defined it in practice.

wking:
>  I don’t believe degraded is an acceptable state for operators ever, especially during upgrade, as we have historically defined it in practice.

I'm not arguing for it to be acceptable, I'm arguing about it being non-blocking.  We will still alert if operators go Degraded=True in the wild.  We can build CI to fail operators that go Degraded=True during a run, if we don't do that already.  But if an operator goes Degraded=True in a customer cluster, it's not clear to me how sticking mid-update is helping the customer resolve that situation, vs. pushing through with the rest of the update, as long as the operator is Available=True, and letting admins sort out the degrading issue orthogonally.

wking:
Picking this one back up for more discussion.

PR #435, opened 2020-08-13
pkg/cvo/sync_worker: TaskNodeComplete events during updates

wking:
Only during updates, because:

* Install-time is a free-for-all, where the CVO doesn't block on anything.  This would be a lot of "node complete" noise about nodes where we had only attempted to push manifests, and that's unlikely to be what event-readers expect `TaskNodeComplete` to imply.

* Reconcile-time hopefully has very few instances where the CVO needs to stomp on changes, block on a recently `Available=False` operator, etc.  Eventing on each completed `TaskNode` would be lots of noise without much interesting signal.

During updates, we have the structured graph and blocking `TaskNodes` described [here](https://github.com/openshift/cluster-version-operator/blob/master/docs/user/reconciliation.md#manifest-graph), and the flow through that graph is what the events from this commit will help shed light on.  You could also achieve this by preserving logs from the CVO pods as they are repositioned throughout an update, but we don't have tooling in CI to do that conveniently today.

The hardcoded name and namespace for `cvoObjectRef` isn't great (for example, it won't work in `pkg/start/start_integration_test.go` , where the ClusterVersion's name and namespace are random).  But it's the pattern we've used since we started eventing in 475e71f334 (#411), so I'm recylcing it for now.

PR #484, opened 2020-12-01
Bug 1903382: pkg/payload/task_graph: Require firstIncompleteNode to have tasks

wking:
Our graph has some nodes without tasks, e.g. nodes which create a choke point between two sets of parallel nodes.  Ideally we give all nodes descriptive names (#435), but until we have that, require at least one entry in `Tasks` before we store a node as `firstIncompleteNode`.
This avoids the chance of panicking during the subsequent:

```go
fmt.Errorf("%d incomplete task nodes, beginning with %s", incompleteCount, firstIncompleteNode.Tasks[0])
```

### DIFF
--- a/pkg/payload/task_graph.go
+++ b/pkg/payload/task_graph.go
@@ -536,7 +536,7 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
 	incompleteCount := 0
 	for i, result := range results {
 		if result == nil {
-			if firstIncompleteNode == nil {
+			if firstIncompleteNode == nil && len(graph.Nodes[i].Tasks) > 0 {
 				firstIncompleteNode = graph.Nodes[i]
 			}
 			incompleteCount++


pkg/payload/task_graph.go
```diff
@@ -536,7 +536,7 @@ func RunGraph(ctx context.Context, graph *TaskGraph, maxParallelism int, fn func
 	incompleteCount := 0
 	for i, result := range results {
 		if result == nil {
-			if firstIncompleteNode == nil {
+			if firstIncompleteNode == nil && len(graph.Nodes[i].Tasks) > 0 {
```
deads2k, line 539: if you do this, the check on line 548 needs to consider the case where firstIncompleteNode is  nil and the incompleteCount is greater than zero.

jottofar:
Why is it preferred to change the selection of `firstIncompleteNode` rather than checking the `Tasks` length in the `fmt.Errorf` and indicating the `firstIncompleteNode` has no tasks? It's still the first incomplete node.

pkg/payload/task_graph.go
wking: [Line 548][1] already has a guard for `firstIncompleteNode != nil`.  If `incompleteCount` is greater than zero, and `firstIncompleteNode` is `nil`, I'm not sure we care, since no meaningful work was left on the table.  But I can check `nextNode` to make sure...

[1]: https://github.com/openshift/cluster-version-operator/blob/dc6f4f07d8986272b538b5d6bed9e8b6412b3ce6/pkg/payload/task_graph.go#L548

wking:
> It's still the first incomplete node...

But without some of the node-naming work from #435, we can't say "the empty node between $GROUP_A and $GROUP_B was canceled", and have to say "some empty node was canceled; no idea how far we got through the graph or what real work was left uncovered".  It seems more useful to try to step forward and figure out what the real work is that we're not getting to.

pkg/payload/task_graph.go
LalatenduMohanty: I am still not sure why the Task i.e. `.Tasks[0]` is nil.

deads2k: but you don't produce the error in that case

wking: > I am still not sure why the Task i.e. `.Tasks[0]` is nil.

We have empty placeholder nodes in [the graph][1], like:

Screenshot 2021-02-01 at 15 11 34

> But I can check `nextNode` to make sure...

I'm pretty sure we're fine with the change I'm proposing, without needing to involve `nextNode`.  Cases:

* All result-less nodes have no tasks.  Then with my change I will never set `firstIncompleteNode`, and we will not inject an `incomplete task nodes, beginning with...` error.  But there were no outstanding tasks, so not injecting that error is an improvement.
* The first result-less node has at least one task.  Then that matches my guard, and we get exactly the same non-panicky `firstIncompleteNode` handling as we have in master today.
* The first result-less node has no tasks, but some later result-less node has at least one task.  Then that later node will match my guard, and we'll mention it with a `incomplete task nodes, beginning with...` error instead of panicking (where we panic today).

[1]: https://github.com/openshift/cluster-version-operator/blob/1d30d9c07c4220d0271a37f5e8999316f67c3ce6/docs/user/reconciliation.md#manifest-graph

PR #464, opened 2020-10-01
Bug 1884334: pkv/cvo/status: Raise Operator leveling grace-period to 40 minutes

sdodson:
Similar to #422, further tune things up so that we can ensure that
our 90th percentile of clusters do not trip over momentary cluster
upgrade failures whenever operators take longer than 20 minutes to
roll out.

PR #486, opened 2020-12-08
Bug 1884334: UpdateError: enhance for ability to determine when upgrade failing

jottofar:
Currently during an upgrade the cluster will be marked as failing when an operator is still updating but not necessarily failing. Add the ability to set whether a given cluster operator error should result in the cluster being set to failing.

The bool  `UpgradeSucceeding` was added to the `UpdateError` structure to indicate whether the error should result in `Failing=True`. The name was chosen such that `!UpgradeSucceeding== Failing`. In this way  the default for a given error is that it results in the cluster being marked as failing and explicit action must be taken by a developer, for example when a new `UpdateError` is added, to determine and set the error such that the cluster is not marked as failing. I chose to stick with a new field in `UpdateError` as opposed to an `UpdateError` sibling to avoid having to change method signatures especially since this first round is only addressing cluster operator errors.

This PR also adds the ability to track each cluster operator's sync'ing time such that those marked as `ClusterOperatorNotAvailable`will not result in cluster `Failing=True` unless sync'ing time has exceeded 40 minutes.

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -96,6 +96,10 @@ func (b *clusterOperatorBuilder) WithModifier(f resourcebuilder.MetaV1ObjectModi
 
 func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
 	os := readClusterOperatorV1OrDie(b.raw)
+
+	// add cluster operator's start time if not already there
+	payload.COUpdateStartTimesEnsureName(os.Name)
+
 	if b.modifier != nil {
 		b.modifier(os)
 	}
@@ -128,10 +132,11 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration, 
 		actual, err := client.Get(ctx, expected.Name)
 		if err != nil {
 			lastErr = &payload.UpdateError{
-				Nested:  err,
-				Reason:  "ClusterOperatorNotAvailable",
-				Message: fmt.Sprintf("Cluster operator %s has not yet reported success", expected.Name),
-				Name:    expected.Name,
+				Nested:       err,
+				UpdateEffect: payload.UpdateEffectNone,
+				Reason:       "ClusterOperatorNotAvailable",
+				Message:      fmt.Sprintf("Cluster operator %s has not yet reported success", expected.Name),
+				Name:         expected.Name,
 			}
 			return false, nil
 		}
@@ -160,10 +165,11 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration, 
 
 			message := fmt.Sprintf("Cluster operator %s is still updating", actual.Name)
 			lastErr = &payload.UpdateError{
-				Nested:  errors.New(lowerFirst(message)),
-				Reason:  "ClusterOperatorNotAvailable",
-				Message: message,
-				Name:    actual.Name,
+				Nested:       errors.New(lowerFirst(message)),
+				UpdateEffect: payload.UpdateEffectNone,
+				Reason:       "ClusterOperatorNotAvailable",
+				Message:      message,
+				Name:         actual.Name,
 			}
 			return false, nil
 		}
@@ -210,31 +216,44 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration, 
 			}
 		}
 
+		nestedMessage := fmt.Errorf("cluster operator %s conditions: available=%v, progressing=%v, degraded=%v",
+			actual.Name, available, progressing, degraded)
+
+		if !available {
+			lastErr = &payload.UpdateError{
+				Nested:       nestedMessage,
+				UpdateEffect: payload.UpdateEffectFail,
+				Reason:       "ClusterOperatorNotAvailable",
+				Message:      fmt.Sprintf("Cluster operator %s is not available", actual.Name),
+				Name:         actual.Name,
+			}
+			return false, nil
+		}
+
 		condition := failingCondition
 		if degradedCondition != nil {
 			condition = degradedCondition
 		}
 		if condition != nil && condition.Status == configv1.ConditionTrue {
-			message := fmt.Sprintf("Cluster operator %s is reporting a failure", actual.Name)
 			if len(condition.Message) > 0 {
-				message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, condition.Message)
+				nestedMessage = fmt.Errorf("cluster operator %s is reporting a message: %s", actual.Name, condition.Message)
 			}
 			lastErr = &payload.UpdateError{
-				Nested:  errors.New(lowerFirst(message)),
-				Reason:  "ClusterOperatorDegraded",
-				Message: message,
-				Name:    actual.Name,
+				Nested:       nestedMessage,
+				UpdateEffect: payload.UpdateEffectFailAfterInterval,
+				Reason:       "ClusterOperatorDegraded",
+				Message:      fmt.Sprintf("Cluster operator %s is degraded", actual.Name),
+				Name:         actual.Name,
 			}
 			return false, nil
 		}
 
 		lastErr = &payload.UpdateError{
-			Nested: fmt.Errorf("cluster operator %s is not done; it is available=%v, progressing=%v, degraded=%v",
-				actual.Name, available, progressing, degraded,
-			),
-			Reason:  "ClusterOperatorNotAvailable",
-			Message: fmt.Sprintf("Cluster operator %s has not yet reported success", actual.Name),
-			Name:    actual.Name,
+			Nested:       nestedMessage,
+			UpdateEffect: payload.UpdateEffectNone,
+			Reason:       "ClusterOperatorNotAvailable",
+			Message:      fmt.Sprintf("Cluster operator %s is updating versions", actual.Name),
+			Name:         actual.Name,
 		}
 		return false, nil
 	}, ctx.Done())
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"strings"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -240,14 +241,18 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 		})
 	}
 
-	progressReason, progressShortMessage, skipFailure := convertErrorToProgressing(config.Status.History, now.Time, status)
+	progressReason, progressMessage, skipFailure := convertErrorToProgressing(config.Status.History, now.Time, status)
 
 	if err := status.Failure; err != nil && !skipFailure {
 		var reason string
-		msg := "an error occurred"
+		msg := progressMessage
 		if uErr, ok := err.(*payload.UpdateError); ok {
 			reason = uErr.Reason
-			msg = payload.SummaryForReason(reason, uErr.Name)
+			if msg == "" {
+				msg = payload.SummaryForReason(reason, uErr.Name)
+			}
+		} else if msg == "" {
+			msg = "an error occurred"
 		}
 
 		// set the failing condition
@@ -304,7 +309,7 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 			case fractionComplete > 0 && skipFailure:
 				reason = progressReason
 				message = fmt.Sprintf("Working towards %s: %d of %d done (%.0f%% complete), %s", version,
-					status.Done, status.Total, math.Trunc(float64(fractionComplete*100)), progressShortMessage)
+					status.Done, status.Total, math.Trunc(float64(fractionComplete*100)), progressMessage)
 			case fractionComplete > 0:
 				message = fmt.Sprintf("Working towards %s: %d of %d done (%.0f%% complete)", version,
 					status.Done, status.Total, math.Trunc(float64(fractionComplete*100)))
@@ -315,7 +320,7 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 				message = fmt.Sprintf("Working towards %s: downloading update", version)
 			case skipFailure:
 				reason = progressReason
-				message = fmt.Sprintf("Working towards %s: %s", version, progressShortMessage)
+				message = fmt.Sprintf("Working towards %s: %s", version, progressMessage)
 			default:
 				message = fmt.Sprintf("Working towards %s", version)
 			}
@@ -348,21 +353,35 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 
 // convertErrorToProgressing returns true if the provided status indicates a failure condition can be interpreted as
 // still making internal progress. The general error we try to suppress is an operator or operators still being
-// unavailable AND the general payload task making progress towards its goal. An operator is given 40 minutes since
-// its last update to go ready, or an hour has elapsed since the update began, before the condition is ignored.
+// unavailable AND the general payload task making progress towards its goal. The error's UpdateEffect determines
+// whether an error should be considered a failure and, if so, whether the operator should be given up to 40 minutes
+// to recover from the error.
 func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time, status *SyncWorkerStatus) (reason string, message string, ok bool) {
-	if len(history) == 0 || status.Failure == nil || status.Reconciling || status.LastProgress.IsZero() {
-		return "", "", false
-	}
-	if now.Sub(status.LastProgress) > 40*time.Minute || now.Sub(history[0].StartedTime.Time) > time.Hour {
+	if len(history) == 0 || status.Failure == nil || status.Reconciling {
 		return "", "", false
 	}
 	uErr, ok := status.Failure.(*payload.UpdateError)
 	if !ok {
 		return "", "", false
 	}
-	if uErr.Reason == "ClusterOperatorNotAvailable" || uErr.Reason == "ClusterOperatorsNotAvailable" {
+	switch uErr.UpdateEffect {
+	case payload.UpdateEffectNone:
 		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+	case payload.UpdateEffectFail:
+		return "", "", false
+	case payload.UpdateEffectFailAfterInterval:
+		var exceeded []string
+		threshold := now.Add(-(40 * time.Minute))
+		for _, name := range strings.Split(uErr.Name, ", ") {
+			if payload.COUpdateStartTimesGet(name).Before(threshold) {
+				exceeded = append(exceeded, name)
+			}
+		}
+		if len(exceeded) > 0 {
+			return uErr.Reason, fmt.Sprintf("wait has exceeded 40 minutes for these operators: %s", strings.Join(exceeded, ", ")), false
+		} else {
+			return uErr.Reason, fmt.Sprintf("waiting up to 40 minutes on %s", uErr.Name), true
+		}
 	}
 	return "", "", false
 }
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -548,6 +548,10 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 	}
 	klog.V(4).Infof("Running sync %s (force=%t) on generation %d in state %s at attempt %d", versionString(desired), work.Desired.Force, work.Generation, work.State, work.Attempt)
 
+	if work.Attempt == 0 {
+		payload.InitCOUpdateStartTimes()
+	}
+
 	// cache the payload until the release image changes
 	validPayload := w.payload
 	if validPayload != nil && validPayload.Release.Image == desired.Image {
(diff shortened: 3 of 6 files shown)


jottofar:
/retitle Bug 1884334: UpdateError: add and use Progressing field

pkg/cvo/internal/operatorstatus.go
```diff
@@ -232,9 +235,10 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
 			Nested: fmt.Errorf("cluster operator %s is not done; it is available=%v, progressing=%v, degraded=%v",
 				actual.Name, available, progressing, degraded,
 			),
-			Reason:  "ClusterOperatorNotAvailable",
-			Message: fmt.Sprintf("Cluster operator %s has not yet reported success", actual.Name),
-			Name:    actual.Name,
+			Progressing: progressing,
```
wking, line 238: This adds sensitivity to the `Progressing` condition, which [we should drop][1].  And maybe that's a reason to name your new property something other than `Progressing`?  Because the issue is less "are we attempting to reconcile divergent state" and more "are we worried about the current state".  E.g. running an outdated operator for a few minutes while it updates is progressing, but should not result in `Failing=True`.  And running a leveled, happy operator is not progressing, but should not result in `Failing=True` either.  Possibly call it `Satisfactory`?  Or call it `Failing` and make it a `*bool` to distinguish folks who aren't setting it?  I dunno...

[1]: https://github.com/openshift/cluster-version-operator/blame/b6cfb178a4cb0ddb735084244643f3bbd7a34c78/docs/user/reconciliation.md#L137

pkg/cvo/status.go
```diff
@@ -347,13 +347,16 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 // unavailable AND the general payload task making progress towards its goal. An operator is given 40 minutes since
 // its last update to go ready, or an hour has elapsed since the update began, before the condition is ignored.
 func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time, status *SyncWorkerStatus) (reason string, message string, ok bool) {
+	uErr, ok := status.Failure.(*payload.UpdateError)
 	if len(history) == 0 || status.Failure == nil || status.Reconciling || status.LastProgress.IsZero() {
+		if ok && uErr.Progressing {
+			return "", "", true
```
wking, line 353: This removes the "...but the second-level operator has been at it for over 40m and still hasn't leveled".  We want to drop that limit eventually, but I'm not sure if we're ready to drop it quite yet.  And see also the places touched in openshift/cluster-version-operator#464 that would need updating if we do decide to drop the timeout entirely.

jottofar: Are you sure? I thought I added it in the least impactful way possible. The 40m check is next so the existing code would have already entered this `if` and returned before my mod.

jottofar:
/retitle Bug 1884334: UpdateError: enhance for ability to determine when upgrade failing

pkg/cvo/internal/operatorstatus.go
```diff
@@ -128,10 +132,11 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
 		actual, err := client.Get(ctx, expected.Name)
 		if err != nil {
 			lastErr = &payload.UpdateError{
-				Nested:  err,
-				Reason:  "ClusterOperatorNotAvailable",
-				Message: fmt.Sprintf("Cluster operator %s has not yet reported success", expected.Name),
-				Name:    expected.Name,
+				Nested:            err,
+				UpgradeSucceeding: true,
```
wking, line 136: This is "Kube API-server down/unreachable or requested ClusterOperator is missing".  That all seems serious enough to be `UpgradeSucceeding: false`, although I guess it might be worth a safety valve to avoid freaking out on temporary networking/Kube-API-server hiccups.  Do we want something more nuanced in error type checking here?

pkg/cvo/internal/operatorstatus.go
```diff
@@ -232,9 +239,10 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
 			Nested: fmt.Errorf("cluster operator %s is not done; it is available=%v, progressing=%v, degraded=%v",
 				actual.Name, available, progressing, degraded,
 			),
-			Reason:  "ClusterOperatorNotAvailable",
-			Message: fmt.Sprintf("Cluster operator %s has not yet reported success", actual.Name),
-			Name:    actual.Name,
+			UpgradeSucceeding: true,
```
wking, line 242: This case is a pretty big bucket, and I'm not sure all of the situations that hit this code are actually succeeding.  Having `Available=False` seems like it deserves `UpgradeSucceeding: false`.  Do we have other cases in here?  `ClusterOperatorNotAvailable` suggests that the original dev was mostly worried about `Available=False`, but we should probably go through the code and see if there are cases like `Available=True` but `Degraded=True` that need a more specific `Reason` and which would qualify for the less-alarming `UpgradeSucceeding: true`.

pkg/cvo/status.go
```diff
@@ -347,18 +352,34 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 // unavailable AND the general payload task making progress towards its goal. An operator is given 40 minutes since
 // its last update to go ready, or an hour has elapsed since the update began, before the condition is ignored.
 func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time, status *SyncWorkerStatus) (reason string, message string, ok bool) {
-	if len(history) == 0 || status.Failure == nil || status.Reconciling || status.LastProgress.IsZero() {
+	if len(history) == 0 || status.Failure == nil || status.Reconciling {
 		return "", "", false
 	}
-	if now.Sub(status.LastProgress) > 40*time.Minute || now.Sub(history[0].StartedTime.Time) > time.Hour {
+	if now.Sub(history[0].StartedTime.Time) > time.Hour {
```
wking, line 358: I think we can drop this "has it been an hour since we started updating?" condition too.  "We are updating, and it's been more than an hour" is not in itself enough of a reason to go `Failing=True`.  The hour bit landed in c2ac20fa172da (#158), along with the Godoc comment "or an hour has elapsed since the update began", which I want to drop along with this block.

pkg/cvo/status.go
```diff
@@ -347,18 +352,34 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
 		return "", "", false
 	}
 	uErr, ok := status.Failure.(*payload.UpdateError)
 	if !ok {
 		return "", "", false
 	}
 	if uErr.Reason == "ClusterOperatorNotAvailable" || uErr.Reason == "ClusterOperatorsNotAvailable" {
-		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+		if !uErr.UpgradeSucceeding {
```
wking, line 366: nit: it's a bit easier for me to read if we avoid the double-negative of `if !something { } else { /* makes me think too hard */ }` and use:

```go
if uErr.UpgradeSucceeding {
  return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
} else {
  // !succeeding logic
}
```

pkg/cvo/internal/operatorstatus.go
jottofar: Do we want to make this UpdateEffectFailAfterInterval?

pkg/cvo/sync_worker.go
```diff
@@ -928,6 +933,15 @@ func newClusterOperatorsNotAvailable(errs []error) error {
 		if len(uErr.Name) > 0 {
 			names = append(names, uErr.Name)
 		}
+		switch uErr.UpdateEffect {
+		case payload.UpdateEffectNone:
+		case payload.UpdateEffectFail:
+			updateEffect = payload.UpdateEffectFail
+		case payload.UpdateEffectFailAfterInterval:
+			if updateEffect != payload.UpdateEffectFail {
+				updateEffect = payload.UpdateEffectFailAfterInterval
```
wking, line 942: This collapsing is going to slightly break the effect.  For example, say in round one, operator A was `UpdateEffectFailAfterInterval`.  In round two, A is `UpdateEffectNone`, but B is `UpdateEffectFailAfterInterval`.  This code collapses them into a single error that names A and B and has `UpdateEffectFailAfterInterval`.  The consuming code sees `UpdateEffectFailAfterInterval`, sees that A has been in the doghouse since the first round, and maybe goes `Failing=True` if it's been long enough.  But still, better than what we have today, and we can always tune more in the future.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -220,21 +226,27 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
 				message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, condition.Message)
 			}
 			lastErr = &payload.UpdateError{
-				Nested:  errors.New(lowerFirst(message)),
-				Reason:  "ClusterOperatorDegraded",
-				Message: message,
-				Name:    actual.Name,
+				Nested:       errors.New(lowerFirst(message)),
+				UpdateEffect: payload.UpdateEffectFail,
+				Reason:       "ClusterOperatorDegraded",
```
wking, line 231: I still think we should ignore ClusterOperator Degraded entirely (#482), but as long as we keep it, can it at least be `UpdateEffectFailAfterInterval` or `UpdateEffectNone` (I prefer the latter, but am ok compromising on the former ;)?

pkg/cvo/status.go
```diff
@@ -347,18 +352,34 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
+		switch uErr.UpdateEffect {
+		case payload.UpdateEffectNone:
+			return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+		case payload.UpdateEffectFail:
+			return "", "", false
+		case payload.UpdateEffectFailAfterInterval:
+			var exceeded []string
+			for _, name := range strings.Split(uErr.Name, ", ") {
+				if name != "" {
+					if now.Sub(payload.ClusterOperatorUpdateStartTimes[name]) > 40*time.Minute {
```
wking, line 372: nit: slightly more efficient and possibly easier to read with:

```go
threshold := now.Sub(40*time.Mintute)
for _, name := range strings.Split(uErr.Name, ", ") {
  if payload.ClusterOperatorUpdateStartTimes[name] < threshold {
    exceeded = append(exceeded, name)
  }
}
```

pkg/cvo/internal/operatorstatus.go
jottofar: This now results in failing if `Available=False`

pkg/cvo/status.go
jottofar: redone

pkg/cvo/internal/operatorstatus.go
jottofar: I can soften this as you're suggesting but shouldn't this also only apply if `Available=True`. So overall it's really only `ClusterOperatorDegraded` if `Available=True` && `Degraded=True`? Otherwise I'll drop through and handle it below resulting in 'ClusterOperatorNotAvailable` ...

jottofar: Changed per above comment

pkg/cvo/status.go
jottofar: agree

pkg/cvo/internal/operatorstatus.go
wking: Yeah, `Available=False` should be `UpdateEffectFail` regardless of `Degraded`.  dee13735 looks good to me.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -96,6 +96,10 @@ func (b *clusterOperatorBuilder) WithModifier(f resourcebuilder.MetaV1ObjectModi
 
 func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
 	os := readClusterOperatorV1OrDie(b.raw)
+
+	if _, ok := payload.ClusterOperatorUpdateStartTimes[os.Name]; !ok {
+		payload.ClusterOperatorUpdateStartTimes[os.Name] = time.Now()
```
wking, line 101: Both [e2e][1] and [e2e-operator][2] died on:

```
level=fatal msg=Bootstrap failed to complete: failed to wait for bootstrapping to complete: timed out waiting for the condition
```

In the log bundle:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/486/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1352370358094663680/artifacts/e2e-agnostic/ipi-install-install/log-bundle-20210121224256.tar | tar -xOz log-bundle-20210121224256/bootstrap/containers/cluster-version-operator-05aa025888b3bee7a1b52ba61bef85baace11a35eae7a3f9e6fe0f0174fbbaaf.log
...
E0121 22:42:32.828219       1 runtime.go:78] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
goroutine 216 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1856060, 0x1c47440)
        /go/src/github.com/openshift/cluster-version-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/src/github.com/openshift/cluster-version-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x89
panic(0x1856060, 0x1c47440)
        /usr/lib/golang/src/runtime/panic.go:969 +0x1b9
github.com/openshift/cluster-version-operator/pkg/cvo/internal.(*clusterOperatorBuilder).Do(0xc001758000, 0x1ca5440, 0xc000d4ecc0, 0xc001758000, 0xc001758000)
        /go/src/github.com/openshift/cluster-version-operator/pkg/cvo/internal/operatorstatus.go:101 +0x489
github.com/openshift/cluster-version-operator/pkg/cvo.(*resourceBuilder).Apply(0xc0019063f0, 0x1ca5440, 0xc000d4ecc0, 0xc001a82be0, 0x3, 0x0, 0x0)
        /go/src/github.com/openshift/cluster-version-operator/pkg/cvo/cvo.go:713 +0xcb
github.com/openshift/cluster-version-operator/pkg/cvo.(*SyncWorker).apply.func1(0x1ca5440, 0xc000d4ecc0, 0xc000ce6368, 0x9, 0x393, 0x0, 0x1a6c757)
        /go/src/github.com/openshift/cluster-version-operator/pkg/cvo/sync_worker.go:747 +0x299
github.com/openshift/cluster-version-operator/pkg/payload.RunGraph.func3(0xc0009ff280, 0xc001e8da40, 0xc001629de0, 0xc001e8daa0, 0x1ca5440, 0xc000d4ecc0, 0x6)
        /go/src/github.com/openshift/cluster-version-operator/pkg/payload/task_graph.go:487 +0xed
created by github.com/openshift/cluster-version-operator/pkg/payload.RunGraph
        /go/src/github.com/openshift/cluster-version-operator/pkg/payload/task_graph.go:477 +0x32f
I0121 22:42:32.828247       1 task_graph.go:486] Running 23 on worker 5
panic: assignment to entry in nil map [recovered]
        panic: assignment to entry in nil map
        ...
```

You have the `make(map...)` call in `sync_worker.go`, but maybe add a fallback here to guard against that `apply()` pathway?  Might be getting called via a parallel thread; I haven't traced it out yet.  Probably worth tracing out in case we need to guard it with a lock...  Actually, we definitely want to guard it with a lock, see [here][3] and [here][4].

There may also be alternatives to a map that are more convenient to concurrency.  Maybe we could inspect the manifest around [here][5] and feed "I'm about to attempt $CLUSTER_VERSION_OPERATOR_NAME" into a local channel that we collect into the map in a parallel goroutine (so only that one goroutine would be writing to the map)?  I dunno.  Maybe we ditch the "40m" thing entirely as "not worth the trouble" ;).

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/486/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1352370358094663680
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/486/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator/1352370358128218112
[3]: https://blog.golang.org/maps#TOC_6.
[4]: https://golang.org/doc/faq#atomic_maps
[5]: https://github.com/openshift/cluster-version-operator/blob/f9c42b254b2103be17ca26519f500490a80db125/pkg/cvo/sync_worker.go#L768

pkg/cvo/internal/operatorstatus.go
```diff
@@ -96,6 +96,11 @@ func (b *clusterOperatorBuilder) WithModifier(f resourcebuilder.MetaV1ObjectModi
 
 func (b *clusterOperatorBuilder) Do(ctx context.Context) error {
 	os := readClusterOperatorV1OrDie(b.raw)
+
+	// add cluster operator's start time if not already there
+	if !payload.COUpdateStartTimesHasName(os.Name) {
+		payload.COUpdateStartTimesAdd(os.Name)
```
wking, line 102: this will create two lock/release cycles.  Can we make this a single `payload.COUpdateStartTimesEnsureName` or something that is add-if-missing?

jottofar: Good catch. I had it like this when I originally used a RW lock, which had its own issues, so yes, make sense to only lock once.

jottofar: Changed in [9298cb3](https://github.com/openshift/cluster-version-operator/pull/486/commits/9298cb355153d4145154c5c37083eda8a01be7d3)

pkg/cvo/internal/operatorstatus.go
```diff
@@ -210,31 +216,40 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
...
-				Message: message,
-				Name:    actual.Name,
+			if condition != nil && condition.Status == configv1.ConditionTrue {
+				message := fmt.Sprintf("Cluster operator %s is reporting a failure", actual.Name)
+				if len(condition.Message) > 0 {
+					message = fmt.Sprintf("Cluster operator %s is reporting a failure: %s", actual.Name, condition.Message)
+				}
+				lastErr = &payload.UpdateError{
+					Nested:       errors.New(lowerFirst(message)),
+					UpdateEffect: payload.UpdateEffectNone,
```
wking, line 232: this is `Available=True, Degraded=True`, right?  Maybe that should be `UpdateEffectFailAfterInterval`?  Possibly by shifting the `updateEffect` definition up before this block and using `updateEffect` here?  I'm personally ok ditching `UpdateEffectFailAfterInterval` and also with not caring about `Degraded`, but while we have both, it seems strange to have `Available=True, Degraded=False, Progressing=True` send us `Failing=True` after the interval, while `Available=True, Degraded=True` stays `Failing=False`.

pkg/cvo/status.go
```diff
@@ -351,18 +356,33 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
-	}
-	if now.Sub(status.LastProgress) > 40*time.Minute || now.Sub(history[0].StartedTime.Time) > time.Hour {
+	if len(history) == 0 || status.Failure == nil || status.Reconciling {
 		return "", "", false
 	}
 	uErr, ok := status.Failure.(*payload.UpdateError)
 	if !ok {
 		return "", "", false
 	}
 	if uErr.Reason == "ClusterOperatorNotAvailable" || uErr.Reason == "ClusterOperatorsNotAvailable" {
```
wking, line 366: We might need to include `ClusterOperator(s)Degraded` in here too?

pkg/cvo/internal/operatorstatus.go
jottofar: reworked per our discussion

pkg/cvo/status.go
jottofar: reworked per our discussion

pkg/cvo/internal/operatorstatus.go
```diff
@@ -210,31 +216,44 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
...
 		lastErr = &payload.UpdateError{
-			Nested: fmt.Errorf("cluster operator %s is not done; it is available=%v, progressing=%v, degraded=%v",
-				actual.Name, available, progressing, degraded,
-			),
-			Reason:  "ClusterOperatorNotAvailable",
-			Message: fmt.Sprintf("Cluster operator %s has not yet reported success", actual.Name),
-			Name:    actual.Name,
+			Nested:       nestedMessage,
+			UpdateEffect: payload.UpdateEffectNone,
+			Reason:       "ClusterOperatorNotAvailable",
```
wking, line 254: We might want a different reason for this, but we can twiddle that in follow-up work.

pkg/payload/task.go
```diff
@@ -112,12 +143,28 @@ func (st *Task) Run(ctx context.Context, version string, builder ResourceBuilder
 	}
 }
 
+// UpdateEffectType defines the effect an update error has on the overall update state.
+type UpdateEffectType string
+
+const (
+	// "None" defines an error as having no affect on the update state.
```
wking, line 150: nit: We should add Go linter.  I expect it would want these godocs to start with the exposed name, so:

```go
// UpdateEffectNone defines an error as having no affect on the update state.
```

and similar for the other constants below.

wking:
/lgtm
/hold

Holding in case you want to address either of the two new nits.  I'm fine punting them both to follow up work as well, in which case, just pull the hold.

jottofar:
/unhold

Let's address them in a follow-up so this gets some soak time before code cutoff.

pkg/cvo/internal/operatorstatus.go
jottofar: added lock

PR #491, opened 2020-12-17
WIP: pkg/cvo/sync_worker: Printf debugging for the initializing -> * transition

wking:
Poking around on [rhbz#1905221](https://bugzilla.redhat.com/show_bug.cgi?id=1905221).


wking:
#500 has the fix.

/close

PR #500, opened 2021-01-19
Bug 1905221: pkg/cvo/sync_worker.go: Ignore work changes during init

jottofar:
Ignore version and overrides changes during payload initialization and log warning.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -222,7 +222,7 @@ func (w *SyncWorker) Update(generation int64, desired configv1.Update, overrides
 		return w.status.DeepCopy()
 	}
 
-	if equalSyncWork(w.work, work) {
+	if equalSyncWork(w.work, work, state) {
 		klog.V(5).Info("Update work is equal to current target; no change required")
 		return w.status.DeepCopy()
 	}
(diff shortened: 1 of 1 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -476,12 +476,24 @@ func equalSyncWork(a, b *SyncWork) bool {
 	overridesEqual := reflect.DeepEqual(a.Overrides, b.Overrides)
 
 	if !sameVersion && !overridesEqual {
+		if state == payload.InitializingPayload {
+			klog.Warning("Ignoring detected version and overrides changes during payload initialization")
+			return true
+		}
 		klog.V(5).Info("Detected version and overrides changes")
```
wking, line 483: Trying to think of a way to DRY this up, how about:

```go
var difference string
if !sameVersion && !overridesEqual {
  difference = "version and overrides changes"
} else if ... {
  ... other cases...
}
if difference != "" {
  if state == payload.InitializingPayload {
    klog.Warningf("Ignoring detected %s during payload initialization", difference)
    return true
  }
  klog.V(5).Infof("Detected %s", difference)
  return false
}
```

jottofar: I can do that - and had considered it- but I think it sacrifices readability.

jottofar: Changed in [213726d](https://github.com/openshift/cluster-version-operator/pull/500/commits/213726df85663fb9ec3e7b7c5671906a3feab6a5)

wking (approved):
/lgtm

PR #507, opened 2021-01-27
Bug 1921277: pkg/cvo/sync_worker.go: Fix log statements to handle arguments

jottofar:


wking:
Fixes a bug from #500.

/lgtm

PR #512, opened 2021-02-04
Bug 1925199: add current profile annotations to CVO manifests

jottofar:
This is matches openshift/enhancements#510 and doesn't change existing behavior

### DIFF
--- a/install/0000_00_cluster-version-operator_00_namespace.yaml
+++ b/install/0000_00_cluster-version-operator_00_namespace.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   name: openshift-cluster-version
   annotations:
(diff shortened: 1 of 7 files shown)


wking (approved):
```console
$ cherry-pick-diff origin/release-4.6..origin/pr/512 origin/master
d4700fb6 -> 15ac2971 add current profile annotations to CVO manifests
```

So a clean pick of the portion of #404 that we want to bring back to 4.6 (if we brought back #8f4d36c4, that would break 4.6->4.5 rollbacks).

/lgtm

rhpmali:
As I am not the QA contact for BZ 1925199, so did not follow the pre-merge workflow. But while testing its failed. Details are in https://bugzilla.redhat.com/show_bug.cgi?id=1925199#c3

PR #518, opened 2021-02-11
WIP: Ensure ClusterVersion gets updated when CVO acquires lease

jottofar:


jottofar:
/retitle Bug 1909875: Ensure ClusterVersion gets updated when CVO acquires lease

pkg/cvo/status.go
```diff
@@ -343,6 +344,11 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 		})
 	}
 
+	// applyClusterVersionStatus will always update ClusterVersion when 'original' is nil
+	if forceCVUpdate {
+		original = nil
```
wking, line 349: Do we have a story for why we'd need this hammer?  We have an `original` and our desired `config`.  There are no early `return`s in `Operator.syncStatus`, so regardless of what we do here, we'll drop down into `applyClusterVersionStatus`.  The only place `original` comes in there is for a `equality.Semantic.DeepEqual` early-`return`.  But if we are bailing out in that early `return`, doesn't it mean our desired content already matches what is in-cluster?  In which case it would be appropriate to not sync the status.

jottofar:
/retitle WIP: Ensure ClusterVersion gets updated when CVO acquires lease

PR #508, opened 2021-01-28
Bug 1921413: pkg/cvo/cvo: Guard Operator.Run goroutine handling from early cancels

wking:
Builds on #501; possibly review that first.

In [rbhz#1921413][1], we had a CVO fail to gracefully release the leader lock, with:

```console
$ grep start.go: namespaces/openshift-cluster-version/pods/cluster-version-operator-968d9fd48-25zkz/cluster-version-operator/cluster-version-operator/logs/current.log
2021-01-26T19:02:48.678276895Z I0126 19:02:48.678120       1 start.go:21] ClusterVersionOperator 4.6.0-202012050130.p0-39a4256
2021-01-26T19:02:49.273117163Z I0126 19:02:49.273074       1 start.go:260] Waiting on 2 outstanding goroutines.
2021-01-26T19:02:49.273512006Z E0126 19:02:49.273474       1 start.go:271] Collected metrics server goroutine: listen tcp 0.0.0.0:9099: bind: address already in use
2021-01-26T19:02:49.273587622Z I0126 19:02:49.273551       1 start.go:260] Waiting on 1 outstanding goroutines.
2021-01-26T19:02:49.273587622Z I0126 19:02:49.273569       1 start.go:264] Run context completed; beginning two-minute graceful shutdown period.
2021-01-26T19:02:49.273587622Z I0126 19:02:49.273577       1 start.go:260] Waiting on 1 outstanding goroutines.
2021-01-26T19:04:49.273938652Z I0126 19:04:49.273877       1 start.go:260] Waiting on 1 outstanding goroutines.
2021-01-26T19:05:39.264746935Z I0126 19:05:39.264690       1 start.go:190] Shutting down due to terminated
```

So the bind failed, which should be rare, but can happen if the new CVO is scheduled before the kernel gets around to reaping the port from a previous CVO pod.  The `start.go` handling notices the metrics server dying, and immediately cancels `runContext`, and we go into the two-minute shutdown grace period.  That happened so early that the context was already done by the time we got into `Operator.Run`, with its:

```go
go wait.UntilWithContext(runContext, func(runContext context.Context) {
 defer close(workerStopCh)
```

From the godocs for `UntilWithContext`:

```console
$ git --no-pager grep -n '// UntilWithContext is syntactic sugar\|f may not be invoked if context\|func JitterUntilWithContext' vendor/k8s.io/apimachinery/pkg/util/wait
[...]

PR #519, opened 2021-02-11
Bug 1927944: pkg/start: Fix shutdown deadlock when die before getting a leader lock

wking:
a9e075a79d (#508) made us more robust to situations where we are canceled after acquiring the leader lock but before we got into `Operator.Run`'s `UntilWithContext`.  However, there was [still a bug][1] from cc1921d186 (#424) where we had not acquired the leader lock.  `postMainContext` is used for metrics, informers, and the leader election loop.  We used to only call `postMainCancel` after reaping the main goroutine, and obviously that will only work if we've launched the main goroutine.  This commit adds a new `launchedMain` to track that.  If `launchedMain` is true, we get the old handling.  If `launchedMain` is still false when `runContext.Done`, we now call `postMainCancel` without waiting to reap a non-existent main goroutine.

There's also a new `postMainCancel` when the shutdown timer expires.  I don't expect us to ever need that, but it protects us from future bugs like this one.

I've added `launchedMain` without guarding it behind a lock, and it is touched by both the main `Options.run` goroutine and the leader-election callback.  So there's a racy chance of:

1. `Options.run` goroutine: `runContext` canceled, so `runContext.Done()`
2. Leader-election goroutine: Leader lock acquired
3. `Options.run` goroutine: `!launchedMain`, so we call `postMainCancel()`
4. Leader-election goroutine: `launchedMain` set true
5. Leader-election goroutine: launches the main goroutine via `CVO.Run(runContext, ...)`

I'm trusting `Operator.Run` to respect `runContext` there and not do anything significant, so the fact that we are already tearing down all
the post-main stuff won't cause problems.  Previous fixes like a9e075a79d will help with that.  But there could still be bugs in `Operator.Run`.  A lock around `launchedMain` that avoided calling `Operator.Run` when `runContext` was already done would protect against that, but seems like overkill in an already complicated goroutine tangle.  Without the lock, we just have to field and fix any future `Operator.Run` `runContext` issues as we find them.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1927944

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -184,6 +184,7 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 	defer shutdownCancel()
 	postMainContext, postMainCancel := context.WithCancel(context.Background()) // extends beyond ctx
 	defer postMainCancel()
+	launchedMain := false
 
 	ch := make(chan os.Signal, 1)
 	defer func() { signal.Stop(ch) }()
@@ -234,6 +235,7 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 			RetryPeriod:     retryPeriod,
 			Callbacks: leaderelection.LeaderCallbacks{
 				OnStartedLeading: func(_ context.Context) { // no need for this passed-through postMainContext, because goroutines we launch inside will use runContext
+					launchedMain = true
 					resultChannelCount++
 					go func() {
 						defer utilruntime.HandleCrash()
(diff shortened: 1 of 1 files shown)


jottofar:
I wouldn't consider adding a lock around `launchedMain` as adding much complexity especially if we think it's at all likely it may introduce a future bug. On the other hand, I know not having the lock could introduce a different logic flow but shouldn't `controllerCtx.CVO.Run` support exiting "cleanly" whenever  `runContext` is cancelled?

jottofar:
/lgtm

Discussed my above comment with Trevor on slack and we agreed that if `controllerCtx.CVO.Run` cannot handle runContext cancellation it most likely is a bug unrelated to launchedMain locking and something we want to fix.

PR #520, opened 2021-02-12
Adjust alert severities

michaelgugino:
The alerts that are currently "critical" are not actually
situations that jeopardize the cluster's immediate health
or ability to run workloads.  No one needs to be paged
in the middle of the night for these alerts.

This commit reduces severity to warning to reflect
degraded state.

### DIFF
--- a/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
+++ b/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
@@ -45,7 +45,7 @@ spec:
         absent(up{job="cluster-version-operator"} == 1)
       for: 10m
       labels:
-        severity: critical
+        severity: warning
     - alert: CannotRetrieveUpdates
       annotations:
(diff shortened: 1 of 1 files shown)


sdodson:
/hold
I don't disagree with this, but in order to reduce churn I'd like to see https://github.com/openshift/enhancements/pull/637 accepted first

wking:
I'm not really interested in getting in the middle here.  @smarterclayton was pushing reasonably firmly for `critical` [back when these all landed][1].  I'll let folks fight out what they think the severity level should be, and focus on trying to reduce the number of times any of these go off ;).

[1]: https://github.com/openshift/cluster-version-operator/pull/232#discussion_r316870985

smarterclayton:
As per discussions in the alert enhancement, and in slack, and a general "when is available and degraded important" - degraded generally means 'still working but not making progress / unable to give full status'.  By that definition degraded =~ warning, and timing should be "in the morning".

By that measure available=false = critical, I am ok (if trevor is) with splitting ClusterOperatorDown to only include available and reachable and be critical, then to make ClusterOperatorDegraded cover the degraded condition and have a longer `for`.

Finally, if we do that, suppressing operation condition alerts until install is complete (until the operator reaches level) may also be useful.

michaelgugino:
These CVO conditions clearly don't meet the criteria of 'get out of bed in the middle of the night.'  Workloads will continue to run, new workloads will continue to be scheduled, existing operators will continue to run.  The CVO being completely down could persist indefinitely with no adverse impact on workloads.

wking:
> ... ClusterOperatorDegraded cover the degraded condition and have a longer `for`.

I've spun this portion out into #554.  I'm more interested in the longer `for`, but I also dropped the severity to `warning`.  Let me know if you'd rather me leave the severity changing to this PR or not.  I haven't addressed `ClusterOperatorDown`'s timing yet, but I'll try to get something in place for that soon, and then folks can tell me if they want me to twiddle severity for it in that PR or leave that to this PR as well.

PR #524, opened 2021-02-19
Update a doc example to match modern OCP

danwinship:
The example of how to make an operator unmanaged makes assumptions that probably haven't been true since 4.1. The _idea_ is still right, but the specifics of it are all wrong...


LalatenduMohanty (approved):
/lgtm

sdodson:
This only changes in tree docs, overriding bugzilla/valid-bug label.

PR #525, opened 2021-02-20
Bug 1931025: pkg/cvo: Use shutdownContext for final status synchronization

wking:
This commit picks c4ddf03bec (#517) back to 4.5.  It's not a clean pick, because we're missing changes like:

* b72e843974 (Bug 1822844: Block z level upgrades if ClusterVersionOverridesSet set, 2020-04-30, #364).
* 1d1de3ba1a (Use context to add timeout to cincinnati HTTP request, 2019-01-15, #410).

which also touched these lines.  But we've gotten this far without backporting rhbz#1822844, and #410 was never associated with a bug in the first place, so instead of pulling back more of 4.6 to get a clean pick, I've just manually reconciled the pick conflicts.

/assign @jottofar

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -314,17 +314,20 @@ func loadConfigMapVerifierDataFromUpdate(update *payload.Update, clientBuilder v
 	return nil, nil, nil
 }
 
-// Run runs the cluster version operator until stopCh is completed. Workers is ignored for now.
-func (optr *Operator) Run(ctx context.Context, workers int) error {
+// Run runs the cluster version operator until runContext.Done() and
+// then attempts a clean shutdown limited by shutdownContext.Done().
+// Assumes runContext.Done() occurs before or simultaneously with
+// shutdownContext.Done().
+func (optr *Operator) Run(runContext context.Context, shutdownContext context.Context, workers int) error {
 	defer optr.queue.ShutDown()
-	stopCh := ctx.Done()
+	stopCh := runContext.Done()
 	workerStopCh := make(chan struct{})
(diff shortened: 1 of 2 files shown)


pkg/cvo/cvo.go
```diff
@@ -314,38 +314,43 @@ func loadConfigMapVerifierDataFromUpdate(update *payload.Update, clientBuilder v
...
 	if !cache.WaitForCacheSync(stopCh, optr.cacheSynced...) {
-		return fmt.Errorf("caches never synchronized: %w", ctx.Err())
+		return fmt.Errorf("caches never synchronized: %w", runContext.Err())
 	}
 
 	// trigger the first cluster version reconcile always
 	optr.queue.Add(optr.queueKey())
 
 	// start the config sync loop, and have it notify the queue when new status is detected
 	go runThrottledStatusNotifier(stopCh, optr.statusInterval, 2, optr.configSync.StatusCh(), func() { optr.queue.Add(optr.queueKey()) })
```
jottofar, line 337: This method was changed in https://github.com/openshift/cluster-version-operator/pull/517 as well. Did you miss it?

wking: In #517, `runThrottledStatusNotifier` is using `runContext`, which is effectively the same as the `stopCh` I use here (see `stopCh := runContext.Done()` above).  In master, `runThrottledStatusNotifier` was ported to use `Context` in #410, which was not associated with a bug, and which I'm not pulling back here.  So while I could backport that portion of #410 here, and leave of the rest of #410, I haven't done that with the current PR tip.

PR #526, opened 2021-02-22
Remove toleration of unschedulable

michaelgugino:
Currently, the CVO tolerates the taint
'node.kubernetes.io/unschedulable'.  This is undersireable as it will
result in the pod potentially being rescheduled onto a cordoned host.

This commit removes this particular toleration from the CVO deployment
so we do not conflict with administrative drains.


wking:
Makes sense to me, but can we land openshift/enhancements#650 first?

michaelgugino:
> Makes sense to me, but can we land [openshift/enhancements#650](https://github.com/openshift/enhancements/pull/650) first?

Sure.  Do you have any more feedback on that enhancement?

wking (approved):
/lgtm
/retest

PR #514, opened 2021-02-04
pkg/cvo/internal/operatorstatus: Change nested message

jottofar:
to a different message for cluster operators that are available, not degraded, but not yet finished updating

PR #527, opened 2021-02-27
pkg/cvo/internal/operatorstatus: Drop deprecated failing/progressing handling

wking:
Also simplify the `undone` type from a map to a slice of strings.  The only consumers are length checks and `nestedMessage` formatting, so we don't need to preserve structured access to the current and expected values.

I've also dropped the old, early "Cluster operator %s is still updating" guard.  If the ClusterOperator is degraded or unavailable, those are much more serious than being behind in version, so we want those guards coming first.  The undone guard is now only the final fallback for available, non-degraded operators that aren't completely satisfactory (and I've pivoted to use `ClusterOperatorUpdating` for that reason).

I'm also now keeping any available condition data, and including that status, reason, and message in the nested message, because it seemed strange to do that for degraded, but not for the more serious unavailable.

And finally, I've dropped the old mode switch, and instead distributed the mode check to each relevant guard.  The "we're completely satisfied" case is now where you end up if none of the "we aren't happy" guards trip.  This avoids the need to have an "if we got this far, it must have been because we have undone versions" assumption for the `ClusterOperatorUpdating` block.

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -2,9 +2,9 @@ package internal
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 	"unicode"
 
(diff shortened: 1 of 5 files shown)


jottofar:
This looks okay to me and unit tests should catch any detail I missed. I know you need to change 2 test cases from using `ClusterOperatorNotAvailable` to `ClusterOperatorUpdating`. I will circle back with the `lgtm` after the tests pass.

wking:
Rebased around #514 and added a fix for the `cluster_operator_up` metric's `version` label with 92b76cb1 -> 5de0be43.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -141,85 +141,66 @@ func waitForOperatorStatusToBeDone(ctx context.Context, interval time.Duration,
...
-		// for incomplete operands.
-		undone := map[string][]string{}
+		if len(expected.Status.Versions) == 0 {
+			lastErr = &payload.UpdateError{
+				UpdateEffect: payload.UpdateEffectNone,
+				Reason:       "ClusterOperatorNotAvailable",
+				Message:      fmt.Sprintf("Cluster operator %s does not declare expected versions", expected.Name),
+				Name:         expected.Name,
+			}
+			return false, nil
```
jottofar, line 151: Why do we not still want to check the operator's conditions? If the operator does not declare expected versions does that also imply that it also does not set conditions?

wking: If the operator does not declare expected versions in its manifest, it is hugely bugged.  This PR is dropping the `Progressing` fallback to these cases, so I went with failing fast here to make sure we never let those bugs through CI.

wking: @jottofar pointed out that for fail-fast, `UpdateEffectNone` should be `UpdateEffectFail`.  I've fixed that and rebased onto master with 4abd57c5 -> 98794a62.

smarterclayton:
I am not comfortable with parts of this change - specifically allowing degraded during initialization and allowing us to exit install in a "ok" state with operators degraded (worse, we report available, then go failed after, when we should install, immediately report "Failed" and "Available" at the same time, and then cause installer to exit as "install failed").

I see no justification in the comment for such a change (the commit message is about code changes, not about the intent change).  If you can link to such discussion I'd appreciate it, but this doesn't pass my smell test which is that degraded during install may be suppressible, but that reporting available (for automation) and then immediately going degraded punts problems downstream that should be caught right up front.  An install completes if it is degraded, but it should not pretend that it is ok.

PR #531, opened 2021-03-16
Bug 1929917: pkg/cvo/sync_worker: Skip precreation of baremetal ClusterOperator

wking:
This is a hack fix for [rhbz#1929917][1], where we have a delay on 4.6->4.7 updates, and on some 4.7 installs, between the very early ClusterOperator precreation and the operator eventually coming up to set its status conditions.  In the interim, there are no conditions, which causes `cluster_operator_up` to be 0, which causes the `critical` `ClusterOperatorDown` to fire.  We'll want a more general fix going forward, this commit is a temporary hack to avoid firing the critical `ClusterOperatorDown` while we build consensus around the general fix.

The downside to dropping precreates for this operator is that we lose the must-gather references when the operator fails to come up.  That was what precreation was designed to address in 2a469e37c1 (#318).  If we actually get a must-gather without the bare-metal bits and we miss them, we can revisit the approach this hack is taking.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1929917

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -742,6 +742,10 @@ func (w *SyncWorker) apply(ctx context.Context, payloadUpdate *payload.Update, w
 					klog.V(4).Infof("Skipping precreation of %s as unmanaged", task)
 					continue
 				}
+				if task.Manifest.Obj.GetName() == "baremetal" {
+					klog.V(4).Infof("Skipping precreation of %s, https://bugzilla.redhat.com/show_bug.cgi?id=1929917", task)
+					continue
+				}
 				if err := w.builder.Apply(ctx, task.Manifest, payload.PrecreatingPayload); err != nil {
 					klog.V(2).Infof("Unable to precreate resource %s: %v", task, err)
 					continue


wking:
I've pushed 7731ee7 -> fdef37d to [fix][1]:

```
pkg/cvo/sync_worker.go:745:21: task.Manifest.Metadata undefined (type *manifest.Manifest has no field or method Metadata)
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/531/pull-ci-openshift-cluster-version-operator-master-images/1371885042271260672#1:build-log.txt%3A77

wking:
[Previous update job][1] had only unrelated failures.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/531/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1372528943906689024

wking:
[MCO got stuck][1], unrelated to the PR

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/531/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1372588195924938752

PR #532, opened 2021-03-17
WIP: pkg/cvo: Add logging to track ClusterVersion

jottofar:
This logging will hopefully help to debug CI testing issue `[sig-cluster-lifecycle] Cluster version operator acknowledges upgrade
`.


vrutkovs:
I don't think we need this anymore, right? Related bug was fixed and this can be closed?

PR #541, opened 2021-03-31
docs/user/status: Mention channel-clearing for VersionNotFound

wking:
Local analog for openshift/openshift-docs#22252.


wking:
Docs change.

PR #545, opened 2021-04-12
Remove k8s.io v1beta1 APIs no longer served

jottofar:
k8s.io v1beta1 APIs are no longer served In kuberentes v1.22. Therefore removing  apiextensions.k8s.io/v1beta1 version of
CustomResourceDefinition, apiregistration.k8s.io/v1beta1 version of APIService, and rbac.authorization.k8s.io/v1beta1 version of ClusterRole, ClusterRoleBinding, Role, and RoleBinding. Kubernetes recommended action is to use the v1 API versions which are already in use in CVO. It supported both versions.

Use of the v1beta1 versions is currently causing the DeprecatedAPIInUse alert to fire in 4.8 clusters.


jottofar:
/retitle Bug 1947797: Remove k8s.io v1beta1 APIs no longer served

sttts:
I don't think this is solution for 4.8 for the problem at hand. The complain in the BZ is about CVO having two CRD manifests in its payload that are v1beta1.

This PR is good, but I would not merge it before 4.9 because during upgrade/downgrade between 4.7 and 4.8 the CVO mifght "see" v1beta1 CRDs manifests that are completely valid.

/hold

jottofar:
/retitle Remove k8s.io v1beta1 APIs no longer served

PR #546, opened 2021-04-13
POC: Exclude featuregate.release.openshift/tech-preview=true manifests

sttts:


pkg/payload/payload.go
```diff
@@ -132,12 +132,20 @@ type metadata struct {
 	Metadata map[string]interface{}
 }
 
-func LoadUpdate(dir, releaseImage, excludeIdentifier, profile string) (*Update, error) {
+func LoadUpdate(dir, releaseImage, excludeIdentifier string, includeTechPreview func() (bool, error), profile string) (*Update, error) {
```
wking, line 135: I don't think this approach is going to work.  Today we only call `InitializeFromPayload` [when we launch the CVO][1], and the other `LoadUpdate` caller is [when we have detected the admin requesting an update][2].  So if someone sets the tech-preview feature gate after that load, the CVO will not notice or react to that change until the next load (after a pod reschedule or requested update).  And we'd have to track the feature-gate-status as part of ClusterVersion's `status.desired` and/or `status.history` so the new CVO pod would be able to distinguish between "I'm continuing the work of my predecessor and can launch in reconciliation mode" and "tech-preview has changed since the last reconciled payload, so I should launch in update mode".  That's... possible.  But it seems like a lot of juggling vs. just deploying the tech-preview operator all the time, and having that tech-preview operator decide if it should enable its operands or not.  For precedent, consider the registry operator, which the CVO always installs, but which can be configured to not deploy its registry operand.

[1]: https://github.com/openshift/cluster-version-operator/blob/31cc2f8f26aa5a4ac12c8e8a9917d2efa05b6540/pkg/start/start.go#L145
[2]: https://github.com/openshift/cluster-version-operator/blob/31cc2f8f26aa5a4ac12c8e8a9917d2efa05b6540/pkg/cvo/sync_worker.go#L560-L586

wking:
[cert-manager ended up being OLM-installed][1], and openshift/enhancements#922 is being considered with more generic knobs for opting in to or out of specific core-release-payload components.  So closing this narrower annotation as obsolete.

/close

[1]: https://github.com/openshift/cert-manager-operator#running-the-operator-in-the-cluster

PR #552, opened 2021-05-04
hack/generate-lib-resources: Use library-go manifests

wking:
Catching the generator up with ce2874f129 (#443).


vrutkovs (approved):
/lgtm

wking:
Twiddling the bug label, because this just catches the generator up with the existing code, and that's not user-facing.

PR #559, opened 2021-05-12
Bug 1881484: Set defaults in deployment

jottofar:
Mostly followed kube defaulting logic however with some exceptions:
-  Currently only CVO deployment uses "KUBERNETES_SERVICE_HOST" env variable to inject internal LB host.This may result in an IP address being returned by API so always assuming returned value is okay.
-  Toleration keys can be duplicated and are so in CVO (see [Bug 1941901](https://bugzilla.redhat.com/show_bug.cgi?id=1941901)) so needed to change logic to not simply exit when first matched key found possibly incorrectly overwrite the wrong key's value.
Also added VolumeSource defaulting in its entirety even though some are not currently used to avoid future hot looping issues.

### DIFF
--- a/lib/resourcemerge/apps_test.go
+++ b/lib/resourcemerge/apps_test.go
@@ -63,6 +63,8 @@ func TestEnsureDeployment(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			defaultDeployment(&test.existing, test.existing)
+			defaultDeployment(&test.expected, test.expected)
 			modified := pointer.BoolPtr(false)
 			EnsureDeployment(modified, &test.existing, test.required)
 			if *modified != test.expectedModified {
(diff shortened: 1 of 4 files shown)


jottofar:
/retitle Bug 1881484: Set defaults in deployment

vrutkovs:
Verified that its no longer happening via cluster-bot ` test e2e openshift/cluster-version-operator#561,openshift/cluster-version-operator#559` task:
```
$ curl -sL https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1400083116516708352/artifacts/launch/pods/openshift-cluster-version_cluster-version-operator-54b7b7b8fb-frm7n_cluster-version-operator.log | grep -oP "Updating \K.+ due to diff" | cut -d' ' -f1 | sort | uniq -c
     10 ClusterServiceVersion
     10 PriorityClass
```

vrutkovs (approved):
/lgtm

wking:
[API-server `TargetDown`][1] is unrelated to this PR.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/559/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1400185777928802304

PR #560, opened 2021-05-13
Bug 1927168: pkg/cvo/internal/operatorstatus: Replace wait-for with single-shot "is it alive now?"

wking:
Like cc9292a061 (#400), but for ClusterOperator.  In that commit message, I'd explain that we'd continue on past imperfect (e.g. Progressing=True) but still happy-enough Deployments and such and later block on the ClusterOperator.  But there's really no need to poll the ClusterOperator while we're blocked on it; we can instead fail that sync loop, take the short cool-off break, and come in again with a new pass at the sync cycle.

This will help reduce delays like [rhbz#1927168][1] where a good chunk of the ~5.5 minute delay was waiting for the network ClusterOperator to become happy.  If instead we give up on that sync cycle and start in again with a fresh sync cycle, we would have been more likely to recreate the ServiceMonitors CRD more quickly.

The downside is that we will now complain about unavailable, degraded, and unleveled operators more quickly, without giving them time to become happy before complaining in ClusterVersion's status.  This is mitigated by most of the returned errors being UpdateEffectNone, which we will render as "waiting on..." Progressing messages.  The exceptions are mostly unavailable, which is a serious enough condition that I'm fine complaining about it aggressively, and degraded, which has the fail-after-interval guard keeping us from complaining about it too aggressively.

I'm also shifting the "does not declare expected versions" check before the `Get` call.  It's goal is still to ensure that we fail in CI before shipping an operator with such a manifest, and there's no point in firing off an API GET before failing on a guard that only needs local information.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1927168#c2

### DIFF
--- a/pkg/cvo/internal/operatorstatus.go
+++ b/pkg/cvo/internal/operatorstatus.go
@@ -5,14 +5,12 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-	"time"
 	"unicode"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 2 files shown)


wking:
[`/run` failure][1] is [rhbz#1959564][2].

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/560/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1392891079665127424
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1959564

wking:
[This update job][1] passed everything except cluster teardown, which is orthogonal to this PR.  Other update presubmits also seem to have failed in unrelated-to-this-PR modes.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/560/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1392891079753207808

PR #561, opened 2021-05-13
Bug 1879184: Log object updates and show existing/required diff

vrutkovs:
Show when object gets created or updated with diff

### DIFF
--- a/lib/resourceapply/apiext.go
+++ b/lib/resourceapply/apiext.go
@@ -10,6 +10,7 @@ import (
 	apiextclientv1beta1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 )
@@ -35,7 +36,7 @@ func ApplyCustomResourceDefinitionv1beta1(ctx context.Context, client apiextclie
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating CRD %s", required.Name)
+	klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
 
 	actual, err := client.CustomResourceDefinitions().Update(ctx, existing, metav1.UpdateOptions{})
 	return actual, true, err
@@ -62,7 +63,7 @@ func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating CRD %s", required.Name)
+	klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
 
 	actual, err := client.CustomResourceDefinitions().Update(ctx, existing, metav1.UpdateOptions{})
 	return actual, true, err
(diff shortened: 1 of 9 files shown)


pkg/cvo/internal/generic.go
```diff
@@ -46,17 +49,37 @@ func applyUnstructured(ctx context.Context, client dynamic.ResourceInterface, re
...
-	existing.SetLabels(required.GetLabels())
-	existing.SetOwnerReferences(required.GetOwnerReferences())
 	skipKeys := sets.NewString("apiVersion", "kind", "metadata", "status")
+
+	// create a copy of required, but copy skipKeys from existing
+	// this would copy skipKeys data into expected from existing
+	expected := required.DeepCopy()
+	for k, v := range existing.Object {
+		if skipKeys.Has(k) {
+			expected.Object[k] = v
```
wking, line 59: This sort of thing is why I think generically logging diffs is brittle.  I'd rather have the `Ensure*` helpers return a slice of patches instead of the current `modified` boolean, so we could log the patches "Updating $PROPERTY to $VALUE to clobber $EXISTING_VALUE" instead of having logic about the things we care about clobbering both here and in the `Ensure*` helpers.

vrutkovs: Logging everytime `modified` is changed may be too verbose. I think we should do that later, on level 4 so that we wouldn't hit master's disks with too much log lines.

wking: if, instead of twiddling `*modified`, these helpers return `[]patch, error`, we can log any `patch` we get at one time, before calling the `UPDATE`.  Should be same number of log lines, and we'd only complain about the in-cluster properties we were trying to stomp.

wking:
Failed to install on [`Total Regional Cores quota`][1]:

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/561/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1402114424143089664#1:build-log.txt%3A66

LalatenduMohanty (changes requested):

LalatenduMohanty:
I am wondering if the current PR will have any performance impact or not. If it is not going to have any performance impact then I am fine to merge it as it is.

lib/resourceapply/apiext.go
```diff
@@ -35,7 +36,7 @@ func ApplyCustomResourceDefinitionv1beta1(ctx context.Context, client apiextclie
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating CRD %s", required.Name)
+	klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
```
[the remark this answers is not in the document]
vrutkovs: I'd rather use the default one, so that we'd see hotloops in CI upgrade jobs. If we move it to v4 we'd have to reconfigure CVO to get them recorded

vrutkovs:
>I am wondering if the current PR will have any performance impact or not. If it is not going to have any performance impact then I am fine to merge it as it is.

It improves performance by fixing hotloops in unstructured resources. Additional log lines might be printed during sync periods, but its effect is negligible

pkg/cvo/internal/generic.go
jottofar: This is only required if we want to see not only that we're hot looping but also why. The latter could be dropped and once we know we're hot looping figuring out the why becomes part of the fix process done off line. Choosing one over the other is a question of how often we believe this is going to occur. If rarely then we probably don't need the "why" in the baseline.

vrutkovs: CVO can start hotlooping on any non-standard configuration - and its complicated to find out details to fix it. 
I'd prefer to include "why" in every run, so that we wouldn't have to request customers to reconfigure CVO to get details.

LalatenduMohanty (approved):
/lgtm

pkg/cvo/internal/generic.go
sdodson: @vrutkovs What sort of non-standard configs would they have which are managed by the CVO that we'd want to look into?

vrutkovs: Users can [sideload additional manifests](https://github.com/openshift/installer/blob/master/docs/user/customization.md#control-plane-with-no-taints) during bootstrap

PR #562, opened 2021-05-13
Bug 1881520: avoid hotlooping on RoleBindings with empty APIGroup

vrutkovs:
`EnsureRoleBinding` and `EnsureClusterRoleBinding` expect `RoleRefs` to be identical. Some manifests are skipping `APIGroup`, so CVO keeps attempting to apply those on every sync.

This PR adds `CompareRoleRefsv1`/`CompareRoleRefsv1beta1` functions, which return true ("identical") if required APIGroup is empty and existing APIGroup is "rbac.authorization.k8s.io"

TODO:
* [x] More idiomatic would be `ensureRoleRefAPIGroup`
* [x] Add test case

### DIFF
--- a/lib/resourcemerge/rbac.go
+++ b/lib/resourcemerge/rbac.go
@@ -9,6 +9,7 @@ import (
 // modified is set to true when existing had to be updated with required.
 func EnsureClusterRoleBinding(modified *bool, existing *rbacv1.ClusterRoleBinding, required rbacv1.ClusterRoleBinding) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	ensureRoleRefDefaultsv1(&required.RoleRef)
 	if !equality.Semantic.DeepEqual(existing.Subjects, required.Subjects) {
 		*modified = true
 		existing.Subjects = required.Subjects
@@ -37,6 +38,7 @@ func EnsureClusterRole(modified *bool, existing *rbacv1.ClusterRole, required rb
 // modified is set to true when existing had to be updated with required.
 func EnsureRoleBinding(modified *bool, existing *rbacv1.RoleBinding, required rbacv1.RoleBinding) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	ensureRoleRefDefaultsv1(&required.RoleRef)
 	if !equality.Semantic.DeepEqual(existing.Subjects, required.Subjects) {
 		*modified = true
 		existing.Subjects = required.Subjects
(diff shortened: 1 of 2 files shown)


wking (approved):
/lgtm

vrutkovs:
/hold

This needs unit tests

vrutkovs:
/test unit

Flake?

vrutkovs:
/hold cancel

Tests added, ready for review

lib/resourcemerge/rbacv1beta1.go
```diff
@@ -1,6 +1,7 @@
 package resourcemerge
```
wking, line 1: We might actually be able to drop this entirely, with:

```console
$ oc adm release extract --to manifests quay.io/openshift-release-dev/ocp-release:4.8.0-fc.3-x86_64
$ grep -r10 v1beta1 manifests/ | grep ClusterRole
manifests/0000_80_machine-config-operator_03_rbac.yaml-kind: ClusterRoleBinding
manifests/0000_30_baremetal-operator_01_baremetalhost.crd.yaml-kind: ClusterRole
```

The MCO has since [bumped their version][1] and the baremetal manifest [doesn't have annotations to be included in any cluster profiles][2].  So I expect we can drop this entirely, instead of fixing v1beta1 hotlooping.

[1]: https://github.com/openshift/machine-config-operator/pull/2572
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1956607

lib/resourcemerge/rbacv1beta1.go
```diff
@@ -1,6 +1,7 @@
...
 
 import (
+	rbacv1 "k8s.io/api/rbac/v1"
```
wking, line 4: No need to mix in `v1`, `v1beta1` has this too:

```console
$ git --no-pager grep 'GroupName =' vendor/k8s.io/api/rbac/v1beta1
vendor/k8s.io/api/rbac/v1beta1/register.go:const GroupName = "rbac.authorization.k8s.io"
```

lib/resourcemerge/rbacv1beta1.go
wking: Part of dropping v1beta1 RBAC will be:

```console
$ git --no-pager diff -U0 hack
diff --git a/hack/generate-lib-resources.py b/hack/generate-lib-resources.py
index 441e253d..294e9885 100755
--- a/hack/generate-lib-resources.py
+++ b/hack/generate-lib-resources.py
@@ -271 +270,0 @@ if __name__ == '__main__':
-    types['k8s.io/api/rbac/v1beta1'] = types['k8s.io/api/rbac/v1']
@@ -282 +280,0 @@ if __name__ == '__main__':
-        'k8s.io/api/rbac/v1beta1': {'package': 'k8s.io/client-go/kubernetes/typed/rbac/v1beta1', 'type': 'RbacV1beta1Client'},
```

although you'll also want #552, and you'll need to drop some remaining v1beta1 RBAC code like this resourcemerge stuff by hand.

vrutkovs: Good idea, dropped v1beta1

wking (approved):
/lgtm
/retest

wking:
[`pods should successfully create sandboxes by other`][1] is not a CVO-RoleBinding-management failure.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/562/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1393263167647256576

PR #563, opened 2021-05-13
Bug 1881481: TargetPort should default to port in ServicePort if unset

vrutkovs:
`TargetPort` is not being set in some required manifests, but returned by API server

### DIFF
--- a/lib/resourcemerge/core.go
+++ b/lib/resourcemerge/core.go
@@ -5,6 +5,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // EnsureConfigMap ensures that the existing matches the required.
@@ -276,6 +277,9 @@ func ensureServicePort(modified *bool, existing *corev1.ServicePort, required co
 func ensureServicePortDefaults(servicePort *corev1.ServicePort) {
 	if servicePort.Protocol == "" {
 		servicePort.Protocol = corev1.ProtocolTCP
+	}
+	if servicePort.TargetPort == intstr.FromInt(0) || servicePort.TargetPort == intstr.FromString("") {
+		servicePort.TargetPort = intstr.FromInt(int(servicePort.Port))
 	}
 }
 
--- a/lib/resourcemerge/core_test.go
+++ b/lib/resourcemerge/core_test.go
@@ -504,8 +504,9 @@ func TestEnsureServicePorts(t *testing.T) {
 				Spec: corev1.ServiceSpec{
 					Ports: []corev1.ServicePort{
 						{
-							Protocol: corev1.ProtocolUDP,
-							Port:     8282,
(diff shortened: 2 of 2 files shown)


lib/resourcemerge/core.go
```diff
@@ -254,16 +255,19 @@ func EnsureServicePorts(modified *bool, existing *[]corev1.ServicePort, required
...
 
-func ensureServicePortDefaults(servicePort *corev1.ServicePort) {
-	if servicePort.Protocol == "" {
-		servicePort.Protocol = corev1.ProtocolTCP
+func ensureServicePortDefaults(required, existing *corev1.ServicePort) {
+	if required.Protocol == "" {
+		required.Protocol = corev1.ProtocolTCP
+	}
+	if required.TargetPort == intstr.FromInt(0) {
+		required.TargetPort = intstr.FromInt(int(required.Port))
```
wking, line 270: Can we use:

```go
	if servicePort.TargetPort == intstr.FromInt(0) || servicePort.TargetPort == intstr.FromString("") {
		servicePort.TargetPort = intstr.FromInt(int(servicePort.Port))
```

to match [the upstream default logic][1]?  Actually vendoring the upstream default logic would be another option, but feels a bit heavy to me, if we plan on moving to server-side-apply soonish.  But copy/pasting the relevant upstream default code seems safer to me than writing something that's similar but not quite identical locally.

[1]: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/core/v1/defaults.go#L122-L124

vrutkovs: Good idea, added this in c37b1c74.

lib/resourcemerge/core.go
```diff
@@ -266,16 +267,19 @@ func EnsureServiceType(modified *bool, existing *corev1.ServiceType, required co
...
 
-func ensureServicePortDefaults(servicePort *corev1.ServicePort) {
-	if servicePort.Protocol == "" {
-		servicePort.Protocol = corev1.ProtocolTCP
+func ensureServicePortDefaults(required, existing *corev1.ServicePort) {
+	if required.Protocol == "" {
+		required.Protocol = corev1.ProtocolTCP
+	}
+	if required.TargetPort == intstr.FromInt(0) || required.TargetPort == intstr.FromString("") {
+		existing.TargetPort = intstr.FromInt(int(required.Port))
```
wking, line 282: This defaulting is just within the required manifest; it doesn't depend on what's in-cluster at the moment.  So you can stick to a single `servicePort` argument in this function, and not pass in both `required` and `existing`.

wking (approved):
/lgtm
/retest

wking:
[`etcdHighNumberOfLeaderChanges`][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/563/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1394288271097335808

PR #565, opened 2021-05-14
Bug 1960554: Remove rbacv1beta1 support

vrutkovs:
All manifests have moved to rbacv1


vrutkovs:
/hold

until https://github.com/openshift/baremetal-operator/pull/147 merges

wking (approved):
The manifests being dropped in openshift/baremetal-operator#147 currently lack profile annotations, so the CVO never cared about them, and we don't need to wait.

/lgtm
/hold cancel

PR #566, opened 2021-05-14
Bug 1960559: drop APIExtensions v1beta1

vrutkovs:
All manifests have moved to v1.

This also bumps openshift/api (and updates k8s deps to 0.21.1 to be uniform) so that all apiextensions/v1beta1 would no longer be used.


vrutkovs:
/hold

I'd prefer [RBACv1beta1 PR](https://github.com/openshift/cluster-version-operator/pull/565) to land first

vrutkovs:
/test unit

Flake

vrutkovs:
/retest

Failed to setup initial cluster in upgrade test

hack/test-prerequisites.go
```diff
@@ -36,12 +36,12 @@ func main() {
 			if err != nil {
 				log.Fatalf("Unable to read %s: %v", path, err)
 			}
-			var crd v1beta1.CustomResourceDefinition
+			var crd v1.CustomResourceDefinition
 			if err := yaml.Unmarshal(data, &crd); err != nil {
```
sttts, line 40: You don't check that the API version matches? That's not going to make you happy when it doesn't.

sttts: ok, test only.

vrutkovs: I don't think we do that in tests. In actual flow non-v1 CRDs would be treated as "unstructered".

@wking @LalatenduMohanty any advice here? I suppose we should be tracking how many manifests are being applied by unstructured handler and attempt to minimize that number. Probably we should be exporting a new metric?

wking: My long-term goal is to completely drop the unstructured handler, and fail folks in CI if they add a new manifest type that we don't support.  Logging in the unstructured handler that we can look for in gathered CI assets would help get us there.

LalatenduMohanty: > if they add a new manifest type that we don't support. Logging in the unstructured handler that we can look for in gathered CI assets would help get us there.

+1

wking (approved):
/lgtm

[` pods should successfully create sandboxes by other` and `etcdHighNumberOfLeaderChanges`][1] are unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/566/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1404700765347385344

wking:
[`The HAProxy router should set Forwarded headers appropriately`][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/566/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1404829352356483072

wking:
This PR is unrelated to the other recent changes, so save some time and money with a manual merge to cut off the retesting.

PR #568, opened 2021-05-18
pkg/cincinnati: Log the proxy/TLS settings used for upstream fetches

wking:
`AppendCertsFromPEM` is [pretty generous][1] about parsing failures:

> It appends any certificates found to s and reports whether any certificates were successfully parsed.

which can mask errors when we succesfully parse at least one cert, but fail to parse some subsequent cert.  Make that a bit easier to detect by logging the number of certs we are using by the time we make the upstream update service request.  It would be more useful to log "does the root CA pool have a CI for the target (proxy) URI?", but that requires more CertPool expertise than I have.  The subject count should be sufficient to determine whether the expected PEM contents are being completely loaded or not, if we have access to the relevant ConfigMap data as well.

We also use the same proxy/TLS approach to configure signature fetches, but the risk of breaking down in the pipeline leading to those signature requests seems small, and the logging I'm adding here will cover the cluster-config -> CVO transmission, so I haven't added logging to that side.  This will leave us uncovered for folks who do not configure a channel, since they may be relying on the signature fetching, but won't be using the upstream polling.  If we end up worried about that, we can add logging for them in follow-up work.

I'm also logging the proxy URI, if configured, because it's a similar config pathway that folks might worry is not working.

[1]: https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM


wking:
/hold

for 4.9

sdodson:
@wking would that strip proxy credentials if the proxy url is of the form `https://user:password@host:3128` ?

wking:
> ...would that strip proxy credentials...

Hrm, no.  And folks could also have a magic token in a query param.  So maybe I should hash the URI?  Or get out of proxy-URI logging altogether?

wking:
Ah, I can parse and just log the host part, since that shouldn't be too sensitive, and folks are probably not running to many proxies on the same host

wking:
Rebased onto master with 38ead0ce -> 4671480b, which also pivots to only logging the proxy host.

vrutkovs:
Fails to build:
```
 Building github.com/openshift/cluster-version-operator (v1.0.0-557-ga6458adc-dirty)
# github.com/openshift/cluster-version-operator/pkg/cincinnati
pkg/cincinnati/cincinnati.go:85:87: undefined: length
pkg/cincinnati/cincinnati.go:85:94: undefined: tlsConfig 
```

wking:
> Fails to build...

Oops.  Fixed with 116b2c5e.

vrutkovs (approved):
/lgtm

vrutkovs:
/override e2e-agnostic-upgrade

sandboxes issue
/hold cancel

PR #570, opened 2021-05-19
Making calculateNext more readable

LalatenduMohanty:


pkg/cvo/sync_worker.go
```diff
@@ -413,18 +413,16 @@ func (w *SyncWorker) calculateNext(work *SyncWork) bool {
...
 		work.State = w.work.State
 		work.Attempt = 0
-	} else if changed {
-		work.State = payload.UpdatingPayload
-		work.Attempt = 0
-	}
-
-	if w.work != nil {
+	} else {
 		work.Desired = w.work.Desired
```
wking, line 426: This could panic if `w.work` is `nil`, because `!work.Empy()` says nothing about the state of `w.work`.

PR #571, opened 2021-05-20
Bug 1962882: pkg/cvo/updatepayload: Set priorityClassName for the version Job

wking:
Avoid [failing][1]:

    [sig-arch] Managed cluster should ensure platform components have system-* priority class associated [Suite:openshift/conformance/parallel]

by setting the property.  From [our conventions][2], `openshift-user-critical` means it is fine for this workload to be preempted by user-workload and OOMKilled.  If we have such dire resource constraints that the version pod gets evicted, it's probably not a good time to be kicking off an update with all of its associated control-plane churn.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-e2e-aws-upgrade-rollback/1395138552144072704
[2]: https://github.com/openshift/enhancements/blob/fe578535d5e6a126a10bb26830b7755266d0f92e/CONVENTIONS.md#priority-classes

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -207,6 +207,7 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 					NodeSelector: map[string]string{
 						nodeSelectorKey: "",
 					},
+					PriorityClassName: "openshift-user-critical",
 					Tolerations: []corev1.Toleration{{
 						Key: nodeSelectorKey,
 					}},


pkg/cvo/updatepayload.go
```diff
@@ -207,6 +207,7 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 					NodeSelector: map[string]string{
 						nodeSelectorKey: "",
 					},
+					PriorityClassName: "openshift-user-critical",
```
LalatenduMohanty, line 210: What is the difference between `system-cluster-critical` vs "openshift-user-critical" ?

LalatenduMohanty: Never mind, I found https://github.com/openshift/enhancements/blob/master/enhancements/scheduling/priority-class.md

LalatenduMohanty (changes requested):

pkg/cvo/updatepayload.go
LalatenduMohanty: After some discussion I think we will be ok with `PriorityClassName: "openshift-user-critical",` as this is an improvement over no priority class. Also we have https://github.com/openshift/cluster-monitoring-operator/blob/6f22c3b33e2af7a61a3d385f0688d270519b49e4/assets/control-plane/prometheus-rule.yaml#L185-L194 and https://github.com/openshift/cluster-monitoring-operator/blob/6f22c3b33e2af7a61a3d385f0688d270519b49e4/assets/control-plane/prometheus-rule.yaml#L25-L40 which will communicate user if the pod is not able to schedule itself.

LalatenduMohanty (approved):
/lgtm

wking:
[API-server availability][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/571/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1395430904557473792

PR #573, opened 2021-05-24
Bug 1927168: pkg/payload/task: Fail fast for UpdateError

wking:
Things like API-server connection errors and patch conflicts deserve some retries before we bubble them up into ClusterVersion conditions.  But when we are able to retrieve in-cluster objects and determine that they are not happy, we should exit more quickly so we can complain about the resource state and start in on the next sync cycle.  For example, see the recent e02d1489a5 (#560) and the older cc9292a061 (#400).  This commit uses the presence of an UpdateError as a marker for "fail fast; no need to retry".

The install-time backoff is from fee2d06871 (#136).  I'm not sure if it really wants the same cap as reconcile and update modes, but I've left them the same for now, and future commits to pivot the backoff settings can focus on motivating those pivots.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -262,6 +262,7 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 			Duration: time.Second * 10,
 			Factor:   1.3,
 			Steps:    3,
+			Cap:      time.Second * 15,
 		},
 		optr.exclude,
 		optr.eventRecorder,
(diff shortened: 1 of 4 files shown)


wking:
/hold

while I work out what is going wrong with the [unit][1] and [e2e-operator][2] jobs.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/573/pull-ci-openshift-cluster-version-operator-master-unit/1397771303145967616
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/573/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator/1397751412221284352

vrutkovs (changes requested):
/lgtm cancel

Not sure why unit/e2e tests are failing. Changing `wait.Backoff.Steps` to 2 in `setupCVOTest` fixes one unit tests, but I don't think its a good idea

wking:
Ended up landing via #577.

PR #585, opened 2021-06-08
Bug 1946479: prevent pod deployment deadlock due to custom SA projected volume injection

stlaz:
In an upgrade scenario:
1. a kube apiserver updates, any pod request that passes its admission
   is injected with SA projected volume
2. OpenShift adds a configMap "openshift-service-ca.crt" among the
   volume sources to the projected volume
3. Any pod that gets created now will be stuck until an updated
   version of KCM gets started
4. If CVO pod gets deleted at this point, the whole upgrade might
   get stuck

Prevent the deadlock by manually configuring the projected volume.
This can be removed in 4.9 when even the older KCM contains the code
creating the openshift-service-ca.crt CMs.

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -20,6 +20,10 @@ spec:
       labels:
         k8s-app: cluster-version-operator
     spec:
+      # prevents circular dependency with OpenShift service-ca CM volume injection by
+      # KAS and creation of the CM by KCM
+      # TODO: remove in 4.9
+      automountServiceAccountToken: false
       containers:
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

sdodson:
@stlaz Does this projection rely on features of either a 4.7 or 4.8 kubelet? Just asking in case this is something that would inhibit a 4.6 kubelet from working properly with a 4.8 cluster.

stlaz:
This relies solely on features from a 4.8 kube-controller-manager and kube-apiserver, kubelet is not involved.

s-urbaniak:
xref https://github.com/openshift/kubernetes/pull/714

PR #586, opened 2021-06-08
[WIP] Bug 1822752: Verify the desired update payload before updating sync worker

arjunrn:
Move the payload verification logic to to the worker update function. This ensures that before the current workload is removed the new payload has been verified.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -539,7 +539,7 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
 	}
 
 	// inform the config sync loop about our desired state
-	status := optr.configSync.Update(config.Generation, desired, config.Spec.Overrides, state)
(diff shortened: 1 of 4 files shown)


arjunrn:
@vrutkovs I'm trying to come up with a fix for this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1822752). When there is a unverified payload the current work being processed by the SyncWorker gets removed. More details in the bug.

pkg/cvo/cvo.go
```diff
@@ -32,6 +32,11 @@ import (
 	clientset "github.com/openshift/client-go/config/clientset/versioned"
 	configinformersv1 "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
```
jottofar, line 34: I guess your formatter is whacking the package ordering

PR #590, opened 2021-06-10
Correctly initialize service host and port when environment is missing

arjunrn:
The init containers where not be correctly initialized with the service host and port. Also fixed tests.


vrutkovs:
/lgtm
/hold

looks pretty straightforward (great find), leaving this for Jack to verify this just to be sure

/cc @jottofar

jottofar:
Haha, cut-and-paste strikes again. And then I guess the unit tests were made (or changed) such that they did not catch the issue.

/lgtm

PR #594, opened 2021-06-11
[WIP] Added golangci-lint for linting and fixed detected linting issues

arjunrn:
Added [golangci-lint](https://golangci-lint.run/) as a make target which can be run as a check.
Also fixed issues which were reported by golangci-lint.

Discovered an issue with `cvo_scenarios_test.go` where the `ClusterVersion` resource was never actually updated in the fake client but always recreated. Fixed this which led to some knock on changes in the tests.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,80 @@
+# This file contains all available configuration options
+# with their default values.
+
+# options for analysis running
+run:
+  # default concurrency is a available CPU number
+  concurrency: 4
+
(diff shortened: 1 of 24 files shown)


LalatenduMohanty (changes requested):
The code changes which fixes specific linting issues should be in a separate commit. For example all `deadcode` fixes should a single commit. All  `gofmt` fixes should in one commit. This will help the review easier . Also will help if someone wants to track the history behind some change.

arjunrn:
Closing in favor of #598

PR #597, opened 2021-06-14
Fixed fake client for CVO scenarios test

arjunrn:
The fake client for the CVO scenarios test uses a switch-case statement to compare the received `Action` pointer type to interfaces for update-action, create-action and get-action. However in go when two interfaces have the same definition(same functions) a concrete type can match against both of them. And because `clientgotesting.CreateAction` and `clientgotesting.UpdateAction` are basically the same interface the action always matched against the first interface. In this case the `CreateAction`. So the `ClusterVersion` object was always recreated and the `ResourceVersion` and `Generation` fields in the result object were reset all the time.

I have fixed the above issue by using the actual structs `CreateActionImpl`, `UpdateActionImpl` and `GetActionImpl` in the switch-case. This means that the test now updates the `ResourceVersion` correctly. The `Generation` field was also not being incremented correctly. The test client now takes the highest value from the received payload and the existing object.
  
I've also added a `verifySingleUpdate` function to ensure that the `ClusterVersion` is updated only once, even if multiple `get` calls are made to it.

### DIFF
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
@@ -38,22 +37,19 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 	})
 	cvs := make(map[string]runtime.Object)
 	client.AddReactor("*", "clusterversions", func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
-		switch strings.ToLower(action.GetVerb()) {
-		case "get":
-			a := action.(clientgotesting.GetAction)
+		switch a := action.(type) {
+		case clientgotesting.GetActionImpl:
 			obj, ok := cvs[a.GetName()]
 			if !ok {
 				return true, nil, errors.NewNotFound(schema.GroupResource{Resource: "clusterversions"}, a.GetName())
 			}
 			return true, obj.DeepCopyObject(), nil
-		case "create":
-			a := action.(clientgotesting.CreateAction)
+		case clientgotesting.CreateActionImpl:
 			obj := a.GetObject().DeepCopyObject().(*configv1.ClusterVersion)
 			obj.Generation = 1
 			cvs[obj.Name] = obj
 			return true, obj, nil
-		case "update":
-			a := action.(clientgotesting.UpdateAction)
+		case clientgotesting.UpdateActionImpl:
 			obj := a.GetObject().DeepCopyObject().(*configv1.ClusterVersion)
 			existing := cvs[obj.Name].DeepCopyObject().(*configv1.ClusterVersion)
 			rv, _ := strconv.Atoi(existing.ResourceVersion)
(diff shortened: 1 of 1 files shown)


pkg/cvo/cvo_scenarios_test.go
```diff
@@ -415,7 +425,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 	// make the image report unverified
 	payloadErr := &payload.UpdateError{
 		Reason:  "ImageVerificationFailed",
-		Message: fmt.Sprintf("The update cannot be verified: some random error"),
```
arjunrn, line 418: There is no formatting to be done so removed the `fmt.Sprintf` call.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -37,19 +38,22 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 	})
 	cvs := make(map[string]runtime.Object)
 	client.AddReactor("*", "clusterversions", func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
-		switch a := action.(type) {
-		case clientgotesting.GetAction:
+		switch strings.ToLower(action.GetVerb()) {
+		case "get":
```
wking, line 42: Good catch.  Checking for precedent upstream, I see them [switching on `GetActionImpl`, etc.][1]  Thoughts on that vs. just using the verb?

[1]: https://github.com/kubernetes/client-go/blob/1bccfc8c60977e2ce3235394daff128996818f7b/testing/fixture.go#L83-L97

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -2864,19 +2906,6 @@ func verifyAllStatus(t *testing.T, ch <-chan SyncWorkerStatus, items ...SyncWork
 	}
 }
 
-func waitFor(t *testing.T, fn func() bool) {
```
wking, line 2867: +1 to dropping this; looks like it was added in 961873d66a (#82) with no consumers then or since.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -2835,6 +2861,22 @@ func TestCVO_VerifyUpdatingPayloadState(t *testing.T) {
 	}
 }
 
+// verifyCVSingleUpdate ensures that the only object to be updated is a ClusterVersion type and it is updated only once
+func verifyCVSingleUpdate(t *testing.T, actions []clientgotesting.Action) {
+	var count int
+	for _, a := range actions {
+		if a.GetResource().Resource != "clusterversions" {
+			t.Fatalf("an resource of type %s was accessed/updated", a.GetResource().Resource)
```
wking, line 2869: nit: "an resource" -> "resource" .

I was wondering if we could also log the name, but [the generic `Action`][1] doesn't have a `GetName()`.  Although I also see the `GetVerb() != "get"` below.  Maybe we should be explicit about the verbs that count as updates?  That would be `create`, `update`, and `delete`?  `patch`?  I dunno.

[1]: https://pkg.go.dev/k8s.io/client-go/testing#Action
And maybe we can log the name as well, to make this a bit easier to debug if it happens?

arjunrn: Anything which is not a `get` would be a mutating change. So I think this is good enough. I've added the entire object into the failure log to help with debugging.

pkg/cvo/cvo_scenarios_test.go
arjunrn: Replaced with the `Impl`s

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -59,7 +59,11 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 			} else {
 				existing.Spec = obj.Spec
 				existing.ObjectMeta = obj.ObjectMeta
-				obj.Generation++
```
arjunrn, line 62: `obj` is the object received in the request payload. The generation is incremented here but this change is not stored back into the `existing` object. So removed the line and changed logic so the highest value of generation is used and then incremented.

wking (approved):
/lgtm

PR #598, opened 2021-06-14
Added golangci-lint with most linters disabled

arjunrn:
Added a [golangci-lint](https://golangci-lint.run/) to code linting with most of the linters/analyzers disabled. Once this is merged I will enable this as a step in openshift/release and subsequently enable each of the linters individually to make the code reviews easier.


vrutkovs (approved):
/lgtm

LalatenduMohanty (approved):
/lgtm

Makefile
```diff
@@ -21,3 +25,12 @@ verify-codegen-crds:
...
 .PHONY: update-codegen-crds update-codegen verify-codegen-crds verify-codegen verify
+
+
+.PHONY: lint
+## Checks the code with golangci-lint
+lint: $(GOLANGCI_LINT_BIN)
+	$(GOLANGCI_LINT_BIN) run -c .golangci.yaml --deadline=30m
+
+$(GOLANGCI_LINT_BIN):
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${BIN_DIR} ${GOLANGCI_LINT_VERSION}
```
wking, line 36: Installing from master seems dicy, since we could break as master becomes more strict.  Especially once we fork off of this branch into long-lived, hopefully quiet release branches, it would be nice to have this pinned to a specific version which we explicitly bump periodically.

Also, downloading a script and running it directly is fairly risky.  Can we use [hashpipe][1] or something similar to ensure we are only running the script we expect to get?

[1]: https://github.com/jbenet/hashpipe

arjunrn: I've moved the installer script into the `hack` directory. Note that the installer script takes a version parameter for binary it downloads. But like you pointed out curl-piping a bad idea. But having the installation script in repo should be good enough. WDYT?

wking: that's still pulling both the checksum and the binary from the same location.  But either we can replace the script with something similar to pull our pinned version and compare with a local hash, or we can include a cache of the checksum file in our repo and twiddle the script to skip the checksum-fetching step so it uses our local copy.

arjunrn: @wking I've changed the script so that the checksums are part of the script and the version is fixed. It only supports `linux` and `darwin` for now but should be enough for the time being.

LalatenduMohanty: This looks good to me.

wking: Looks good to me to.

/hold cancel

PR #600, opened 2021-06-14
Enable errcheck linter

arjunrn:
Depends on #598 

The [errcheck](https://github.com/kisielk/errcheck) linter ensures that all functions which return errors have their return value checked


vrutkovs (approved):
/lgtm

PR #601, opened 2021-06-14
Enable gofmt

arjunrn:
Depends on #598

[gofmt](https://golang.org/cmd/gofmt/) was already being used so this linter has not detected many issue. The only one detected is for the redundant type declaration.


vrutkovs (approved):
/lgtm

PR #604, opened 2021-06-14
Enable staticcheck

arjunrn:
Depends on #598 and #602

[staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck) is a code analysis tool which detects common mistakes in go.

### DIFF
--- a/cmd/image.go
+++ b/cmd/image.go
@@ -31,5 +31,5 @@ func runImageCmd(cmd *cobra.Command, args []string) {
 	if err != nil {
 		klog.Fatalf("error: %v", err)
 	}
-	fmt.Printf(image)
+	fmt.Print(image)
 }
--- a/pkg/cvo/egress.go
+++ b/pkg/cvo/egress.go
@@ -21,14 +21,12 @@ func (optr *Operator) getHTTPSProxyURL() (*url.URL, error) {
 		return nil, err
 	}
 
-	if &proxy.Spec != nil {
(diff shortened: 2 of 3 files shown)


cmd/image.go
```diff
@@ -31,5 +31,5 @@ func runImageCmd(cmd *cobra.Command, args []string) {
 	if err != nil {
 		klog.Fatalf("error: %v", err)
 	}
-	fmt.Printf(image)
+	fmt.Print(image)
```
arjunrn, line 34: There is no formatting of the strings required here.

pkg/cvo/egress.go
```diff
@@ -21,14 +21,12 @@ func (optr *Operator) getHTTPSProxyURL() (*url.URL, error) {
 		return nil, err
 	}
 
-	if &proxy.Spec != nil {
```
arjunrn, line 24: `proxy` is referenced and compared to nil. Since proxy is a value type this check will never be false.

vrutkovs: Looks good, but I'm not entirely sure if these are equivalent.

/cc @jottofar @wking

arjunrn: I didn't change the check. I removed it and what you see now is the inner-check. The outer-check was superfluous for the reason mentioned above.

LalatenduMohanty (approved):
/lgtm

PR #611, opened 2021-06-17
pkg/cvo/upgradeable: Soften "cannot be upgraded" to "should not be upgraded"

wking:
Because folks can force to override the CVO's precondition guards, including the Upgradeable=False guard.  The "cannot" wording is from way back in 04528144fe (#243).

Generated with:

```console
$ sed -i 's/cannot be upgraded/should not be upgraded/g' $(git grep -l 'cannot be upgraded')
```


vrutkovs (approved):
/lgtm

vrutkovs:

Text-only change

PR #613, opened 2021-06-18
Added waitgroups for autoupdate workers to complete before stopping

arjunrn:
The `wait.UntilWithContext` will repeatedly run the function passed to it
till the context also passed to it is done. At which point it will
return. Adding a wait group after the UntilWithContext ensures that
the top level goroutine does not return till all the spawned goroutines
terminate.

### DIFF
--- a/pkg/autoupdate/autoupdate.go
+++ b/pkg/autoupdate/autoupdate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/blang/semver/v4"
@@ -92,13 +93,17 @@ func (ctrl *Controller) Run(ctx context.Context, workers int) error {
 	if !cache.WaitForCacheSync(ctx.Done(), ctrl.cacheSynced...) {
 		return fmt.Errorf("caches never synchronized: %w", ctx.Err())
 	}
-
+	var wg sync.WaitGroup
+	wg.Add(workers)
 	for i := 0; i < workers; i++ {
-		// FIXME: actually wait until these complete if the Context is canceled.  And possibly add utilruntime.HandleCrash.
-		go wait.UntilWithContext(ctx, ctrl.worker, time.Second)
+		go func() {
+			defer wg.Done()
+			defer utilruntime.HandleCrash()
+			wait.UntilWithContext(ctx, ctrl.worker, time.Second)
+		}()
 	}
 
-	<-ctx.Done()
+	wg.Wait()
 	return nil
 }
 


pkg/autoupdate/autoupdate.go
```diff
@@ -93,13 +94,17 @@ func (ctrl *Controller) Run(ctx context.Context, workers int) error {
 	if !cache.WaitForCacheSync(ctx.Done(), ctrl.cacheSynced...) {
 		return fmt.Errorf("caches never synchronized: %w", ctx.Err())
 	}
-
+	var wg sync.WaitGroup
+	wg.Add(workers)
 	for i := 0; i < workers; i++ {
-		// FIXME: actually wait until these complete if the Context is canceled.  And possibly add utilruntime.HandleCrash.
```
arjunrn, line 98: A `runtime.HandleCrash()` is not required here because it's already part of the [JitterUntil](https://github.com/kubernetes/apimachinery/blob/e1b4d3faae10bda8d956188103745b1fecd462fe/pkg/util/wait/wait.go#L154) function.

wking: My impression was that `HandleCrash()` would ideally be called whenever we launch a new goroutine.  Walking the stack, I don't see any goroutines being launched before our `UntilWithContext` call and the underlying `JitterUntil`, but I dunno if we want to rely on that current lack of intermediate goroutines as part of the library's committed API.  Does it hurt to have an extra call here?

arjunrn: Adding a `HandleCrash()` here would recover from any panics in the apimachinery code, because the user function is already covered by the inner `HandleCrash()`. This is an issue because the `wait.UntilWithContext` would return and the autoupdate worker would terminate. If this happens in all workers then basically autoupdate would stop working. It would instead be better if program panicked so that the pod would restart. We can expect anything in the apimachinery code to be well tested and shouldn't be guarding against it.

wking: From [the docs][1]:

> HandleCrash actually crashes, after calling the handlers and logging the panic message.

So I expect we will still exit and get restarted even if we have a `HandleCrash` in here.

[1]: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/runtime#HandleCrash

arjunrn: So then the only thing `HandleCrash()` does(without any additional handlers) is to `klog` the stack trace. 🤷🏼 
Will add the `HandleCrash()` here as well.

pkg/autoupdate/autoupdate.go
```diff
@@ -92,13 +93,18 @@ func (ctrl *Controller) Run(ctx context.Context, workers int) error {
...
-		go wait.UntilWithContext(ctx, ctrl.worker, time.Second)
+		go func() {
+			defer wg.Done()
+			defer utilruntime.HandleCrash()
+			wait.UntilWithContext(ctx, ctrl.worker, time.Second)
+		}()
 	}
 
 	<-ctx.Done()
```
wking, line 106: now that we have a wait-group blocking on our child goroutines, we can probably drop this direct block on the passed `ctx`.

wking (approved):
/lgtm

PR #614, opened 2021-06-21
Bug 1970421: Do not check if Job has Failed count set

arjunrn:
When a job pod fails because kubelet is unable to pull the image, the pod is never considered *active* and subsequently after timing out it is never considered as `Failed` and the `Failed` count in the job status is never incremented.

### DIFF
--- a/lib/resourcebuilder/batch.go
+++ b/lib/resourcebuilder/batch.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	batchclientv1 "k8s.io/client-go/kubernetes/typed/batch/v1"
@@ -46,15 +47,40 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
 	}
 
 	// Since we have filled in "activeDeadlineSeconds",
-	// the Job will 'Active == 0' if and only if it exceeds the deadline.
+	// the Job will 'Active == 0' if and only if it exceeds the deadline or if the update image could not be pulled.
 	// Failed jobs will be recreated in the next run.
-	if j.Status.Active == 0 && j.Status.Failed > 0 {
-		reason := "DeadlineExceeded"
-		message := "Job was active longer than specified deadline"
-		if len(j.Status.Conditions) > 0 {
-			reason, message = j.Status.Conditions[0].Reason, j.Status.Conditions[0].Message
+	if j.Status.Active == 0 {
+		klog.V(4).Infof("No active pods for job %s in namespace %s", job.Name, job.Namespace)
+		failed, reason, message := hasJobFailed(job)
+		// If there is more than one failed job pod then get the cause for failure
+		if j.Status.Failed > 0 {
+			failureReason := "DeadlineExceeded"
+			failureMessage := "Job was active longer than specified deadline"
+			if failed {
+				failureReason, failureMessage = reason, message
+			}
+			return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", failureReason, failureMessage)
 		}
-		return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
+
+		// When the update image cannot be pulled then the pod is not marked as Failed, but the status condition is set
+		// after the job deadline is exceeded.
+		if failed {
+			if reason == "DeadlineExceeded" {
+				return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
+			} else {
+				klog.V(4).Infof("Ignoring job %s in namespace %s with condition Failed=True because %s: %s", job.Name, job.Namespace, reason, message)
+			}
+		}
 	}
+
 	return false, nil
+}
+
+func hasJobFailed(job *batchv1.Job) (bool, string, string) {
+	for _, condition := range job.Status.Conditions {
+		if condition.Type == batchv1.JobFailed && condition.Status == corev1.ConditionTrue {
+			return true, condition.Reason, condition.Message
+		}
+	}
+	return false, "", ""
 }


lib/resourcebuilder/batch.go
```diff
@@ -48,7 +48,7 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
 	// Since we have filled in "activeDeadlineSeconds",
 	// the Job will 'Active == 0' if and only if it exceeds the deadline.
 	// Failed jobs will be recreated in the next run.
-	if j.Status.Active == 0 && j.Status.Failed > 0 {
+	if j.Status.Active == 0 && len(j.Status.Conditions) > 0 {
```
vrutkovs, line 51:
```suggestion
	if j.Status.Active == 0{
```
Perhaps we should leave out conditions check and have it checked later?

vrutkovs (changes requested):

lib/resourcebuilder/batch.go
arjunrn: I messed up the implementation. I've fixed it now. PTAL.

lib/resourcebuilder/batch.go
```diff
@@ -46,15 +46,25 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
...
+			}
+			return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
+		}
+
+		// When the update image cannot be pulled then the pod is not marked as Failed, but the status condition is set
+		// after the job deadline is exceeded.
+		if len(j.Status.Conditions) > 0 && j.Status.Conditions[0].Reason == "DeadlineExceeded" {
+			return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", j.Status.Conditions[0].Reason, j.Status.Conditions[0].Message)
 		}
-		return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
```
vrutkovs, line 57: Lets add "unknown error" to the case where Active=0 and no Conditions set (that probably means some serious kube-api/scheduler issue)

arjunrn: That condition will be true for a short period of time between the job being created and the first pod being created. So it might give false positives.

vrutkovs: Lets add some logging to this event, seems several occurrences might indicate a problem with the platform - I'd rather have CVO find it. Setting status for false positives is a bad idea indeed, but logging won't hurt

arjunrn:
CVO can detect and report failure to download. Example:

```
status:
  availableUpdates: null
  conditions:
  - lastTransitionTime: "2021-06-22T09:18:11Z"
    message: The update channel has not been configured.
    reason: NoChannel
    status: "False"
    type: RetrievedUpdates
  - lastTransitionTime: "2021-06-22T09:47:31Z"
    message: Done applying 4.8.0-0.ci.test-2021-06-22-091115-ci-ln-jr6tvvb-latest
    status: "True"
    type: Available
  - lastTransitionTime: "2021-06-22T10:41:01Z"
    message: 'Unable to download and prepare the update: timed out waiting for the condition'
    reason: UpdatePayloadRetrievalFailed
    status: "True"
    type: Failing
  - lastTransitionTime: "2021-06-22T10:26:32Z"
    message: 'Unable to apply awesome.registry.cz/image/tag:latest: could not download the update'
    reason: UpdatePayloadRetrievalFailed
    status: "True"
    type: Progressing
  desired:
    image: awesome.registry.cz/image/tag:latest
    version: ""
  history:
  - completionTime: null
    image: awesome.registry.cz/image/tag:latest
    startedTime: "2021-06-22T10:26:32Z"
    state: Partial
    verified: false
    version: ""
  - completionTime: "2021-06-22T09:47:31Z"
    image: registry.build01.ci.openshift.org/ci-ln-jr6tvvb/release@sha256:36d7aaaf2141d703198e64cb3ccd3304833b7fc37a08df1e1d82676584c2d882
    startedTime: "2021-06-22T09:18:11Z"
    state: Completed
    verified: false
    version: 4.8.0-0.ci.test-2021-06-22-091115-ci-ln-jr6tvvb-latest
  observedGeneration: 2
  versionHash: 4G-siGb0hW0=
```

lib/resourcebuilder/batch.go
arjunrn: Added the log

vrutkovs (approved):
/lgtm

lib/resourcebuilder/batch.go
```diff
@@ -46,15 +46,26 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
...
+			message := "Job was active longer than specified deadline"
+			if len(j.Status.Conditions) > 0 {
+				reason, message = j.Status.Conditions[0].Reason, j.Status.Conditions[0].Message
+			}
+			return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
+		}
+
+		// When the update image cannot be pulled then the pod is not marked as Failed, but the status condition is set
+		// after the job deadline is exceeded.
+		if len(j.Status.Conditions) > 0 && j.Status.Conditions[0].Reason == "DeadlineExceeded" {
```
wking, line 65: I see we were using `Conditions[0]` before as well, but ideally we'd be looking for specific condition types.  Because maybe sometime in the future, the job controller adds a `OptimisticThatTheNextAttemptWillPass=True` condition, and we don't want to use that when we complain about it being broken so far.

lib/resourcebuilder/batch.go
```diff
@@ -46,15 +47,36 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
...
+			failureMessage := "Job was active longer than specified deadline"
+			if failed {
+				failureReason, failureMessage = reason, message
+			}
+			return false, fmt.Errorf("deadline exceeded, failureReason: %q, message: %q", failureReason, failureMessage)
+		}
+
+		// When the update image cannot be pulled then the pod is not marked as Failed, but the status condition is set
+		// after the job deadline is exceeded.
+		if failed && reason == "DeadlineExceeded" {
```
wking, line 67: I don't think we need the `reason` check here.  Any reason out of `hasJobFailed`'s `Failed=True` condition should be fine for complaining about, right?

arjunrn: I'm not sure. Ideally all the conditions should be present and the `Failed` condition should only be set when the Job has actually failed. But the conditions on the `Job` don't seem very consistent compared to other types of resources. I'm trying to keep the logic the same as it was before and not report a failure when there is chance that the job would be successful at some point in the future.

lib/resourcebuilder/batch.go
```diff
@@ -46,15 +47,36 @@ func checkJobHealth(ctx context.Context, client batchclientv1.JobsGetter, job *b
...
+		klog.V(4).Infof("No active pods for job %s in namespace %s", job.Name, job.Namespace)
+		failed, reason, message := hasJobFailed(job)
+		// If there is more than one failed job pod then get the cause for failure
+		if j.Status.Failed > 0 {
+			failureReason := "DeadlineExceeded"
+			failureMessage := "Job was active longer than specified deadline"
+			if failed {
+				failureReason, failureMessage = reason, message
+			}
+			return false, fmt.Errorf("deadline exceeded, failureReason: %q, message: %q", failureReason, failureMessage)
```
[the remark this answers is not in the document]
arjunrn: Updated

lib/resourcebuilder/batch.go
wking: can we at least log the `Failed=True` reason/message we're going to ignore, then?  Something like:

```go
if failed {
  if reason == "DeadlineExceeded" {
    return false, fmt.Errorf("deadline exceeded, reason: %q, message: %q", reason, message)
  } else {
    klog.V(4).Infof("Ignoring job %s in namespace %s setting Failed=True with %s: %s", job.Name, job.Namespace, reason, message)
   }
 }
 
 return false, nil
 ```

arjunrn: Done

wking (approved):
/lgtm

PR #616, opened 2021-06-24
Fix more golangci-lint issues

arjunrn:
Fixed linting issues with redundant type declaration and the use of `fmt.Sprintf()` when there is no string formatting required.


LalatenduMohanty (approved):
/lgtm

vrutkovs (approved):
/lgtm

PR #617, opened 2021-06-25
Added ownerReference to all objects created and managed by CVO

arjunrn:
The operator object no has a metadata modifier which injects an owner reference if there is no existing owner references. Also updates the ownerReference if it has been changed due to recreation

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	informerscorev1 "k8s.io/client-go/informers/core/v1"
@@ -148,6 +149,7 @@ type Operator struct {
 	exclude string
 
 	clusterProfile string
+	uid            types.UID
 }
 
 // New returns a new cluster version operator.
@@ -253,7 +255,7 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	// which will consume the verifier
 	optr.configSync = NewSyncWorkerWithPreconditions(
 		optr.defaultPayloadRetriever(),
-		NewResourceBuilder(restConfig, burstRestConfig, &dummyContextOperatorGetter{wrapped: optr.coLister}),
+		NewResourceBuilder(restConfig, burstRestConfig, &dummyContextOperatorGetter{wrapped: optr.coLister}, optr.ownerReferenceModifier),
 		optr.defaultPreconditionChecks(),
 		optr.minimumUpdateCheckInterval,
 		wait.Backoff{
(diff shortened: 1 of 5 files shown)


pkg/cvo/cvo.go
```diff
@@ -269,6 +271,24 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	return nil
 }
 
+func (optr *Operator) ownerReferenceModifier(object metav1.Object) {
+	var found bool
+	for _, ownerRef := range object.GetOwnerReferences() {
+		if ownerRef.Kind == "ClusterVersion" && ownerRef.Name == optr.name {
```
jottofar, line 277: We should be careful adding CVO ownerReference to an object that already has an owner.  If the object has an owner because it was part of a cascading delete set-up then adding CVO would interfere with that (See discussions in the [Spike](https://issues.redhat.com/browse/OTA-403)).

I thought it would be worthwhile to log those objects that already had an owner to do some spot checking. May be a non-issue but not sure.

arjunrn: I've updated the logic to only set the CV owner reference when there is no existing owner reference which is not a CV resource.

pkg/cvo/cvo.go
```diff
@@ -269,6 +271,40 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
...
+			found = true
+			break
+		}
+	}
+	if found {
+		// If there are ownerReferences which for non ClusterVersion resource then remove all ClusterVersion
+		// ownerReferences to them
+		var nonCVORefs []metav1.OwnerReference
+		for _, ref := range object.GetOwnerReferences() {
+			if ref.Kind != "ClusterVersion" && ref.Name != optr.name {
```
wking, line 289: this seems a bit redundent with the earlier loop setting `found`.  How about:

```go
modified := false
var otherRefs []metav1.OwnerReference
for _, ref := range object.GetOwnerReferences() {
  if ref.Kind != "ClusterVersion" || ref.Name != optr.name) {
    otherRefs = append(otherRefs, ref)
  }
  if len(otherRefs) > 0 {
    // FIXME: Log that we are defering to these other owners, despite us continuing to semi-manage this resource
    object.SetOwnerReferences(otherRefs)
  } else {
    object.SetOwnerReferences([]metav1.OwnerReference{{
      APIVersion: configv1.GroupVersion.Identifier(),
      Kind:       "ClusterVersion",
      Name:       optr.name,
      UID:        optr.uid,
    }})
  }
}
```

arjunrn: Updated as per suggestion.

pkg/cvo/cvo.go
```diff
@@ -269,6 +271,31 @@ func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestCo
 	return nil
 }
 
+// ownerReferenceModifier sets the owner reference to the current CV resource if no other reference exists
+func (optr *Operator) ownerReferenceModifier(object metav1.Object) {
+	var nonCVORefs []metav1.OwnerReference
+	// find any existing owner references which are not ClusterVersion resource types
```
jottofar, line 277: This is necessary I suppose because as CVO reconciles resources one which previously had CV set as the owner now has non-CV owners so we must remove CV? Can we put a comment in here to that effect or explain why this is necessary.

PR #618, opened 2021-06-25
Bug 1976307: add ImageStreams manifest delete annotation logic

jottofar:


jottofar:
This PR is needed by 
https://github.com/openshift/cluster-samples-operator/pull/380

wking (approved):
/lgtm

LalatenduMohanty (changes requested):
Can we add an unit test for this?

jottofar:
> Can we add an unit test for this?

There are no unit tests specific to resource type, e.g. imagestreams vs. deployments. There's only the generic delete annotation unit tests in `lib/resourcedelete/helper_test.go` which this mod doesn't affect.

jottofar:
> Can we add an unit test for this?

Actually,  thinking about this more, you're probably asking "can we add a unit test that would've caught this issue". Probably so but I don't think there's anything there now to "build on" but I'll look again to see what can be done.

jottofar:
> you're probably asking "can we add a unit test that would've caught this issue"

I looked at this when I originally added the `resourcedelete` package. There were no Unit tests for `resourcebuilder` and what I wanted for a Unit test then, and needs to be tested now to have caught this issue, is a test at the `resourcebuilder` Do method level. But I don't see a good way to do that, i.e. one which does not involve running on a cluster.

PR #626, opened 2021-07-08
 Bug 1927168: Fail fast when applying resources with no definition

arjunrn:
The CVO attempts to retry creation of resources when there are failures. But this approach prevents CVO from advancing when the resource defition for a resource is missing. The CVO should instead no rety and proceed with the creation of subsequent resources so that the definition could be applied later.

### DIFF
--- a/pkg/cvo/internal/generic.go
+++ b/pkg/cvo/internal/generic.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/openshift/cluster-version-operator/lib/resourceapply"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
--- a/pkg/cvo/internal/generic_test.go
+++ b/pkg/cvo/internal/generic_test.go
@@ -2,15 +2,16 @@ package internal
 
 import (
 	"context"
+	"encoding/json"
 	"testing"
 
-	"k8s.io/apimachinery/pkg/runtime/schema"
-
-	"k8s.io/apimachinery/pkg/runtime"
-
+	"github.com/openshift/library-go/pkg/manifest"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic/fake"
+	clientgotesting "k8s.io/client-go/testing"
 )
 
 func TestCreateOnlyCreate(t *testing.T) {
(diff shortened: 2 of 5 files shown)


pkg/cvo/sync_test.go
```diff
@@ -41,74 +41,77 @@ func Test_SyncWorker_apply(t *testing.T) {
 
 		check   func(*testing.T, []action)
 		wantErr bool
-	}{{
```
arjunrn, line 44: Note to Reviewers: Although it looks like a lot of this test has changed, the only non-formatting change is the change in number of `Actions`.

wking: easy way to help review with that sort of thing is to isolate the no-op format changes in a separate commit, so folks looking at the full PR can focus on the semantic changes.

arjunrn: Moved it into a separate commit

pkg/cvo/internal/generic_test.go
```diff
@@ -87,3 +87,71 @@ func TestCreateOnlyUpdate(t *testing.T) {
...
+}`,
+		},
+		{
+			name: "delete annotation",
+			resource: `{
+  "kind": "UnknownType",
+  "apiVersion": "config.openshift.io/v1",
+  "metadata": {
+     "name": "cluster",
+	 "annotations": {
```
wking, line 113: nit: seems like you're adding an additional level of indentation here?

pkg/cvo/internal/generic_test.go
```diff
@@ -87,3 +87,71 @@ func TestCreateOnlyUpdate(t *testing.T) {
...
+		 "release.openshift.io/delete": "true"
+	 }
+  }
+}`,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := context.Background()
+			obj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, []byte(tc.resource))
```
wking, line 122: you can probably avoid some of these runtime gymnastics by including `github.com/openshift/library-go/pkg/manifest` and using `json.Unmarshal` like we do [here][1].

[1]: https://github.com/openshift/cluster-version-operator/blob/887aa47093d740ce827e3046848d4f0c1aca374f/pkg/cvo/sync_test.go#L116-L118

pkg/cvo/internal/generic_test.go
arjunrn: Mixing tabs and spaces got me there. Fixed now.

pkg/cvo/internal/generic_test.go
arjunrn: Updated as per recommendation.

pkg/cvo/internal/generic.go
```diff
@@ -42,6 +43,12 @@ func deleteUnstructured(ctx context.Context, client dynamic.ResourceInterface, r
 		Name:      required.GetName(),
 	}
 	existing, err := client.Get(ctx, required.GetName(), metav1.GetOptions{})
+
+	// If there is a NoMatchError then, it's guaranteed that the resource does not exist.
+	if err != nil && meta.IsNoMatchError(err) {
+		return false, err
```
wking, line 49: hmm, but if this is a delete manifest, this isn't an error, this is "hooray, success", right?  Maybe `GetDeleteProgress` needs to be taught to distinguish this case?

jottofar: Agree, it should be handled in `GetDeleteProgress` which is intended to handle all returns from the `Get`. [Here](https://github.com/openshift/cluster-version-operator/blob/3a68652568e9075c23f491bc8c037942bd67ec82) and [here](https://github.com/openshift/cluster-version-operator/blob/3a68652568e9075c23f491bc8c037942bd67ec82/lib/resourcedelete/helper.go#L124) as :
```
if apierrors.IsNotFound(getError) || meta.IsNoMatchError(getError) {
```

arjunrn: Updated to handle it in `GetDeleteProgress`.

pkg/cvo/internal/generic_test.go
```diff
@@ -87,3 +88,74 @@ func TestCreateOnlyUpdate(t *testing.T) {
...
+			}
+			err = builder.Do(ctx)
+			if tc.expectErr {
+				if err == nil {
+					t.Fatal("expected error, got none")
+				}
+				if !meta.IsNoMatchError(err) {
+					t.Fatalf("expected error of type IsNoMatchError, instead got: %T", err)
+				}
+			}
```
wking, line 158: To handle the "we didn't expect an error, but got one" case:

```go
} else if err != nil {
  t.Fatal(err)
}
```

PR #628, opened 2021-07-13
Bug 1984414: Log resource diffs on update only in reconcile mode

arjunrn:
Logging of resource diffs was introduced to detect [hotlooping in the CVO](https://github.com/openshift/cluster-version-operator/pull/561). But this increases the verbosity of logging during updates when the resources have changed which is not useful. The logging of the diffs is more useful in the reconcile mode. Also the diff is generated with [ObjectDiff](https://github.com/kubernetes/apimachinery/blob/235edae7dd90601011bbe3bcd6f84f7dc857b034/pkg/util/diff/diff.go#L57) which is deprecated. The `ObjectDiff` function now just uses [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) internally and all instances of `ObjectDiff` have been replaced with `cmp.Diff`.

### DIFF
--- a/lib/resourceapply/apiext.go
+++ b/lib/resourceapply/apiext.go
@@ -3,17 +3,17 @@ package resourceapply
 import (
 	"context"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextclientv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 )
 
-func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, required *apiextv1.CustomResourceDefinition) (*apiextv1.CustomResourceDefinition, bool, error) {
+func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, required *apiextv1.CustomResourceDefinition, reconciling bool) (*apiextv1.CustomResourceDefinition, bool, error) {
 	existing, err := client.CustomResourceDefinitions().Get(ctx, required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		klog.V(2).Infof("CRD %s not found, creating", required.Name)
@@ -34,7 +34,9 @@ func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
+	if reconciling {
+		klog.V(4).Infof("Updating CRD %s due to diff: %v", required.Name, cmp.Diff(existing, required))
+	}
 
 	actual, err := client.CustomResourceDefinitions().Update(ctx, existing, metav1.UpdateOptions{})
 	return actual, true, err
--- a/lib/resourceapply/cv.go
+++ b/lib/resourceapply/cv.go
@@ -3,42 +3,16 @@ package resourceapply
 import (
 	"context"
 
+	"github.com/google/go-cmp/cmp"
 	configv1 "github.com/openshift/api/config/v1"
 	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 )
-
-func ApplyClusterVersionv1(ctx context.Context, client configclientv1.ClusterVersionsGetter, required *configv1.ClusterVersion) (*configv1.ClusterVersion, bool, error) {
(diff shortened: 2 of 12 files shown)


lib/resourceapply/apiext.go
```diff
@@ -3,17 +3,17 @@ package resourceapply
...
 	apiextclientv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 )
 
-func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, required *apiextv1.CustomResourceDefinition) (*apiextv1.CustomResourceDefinition, bool, error) {
+func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, required *apiextv1.CustomResourceDefinition, reconciling bool) (*apiextv1.CustomResourceDefinition, bool, error) {
```
wking, line 16: Can we pass in the mode instead of passing in a `reconciling` bool?  Like we do for ClusterOperators [here][1]?  That way we don't have to bump cluster signatures if we need to make some other mode distinction later on.

[1]: https://github.com/openshift/cluster-version-operator/blob/3b1047de4d2a82d2a1f4f92c224a1c46eba72ca9/pkg/cvo/internal/operatorstatus.go#L123-L126

lib/resourceapply/cv.go
```diff
@@ -62,7 +36,7 @@ func ApplyClusterVersionFromCache(ctx context.Context, lister configlistersv1.Cl
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating ClusterVersion %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
+	klog.V(2).Infof("Updating ClusterVersion %s due to diff: %v", required.Name, cmp.Diff(existing, required))
```
wking, line 39: no mode guard here?

lib/resourceapply/imagestream.go
```diff
@@ -11,7 +11,7 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-func ApplyImageStreamv1(ctx context.Context, client imageclientv1.ImageStreamsGetter, required *imagev1.ImageStream) (*imagev1.ImageStream, bool, error) {
+func ApplyImageStreamv1(ctx context.Context, client imageclientv1.ImageStreamsGetter, required *imagev1.ImageStream, reconciling bool) (*imagev1.ImageStream, bool, error) {
 	existing, err := client.ImageStreams(required.Namespace).Get(ctx, required.Name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		actual, err := client.ImageStreams(required.Namespace).Create(ctx, required, metav1.CreateOptions{})
```
wking, line 17: no diff log before updating?

lib/resourceapply/apiext.go
```diff
@@ -34,7 +34,9 @@ func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.
 		return existing, false, nil
 	}
 
-	klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, diff.ObjectDiff(existing, required))
+	if reconciling {
+		klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, cmp.Diff(existing, required))
```
vrutkovs, line 38: @LalatenduMohanty argued that we should use `v(4)` to avoid logging too many lines on production clusters (during development we could manually bump level and check for hotloops, as it doesn't need to be done too often)

lib/resourceapply/apiext.go
arjunrn: I considered that and decided against it because it would leak abstractions. `Mode` is part of the `resourcebuilder` package, where as the resource creation/update logic is in the `internal`. If in the future the full mode is required more refactoring can be done where the mode is moved to a common package and then it can be passed to the apply function.

lib/resourceapply/cv.go
arjunrn: Unlike the other resources, the CV is created/updated from multiple controllers and the mode is not present there. So it cannot be used.

lib/resourceapply/apiext.go
arjunrn: I could reduce the log level but the logs will now be emitted only when there are resource updates when in reconcile mode. And this would be indicative of a hot loop. So these logs could help with diagnostics in production clusters. Changing the log level to 4 would be mean that before every release someone would have to increase the log level and check for hot loops. This would also not help in clusters where there are hot loops due to configuration which is not tested in CI.
In summary, this change basically eliminates all the logs which is considered verbose in the bug report. While at the same time keeping the logs for debugging in production clusters.

lib/resourceapply/imagestream.go
arjunrn: Added now.

lib/resourceapply/apiext.go
vrutkovs: I'm fine with moving this to a different PR.

Lets file a bug for this change so that we could backport it? Other than that LGTM

vrutkovs (approved):
/lgtm

lib/resourceapply/apiext.go
LalatenduMohanty: > This would also not help in clusters where there are hot loops due to configuration which is not tested in CI.

This is not true as we are only concerned about hotloops because of manifests in release payload which CVO is responsible for. We can not track every hotloops in the cluster.

LalatenduMohanty: /hold

> klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, cmp.Diff(existing, required))

This will only catch hotloops from manifests which is managed by CVO. So it wont help if there is a hot loop somewhere else. Even if there is a hot loop in a production cluster it does not mean that the cluster's availability is impacted or this is a serious issue which needs to be fixed ASAP. Also I do not see much use for customers from this information because they can not change the manifests from release payload.

LalatenduMohanty: However if we move the log level to V(4) then we can easily get it in our CI (by bumping the log level). So IMHO it should be atleast V(4).

arjunrn: Updated to `V(4)`

LalatenduMohanty (approved):
/lgtm

PR #629, opened 2021-07-14
Bug 1982046: lib/resourcedelete: Always check delete progress

jottofar:
Previously delete methods would simply return when not in UpdateMode. This meant that if a delete was started in UpdateMode but had not completed while in UpdateMode the delete would never be marked as completed and minor upgrades would remain blocked. With this change deletes are continuously checked which also means a deleted resource that is recreated is logged as such by CVO even when not in UpdateMode. Deletes are still only requested when in UpdateMode.

### DIFF
--- a/lib/resourcedelete/apiext.go
+++ b/lib/resourcedelete/apiext.go
@@ -10,25 +10,23 @@ import (
 )
 
 // DeleteCustomResourceDefinitionv1 checks the given resource for a valid delete annotation. If found
-// and in UpdatingMode it will issue a delete request or provide status of a previousily issued delete request.
-// If not in UpdatingMode it simply returns an indication that the delete annotation was found. An error is
-// returned if an invalid annotation is found or an error occurs during delete processing.
+// it checks the status of a previousily issued delete request. If delete has not been
+// requested and in UpdatingMode it will issue a delete request.
 func DeleteCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter,
 	required *apiextv1.CustomResourceDefinition, updateMode bool) (bool, error) {
 
 	if delAnnoFound, err := ValidDeleteAnnotation(required.Annotations); !delAnnoFound || err != nil {
 		return delAnnoFound, err
-	} else if !updateMode {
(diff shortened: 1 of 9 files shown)


jottofar:
/retitle Bug 1982046: lib/resourcedelete: Always check delete progress

lib/resourcedelete/apiext.go
```diff
@@ -10,25 +10,23 @@ import (
...
+// requested and in UpdatingMode it will issue a delete request.
 func DeleteCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter,
 	required *apiextv1.CustomResourceDefinition, updateMode bool) (bool, error) {
 
 	if delAnnoFound, err := ValidDeleteAnnotation(required.Annotations); !delAnnoFound || err != nil {
 		return delAnnoFound, err
-	} else if !updateMode {
-		return true, nil
 	}
+	existing, err := client.CustomResourceDefinitions().Get(ctx, required.Name, metav1.GetOptions{})
```
vrutkovs, line 21: There's no real reason to move this higher, right?

vrutkovs (commented):
/lgtm
/hold
/cc @wking @LalatenduMohanty 

Loots fine to me (except the comment below to improve the diff), feel free to pull the hold

lib/resourcedelete/apiext.go
jottofar: Not required but I did it so this matches `resourcedelete` delete methods. Also, a minor thing, but now `resource` is declared closer to where it's used.

LalatenduMohanty (approved):
/lgtm

vrutkovs:
/lgtm cancel

```
 Building github.com/openshift/cluster-version-operator (v1.0.0-627-g98177c71-dirty)
# github.com/openshift/cluster-version-operator/pkg/cvo/internal
pkg/cvo/internal/generic.go:124:2: duplicate field mode
pkg/cvo/internal/generic.go:137:3: ambiguous selector b.mode 
```
Must be a bad rebase

vrutkovs (approved):
/lgtm

PR #631, opened 2021-07-23
tombstone previous docs location with link to openshift/enhancements.

deads2k:
Our enhancements repo exists to coordinate between our teams.  The CVO
organizes the second level operators, so all thirty-ish of those
components need to see these docs.  The information about local
development remains local to this repo.

The replacements are created here: https://github.com/openshift/enhancements/pull/695


LalatenduMohanty (approved):
/lgtm

PR #633, opened 2021-07-27
Bug 1978376: Add admin ack Upgradeable condition gate

jottofar:
This PR adds a generic upgrade blocking mechanism for blocking minor level upgrades. It is the general implementation of the [Add admin ack upgrade gate enhancement](https://github.com/openshift/enhancements/pull/828). Therefore since 4.9 does not contain any active admin ack gates the check `clusterAdminAcksCompletedUpgradeable` is not added to `defaultUpgradeableChecks`. It is added as part of unit test and tested.

This PR will be backported to 4.8.z and the 4.8 specific implementation details will be added, i.e. the `ack-4.8-kube-122-api-removals-in-4.9` gate will be added by adding check `clusterAdminAcksCompletedUpgradeable` to `defaultUpgradeableChecks`.

Once active the  `clusterAdminAcksCompletedUpgradeable` Upgradeable check implementation differs from other Upgradeable checks by also being synchronously triggered when either configmap admin-gates or admin-acks is added, updated, or deleted. This will be achieved by adding an event handler to the `cmConfigInformer` and `cmConfigManagedInformer` Informers. The event handlers will simply invoke `setUpgradeableConditions` to ensure all conditions are properly evaluated. Since the check will be added as a default Upgradeable check when active it will also be evaluated during the existing periodic polling of Upgradeable conditions.
 
I also removed the unused parameter `config` from `cvo.syncUpgradeable`.

### DIFF
--- a/install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
+++ b/install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: admin-acks
+  namespace: openshift-config
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
--- a/install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
+++ b/install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: admin-gates
+  namespace: openshift-config-managed
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -200,6 +200,8 @@ func New(
 	}
 
 	cvInformer.Informer().AddEventHandler(optr.eventHandler())
+	cmConfigInformer.Informer().AddEventHandler(optr.adminAcksEventHandler())
+	cmConfigManagedInformer.Informer().AddEventHandler(optr.adminGatesEventHandler())
 
 	optr.coLister = coInformer.Lister()
 	optr.cacheSynced = append(optr.cacheSynced, coInformer.Informer().HasSynced)
@@ -613,7 +615,7 @@ func (optr *Operator) upgradeableSync(ctx context.Context, key string) error {
 		return nil
 	}
 
-	return optr.syncUpgradeable(config)
+	return optr.syncUpgradeable()
 }
 
 // isOlderThanLastUpdate returns true if the cluster version is older than
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -26,8 +26,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/informers"
 	kfake "k8s.io/client-go/kubernetes/fake"
+	clienttesting "k8s.io/client-go/testing"
 	ktesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
@@ -2708,15 +2713,36 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 	}
 }
 
+// waits for admin ack configmap changes
+func waitForCm(t *testing.T, cmChan chan *corev1.ConfigMap) {
+	select {
+	case cm := <-cmChan:
+		t.Logf("Got configmap from channel: %s/%s", cm.Namespace, cm.Name)
+	case <-time.After(wait.ForeverTestTimeout):
+		t.Error("Informer did not get the configmap")
+	}
+}
+
 func TestOperator_upgradeableSync(t *testing.T) {
 	id := uuid.Must(uuid.NewRandom()).String()
 
+	var defaultGateCm = corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+			Namespace: "test"},
+	}
+	var defaultAckCm = corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "admin-acks",
+			Namespace: "test"},
+	}
 	tests := []struct {
-		name    string
-		key     string
-		optr    *Operator
-		wantErr func(*testing.T, error)
-		want    *upgradeable
+		name           string
+		key            string
+		optr           *Operator
+		cm             corev1.ConfigMap
+		gateCm         *corev1.ConfigMap
+		ackCm          *corev1.ConfigMap
+		wantErr        func(*testing.T, error)
+		expectedResult *upgradeable
 	}{
 		{
 			name: "when version is missing, do nothing (other loops should create it)",
@@ -2728,6 +2754,10 @@ func TestOperator_upgradeableSync(t *testing.T) {
 				namespace: "test",
 				name:      "default",
 				client:    fake.NewSimpleClientset(),
+			},
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
 			},
 		},
 		{
@@ -2758,7 +2788,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -2807,7 +2841,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -2864,7 +2902,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -2924,7 +2966,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -2986,7 +3032,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -3051,7 +3101,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{
 				Conditions: []configv1.ClusterOperatorStatusCondition{{
 					Type:    configv1.OperatorUpgradeable,
 					Status:  configv1.ConditionFalse,
@@ -3098,7 +3152,11 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{},
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{},
 		},
 		{
 			name: "no error conditions",
@@ -3130,15 +3188,19 @@ func TestOperator_upgradeableSync(t *testing.T) {
 					},
 				),
 			},
-			want: &upgradeable{},
+			cm: corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "admin-gates",
+					Namespace: "test"},
+			},
+			expectedResult: &upgradeable{},
 		},
 		{
-			name: "no error conditions",
+			name: "no error conditions and admin ack gate does not apply to version",
 			optr: &Operator{
 				defaultUpstreamServer: "http://localhost:8080/graph",
 				release: configv1.Release{
-					Version: "v4.0.0",
-					Image:   "image/image:v4.0.1",
+					Version: "v4.9.0",
+					Image:   "image/image:v4.9.1",
 				},
 				namespace: "test",
 				name:      "default",
@@ -3156,7 +3218,8 @@ func TestOperator_upgradeableSync(t *testing.T) {
 						},
 						Status: configv1.ClusterVersionStatus{
 							History: []configv1.UpdateHistory{
-								{Image: "image/image:v4.0.1"},
+								{Version: "4.9.0"},
+								{Image: "image/image:v4.9.1"},
 							},
 						},
 					},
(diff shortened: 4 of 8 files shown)


pkg/cvo/upgradeable.go
```diff
@@ -2,24 +2,33 @@ package cvo
...
+	listerscorev1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-version-operator/lib/resourcedelete"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
+	"github.com/openshift/cluster-version-operator/pkg/internal"
+	"github.com/openshift/cluster-version-operator/pkg/payload/precondition/clusterversion"
 )
 
+const adminAckGateFmt string = "^ack-[4-5].([0-9]{1,})"
```
jianlinliu, line 28: How about escape the `.` to be more restrict?

jottofar:
/retitle WIP: Bug 1978376: Add admin ack Upgradeable condition gate

jottofar:
/retitle Bug 1978376: Add admin ack Upgradeable condition gate

install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: admin-acks
+  namespace: openshift-cluster-version
+  annotations:
+    release.openshift.io/create-only: "true"
```
wking, line 7: and our usual `include.release.openshift.io/self-managed-high-availability: "true"
` profile annotation?

install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: admin-gates
+  namespace: openshift-cluster-version
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+data:
+  ack-4.8-kube-122-api-removals-in-4.9: |
+    Detailed description of what admin needs to do...Read documentation here https://
```
wking, line 10: probably needs to stay WIP until we have some content to put here?

pkg/cvo/cvo.go
```diff
@@ -210,6 +212,7 @@ func New(
 	optr.proxyLister = proxyInformer.Lister()
 	optr.cmConfigLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigNamespace)
 	optr.cmConfigManagedLister = cmConfigManagedInformer.Lister().ConfigMaps(internal.ConfigManagedNamespace)
+	optr.cmConfigAdminGatesLister = cmConfigAdminGatesInformer.Lister().ConfigMaps(internal.ClusterVersionNamespace)
```
wking, line 215: admin gates are what we're using this for now, but the actual informer is covering all ConfigMaps in our namespace, of which there may be more later.  So maybe `optr.cmClusterVersion` or some such instead?

install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
jottofar: In the enhancement we said we were not going to have CVO manage this one in order to allow an admin to simply delete it.

pkg/cvo/cvo.go
wking: or maybe we move `admin-gates` into `openshift-config-managed` and `admin-acks` into `openshift-config`?  One benefit for us is that we already have informers for those, but another benefit is that admins touching some, but not all, ConfigMaps in `openshift-config` sounds better to me than training folks to touch some, but not all, ConfigMaps in `openshift-cluster-version`.  Would probably need an enhancement bump, if we make this pivot.

install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
jottofar: Actually I need to pull the data out since we have no active gate until back port to 4.8.

jottofar: Removed the data altogether since there are no active gates in 4.9.

pkg/cvo/cvo.go
jottofar: Renamed relevant names changing `AdminGates` -> `ClusterVersion`.

install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
```
wking, line 2: Checking up on how we manage ConfigMaps, we currently [use `mergeMap` on `data`][1], and [`mergeMap`][2] allows other actors to append keys beyond those we list in the manifest.  That's unlikely to cause a problem, because actors with the auth to update this map are likely to also have the auth to ack any keys they inject here, but I thought I'd point it out in case others felt differently.

[1]: https://github.com/openshift/cluster-version-operator/blob/3b1047de4d2a82d2a1f4f92c224a1c46eba72ca9/lib/resourcemerge/core.go#L13-L19
[2]: https://github.com/openshift/cluster-version-operator/blob/3b1047de4d2a82d2a1f4f92c224a1c46eba72ca9/lib/resourcemerge/meta.go#L28-L41

install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
wking: In this PR's [e2e run][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/633/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1423306540869226496/artifacts/e2e-agnostic/gather-extra/artifacts/configmaps.json | jq -r '.items[].metadata | select(.name == "admin-acks")'
{
  "annotations": {
    "release.openshift.io/create-only": "true"
  },
  "creationTimestamp": "2021-08-05T15:56:27Z",
  "name": "admin-acks",
  "namespace": "openshift-cluster-version",
  "resourceVersion": "68",
  "uid": "d611d1f7-ba10-46b0-a0e9-9e4f2b4a4abf"
}
```

So yeah, something seems to be creating this manifest...  The CVO loads manifests [from two directories][2], [`manifests` for CVO manifests, and `release-manifests` for everyone else's manifests][3].  Both directories are [filtered by profile][4].  15:56:27Z is very early in the install, while [the installer is still creating Terraform resources][5]:

```
time="2021-08-05T15:59:28Z" level=info msg="Waiting up to 20m0s for the Kubernetes API at https://api.ci-op-345d5w3h-3302f.ci.azure.devcluster.openshift.com:6443..."
time="2021-08-05T15:59:28Z" level=info msg="API v1.21.1-1423+8268f88fb9294f-dirty up"
```

Poking at audit logs:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/633/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1423306540869226496/artifacts/e2e-agnostic/gather-audit-logs/artifacts/audit-logs.tar | tar xz --strip-components=1
$ zgrep -h '"resource":"configmaps"' audit_logs/*/*.gz 2>/dev/null | jq -r 'select(.objectRef.name // "" | contains("admin-"))  | .stageTimestamp + " " + (.responseStatus.code | tostring) + " " + .verb + " " + .objectRef.name + " " + .user.username' | sort
2021-08-05T16:09:44.035961Z 200 get admin-gates system:serviceaccount:kube-system:generic-garbage-collector
2021-08-05T16:13:56.999908Z 200 get admin-gates system:serviceaccount:kube-system:generic-garbage-collector
...
2021-08-05T16:44:03.026572Z 200 get admin-gates system:serviceaccount:openshift-cluster-version:default
2021-08-05T16:50:50.925720Z 200 get admin-gates system:serviceaccount:openshift-cluster-version:default
```

So I guess we don't have audit logs for stuff that happened on the bootstrap host.  Hmm, I wonder if we account for cluster-profile when [rendering our manifests on the bootstrap host][6].  We're [picking up the env var][7].  But then [we are only using it for templating, not for filtering][8].  So we need to get a `shouldExclude` call somewhere under `Render`.

And while leaving off profile annotations will work on bootstrap with the current rendering code, it will not work on updates from previous versions to this new version (which we should be able to see in CI here), because outside of the rendered manifests, the CVO will ignore this profile-less file.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/633/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1423306540869226496
[2]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/pkg/payload/payload.go#L301-L309
[3]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/pkg/payload/payload.go#L93-L94
[4]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/pkg/payload/payload.go#L182-L189
[5]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/633/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1423306540869226496/artifacts/e2e-agnostic/ipi-install-install/artifacts/.openshift_install.log
[6]: https://github.com/openshift/installer/blob/09d231228f974e00a8c18946add3a3ff32fbd2b6/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L80-L105
[7]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/cmd/render.go#L43-L54
[8]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/pkg/payload/render.go#L17-L105

jottofar: IIUC seems the safer way to go is to go ahead and include the `self-managed-high-availability: "true" `. The only "side-affect" is that if an admin decides to delete the configmap CVO will re-create it in its original form but there's really no reason to delete it. If the gates don't apply the code never checks admin-acks and if any gates do apply removing the configmap still results in Upgradeable=false with an admin-acks NotFound error.

install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
jottofar: The code only checks the hardcoded keys so any additional data added is ignored.

jottofar: Verified through testing that he code only checks the hardcoded keys so any additional data added is ignored.

install/0000_00_cluster-version-operator_01_adminack_configmap.yaml
LalatenduMohanty: @jottofar Were you able to confirm if the manifests in cvo/install directory goes in to the CVO manifest directory?

LalatenduMohanty: Will making it CVO managed can cause any issue when we move from 4.7 to 4.8?

jottofar: > Were you able to confirm if the manifests in cvo/install directory goes in to the CVO manifest directory?

I will check but they should end up in the same place as the existing CVO manifests in `install`.

>  Will making it CVO managed can cause any issue when we move from 4.7 to 4.8?

I can't think of any issue. It's a new manifest that comes in as part of some 4.8.z and 4.9 and the annotation simply causes CVO to NOT ignore it when`sync'ing`. The CI upgrade test has passed before but I'll look closer at it. Is there something specific you have in mind or are you thinking there's something different about 4.7->4.8 vs 4.8->4.9?

jottofar: > Were you able to confirm if the manifests in cvo/install directory goes in to the CVO manifest directory?

Confirmed that the two new configmap manifests are in /manifests directory of the CVO pod.

jottofar: > Will making it CVO managed can cause any issue when we move from 4.7 to 4.8?

@LalatenduMohanty see comment/question regarding this above

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+		parts := strings.Split(ackVersion, "-")
+		if strings.HasPrefix(currentVersion, parts[1]) {
+			applicable = true
+		}
+	}
+	return applicable, nil
+}
+
+func checkAdminGate(gate string, currentVersion string,
+	gatesConfigmap *corev1.ConfigMap, acksConfigmap *corev1.ConfigMap,
```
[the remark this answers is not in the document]
jottofar: It can defines >1 gates. Same with the acksConfigMap.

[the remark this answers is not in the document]
LalatenduMohanty: In that case it should be `gateConfigmaps`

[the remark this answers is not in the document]
jottofar: It's 1 `configmap` with multiple `gates` so `gatesConfigmap` makes sense to me but if you really prefer `gateConfigmap` and `ackConfigmap` I can change them.

pkg/cvo/cvo_test.go
```diff
@@ -3173,42 +3236,522 @@ func TestOperator_upgradeableSync(t *testing.T) {
...
+			cv:       "4.8.2",
+			wantErr:  false,
+			want:     true,
+		},
+		{
+			name:     "gate name no match",
+			gateName: "ack-4.8-foo",
+			cv:       "4.9.1",
+			wantErr:  false,
+			want:     false,
```
LalatenduMohanty, line 3737: s/wanted/expectedResult/

pkg/cvo/cvo_test.go
```diff
@@ -3173,42 +3236,522 @@ func TestOperator_upgradeableSync(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		{
+			t.Run(tt.name, func(t *testing.T) {
+				applies, err := gateApplicableToCurrentVersion(tt.gateName, tt.cv)
```
LalatenduMohanty, line 3743: s/applies/isApplicable/

LalatenduMohanty (commented):
Adding them as comments as this should not block the PRs from getting merged. I understand that this test just continued some of the naming convention existing in this file.

pkg/cvo/upgradeable.go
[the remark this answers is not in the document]
LalatenduMohanty: @jottofar what about `configmapGates` and `configmapAcks` ?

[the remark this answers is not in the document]
jottofar: Changed to `gateConfigmap` and `ackConfigmap`.

pkg/cvo/cvo_test.go
jottofar: done

pkg/cvo/cvo_test.go
jottofar: done

pkg/internal/constants.go
```diff
@@ -1,8 +1,11 @@
 package internal
 
 const (
-	ConfigNamespace        = "openshift-config"
-	ConfigManagedNamespace = "openshift-config-managed"
-	InstallerConfigMap     = "openshift-install"
-	ManifestsConfigMap     = "openshift-install-manifests"
+	ClusterVersionNamespace = "openshift-cluster-version"
```
wking, line 4: we don't need this now that the ConfigMaps have moved.

LalatenduMohanty (approved):
/lgtm

pkg/internal/constants.go
LalatenduMohanty: /hold to address this comment

pkg/cvo/upgradeable.go
```diff
@@ -2,34 +2,50 @@ package cvo
...
+	listerscorev1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-version-operator/lib/resourcedelete"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
+	"github.com/openshift/cluster-version-operator/pkg/internal"
+	"github.com/openshift/cluster-version-operator/pkg/payload/precondition/clusterversion"
 )
 
+const adminAckGateFmt string = "^ack-[4-5]\\.([0-9]{1,})"
```
wking, line 28: nit: you can use backticked [raw string literals][1] to avoid the doubled escape:

```go
const adminAckGateFmt string = `^ack-[4-5]\.([0-9]{1,})`
```

Although I often just use a range:

```go
const adminAckGateFmt string = "^ack-[4-5][.]([0-9]{1,})"
```

[1]: https://golang.org/ref/spec#String_literals

wking: Also, if this is only feeding the `MustCompile`, we can probably drop this `const` and inline:

```go
var adminAckGateRegexp = regexp.MustCompile("^ack-[4-5][.]([0-9]{1,})")
```

wking: Hmm, also, should the regexp include a trailing `-`?  `^ack-[4-5][.]([0-9]{1,})-`

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
 	return nil
 }
 
+func gateApplicableToCurrentVersion(gateName string, currentVersion string) (bool, error) {
+	var applicable bool
+	if ackVersion := adminAckGateRegexp.FindString(gateName); ackVersion == "" {
+		return false, fmt.Errorf("%s configmap gate name %s has invalid format; must begin with 'ack-X.Y'",
```
wking, line 253: If the regexp requires a `-` after the 4.y, we should mention that here^ too

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+	if applies, err := gateApplicableToCurrentVersion(gate, currentVersion); err == nil {
+		if !applies {
+			return nil
+		}
+	} else {
+		klog.Error(err)
+		return &configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
+			Status:  configv1.ConditionFalse,
+			Reason:  "AdminAckConfigMapIssue",
```
wking, line 277: `Issue` -> `Error`?  Because we are failing out here, so no need to soft-pedal things.

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+			Message: err.Error(),
+		}
+	}
+	var errorMessage string
+	var gateMessage string
+	if val, ok := gateConfigmap.Data[gate]; !ok {
+		errorMessage = fmt.Sprintf("%s configmap does not contain gate %s.", internal.AdminGatesConfigMap, gate)
+	} else if val == "" {
+		errorMessage = fmt.Sprintf("%s configmap gate %s has no value.", internal.AdminGatesConfigMap, gate)
```
wking, line 286: nit: "no value" -> "an empty value"?  Or "is not $WHATEVER_WE_NEED_FOR_AN_ACK"?

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+	} else {
+		gateMessage = val
+	}
+	if ackConfigmapErr != nil {
+		if apierrors.IsNotFound(ackConfigmapErr) {
+			errorMessage = strings.Trim(errorMessage+" "+fmt.Sprintf("%s configmap not found.", internal.AdminAcksConfigMap), " ")
+		} else if ackConfigmapErr != nil {
+			errorMessage = strings.Trim(errorMessage+" "+
+				fmt.Sprintf("Unable to access configmap %s, err=%v.", internal.AdminAcksConfigMap, ackConfigmapErr), " ")
```
wking, line 295: This feels like it should be handled in the caller, or the ConfigMap fetch should be pushed down into this function.  Is there a reason to have the fetcher in one location, and the failed-fetch handlers in another?

pkg/cvo/upgradeable.go
wking: Oh, wait, this is the thing we control.  Probably want some different reason for that, because it should never happen, and we want it to be really obvious in Telemetry if that's going on.

pkg/cvo/upgradeable.go
jottofar: I did it this way so gates can be added easily:

`conds []configv1.ClusterOperatorStatusCondition`
`conds[0] = checkAdminGate("ack-4.8-kube-122-api-removals-in-4.9", currentVersion, gateCm, ackCm,err)`
`... = checkAdminGate("ack-4.9-foo", currentVersion, gateCm, ackCm, err)`

pkg/cvo/upgradeable.go
jottofar: Changed to "...must contain a non-empty value".

pkg/cvo/upgradeable.go
jottofar: Done

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,126 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
 }
 
+func gateApplicableToCurrentVersion(gateName string, currentVersion string) (bool, error) {
+	var applicable bool
+	if ackVersion := adminAckGateRegexp.FindString(gateName); ackVersion == "" {
+		return false, fmt.Errorf("%s configmap gate name %s has invalid format; must begin with 'ack-X.Y'",
+			internal.AdminGatesConfigMap, gateName)
+	} else {
+		parts := strings.Split(ackVersion, "-")
+		if strings.HasPrefix(currentVersion, parts[1]) {
```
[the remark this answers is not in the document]
jottofar: yep, fixed it

pkg/cvo/upgradeable.go
jottofar: It's only checking for `ack-x.y` and ensuring we have a major/minor version number.

pkg/cvo/upgradeable.go
jottofar: changed

pkg/internal/constants.go
jottofar: removed

pkg/cvo/upgradeable.go
jottofar: Didn't see your second comment until after the change above. Right, it should not in the field since this will cause upgrade CI test to fail once this blocker is active (in 4.8).

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,128 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+		klog.Error(message)
+		return &configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
+			Status:  configv1.ConditionFalse,
+			Reason:  "UnableToAccessAdminGatesConfigMap",
+			Message: message,
+		}
+	}
+	ackCm, err := check.adminAcksLister.Get(internal.AdminAcksConfigMap)
+	return checkAdminGate("ack-4.8-kube-122-api-removals-in-4.9", currentVersion, gateCm, ackCm, err)
```
wking, line 367: Hmm.  Are we hard-coding these keys here to avoid the risk of losing a race between admin removing a key from the gate ConfigMap and the manifest-reconciliation loop noticing and stuffing that key back in?

pkg/cvo/upgradeable.go
wking: Can't we:

```go
ackCm, err := check.adminAcksLister.Get(internal.AdminAcksConfigMap)
// handle err, and fast-fail with these messages as needed.
ackErrors := []string{}
for _, ack := range []string{"well", "known", "acks", ...} {
  if err := checkAdminGate(ack, currentVersion, gateCm, ackCm); err != nil {
    ackErrors = append(ackErrors, err)
  }
}
return someAggregatingThing(ackErrors)
```

pkg/cvo/upgradeable.go
jottofar: We hard-coded to avoid added complexity of making them configurable per enhancement.

pkg/cvo/upgradeable.go
jottofar: Sure. I was just trying to show why I split the logic up as I did not that I would necessarily implement it exactly like that.

pkg/cvo/upgradeable.go
sdodson: I wasn't expecting to have the gates hardcoded here, if we add another we'd have to update this code?
I thought we were going to match `ack-4.8` against the current major.minor and anything that matched would get checked for a corresponding admin ack.
We won't need a new enhancement for each gate, we filed an enhancement for the general pattern with a specific example for context.

jottofar: Sorry, hard-coded means "in the code" to me but I put a hold on this. So anything added to the gates cm, along as it fits the required naming pattern, is valid? And if it's not in the ack cm Upgradeable=False.

jottofar: Re-worked to remove the hard-coded gate names from CVO go code.

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,131 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+			Status:  configv1.ConditionFalse,
+			Reason:  reasons[0],
+			Message: messages[0],
+		}
+	} else if len(reasons) > 1 {
+		return &configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
+			Status:  configv1.ConditionFalse,
+			Reason:  "MultipleReasons",
+			Message: strings.Join(messages, " "),
```
wking, line 367: `return "AdminAckRequired", gateValue` in the "you haven't ack'ed this gate yet" case could be fairly long, right?  Maybe we want `fmt.Sprintf("Multiple admin acks failed:\n* %s", strings.Join(messages, "\n* "))` here, like we have for [collecting multiple precondition check errors][1]?  Although there's a risk that life gets complicated while some higher level applies its own concatenation logic to glue together a bunch of `Upgradeable=False` reasons...

[1]: https://github.com/openshift/cluster-version-operator/blob/77a24f77b8f6e9975b5f41a787d6368ff609e391/pkg/payload/precondition/precondition.go#L84

pkg/cvo/upgradeable.go
```diff
@@ -234,6 +247,131 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
+			Status:  configv1.ConditionFalse,
+			Reason:  reasons[0],
+			Message: messages[0],
+		}
+	} else if len(reasons) > 1 {
+		return &configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
+			Status:  configv1.ConditionFalse,
+			Reason:  "MultipleReasons",
```
wking, line 366: Ideally if we have `len(reasons) > 1` but all the reasons are identical (e.g. `AdminAckRequired`), we'd collapse to a single value here.  Maybe we should collect a map of `reasons`, and then here in the `len > 1` case collapse that to a sorted slice?

```go
exists := struct{}{}
reasons := map[string]struct{}{}
for k, v := range gateCm.Data {
	if reason, message := checkAdminGate(k, v, currentVersion, ackCm); reason != "" {
		reasons[reason] = exists
		messages = append(messages, message)
	}
}
if len(reasons) == 1 {
	condition := ...
	for reason in :=range reasons {
		condition.Reason = reason // or is there a more convenient way to get a key out of a single-member map?
	}
	return condition
} else if len(reasons) > 1 {
	...
}

Comment (jottofar) [reply pkg/cvo/upgradeable.go:372]: Unique `Reason` collapsing done above and then check for multi-reasons and join/sort messages.

File: pkg/cvo/upgradeable.go (hunk 17 above)
Comment (jottofar) [reply pkg/cvo/upgradeable.go:373]: Collapse down to single Reason with sorted messages.

File: pkg/cvo/upgradeable.go [hunk 19]
```diff
@@ -234,6 +247,137 @@ func (check *clusterManifestDeleteInProgressUpgradeable) Check() *configv1.Clust
...
+	cvoName          string
+}
+
+func (check *clusterAdminAcksCompletedUpgradeable) Check() *configv1.ClusterOperatorStatusCondition {
+	cv, err := check.cvLister.Get(check.cvoName)
+	if meta.IsNoMatchError(err) || apierrors.IsNotFound(err) {
+		message := fmt.Sprintf("Unable to get ClusterVersion, err=%v.", err)
+		klog.Error(message)
+		return &configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired"),
```
wking, line 301: We have two ~internal type constants already:

```console
$ git --no-pager grep ClusterStatusConditionType pkg/cvo/status.go 
pkg/cvo/status.go:      ClusterStatusFailing = configv1.ClusterStatusConditionType("Failing")
pkg/cvo/status.go:const ClusterVersionInvalid configv1.ClusterStatusConditionType = "Invalid"
```

Can we follow one of those approaches for `UpgradeableAdminAckRequired` too?  That way the compiler will catch us if we accidentally typo the string in of of these consumers.

jottofar: Done

jottofar:
Tightened up gate regex to require one and only one following dash followed by something, i.e. the description. Changed error message to simply spit out "must comply with" and the regex expression itself since it's too hard to explain and these errors should only be consumed by developers. If an admin does add a gate in the field I don't think it's too much to ask that they understand regex.

wking (approved):
/lgtm
/retest

wking:
[The update job][1] had terrible API-server / ingress connectivity.  That's being tracked in [rhbz#1955333][2] and/or [rhbz#1997057][3], and is orthogonal to this PR, so:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/633/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1430314797114593280
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1955333
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1997057

PR #637, opened 2021-08-05
Bug 1997347: Ensure recent etcd backup before minor-version updates

jottofar:


pkg/cvo/cvo.go
```diff
@@ -159,6 +162,8 @@ func New(
 	overridePayloadDir string,
 	minimumInterval time.Duration,
 	cvInformer configinformersv1.ClusterVersionInformer,
+	etcdInformer informers.GenericInformer,
+	etcdClient v1helpers.StaticPodOperatorClient,
```
wking, line 166: We're just looking at etcd's ClusterOperator, right?  I'd expected us to be able to use the existing `coLister` / `coInformer`, etc., without having to add a new informer/client here.

jottofar: Maru said they were updating conditions in "the resource will be apiVersion: operator.openshift.io/v1 kind: Etcd name: cluster"
`etcdInformer` is no longer needed I just hasn't removed it yet.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -111,18 +113,93 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 // Name returns Name for the precondition.
 func (pf *Upgradeable) Name() string { return "ClusterVersionUpgradeable" }
 
+//  RecentEtcdBackup checks if a recent etcd backup has been taken.
+type RecentEtcdBackup struct {
+	key      string
+	lister   configv1listers.ClusterVersionLister
```
wking, line 119: nit: now that we have two listers, can this one be `cvLister`?

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -111,18 +113,93 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+func NewRecentEtcdBackup(lister configv1listers.ClusterVersionLister, coLister configv1listers.ClusterOperatorLister) *RecentEtcdBackup {
+	return &RecentEtcdBackup{
+		key:      "version",
+		lister:   lister,
+		coLister: coLister,
+	}
+}
+
+func recentEtcdBackupCondition(lister configv1listers.ClusterOperatorLister) (string, string) {
+	var reason, message string
```
wking, line 133: nit: the function signature would be a bit more readable if we dropped these names in there, which would also allow you to drop this line:

```go
func recentEtcdBackupCondition(lister configv1listers.ClusterOperatorLister) (reason string, message string) {
```

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -111,18 +113,93 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+	ops, err := lister.Get("etcd")
+	if err == nil {
+		backupCondition := resourcemerge.FindOperatorStatusCondition(ops.Status.Conditions, "RecentBackup")
+		if backupCondition == nil {
+			reason = "EtcdRecentBackupNotSet"
+			message = "RecentBackup: Etcd backup condition is not set."
```
wking, line 139: nit: "Etcd" -> "etcd".  They prefer downcased, even when they're the first word in a sentence, e.g. [here][1].

[1]: https://github.com/etcd-io/etcd#etcd

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -111,18 +113,93 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+		} else if backupCondition.Status != configv1.ConditionTrue {
+			reason = backupCondition.Reason
+			message = backupCondition.Message
```
wking, line 142: The message needs to continue to contain `RecentBackup`, so etcd knows we still want a backup.  Can we `message = fmt.Sprintf("RecentBackup: %s", backupCondition.Message)` or `fmt.Sprintf("%s=%s: %s", backupCondition.Type, backupCondition.Status, backupCondition.Message)` or some such here?

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: > which would also allow you to drop this line:

Not sure what you want? I can simply drop the function altogether and in-line it. Is that what you mean?

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: done

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: done

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: done

hexfusion: For context: The root is derived from the /etc directory.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -111,18 +115,96 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+				Message: message,
+				Name:    pf.Name(),
+			}
+		}
+	}
+	return nil
+}
+
+// Name returns Name for the precondition.
+func (pf *RecentEtcdBackup) Name() string { return "RecentEtcdBackup" }
```
wking, line 190: If we name this `EtcdRecentBackup`, that name will be added to the `Failing=True` message, and we'll have redundant coverage for the "`message` contains a `RecentBackup` substring" that the etcd operator is looking for.

jottofar: done

wking:
Looks good to me, but the commit and pull request both need the WIP dropped from them.

jottofar:
/retitle Bug 1997347: Ensure recent etcd backup before minor-version updates

wking (approved):
[The update job][1] had terrible API-server / ingress connectivity.  That's being tracked in [rhbz#1955333][2] and/or [rhbz#1997057][3], and is orthogonal to this PR, so:

/lgtm

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/637/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1430218723452522496
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1955333
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1997057

wking:
Prow seems to have missed my previous comment, so here are the important bits again:

/lgtm

PR #550, opened 2021-04-26
Bug 1834551: pkg/cvo/metrics: Ignore Degraded for cluster_operator_up

wking:
`ClusterOperatorDown` is based on `cluster_operator_up`, but we also have `ClusterOperatorDegraded` based on `cluster_operator_conditions{condition="Degraded"}`.  Firing `ClusterOperatorDown` for operators which are `Available=True` and `Degraded=True` [confuses users][1].  `ClusterOperatorDegraded` will also be firing.  With this commit, I'm adjusting `cluster_operator_up` to only care about `Available`, to decouple the two alerts and bring them in line with their existing "has not been available" and "has been degraded" descriptions.

However, `IsOperatorStatusConditionTrue` requires the condition to be present, and `cluster_operator_conditions` only creates entries when the conditions are present.  To guard against the Degraded-unset condition in `ClusterOperatorDegraded`, I'm covering with an [`or`][2] and [`group by`][3] guard.  So we should have the following cases:

* `Available` unset or `!= True`: `ClusterOperatorDown` will fire.
* `Available=True`, `Degraded` unset or `!= False`: `ClusterOperatorDegraded` will fire.  Firing on unset is new in this commit.  Not firing `ClusterOperatorDown` here is new in this commit.
* `Available=True`, `Degraded=False`: Neither alert fires.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1834551#c0
[2]: https://prometheus.io/docs/prometheus/latest/querying/operators/#logical-set-binary-operators
[3]: https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

PR #638, opened 2021-08-06
Bug 1991010: pkg/cvo/metrics: Ignore Degraded for cluster_operator_up

wking:
Bringing #550 back to 4.7.  By hand, since there are neighbor-line conflicts with #587, which landed first in 4.7, but is backported from changes that landed later in master/4.8.


wking:
Failures should be fixed by openshift/release#20978.

/retest

PR #641, opened 2021-08-20
pkg/cvo/availableupdates: Expose upstream URI in RetrievedUpdates=True

wking:
So folks can easily see where we're getting the data from, even if `spec.upstream` is unset.  I'm not including the proxy information, because that's all calculated in other layers, and it may contain sensitive information like proxy-connection credentials.  We don't have to worry about sensitive information in the upstream URI, because our default value is in public code, and any sensitive information from `spec.upstream` is already in the ClusterVersion object where we're putting this message.

### DIFF
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -2540,6 +2540,7 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 				Condition: configv1.ClusterOperatorStatusCondition{
 					Type:   configv1.RetrievedUpdates,
 					Status: configv1.ConditionTrue,
+					Reason: "AsExpected",
 				},
 			},
 		},
@@ -2603,6 +2604,7 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 				Condition: configv1.ClusterOperatorStatusCondition{
 					Type:   configv1.RetrievedUpdates,
 					Status: configv1.ConditionTrue,
+					Reason: "AsExpected",
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (changes requested):
We should also print the URL in case of errors  e.g in `Message: fmt.Sprintf("Unable to retrieve available updates: %s", updateError.Message),`

pkg/cvo/cvo_test.go
```diff
@@ -2671,6 +2673,9 @@ func TestOperator_availableUpdatesSync(t *testing.T) {
 				if optr.availableUpdates != nil && optr.availableUpdates.Upstream == "http://localhost:8080/graph" {
 					optr.availableUpdates.Upstream = s.URL
 				}
+				if tt.wantUpdates != nil && tt.wantUpdates.Condition.Status == configv1.ConditionTrue {
+					tt.wantUpdates.Condition.Message = fmt.Sprintf("Successfully retrieved updates from %s", s.URL)
+				}
```
LalatenduMohanty, line 2678: I am not sure how is this going help?

wking:
> We should also print the URL in case of errors...

Don't we?  For example, bringing [this block][1] into [a play][2] shows the error we're bubbling up looks like:

```console
Get "https://example.com/does-not-exist": dial tcp: lookup example.com on 169.254.169.254:53: dial udp 169.254.169.254:53: connect: no route to host
```

Is there a particular error case where you expect to see the upstream URI today, but are not?

[1]: https://github.com/openshift/cluster-version-operator/blob/256eea42a0b99617f93902ac4e3f13a37a4b9380/pkg/cincinnati/cincinnati.go#L102-L104
[2]: https://play.golang.org/p/LoTvHsYn5lE

pkg/cvo/cvo_test.go
wking: The error message includes the upstream URI.  In these tests, we construct the slice of test-cases before we create the upstream service, so at this point, now that we know the upstream URI and can construct the expected message, I'm inserting that message in the test-case so that the later "does the actual message match the test-case's expected message?" logic can confirm that the upstream URI is included.

PR #645, opened 2021-08-27
Bug 1978376: pkg/cvo/upgradeable: Enable admin-ack logic

wking:
519b466793 (#633) had these commented out, because 4.9 has no built-in acks.  But with the code commented out, it's [hard to verify][1] that the logic works in 4.9 before backporting to 4.8.  Enabling these checks should be a no-op outside of verification, because admins are unlikely to inject additional keys in the `openshift-config-managed` namespace's `admin-gates` ConfigMap.  And it allows us to verify the logic in 4.9 and cook there with live code before approving the 4.8 backports.  It's also one less thing we might forget before enabling new admin acks in future versions, like 4.10 or later.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1978376#c19

### DIFF
--- a/pkg/cvo/cvo_test.go
+++ b/pkg/cvo/cvo_test.go
@@ -3756,12 +3756,6 @@ func TestOperator_upgradeableSync(t *testing.T) {
 				optr.cmConfigLister = cmInformerLister.Lister().ConfigMaps("test")
 
 				optr.upgradeableChecks = optr.defaultUpgradeableChecks()
-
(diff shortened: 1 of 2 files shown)


wking:
[`KubePersistentVolumeErrors` going off][1] is orthogonal.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/645/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1432407830538752000

PR #653, opened 2021-09-13
Adding seed to make rand.Float64() return random values

LalatenduMohanty:
Without different seed values in rand.Seed(), rand.Float64() does not return random values

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"context"
+	"math/rand"
+	"time"
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
@@ -19,6 +21,9 @@ func init() {
 		Run: func(cmd *cobra.Command, args []string) {
 			// To help debugging, immediately log version
 			klog.Info(version.String)
+
+			//Setting seed for all the calls to rand methods in the code
+			rand.Seed(time.Now().UnixNano())
 
 			if err := opts.Run(context.Background()); err != nil {
 				klog.Fatalf("error: %v", err)


pkg/start/start.go
```diff
@@ -337,6 +337,7 @@ func createResourceLock(cb *ClientBuilder, namespace, name string) (*resourceloc
 
 func resyncPeriod(minResyncPeriod time.Duration) func() time.Duration {
 	return func() time.Duration {
+		rand.Seed(time.Now().UnixNano())
```
wking, line 340: seeding makes sense to me.  Can we put this in `cmd/main.go`'s `init` or something, so it only gets called once per CVO launch?

LalatenduMohanty: Fixed in https://github.com/openshift/cluster-version-operator/commit/a8a347840c4b313d6bf0dc6fe6bca1d1be2fee0a

wking (approved):
/lgtm

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/653/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1444082967969271808/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-6d69cdd485-shk4v_cluster-version-operator.log | grep 'Starting ClusterVersionOperator with minimum reconcile period'
I1002 00:06:10.137305       1 cvo.go:331] Starting ClusterVersionOperator with minimum reconcile period 2m40.588545509s
```

PR #656, opened 2021-09-15
OWNERS: Prune crawford

wking:
Alex has left Red Hat :(.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,7 +6,6 @@ approvers:
   - jottofar
   - LalatenduMohanty
 emeritus_approvers:
-  - crawford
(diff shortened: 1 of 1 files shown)


wking:
Not that I have a problem leaving him in `emeritus_approvers`, just that he probably doesn't want anyone pestering him about this repo if all of the current approvers are missing for some reason (which is [what `emeritus_approvers` is for][1]).

[1]: https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#emeritus

sdodson:
Need to chase all who have left but since these PRs are already open I'll merge.

PR #577, opened 2021-05-27
pkg/cvo/sync_worker: Consolidate all ClusterOperator errors by reason

wking:
Builds on #575; you may want to review that first.

`newClusterOperatorsNotAvailable` is from c2ac20fa17 (#158).  And the not-available filtering is from bdd4545b1d (#192). But in ce1eda1095 (#514), we moved "waiting on `status.versions`" from `ClusterOperatorNotAvailable` to `ClusterOperatorUpdating`.  And we want to avoid uncollapsed errors like:

```
Multiple errors are preventing progress:
* Cluster operator machine-api is updating versions
* Cluster operator openshift-apiserver is updating versions
```

where we are waiting on multiple ClusterOperator which are in similar situations.  This commit drops the filtering, because cluster operators are important.  It does sort those errors to the end of the list though, so the first error is the non-ClusterOperator error.

PR #662, opened 2021-09-22
Bug 1951835: Propagate Degraded to update status

jottofar:


jottofar:
Failure not related to this PR and test has passed before.

/override e2e-agnostic-upgrade

jottofar:
Failure not related to this PR and test has passed before.

pkg/cvo/status.go
```diff
@@ -367,6 +368,8 @@ func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time,
 	switch uErr.UpdateEffect {
 	case payload.UpdateEffectNone:
 		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+	case payload.UpdateEffectReport:
+		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), false
```
wking, line 372: We're not actually waiting on these, right?  If our most serious error was a report-level issue, we completed the sync loop and should go `Progressing=False`.

pkg/cvo/internal/operatorstatus.go
```diff
@@ -203,14 +203,21 @@ func checkOperatorHealth(ctx context.Context, client ClusterOperatorsGetter, exp
...
 
-	// during initialization we allow degraded
-	if degraded && mode != resourcebuilder.InitializingMode {
+	if degraded {
 		if degradedCondition != nil && len(degradedCondition.Message) > 0 {
 			nestedMessage = fmt.Errorf("cluster operator %s is %s=%s: %s, %s", actual.Name, degradedCondition.Type, degradedCondition.Status, degradedCondition.Reason, degradedCondition.Message)
 		}
+		var updateEffect payload.UpdateEffectType
+
+		// during initialization we
```
wking, line 212: > ... we report Degraded=True ClusterOperators, but do not block on that condition.

pkg/cvo/status.go
```diff
@@ -367,6 +368,8 @@ func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time,
 	switch uErr.UpdateEffect {
 	case payload.UpdateEffectNone:
 		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+	case payload.UpdateEffectReport:
+		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
```
wking, line 372: As mentioned [here][1], we don't want to actually wait on these.  You've changed the returned `ok` from `false` to `true`, which is good, but still have `waiting on ...`.  We probably want to just return `uErr.Error()` here for `message` or something.

[1]: https://github.com/openshift/cluster-version-operator/pull/662#discussion_r720535306

pkg/cvo/internal/operatorstatus.go
```diff
@@ -203,14 +203,20 @@ func checkOperatorHealth(ctx context.Context, client ClusterOperatorsGetter, exp
...
-	// during initialization we allow degraded
-	if degraded && mode != resourcebuilder.InitializingMode {
+	if degraded {
 		if degradedCondition != nil && len(degradedCondition.Message) > 0 {
 			nestedMessage = fmt.Errorf("cluster operator %s is %s=%s: %s, %s", actual.Name, degradedCondition.Type, degradedCondition.Status, degradedCondition.Reason, degradedCondition.Message)
 		}
+		var updateEffect payload.UpdateEffectType
+
+		if mode == resourcebuilder.InitializingMode {
+			updateEffect = payload.UpdateEffectReport
```
wking, line 213: In order to make this non-blocking (which we want), I think we'll also need code around [here][1] like:

```go
if err := task.Run(ctx, payloadUpdate.Release.Version, w.builder, work.State); err != nil {
  if uErr, ok := err.(*payload.UpdateError); ok && uErr.UpdateEffect == payload.UpdateEffectReport {
    // do not fail the task on this manifest, just record it for later complaining
    reportEffectErrors = append(reportEffectErrors, err)
  } else {
    return err
  }
}
```

And then down around [here][2], something like:

```go
// update the status
cr.Complete()
return errors.NewAggregate(reportEffectErrors)
```

[1]: https://github.com/openshift/cluster-version-operator/blob/d3f4ca856badfe3db7a74415bd4a418d128999b1/pkg/cvo/sync_worker.go#L762
[2]: https://github.com/openshift/cluster-version-operator/blob/d3f4ca856badfe3db7a74415bd4a418d128999b1/pkg/cvo/sync_worker.go#L778

pkg/cvo/status.go
wking: oh, wait, I think we do want `false` here for `ok` / `skipFailure`.  Because we do want to go `Failing=True` on these (that's the explicit as in [rhbz#1951835][1]).  We just don't want them to block the transition out of the installing state.  And that transition is happening in the `cr.Complete()` call discussed [here][2], and not in this `convertErrorToProgressing` call chain.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1951835
[2]: https://github.com/openshift/cluster-version-operator/pull/662#discussion_r732424001

LalatenduMohanty:
It depends on https://github.com/openshift/cluster-version-operator/pull/577

jottofar:
`TestCVO_InitImplicitlyEnabledCaps` still failing unit test even after addition of a `sort` but unrelated to this PR's change.

/override ci/prow/unit

pkg/cvo/status.go
```diff
@@ -367,6 +368,8 @@ func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time,
 	switch uErr.UpdateEffect {
 	case payload.UpdateEffectNone:
 		return uErr.Reason, fmt.Sprintf("waiting on %s", uErr.Name), true
+	case payload.UpdateEffectReport:
```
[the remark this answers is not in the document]
wking: #577 has landed, so marking this thread resolved :)

pkg/cvo/sync_worker.go
```diff
@@ -1184,6 +1191,8 @@ func condenseClusterOperators(errs []error) []error {
 			case payload.UpdateEffectNone:
 			case payload.UpdateEffectFail:
 				updateEffect = payload.UpdateEffectFail
+			case payload.UpdateEffectReport:
+				updateEffect = payload.UpdateEffectReport
```
wking, line 1195: In order of increasing severity:

* `None`
* `Report`
* `FailAfterInterval`
* `Fail`

So we should have a guard here like we already have for `FailAfterInterval` that says "if we have an error like this, and the aggregated `updateEffect` is not already worse, set the aggregated `updateEffect`:

```go
case payload.UpdateEffectReport:
  if updateEffect != payload.UpdateEffectFail && updateEffect != payload.UpdateEffectFailAfterInterval {
    updateEffect = payload.UpdateEffectReport
  }
```

pkg/cvo/status.go
```diff
@@ -443,6 +444,12 @@ func convertErrorToProgressing(history []configv1.UpdateHistory, now time.Time,
...
+		if len(uErr.Names) > 0 {
+			return uErr.Reason, fmt.Sprintf("waiting on these operators: %s", strings.Join(uErr.Names, ", ")), false
+		} else {
+			return uErr.Reason, fmt.Sprintf("waiting on operator %s", uErr.Name), false
```
wking, line 451: we aren't actually waiting in the report case, right?  And if report-level errors are our most serious, we probably completed the update, and want to be `Progressing=False`, with the report-level errors in the `Failing=True` message?

pkg/cvo/sync_worker.go
wking: Hrm, or maybe `None` is more severe than `Report`?  It's hard to rank one effect that blocks reconciliation completion but does not get inserted into `Failing` (`None`) vs. another that does not block reconciliation completion but does get inserted into `Failing` (`Report`).  We might need to bump `condenseClusterOperators` to return `(blocking, report error)` or something.

pkg/cvo/status.go
jottofar: Changed them to `degraded...` since for now that's the only time we use that `effect` and I couldn't come up with a good generic message.

pkg/cvo/sync_worker.go
jottofar: Made `UpdateEffectReport` least severity.

wking (approved):
/lgtm

[Disruption limits][1] are not scaled for our A->B->A testing.

/override e2e-agnostic-upgrade

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/662/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1559977501667627008

Issue #666, opened 2021-09-25
Missing the specification for usage of --serving-key-file if --listen option is not unset.


apoorvajagtap:
While trying to run the CVO locally as per the [doc/dev](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/run-cvo-locally.md#how-to-build-and-run-cvo-locally), the [section](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/run-cvo-locally.md#steps-to-run-cvo-locally) specifies the following command:
`./_output/linux/amd64/cluster-version-operator -v5 start --release-image 4.4.0-rc.4`

But, as the `--listen` is set to "0.0.0.0:9099" by default, the command fails with the following error unless we append `--listen=""` to the command:
`F0926 00:28:13.708624   62174 start.go:24] error: --listen was not set empty, so --serving-cert-file must be set`

The documentation should be updated to either specify the `--serving-cert-file` or at least to unset the listen option `--listen=""`.

PR #667, opened 2021-09-25
Unset listen option while running CVO locally [doc/dev]

apoorvajagtap:
Reference: https://github.com/openshift/cluster-version-operator/issues/666

As I am a newbie and trying to learn the CVO's flow, let me know, in case unsetting the `listen` doesn't make sense. Will be glad to hear some inputs.


jottofar:
Yes I believe it does make sense since we've added the checks on listen address https://github.com/openshift/cluster-version-operator/blob/26a6ca6b281d83cbf81e7c414bfaca7d00d4669b/pkg/start/start.go#L116-L121

apoorvajagtap:
Thanks for confirming @jottofar!

PR #668, opened 2021-10-01
podsecurity: enforce privileged for openshift-cluster-version namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-cluster-version ` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz

### DIFF
--- a/install/0000_00_cluster-version-operator_00_namespace.yaml
+++ b/install/0000_00_cluster-version-operator_00_namespace.yaml
@@ -10,3 +10,6 @@ metadata:
     name: openshift-cluster-version
     openshift.io/run-level: "1"
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged


install/0000_00_cluster-version-operator_00_namespace.yaml
```diff
@@ -10,3 +10,6 @@ metadata:
     name: openshift-cluster-version
     openshift.io/run-level: "1"
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
```
wking, line 15: It's not clear to me why the CVO would need to be `privileged`.  The CVO isn't trying to do anything particularly fancy at the container-security level, and it would be nice if we could fit ourselves into the [`restricted` policy][1].  We do have [two `hostPath` mounts][2] today, but one of those if for mounting TLS certs, and surely there is some way to get at those without requiring all the access `privileged` gets us?  I expect we could also do something similar with volumes for `etc-cvo-updatepayloads` to pass payloads from one pod to future pods.  Is there some discussion somewhere about collaborating on changes needed to move to `restricted`?  At least the CA-mounting thing seem like they'll be common problems.

[1]: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
[2]: https://github.com/openshift/cluster-version-operator/blob/e816c118ac608f131d24b28d617e91d9d5cc34a6/install/0000_00_cluster-version-operator_03_deployment.yaml#L90-L95

s-urbaniak: @wking the concrete violation is as follows:
```
$ jq  '. | select((.objectRef.resource == "deployments") and (.annotations | has("pod-security.kubernetes.io/audit")) and (.objectRef.namespace == "openshift-cluster-version")) | {name: .objectRef.name, resource: .objectRef.resource, podSecurityReason: .annotations["pod-security.kubernetes.io/audit"] }' audit.log | jq -C -s --sort-keys '. | unique'
[
  {
    "name": "cluster-version-operator",
    "podSecurityReason": "host namespaces (hostNetwork=true), hostPath volumes (volumes \"etc-ssl-certs\", \"etc-cvo-updatepayloads\")",
    "resource": "deployments"
  }
]
```

s-urbaniak: so you are declaring
- host networking
- host paths
which violates baseline policy

s-urbaniak: if you can get rid of those you'd at least baseline policy.

s-urbaniak: Note: there is no level between baseline and restricted.

s-urbaniak: >  surely there is some way to get at those without requiring all the access privileged gets us?

unfortnately, with the current Pod Security admission, no.

s-urbaniak:
https://github.com/openshift/enhancements/pull/899 is merged, is there anything I can help to proceed here?

install/0000_00_cluster-version-operator_00_namespace.yaml
wking, line 15: Reading [the KEP][1], it's not clear to me what the benefit of setting `warn` and `enforce` is.  Perhaps it's something that's useful in the short term while [the web-hook is in audit-only mode][2]?  But something we can drop once `enforce` grows some teeth?

[1]: https://github.com/kubernetes/enhancements/tree/3133297720aef516aa07a6553b314de2dd340d72/keps/sig-auth/2579-psp-replacement#api
[2]: https://github.com/openshift/enhancements/pull/899/files#diff-d6c490f9ea419e79a03c81cfe8ff642e110a202bdeeac5574d576dab666717ebR69

PR #669, opened 2021-10-07
Bug 1990635: Fixing the sync issue when desired version and channel changed at the same time

LalatenduMohanty:


pkg/cvo/cvo.go
```diff
@@ -589,9 +589,6 @@ func (optr *Operator) availableUpdatesSync(ctx context.Context, key string) erro
 	if err != nil {
 		return err
 	}
-	if errs := validation.ValidateClusterVersion(config); len(errs) > 0 {
```
wking, line 592: Completely dropping this guard means we will run `syncAvailableUpdates` even in situations like "`upstream` is not a valid URI".  I don't think we want that, because it might blow away an earlier cached result and replace it with emptiness from a failed attempt to hit that invalid `upstream`.  Can you give more details on your motivation for dropping the guard?

LalatenduMohanty: The intention behind using the `ValidateClusterVersion`  in availableUpdatesSync() is to move to the next reconciliation loop if something is wrong with the version or even the upstream URL.  So even if we remove ValidateClusterVersion it does not change that anyway. Also `ValidateClusterVersion` is used in `func (optr *Operator) upgradeableSync()` [1]  which is a better place for it. 
I think when  "upstream is not a valid URI" we should blow away an earlier cached result and replace it with emptiness from a failed attempt to hit that invalid upstream. Because if user is changing to a new upstream URL why would we want to carry the old cached information which is specific to an Upstream URL.

Also both `upgradeableSync` and `availableUpdatesSync` are part of Run method incvo.go and called in two separate go routines [1]


[1] https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/cvo.go#L359-L372

wking (approved):
/lgtm

PR #671, opened 2021-10-07
Bug 2011951: pkg/cvo/upgradeable: Include messages for multiple-reason Upgradeable=False

wking:
Backporting #670 to 4.9.


LalatenduMohanty (approved):
/lgtm

sdodson:
This is not a blocker but it's a downward bound UX improvement that we'd like to get into 4.8 ASAP, adding staff-engineer-approved label.

PR #682, opened 2021-10-27
bootstrap/bootstrap-pod: Pivot from --enable-default-cluster-version to --wait-for-cluster-version

wking:
We've been using `--enable-default-cluster-version` to avoid a cluster-bootstrap vs. cluster-version operator race since d7760cec0f (#238). But openshift/enhancements#922 is considering passing component configuration to the CVO via ClusterVersion that would affect the manifests getting applied.  So this commit pivots to a new option that blocks the whole sync cycle on the existence of a ClusterVersion resource.  That way, we know we have up-to-date data before reconciling any manifests (although it might slow down the bootstrapping process a bit).

### DIFF
--- a/bootstrap/bootstrap-pod.yaml
+++ b/bootstrap/bootstrap-pod.yaml
@@ -14,7 +14,7 @@ spec:
       - "start"
       - "--release-image={{.ReleaseImage}}"
       - "--enable-auto-update=false"
-      - "--enable-default-cluster-version=false"
+      - "--wait-for-cluster-version"
       - "--listen="
       - "--v=5"
       - "--kubeconfig=/etc/kubernetes/kubeconfig"
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -31,7 +31,7 @@ func init() {
 	cmd.PersistentFlags().StringVar(&opts.Kubeconfig, "kubeconfig", opts.Kubeconfig, "Kubeconfig file to access a remote cluster (testing only)")
 	cmd.PersistentFlags().StringVar(&opts.NodeName, "node-name", opts.NodeName, "kubernetes node name CVO is scheduled on.")
 	cmd.PersistentFlags().BoolVar(&opts.EnableAutoUpdate, "enable-auto-update", opts.EnableAutoUpdate, "Enables the autoupdate controller.")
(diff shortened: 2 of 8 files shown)


wking:
I haven't run this multiple times to know how reliable it is, but the presubmit that excercises the new CVO code during install [took 1h12m to install this cluster][1], while the presubmit that used already-landed CVO code during install [took 39 minutes][2].  The difference is almost entirely in the wait-for-bootstrap-complete leg:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/682/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1453472422291312640/artifacts/e2e-agnostic-upgrade/ipi-install-install-stableinitial/artifacts/.openshift_install.log | tail | grep bootstrap
time="2021-10-27T22:08:22Z" level=debug msg="         bootstrap: 5m22s"
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/682/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1453472422211620864/artifacts/e2e-agnostic/ipi-install-install/artifacts/.openshift_install.log | tail | grep bootstrap
time="2021-10-27T22:42:00Z" level=debug msg="         bootstrap: 27m35s"
```

so that's at least consistent with this PR being the reason for the slowdown.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/682/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1453472422211620864#1:build-log.txt%3A33
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/682/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1453472422291312640#1:build-log.txt%3A35

wking:
Thinking this over, the existing [`No ClusterVersion object and defaulting not enabled, waiting for one`][1] guard will keep us idle until we get an install-time ClusterVersion, so no need for this PR.

[1]: https://github.com/openshift/cluster-version-operator/blob/b133d822ea2de7ccf7f9b6ee5530da703f3d0b2c/pkg/cvo/cvo.go#L529-L532

PR #683, opened 2021-10-28
Bug 1822752: pkg/cvo: Separate payload load from payload apply

jottofar:
Previously CVO used a single goroutine `SyncWorker.Start` to both load an update (if available) and to apply the currently loaded release. If after loading an update preconditions are not met (e.g. verification failure) application of the currently loaded release is blocked and therefore any changes that might occur to the currently loaded release are also not applied.

This change remedies that by breaking out the "load update" logic into a new mew method `loadUpdatedPayload`. This new method is invoked from the goroutine `SyncWorker.Update`. Previously when this routine recognized that an update was available it would sync with `SyncWorker.Start`which would then load the update and apply it (if no errors occur). With this change `SyncWorker.Update` will first load the update and only if no errors occur syncs with `SyncWorker.Start` which causes it to start applying the update. Otherwise `SyncWorker.Start` continues applying the currently loaded release.

Also changed the payload updating event and report messaging to accurately reflect what is really being done at each step.

### DIFF
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -242,7 +242,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 			Actual:     configv1.Release{Version: "1.0.0-abc", Image: "image/image:1"},
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.0-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.0-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.0-abc", Image: "image/image:1"},
 			},
@@ -631,7 +631,7 @@ func TestCVO_StartupAndSyncUnverifiedPayload(t *testing.T) {
 			Generation: 1,
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.0-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.0-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.0-abc", Image: "image/image:1"},
 			},
@@ -996,7 +996,7 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 			Generation: 1,
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.0-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.0-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.0-abc", Image: "image/image:1"},
 			},
@@ -1294,7 +1294,7 @@ func TestCVO_UpgradeUnverifiedPayload(t *testing.T) {
 			Generation: 1,
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.1-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.1-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.1-abc", Image: "image/image:1"},
 			},
@@ -1542,7 +1542,7 @@ func TestCVO_UpgradeUnverifiedPayloadRetrieveOnce(t *testing.T) {
 			Generation: 1,
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.1-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.1-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.1-abc", Image: "image/image:1"},
 			},
@@ -1863,7 +1863,7 @@ func TestCVO_UpgradePreconditionFailing(t *testing.T) {
 			Generation: 1,
 			loadPayloadStatus: LoadPayloadStatus{
 				Step:               "RetrievePayload",
-				Message:            "Retrieving payload version=\"1.0.1-abc\" image=\"image/image:1\"",
+				Message:            "Retrieving and verifying payload version=\"1.0.1-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
 				Release:            configv1.Release{Version: "1.0.1-abc", Image: "image/image:1"},
 			},
@@ -1888,42 +1888,11 @@ func TestCVO_UpgradePreconditionFailing(t *testing.T) {
 		t.Fatal(err)
 	}
 	actions = client.Actions()
-	if len(actions) != 3 {
+	if len(actions) != 1 {
 		t.Fatalf("%s", spew.Sdump(actions))
 	}
 	expectGet(t, actions[0], "clusterversions", "", "version")
 	actual := cvs["version"].(*configv1.ClusterVersion)
-	expectUpdateStatus(t, actions[2], "clusterversions", "", &configv1.ClusterVersion{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:            "version",
-			ResourceVersion: "2",
-			Generation:      1,
-		},
-		Spec: configv1.ClusterVersionSpec{
-			ClusterID:     clusterUID,
-			Channel:       "fast",
-			DesiredUpdate: &configv1.Update{Version: desired.Version, Image: desired.Image},
-		},
-		Status: configv1.ClusterVersionStatus{
-			// Prefers the image version over the operator's version (although in general they will remain in sync)
-			ObservedGeneration: 1,
-			Desired:            desired,
-			VersionHash:        "DL-FFQ2Uem8=",
-			History: []configv1.UpdateHistory{
-				{State: configv1.PartialUpdate, Image: "image/image:1", Version: "1.0.1-abc", StartedTime: defaultStartedTime},
-				{State: configv1.CompletedUpdate, Image: "image/image:0", Version: "1.0.0-abc", Verified: true, StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
-			},
-			Conditions: []configv1.ClusterOperatorStatusCondition{
-				{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue, Message: "Done applying 1.0.0-abc"},
-				// cleared failing status and set progressing
-				{Type: ClusterStatusFailing, Status: configv1.ConditionFalse},
-				{Type: configv1.OperatorProgressing, Status: configv1.ConditionTrue, Message: "Working towards 1.0.1-abc"},
-				{Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse},
-				{Type: DesiredReleaseAccepted, Status: configv1.ConditionFalse, Reason: "PreconditionChecks",
-					Message: "Preconditions failed for payload loaded version=\"1.0.1-abc\" image=\"image/image:1\": Precondition \"TestPrecondition SuccessAfter: 3\" failed because of \"CheckFailure\": failing, attempt: 2 will succeed after 3 attempt"},
-			},
-		},
-	})
 
 	// Step 2: Set allowUnverifiedImages to true, trigger a sync and the operator should apply the payload
 	//
@@ -2049,7 +2018,7 @@ func TestCVO_UpgradePreconditionFailing(t *testing.T) {
 	expectUpdateStatus(t, actions[1], "clusterversions", "", &configv1.ClusterVersion{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            "version",
-			ResourceVersion: "4",
+			ResourceVersion: "3",
 			Generation:      1,
 		},
 		Spec: configv1.ClusterVersionSpec{
@@ -2135,10 +2104,11 @@ func TestCVO_UpgradeVerifiedPayload(t *testing.T) {
 	worker := o.configSync.(*SyncWorker)
 	retriever := worker.retriever.(*fakeDirectoryRetriever)
 	retriever.Set(PayloadInfo{}, payloadErr)
+	retriever.Set(PayloadInfo{Directory: "testdata/payloadtest-2", Verified: true}, nil)
 
 	go worker.Start(ctx, 1, o.name, o.cvLister)
 
-	// Step 1: The operator should report that it is blocked on unverified content
+	// Step 1: Simulate a verified payload being retrieved and ensure the operator sets verified
 	//
 	client.ClearActions()
 	err := o.sync(ctx, o.queueKey())
@@ -2150,7 +2120,6 @@ func TestCVO_UpgradeVerifiedPayload(t *testing.T) {
 		t.Fatalf("%s", spew.Sdump(actions))
 	}
 	expectGet(t, actions[0], "clusterversions", "", "version")
-	actual := cvs["version"].(*configv1.ClusterVersion)
 	expectUpdateStatus(t, actions[2], "clusterversions", "", &configv1.ClusterVersion{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            "version",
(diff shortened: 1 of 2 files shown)


vrutkovs:
/hold

Needs rebase I guess:
```
 # github.com/openshift/cluster-version-operator/pkg/cvo [github.com/openshift/cluster-version-operator/pkg/cvo.test]
pkg/cvo/cvo_test.go:304:5: cannot use &fakeSyncRecorder{...} (type *fakeSyncRecorder) as type ConfigSyncWorker in field value:
	*fakeSyncRecorder does not implement ConfigSyncWorker (missing LoadUpdatedPayload method)
pkg/cvo/cvo_test.go:379:5: cannot use &fakeSyncRecorder{...} (type *fakeSyncRecorder) as type ConfigSyncWorker in field value:
	*fakeSyncRecorder does not implement ConfigSyncWorker (missing LoadUpdatedPayload method)
pkg/cvo/cvo_test.go:455:5: cannot use &fakeSyncRecorder{...} (type *fakeSyncRecorder) as type ConfigSyncWorker in field value:
	*fakeSyncRecorder does not implement ConfigSyncWorker (missing LoadUpdatedPayload method)
pkg/cvo/cvo_test.go:2269:21: cannot use &fakeSyncRecorder{...} (type *fakeSyncRecorder) as type ConfigSyncWorker in assignment:
	*fakeSyncRecorder does not implement ConfigSyncWorker (missing LoadUpdatedPayload method)
pkg/cvo/cvo_test.go:2288:30: impossible type assertion:
	*fakeSyncRecorder does not implement ConfigSyncWorker (missing LoadUpdatedPayload method)
FAIL	github.com/openshift/cluster-version-operator/pkg/cvo [build failed] 
```

jottofar:
/retitle WIP: Bug 1822752: pkg/cvo: Separate payload load from payload apply

Yeah, this is not ready

pkg/cvo/cvo.go
```diff
@@ -511,9 +511,9 @@ func handleErr(ctx context.Context, queue workqueue.RateLimitingInterface, err e
 // It returns an error if it could not update the cluster version object.
 func (optr *Operator) sync(ctx context.Context, key string) error {
 	startTime := time.Now()
-	klog.V(4).Infof("Started syncing cluster version %q (%v)", key, startTime)
+	klog.V(4).Infof("sync: Started syncing cluster version %q (%v)", key, startTime)
 	defer func() {
-		klog.V(4).Infof("Finished syncing cluster version %q (%v)", key, time.Since(startTime))
+		klog.V(4).Infof("sync: Finished syncing cluster version %q (%v)", key, time.Since(startTime))
```
wking, line 516: what's the motivation for this `sync:` prefix?

pkg/cvo/sync_worker.go
```diff
@@ -535,128 +674,13 @@ func (w *SyncWorker) Status() *SyncWorkerStatus {
...
-		w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeNormal, "PayloadLoaded", "payload loaded version=%q image=%q", desired.Version, desired.Image)
-		klog.V(4).Infof("Payload loaded from %s with hash %s", desired.Image, payloadUpdate.ManifestHash)
-	}
-
-	return w.apply(ctx, w.payload, work, maxWorkers, reporter)
+// syncOnce applies the current payload to the server returning an error if
+// the update could not be completely applied. The status is updated as we
+// progress. Cancelling the context will abort the execution of syncOnce.
+func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers int, reporter StatusReporter) error {
+	klog.V(4).Infof("syncOnce: %s on generation %d in state %s at attempt %d", work.Desired.Version, work.Generation, work.State, work.Attempt)
```
wking, line 681: if this log line is the only remaining portion of `syncOnce` that's not in `apply`, maybe we should push it down into `apply` and drop `syncOnce`?

pkg/cvo/sync_worker.go
```diff
@@ -200,12 +200,149 @@ func (w *SyncWorker) StatusCh() <-chan SyncWorkerStatus {
...
 
+func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork, reporter StatusReporter, clusterVersion *configv1.ClusterVersion) error {
+	desired := configv1.Release{
+		Version: work.Desired.Version,
+		Image:   work.Desired.Image,
+	}
+	klog.V(4).Infof("syncPayload: %s (force=%t)", versionString(desired), work.Desired.Force)
+
+	if work.Attempt == 0 {
+		payload.InitCOUpdateStartTimes()
```
wking, line 211: I think this is an `apply` thing, and not something that `syncPayload` needs to care about

pkg/cvo/cvo.go
jottofar: I found it made it easier to follow the logging w/o having to depend on the line numbers to see which method is being called. There's a lot of `sync'ing` going on in CVO. But I can take it or leave it.

wking: I'm... not opposed, but I'd feel better if "we could use function names for context" was implemented generically in the logger.  Unfortunately, [klog throws out the program counter return value][1] of its [`Caller`][2] invocation, so there's nothing to feed into [`FuncForPC`][3] or similar to get the function name.

Instead of prefixing `sync` (which, like `cvo.go`, is not all that specific), maybe we can expand the wording to be more clear?  `Started syncing cluster version %q (%v), looking for spec changes and updating status` or some such, so it's clear that this is the ClusterVersion action to communicate with the cluster admin and other higher-level stuff, and not the CVO trying to sync its operands or other lower-level stuff?

[1]: https://github.com/kubernetes/klog/blob/main/klog.go#L599
[2]: https://pkg.go.dev/runtime#Caller
[3]: https://pkg.go.dev/runtime#FuncForPC

jottofar: I dropped it since it's not standard and there's not a generic way to log it. Added more description to the existing log entry and noted that this now also sync's payload..

pkg/cvo/sync_worker.go
```diff
@@ -200,12 +200,143 @@ func (w *SyncWorker) StatusCh() <-chan SyncWorkerStatus {
...
+				Actual:      desired,
+				Verified:    info.Verified,
+			})
+			if err := precondition.Summarize(w.preconditions.RunAll(ctx, precondition.ReleaseContext{DesiredVersion: payloadUpdate.Release.Version}, clusterVersion)); err != nil {
+				if work.Desired.Force {
+					klog.V(4).Infof("Forcing past precondition failures: %s", err)
+					w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeWarning, "PreconditionsForced", "preconditions forced for payload loaded version=%q image=%q failures=%v", desired.Version, desired.Image, err)
+				} else {
+					w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeWarning, "PreconditionsFailed", "preconditions failed for payload loaded version=%q image=%q failures=%v", desired.Version, desired.Image, err)
+					reporter.Report(SyncWorkerStatus{
```
wking, line 296: I'm a bit concerned about how the status consumer will handle two parallel streams of `Report` callers, now that we can have the payload-application loop and this update-acceptance loop going in parallel.  Maybe we need a new property in `SyncWorkerStatus` and then separate `status.condition` types for reporting?  I could see a `DesiredUpdateAccepted` or similar property that is unset when `desiredUpdate` is unset or matches our current target, goes `Unknown` when someone requests a new target and we're forming an opinion, goes `False` if we don't like the target, and goes away if we do like the target.  Or something like that.  I guess we could switch on `Step` without adding a new `SyncWorkerStatus` property, but something like `StepGroup` or similar might be more robust.

pkg/cvo/sync_worker.go
```diff
@@ -99,6 +106,8 @@ type SyncWorkerStatus struct {
 
 	Actual   configv1.Release
 	Verified bool
```
wking, line 108: Probably drop this, since `LoadPayloadStatus.Verified` should cover it.  There shouldn't be any verification in the manifest-reconciliation loop.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1218,18 +1196,25 @@ func TestCVO_UpgradeUnverifiedPayload(t *testing.T) {
...
+	// wait until the new payload is applied
+	count = 0
+	for {
+		var status SyncWorkerStatus
+		select {
+		case status = <-worker.StatusCh():
+		case <-time.After(3 * time.Second):
+			t.Fatalf("never saw expected apply event")
+		}
+		if status.Step == "ApplyResources" {
```
wking, line 1208: you're dropping the `ApplyResources` step.  Makes sense to me, but we'll need some logic change here.

pkg/cvo/sync_worker.go
jottofar: I left it in there and added a comment indicating it is the verification status of `Actual` (current) release. It ends being stored in history https://github.com/openshift/cluster-version-operator/blob/8219252a29a9e244744c87fd96a2829f0d360faf/pkg/cvo/status.go#L198 once a release has been accepted https://github.com/openshift/cluster-version-operator/blob/8219252a29a9e244744c87fd96a2829f0d360faf/pkg/cvo/sync_worker.go#L598.
`LoadPayloadStatus.Verified` is verification status of the release currently being loaded or last loaded so it could be transitory. In fact, I'm not currently using it but it seems like something that should be kept around.

jottofar:
/retitle Bug 1822752: pkg/cvo: Separate payload load from payload apply

pkg/cvo/cvo_scenarios_test.go
wking: `ApplyResources` is completely gone in bc3998ad51474fb; marking resolved.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -2913,6 +2967,55 @@ func verifyAllStatus(t *testing.T, ch <-chan SyncWorkerStatus, items ...SyncWork
...
+		}
+
+		if !reflect.DeepEqual(expect, actual) {
+			t.Fatalf("unexpected status item %d\nExpected: %#v\nActual: %#v", i, expect, actual)
+		}
+	}
+}
+
+// Since timing can be non-deterministic, use this instead of verifyAllStatus when
+// don't know or care how many are done.
```
wking, line 2985: :joy:

pkg/cvo/sync_worker_test.go
```diff
@@ -113,16 +113,16 @@ func Test_statusWrapper_ReportGeneration(t *testing.T) {
 		next     SyncWorkerStatus
 		want     int64
 	}{{
-		previous: SyncWorkerStatus{Generation: 1, Step: "Apply", Done: 10, Total: 100},
-		next:     SyncWorkerStatus{Step: "RetreivePayload"},
-		want:     1,
+		previous: SyncWorkerStatus{Generation: 1, Done: 10, Total: 100},
+		//next:     SyncWorkerStatus{},
```
wking, line 117: nit: if we don't need `next`, probably just drop this line instead of commenting it out.

pkg/cvo/status.go
```diff
@@ -310,14 +316,6 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
 				reason = progressReason
 				message = fmt.Sprintf("Working towards %s: %d of %d done (%.0f%% complete), %s", version,
 					status.Done, status.Total, math.Trunc(float64(fractionComplete*100)), progressMessage)
-			case fractionComplete > 0:
-				message = fmt.Sprintf("Working towards %s: %d of %d done (%.0f%% complete)", version,
-					status.Done, status.Total, math.Trunc(float64(fractionComplete*100)))
```
wking, line 315: We've had the separate positive-fraction cases for `skipFailure` and without since `skipFailure` landed in c2ac20fa17 (#158).  If the idea is to always show `progressMessage`, perhaps we should drop the `&& skipFailure` guard in the previous case?  Either way, it would be useful to explain the motivation for this pivot in the commit message.

pkg/cvo/status.go
```diff
@@ -351,6 +349,26 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
+			Type:               DesiredReleaseAccepted,
+			Status:             configv1.ConditionTrue,
+			Reason:             status.Step,
+			Message:            status.Message,
+			LastTransitionTime: now,
+		})
+	} else if status.Step != "" && status.Failure != nil {
+		resourcemerge.SetOperatorStatusCondition(&config.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:               DesiredReleaseAccepted,
+			Status:             configv1.ConditionFalse,
```
wking, line 364: Is it worth trying to make some distinction between the "we're still working through this, but no serious issues yet, patience" where we'd go `Unknown`, and "last time we hit a fairly serious issue, but we're continually retrying, and maybe that issue will go away and this time we'll succeed" where we'd go `False`?

pkg/cvo/cvo_scenarios_test.go
jottofar: done

pkg/cvo/status.go
jottofar: hmm, must have pulled during testing/POC but it should remain so added back.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -975,13 +1115,16 @@ func TestCVO_StartupAndSyncPreconditionFailing(t *testing.T) {
 			},
 			VersionHash: "DL-FFQ2Uem8=",
 			History: []configv1.UpdateHistory{
+				//{State: configv1.PartialUpdate, Image: "image/image:1", Version: "1.0.0-abc", StartedTime: defaultStartedTime},
```
wking, line 1118: nit, drop this line?

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1471,59 +1633,47 @@ func TestCVO_UpgradeUnverifiedPayloadRetrieveOnce(t *testing.T) {
...
 		SyncWorkerStatus{
 			Total:       3,
-			Step:        "ApplyResources",
 			VersionHash: "DL-FFQ2Uem8=",
 			Actual: configv1.Release{
 				Version: "1.0.1-abc",
 				Image:   "image/image:1",
 				URL:     configv1.URL("https://example.com/v1.0.1-abc"),
 			},
+			//LastProgress: time.Unix(1, 0),
```
wking, line 1663: nit: another commented-out line to drop?

pkg/cvo/sync_worker.go
```diff
@@ -393,8 +557,19 @@ type statusWrapper struct {
 	previousStatus *SyncWorkerStatus
 }
 
+func (w *statusWrapper) ValidPayloadStatus(release configv1.Release) bool {
+	return w.previousStatus.loadPayloadStatus.Release.Image == release.Image
```
wking, line 561: I expect we want to use `return equalDigest(w.previousStatus.loadPayloadStatus.Release.Image, release.Image)` here, to avoid invalidating our status if someone moves `registryA@sha256:123...` to `registryB@sha256:123...`, like we sometimes see in QE.

wking:
Poking at [update CI][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/683/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1489735091872600064/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/clusterversion.json | jq -r '.items[].status.conditions[] | .lastTransitionTime + " " + .type + "=" + .status + " " + .reason + ": " + .message'
2022-02-04T23:45:56Z Available=True : Done applying 4.11.0-0.ci.test-2022-02-04-230202-ci-op-lb302rqg-latest
2022-02-05T00:02:13Z Failing=False : 
2022-02-05T00:42:34Z Progressing=False : Cluster version is 4.11.0-0.ci.test-2022-02-04-230202-ci-op-lb302rqg-latest
2022-02-04T23:22:03Z RetrievedUpdates=False NoChannel: The update channel has not been configured.
2022-02-04T23:48:28Z ReleaseAccepted=True PayloadLoaded: Payload loaded version="4.11.0-0.ci.test-2022-02-04-230202-ci-op-lb302rqg-latest" image="registry.build03.ci.openshift.org/ci-op-lb302rqg/release@sha256:5f34f9aecdca85102b837565864d3c48b33956b90440c57ce0eb3c0c0e47ab33"
```

So there's our `ReleaseAccepted=True` with a nice message.  It's a bit after we actually accepted the release, because it's the incoming CVO in this pre-patch -> post-patch CI job:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/683/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1489735091872600064/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/events.json | jq -r '[.items[] | select(.metadata.namespace == "openshift-cluster-version")] | sort_by(.metadata.creationTimestamp)[] | .metadata.creationTimestamp + " " + (.count | tostring) + " " + .involvedObject.name + " " + .reason + ": " + .message' | grep 'Leader\|Scheduled'
2022-02-04T23:21:47Z 1 version LeaderElection: ci-op-lb302rqg-7ee27-mdrjg-bootstrap_0bce9a2c-e462-46f0-90e5-bec03e81d7bb became leader
2022-02-04T23:26:16Z null cluster-version-operator-cfcfb8d8d-hjwn2 Scheduled: Successfully assigned openshift-cluster-version/cluster-version-operator-cfcfb8d8d-hjwn2 to ci-op-lb302rqg-7ee27-mdrjg-master-0
2022-02-04T23:36:55Z 1 version LeaderElection: ci-op-lb302rqg-7ee27-mdrjg-master-0_cbe9fc42-b785-4420-b98a-b6ce88215312 became leader
2022-02-04T23:48:23Z null cluster-version-operator-67c489c976-rbkjz Scheduled: Successfully assigned openshift-cluster-version/cluster-version-operator-67c489c976-rbkjz to ci-op-lb302rqg-7ee27-mdrjg-master-1
2022-02-04T23:48:27Z 1 version LeaderElection: ci-op-lb302rqg-7ee27-mdrjg-master-1_a55d499a-0ec2-4b0f-b52f-18d7108339fe became leader
2022-02-05T00:30:26Z null cluster-version-operator-67c489c976-wnc9r Scheduled: Successfully assigned openshift-cluster-version/cluster-version-operator-67c489c976-wnc9r to ci-op-lb302rqg-7ee27-mdrjg-master-0
2022-02-05T00:30:28Z 1 version LeaderElection: ci-op-lb302rqg-7ee27-mdrjg-master-0_120a9366-3aef-40cc-be47-bba1775fd19b became leader
2022-02-05T00:34:30Z null cluster-version-operator-67c489c976-sf6vm Scheduled: Successfully assigned openshift-cluster-version/cluster-version-operator-67c489c976-sf6vm to ci-op-lb302rqg-7ee27-mdrjg-master-1
2022-02-05T00:34:33Z 1 version LeaderElection: ci-op-lb302rqg-7ee27-mdrjg-master-1_47b47a2d-bdce-4f4d-8c4a-8cfa887123fd became leader
```

I don't think we're too worried about a few minutes offset as folks transition into this new code.  Anyhow, looks good to me, although it will also be good to see logs from the CVO as we update out of the patched release into other places.

/lgtm
/hold

Holding in case we want to go address the three very minor nits I mentioned above.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/683/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1489735091872600064

wking:
/bugzilla refresh

[availability flakes][1] are orthogonal.

I'm not sure we'll backport this, but if we do, I don't mind leaving those off the backport, and we can always attach additional commits to backports if we change our minds there.  This PR has been open long enough, that I don't think it needs another weekend and another reroll ;)

/hold

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/683/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1489735091872600064

wking:
oops, I meant:

/hold cancel

wking:
Nothing remotely close to this code has landed since the previous round of CI:

jottofar:
> don't think we're too worried about a few minutes offset as folks transition into this new code. Anyhow, looks good to me, although it will also be good to see logs from the CVO as we update out of the patched release into other places

From loki logs, we loaded new payload at 23:48:21 then that CVO instance shutdown as expected and new CVO running `4.11.0-0.ci.test-2022-02-04-230202-ci-op-lb302rqg-latest` came up. The ReleaseAccepted is really when the "current" CVO accepted the release but even so, it only differs by seconds.

`2022-02-04 23:48:21 | I0204 23:48:21.604372       1 upgradeable.go:122] Cluster current version=4.11.0-0.ci.test-2022-02-04-225838-ci-op-lb302rqg-initial
...
2022-02-04 23:48:21 | I0204 23:48:21.604546       1 event.go:285] Event(v1.ObjectReference{Kind:"ClusterVersion", Namespace:"openshift-cluster-version", Name:"version", UID:"", APIVersion:"config.openshift.io/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'PreconditionsPassed' preconditions passed for payload loaded version="4.11.0-0.ci.test-2022-02-04-230202-ci-op-lb302rqg-latest"...
`

pkg/cvo/cvo_scenarios_test.go
jottofar: Addressed in https://github.com/openshift/cluster-version-operator/pull/745

pkg/cvo/cvo_scenarios_test.go
jottofar: https://github.com/openshift/cluster-version-operator/pull/745

pkg/cvo/sync_worker.go
jottofar: https://github.com/openshift/cluster-version-operator/pull/745

PR #684, opened 2021-11-03
Bug 1878925: pkg/cvo/updatepayload: Drop history from target pullspec lookup

wking:
The logic I'm removing is a convenience from 558ae021bd (#63).  But the current consensus is that:

* Rollbacks are risky enough that we don't want to make them that convenient.
* Rollbacks have never been supported, so removing the convenient pullspec lookup is a backwards-compatible-enough change.

This also brings the ClusterVersion operator's approach into line with how `oc adm upgrade --to ...` has always worked; openshift/oc#566.


vrutkovs (approved):
/lgtm

Issue #688, opened 2021-11-11
getOverrideForManifest does not check manifest.GVK.Group


mbarnes:
We have the following override in our `ClusterVersion`:
```
    - group: imageregistry.operator.openshift.io
      kind: Config
      name: cluster
      namespace: ""
      unmanaged: true
```

This is causing cluster provisioning to fail, because when the operator encounters this manifest...

**0000_30_config-operator_01_operator.cr.yaml**
```
apiVersion: operator.openshift.io/v1
kind: Config
metadata:
  name: cluster
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
    release.openshift.io/create-only: "true"
spec:
  managementState: Managed
```

... the [getOverrideForManifest function](https://github.com/openshift/cluster-version-operator/blob/4c3a08036da8a96175b7c0445de83b58d0ea5515/pkg/cvo/sync_worker.go#L1060-L1071) is improperly matching it to the above `imageregistry.operator.openshift.io` override because it disregards the Group in its comparison (`imageregistry.operator.openshift.io` != `operator.openshift.io`).

As a result, the [cluster-config-operator](https://github.com/openshift/cluster-config-operator) has no custom resource to act on and it blocks the **cluster-version-operator** from ever completing:
```
$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version             False       True          3h18m   Working towards 4.9.7: 725 of 735 done (98% complete), waiting on config-operator
```

PR #689, opened 2021-11-11
Bug 2022509: cvo: Compare manifest group in getOverrideForManifest

mbarnes:
Fixes https://github.com/openshift/cluster-version-operator/issues/688

`getOverrideForManifest` needs unit tests, but I can follow up separately since this is urgent for ARO.


wking (approved):
/retitle Bug 2022509: cvo: Compare manifest group in getOverrideForManifest
/lgtm

PR #696, opened 2021-11-23
Bug 2025955: Removing the extra indentation

LalatenduMohanty:
From admingate_configmap.yaml

### DIFF
--- a/install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
+++ b/install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 data:
-  ack-4.8-kube-1.22-api-removals-in-4.9: |
+  ack-4.8-kube-1.22-api-removals-in-4.9: >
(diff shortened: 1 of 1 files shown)


install/0000_00_cluster-version-operator_01_admingate_configmap.yaml
```diff
@@ -2,8 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 data:
   ack-4.8-kube-1.22-api-removals-in-4.9: |
```
wking, line 4: From [the YAML spec][1]:

> Scalar content can be written in block notation, using a literal style (indicated by “|”) where all line breaks are significant. Alternatively, they can be written with the folded style (denoted by “>”) where each line break is folded to a space unless it ends an empty or a more-indented line.

So you could leave the following two lines alone, and move to `...: >` here, to get that internal line break folded into a space. 

[1]: https://yaml.org/spec/1.2.2/#23-scalars

LalatenduMohanty: I like the suggestion. Fixed it

wking (approved):
/lgtm

PR #698, opened 2021-12-01
Bug 2028217: lib/resourcemerge/apps: Default Deployment replicas to one

wking:
@tdosek and @vrutkovs noticed that, when a Deployment manifest leaves `replicas` unset, the CVO ignores the property.  This means that cluster admins can scale those Deployments up or, worse, down to 0, and the CVO will happily continue on without stomping them.  Auditing 4.9.10:

```console
$ oc adm release extract --to manifests quay.io/openshift-release-dev/ocp-release:4.9.10-x86_64
Extracted release payload from digest sha256:e1853d68d8ff093ec353ca7078b6b6df1533729688bb016b8208263ee7423f66 created at 2021-12-01T09:19:24Z
$ for F in $(grep -rl 'kind: Deployment' manifests); do yaml2json < "${F}" | jq -r '.[] | select(.kind == "Deployment" and .spec.replicas == null).metadata | .namespace + " " + .name'; done | sort | uniq
openshift-cluster-machine-approver machine-approver
openshift-insights insights-operator
openshift-network-operator network-operator
```

Those are all important operators, and I'm fairly confident that none of their maintainers expect "cluster admin scales them down to 0" to be a supported UX.  We should have the CVO default Deployment `replicas` to 1 ([the type's default][1]), so admins who decide they don't want a network operator pod, etc., have to use some more explicit, alarming API to remove those pods (e.g. setting `spec.overrides` in the ClusterVersion object to assume control of the resource themselves).

Also replace `int32Pointer` with `pointer.Int32`.  Not clear to me why 80dc893514 (#256) had gone with a
local function.

[1]: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec

### DIFF
--- a/lib/resourcemerge/apps.go
+++ b/lib/resourcemerge/apps.go
@@ -3,6 +3,7 @@ package resourcemerge
 import (
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/utils/pointer"
 )
 
 // EnsureDeployment ensures that the existing matches the required.
@@ -10,7 +11,8 @@ import (
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas {
+	ensureReplicasDefault(&required)
+	if existing.Spec.Replicas == nil || *required.Spec.Replicas != *existing.Spec.Replicas {
 		*modified = true
 		existing.Spec.Replicas = required.Spec.Replicas
 	}
@@ -30,6 +32,12 @@ func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required apps
 	}
 
 	ensurePodTemplateSpec(modified, &existing.Spec.Template, required.Spec.Template)
+}
+
+func ensureReplicasDefault(required *appsv1.Deployment) {
+	if required.Spec.Replicas == nil {
+		required.Spec.Replicas = pointer.Int32(1)
+	}
 }
 
 // EnsureDaemonSet ensures that the existing matches the required.
--- a/lib/resourcemerge/apps_test.go
+++ b/lib/resourcemerge/apps_test.go
@@ -23,29 +23,39 @@ func TestEnsureDeployment(t *testing.T) {
 			name: "different replica count",
 			existing: appsv1.Deployment{
 				Spec: appsv1.DeploymentSpec{
-					Replicas: int32Pointer(2)}},
+					Replicas: pointer.Int32(2)}},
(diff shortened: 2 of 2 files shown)


lib/resourcemerge/apps.go
```diff
@@ -3,14 +3,16 @@ package resourcemerge
...
 )
 
 // EnsureDeployment ensures that the existing matches the required.
 // modified is set to true when existing had to be updated with required.
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas {
+	ensureReplicasDefault(&required)
+	if existing.Spec.Replicas == nil || (required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas) {
```
jottofar, line 15: > required.Spec.Replicas != nil

All these can be dropped now with `ensureReplicasDefault`

wking: Thanks; I've dropped the `required.Spec.Replicas != nil` guard with f4b194af -> dd827544.

lib/resourcemerge/apps.go
```diff
@@ -3,14 +3,16 @@ package resourcemerge
...
 )
 
 // EnsureDeployment ensures that the existing matches the required.
 // modified is set to true when existing had to be updated with required.
 func EnsureDeployment(modified *bool, existing *appsv1.Deployment, required appsv1.Deployment) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
 
-	if required.Spec.Replicas != nil && *required.Spec.Replicas != *existing.Spec.Replicas {
+	ensureReplicasDefault(&required)
+	if existing.Spec.Replicas == nil || *required.Spec.Replicas != *existing.Spec.Replicas {
```
vrutkovs, line 15: You can have nil replicas in `required`, but I'm pretty sure API will always return at least `0` in `existing`. We probably don't need to update this line then

vrutkovs: Wait, nevermind, this would leave a risky `*existing.Spec.Replicas`, so it LGTM

vrutkovs (approved):
/lgtm

lib/resourcemerge/apps.go
wking: the `nil` guard for `existing` is mostly because we don't bother populating this in the unit-test mocks.  I would be very surprised if an API-server actually returned a resource with nil `replicas`.

wking:
Ingress and etcd were very sad in [the update job][1], but that's orthogonal to this commit.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/698/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1466420547935014912

PR #706, opened 2021-12-07
pkg/start: Log and continue when we fail to retrieve the feature gate

wking:
To make it easier to differentiate between "cluster version operator is confused about the cluster's tech-preview-ness", "cluster-version operator has broken manifest exclusion logic", and "manifest is mis-setting an annotation".

WIP until we remove the `pkg/payload/payload.go` logging, which I'm using to help debug openshift/cluster-capi-operator#20.

### DIFF
--- a/pkg/featurechangestopper/techpreviewchangestopper.go
+++ b/pkg/featurechangestopper/techpreviewchangestopper.go
@@ -94,7 +94,7 @@ func (c *TechPreviewChangeStopper) Run(ctx context.Context, shutdownFn context.C
 	}()
 	c.shutdownFn = shutdownFn
 
-	klog.Infof("Starting stop-on-techpreview-change controller")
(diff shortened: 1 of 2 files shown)


wking:
27c7c4ae -> 90b14542 drops the debug logging, and adds a new commit that turns 18dd189635 (#694)'s hard-fail on initial-feature-gate-fetch into a log-and-continue, to make bootstrapping a bit quicker with the CVO working on GA manifests while the feature-gate hosting business is coming up.

jottofar:
Is this "log and continue" change valid because the error usually occurs during start-up? So ignoring it just means we continue with `includeTechPreview = false` but eventually the `Get` succeeds and we'll restart if the value is actually `true`?

wking:
[sandbox and connectivity issues][1] are unrelated:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/706/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1468126571909877760

PR #708, opened 2021-12-08
pkg/payload/precondition: File shuffling, drop ClusterVersion argument, etc.

wking:
Assorted tidying in preparation for populating `status.history[].acceptedRisks`.  No semantic changes here.  Details in the commit messages.


pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -51,7 +51,7 @@ func ClusterVersionOverridesCondition(cv *configv1.ClusterVersion) *configv1.Clu
 // Run runs the Upgradeable precondition.
 // If the feature gate `key` is not found, or the api for clusterversion doesn't exist, this check is inert and always returns nil error.
 // Otherwise, if Upgradeable condition is set to false in the object, it returns an PreconditionError when possible.
-func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.ReleaseContext, clusterVersion *configv1.ClusterVersion) error {
+func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.ReleaseContext) error {
```
jottofar, line 51: Test this carefully. The reason I originally added `clusterVersion`, pulled in `Start`, was that it was always set to the true, "current" (Completed) version, whereas `cv`, pulled during precondition evaluation, was showing up as the "desired" target version. Not to say things have not changed or that my initial testing that lead down that path was not flawed somehow but worth a word of caution.

pkg/cvo/sync_worker.go
```diff
@@ -648,6 +646,8 @@ func (w *SyncWorker) syncOnce(ctx context.Context, work *SyncWork, maxWorkers in
 						Verified:    info.Verified,
 					})
 					return err
+				} else {
+					w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeWarning, "PreconditionWarn", "precondition warning for payload loaded version=%q image=%q: %v", desired.Version, desired.Image, err)
```
jottofar, line 650: Not sure if/how the semantics of `force` changes once we have warnings but I think this somewhat obfuscates the use of `force`. I'm thinking we should still say that we're "forcing past" everything.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -91,7 +88,7 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 	// if no cluster overrides have been set
 	if currentMinor == desiredMinor {
 		klog.V(4).Infof("Precondition %q passed: minor from the current %s matches minor from the target %s (both %s).", pf.Name(), currentVersion, releaseContext.DesiredVersion, currentMinor)
-		if condition := ClusterVersionOverridesCondition(clusterVersion); condition != nil {
+		if condition := ClusterVersionOverridesCondition(cv); condition != nil {
```
wking, line 91: Shifting [this][1] thread over.  @jottofar, I think this is the only call-site that was consuming the `clusterVersion` argument I'm removing in this PR.  We added it in b72e843974 (#364).  But [your concerns][1] seem to be about `history`, and `ClusterVersionOverridesCondition` is just looking at `spec.overrides`.  Does that address your concerns?

[1]: https://github.com/openshift/cluster-version-operator/pull/708#discussion_r766168966

pkg/payload/precondition/clusterversion/upgradeable.go
wking: moving this conversation from the function signature [to the one consumer][1], so it's more clear how the value is being used (or was being used, before my proposed change in this PR).

[1]: https://github.com/openshift/cluster-version-operator/pull/708#discussion_r766334993

pkg/cvo/sync_worker.go
wking: I've pushed ff739b2e -> 077d0794 to add a `Forced through blocking failures: ` prefix to the error message when we force through a blocking error.  Does that seem like it would be a clear enough signal?

pkg/payload/precondition/clusterversion/upgradeable.go
jottofar: NM, you're right, I'm confusing it with `currentVersion`. Not sure why I didn't originally use `cv` and instead added that param other than it being an oversight.

wking:
[sandboxes][1] are unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/708/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1469320696386555904

PR #416, opened 2020-07-23
WIP: vendor/github.com/openshift/build-machinery-go: Pull in generic helpers

wking:
Generated by writing the `Makefile` and `pkg/dependencymagnet/doc.go` changes by hand, and then running:

```console
$ go mod tidy
go: downloading github.com/openshift/build-machinery-go v0.0.0-20200211121458-5e3d6e570160
$ go mod vendor
```

with:

```console
$ go version
go version go1.14.4 linux/amd64
```

WIP because:

```console
$ make help
Makefile:12: warning: overriding recipe for target `build'
vendor/github.com/openshift/build-machinery-go/make/targets/golang/build.mk:13: warning: ignoring old recipe for target `build'
The following make targets are available:
all
build
...
images
Makefile
Makefile:12
test
test-unit
...
```

So we need to figure out how to unify our build script with this build-machinery tooling, and I want to understand what the `Makefile` and `Makefile:12` targets are about.

PR #710, opened 2021-12-10
vendor: Bump library-go

wking:
Generated with:

```console
$ go get github.com/openshift/library-go
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

using:

```console
$ go version
go version go1.17.3 linux/amd64
```

I was originally planning on using this to pull in `LeaderElectionSNOConfig` from openshift/library-go#1242, but due to the size of the diff, I figured I might as well file the vendor bump as a stand-alone PR.


wking:
[Hmm][1]:

```
panic: cannot handle unexported field at {*payload.Update}.Manifests[2].id:
```

From our `diff.ObjectReflectDiff` call in `pkg/payload/payload_test.go` vs. [this][2].  I guess we should find another differ...

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/710/pull-ci-openshift-cluster-version-operator-master-unit/1469215832138584064#1:build-log.txt%3A36
[2]: https://github.com/openshift/library-go/pull/1120/files#diff-c2df3d8f17e8cbab0d6ea3d15e4b1588d40d14f32b4066c4a144f7ae9c2e4dd9R48

vrutkovs:
kube [recommends](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/diff/diff.go#L70) using `github.com/google/go-cmp/cmp.Diff` instead

damemi:
Hi, just checking what is the status of this?

vrutkovs:
imo this needs to be closed in favor of #716, which implements a new comparison

wking:
I've just pushed b1a211a3, bringing the `github.com/google/go-cmp/cmp.Diff` bit over from #416, but leaving the `pkg/start` pivot to pull lease defaults out for now.  This PR is just the library-go vendor bump, and the minimal changes we need locally to work with that bump.

wking:
#716 picked up my commits and builds on top, so no need for this separate PR.

/close

PR #716, opened 2021-12-16
Bug 2034484: Library go bump

eggfoobar:
Bumping the library-go dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242) and performance improvements in this [PR](https://github.com/openshift/library-go/pull/1235).

Changes:
  - updated library-go to latest
  - added logic to make use of leader election configs from library-go


jottofar:
This appears to be a dup (at least partially) of https://github.com/openshift/cluster-version-operator/pull/710.
/hold
/cc @wking

qJkee:
@jottofar 
This PR brings the latest changes of library-go with OCPVE team features, while #710 is the old one

vrutkovs:
/approve
/lgtm
/retest

Trevor, please pull the hold if this looks good to you as well

wking:
f81233ddc74d73e7c64c46f76230f4d601ae6088 touches `go.mod` and `go.sum`, but doesn't seem to change any vendored code.  Can we drop that commit or squash it into 577e121fae1a5306a7f0136756eecef6372cce56 or something?

wking:
edc3628b39ace596c2b01e03a6264e04c3fd9e81 looks like it could get squashed into b92dee32 as well.

wking:
Ok, I think I've wrapped my head around this, and the changes break down like:

1. Bump library-go, as in #710.
2. Fix `Handler` -> `ProbeHandler`, as in #710.
3. Move from `ObjectReflectDiff` to `github.com/google/go-cmp/cmp.Diff`, which was the bit I was missing in #710.  I've just updated #710 with b859c467b7dcbe3326189e440df6cee31ae182a6 to bring over this part.
4. The `pkg/start` changes, which is the point of the single-node effort.

I think those should be separate commits, and my just-updated #710 covers the first three points as a no-functional-change vendor bump.  That would leave this pull request for the final `pkg/start` changes.  Or feel free to generate your own commits for all of those steps in this PR, if you'd rather not use the ones from #710.  But I'd rather not have `vendor/` changes mixed into the same commit that lands the `pkg/start` changes, because then it's harder to find the "a human carefully crafted $CHANGES because $REASONS`" portion inside the "and then the robots bumped all of these `vendor/` files too" noise ;).

eggfoobar:
Thanks for the break down @wking , I'm all for making things easier to reason about. So if we want to merge in #710 I can rebase off of that and we can have this be a much simpler PR. I did pull in your changes into this PR, just incase, if we want to take that route then this PR is ready to go.

pkg/start/start.go
```diff
@@ -417,6 +417,27 @@ func useProtobuf(config *rest.Config) {
...
+	// https://github.com/openshift/enhancements/pull/832/files#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fR183
+	// see rhbz#1986477 for more detail
+	defaultLeaderElection := libgoleaderelection.LeaderElectionDefaulting(
+		configv1.LeaderElection{},
+		"", "",
+	)
+
+	if infra, err := clusterstatus.GetClusterInfraStatus(ctx, restcfg); err == nil && infra != nil {
+		if infra.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
+			return libgoleaderelection.LeaderElectionSNOConfig(defaultLeaderElection)
```
wking, line 432: We're not all that particular about our lease settings, and we do have reasonably solid graceful shutdown logic where we release the leases when termination gives us enough time to do that.  So I'm personally fine simplifying all of this and just using `libgoleaderelection.LeaderElectionSNOConfig` all the time, regardless of the Infrastructure configuration.  Thoughts?

wking (approved):
/hold cancel
/lgtm

[etcd leader election and PDB issues][1] are unrelated:

[`KubePersistentVolumeErrors`][2] are unrelated:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/716/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1479378236717666304
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/716/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1479378236633780224

PR #717, opened 2021-12-17
WIP: pkg/cvo/availableupdates: logging for RecentlyChanged guard

wking:
Trying to figure out why this doesn't seem to be tripping.


wking:
/close

#718 fixes some default comparisons, and adds similar detailed logging.

PR #718, opened 2021-12-17
Bug 2033745: pkg/cvo/availableupdates: Acount for default upstream in recent-change throttling

wking:
Since openshift/installer#4112, clusters born in 4.7 or later default to not having an explicit spec.upstream, so they rely on the CVO's internal default.  However, in that case, `u.Upstream` will be an empty string, while by this point in `syncAvailableUpdates`, `upstream` will have been set to the default value.  This commit splits up the "do we really need to check?" logic into a number of distinct cases, and gives them more specific logging, to make it easier to understand and confirm the desired behavior:

a. If we have no cached data, we need to pull a graph.
b. If it's been over `minimumUpdateCheckInterval` since our last check,  we need to pull a graph.  Even if nothing has changed on our side, our data is sufficiently stale to need a refresh.
c. If the `channel` has changed, we have different interests, and we need to pull a graph to hear what the `upstream` recommends for this new set of interests.
d. If the `upstream` hasn't changed, because:
   i. The current `upstream` (explicitly or by default) matches the old explicit `upstream`, or
   ii. The current `upstream` (explicitly or by default) matches the default, and the old `upstream` was unset.
then everything's the same on our side, and our cached graph is recent, so we don't need to do anything.
e. Otherwise, the `upstream` has changed, and we need to pull a graph to see what our new guide has to suggest.

Cases for upstream:

* A -> A: Handled by d.i.
* A -> B: Handled by e.
* A -> unset (defaulted) or default: Handled by e.
* Unset or default -> A: Handled by e.
* Default -> default: Handled in d.i.
* Default -> unset (defaulted): Handled in d.i.
* Unset -> default: Handled by d.ii, new in this commit, previously resulted in an excessive pull.
* Unset -> unset (defaulted): Handled by d.ii, new in this commit, previously resulted in an excessive pull.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -40,9 +40,17 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 
 	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
 	u := optr.getAvailableUpdates()
-	if u != nil && u.Upstream == upstream && u.Channel == channel && u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
-		klog.V(4).Infof("Available updates were recently retrieved, will try later.")
+	if u == nil {
+		klog.V(4).Info("First attempt to retrieve available updates")
+	} else if !u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
+		klog.V(4).Infof("Retrieving available updates again, because more than %s has elapsed since %s", optr.minimumUpdateCheckInterval, u.LastAttempt)
+	} else if channel != u.Channel {
+		klog.V(4).Infof("Retrieving available updates again, because the channel has changed from %q to %q", u.Channel, channel)
+	} else if upstream == u.Upstream || (upstream == optr.defaultUpstreamServer && u.Upstream == "") {
+		klog.V(4).Infof("Available updates were recently retrieved, with less than %s elapsed since %s, will try later.", optr.minimumUpdateCheckInterval, u.LastAttempt)
 		return nil
+	} else {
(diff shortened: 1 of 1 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -40,9 +40,17 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 
 	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
 	u := optr.getAvailableUpdates()
-	if u != nil && u.Upstream == upstream && u.Channel == channel && u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
-		klog.V(4).Infof("Available updates were recently retrieved, will try later.")
+	if u == nil {
+		klog.V(4).Info("First attempt to retrieve available updates")
```
LalatenduMohanty, line 44: Why not add upstream url for logging?

LalatenduMohanty (changes requested):

LalatenduMohanty (approved):
/lgtm

pkg/cvo/availableupdates.go
wking: We log that later, when we actually make the request:

```console
$ git --no-pager grep 'to request updates from' pkg/cincinnati
pkg/cincinnati/cincinnati.go:                   klog.V(5).Infof("Using a root CA pool with 0 root CA subjects to request updates from %s", uri)
pkg/cincinnati/cincinnati.go:                   klog.V(5).Infof("Using a root CA pool with %n root CA subjects to request updates from %s", len(c.transport.TLSClientConfig.RootCAs.Subjects()), uri)
pkg/cincinnati/cincinnati.go:                   klog.V(5).Infof("Using proxy %s to request updates from %s", proxy.Host, uri)
```

In between this line and that, we can still bail out for reasons like "channel is empty"

PR #725, opened 2022-01-11
Bug 2039119: WIP: lib/resourceapply: Avoid Service hotlooping on FIXME

wking:
From [rhbz#2039119][1]:

```console
$ grep -o 'Updating .*due to diff' cvo.log | sort | uniq -c
     18 Updating CronJob openshift-operator-lifecycle-manager/collect-profiles due to diff
     12 Updating Service openshift-monitoring/cluster-monitoring-operator due to diff
```

Adding some debug logging for now, but I'll update once I understand the source of the hotlooping.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2039119#c0

### DIFF
--- a/lib/resourceapply/core.go
+++ b/lib/resourceapply/core.go
@@ -62,16 +62,21 @@ func ApplyServicev1(ctx context.Context, client coreclientv1.ServicesGetter, req
 	if IsCreateOnly(required) {
 		return nil, false, nil
 	}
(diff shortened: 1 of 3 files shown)


wking:
I'm pretty sure this is [monitoring asking us to manage the service and then continuing to try to manage the service directly][1].  If so, nothing to fix on the CVO side.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2039119#c1

PR #726, opened 2022-01-12
Bug 2027342: Add notification for certificate changes

jottofar:
Add a file watcher to the metrics server using the fsnotify package, to watch for changes to the TLS certificate for the metrics endpoint. Prior to this change when changes occurred to the metrics certificate CVO would have to be restarted to pick up the new certificate. With this change the HTTP server is shutdown and then recreated with the new TLS configuration.

Much of the code is reuse from [openshift/cluster-node-tuning-operator#297](https://github.com/openshift/cluster-node-tuning-operator/pull/297).

I tested this by simply running `oc delete secret/signing-key -n openshift-service-ca` as suggested in the bug.

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -1,11 +1,18 @@
 package cvo
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
 	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -20,6 +27,9 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	"github.com/openshift/cluster-version-operator/pkg/internal"
+	"github.com/openshift/library-go/pkg/crypto"
+
+	"gopkg.in/fsnotify.v1"
 )
 
 // RegisterMetrics initializes metrics and registers them with the
@@ -105,50 +115,111 @@ type asyncResult struct {
 	error error
 }
 
-// RunMetrics launches an server bound to listenAddress serving
-// Prometheus metrics at /metrics over HTTPS.  Continues serving
-// until runContext.Done() and then attempts a clean shutdown
-// limited by shutdownContext.Done().  Assumes runContext.Done()
-// occurs before or simultaneously with shutdownContext.Done().
-func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress string, tlsConfig *tls.Config) error {
-	if tlsConfig == nil {
-		return errors.New("TLS configuration is required to serve metrics")
-	}
-
+func createHttpServer() *http.Server {
 	handler := http.NewServeMux()
 	handler.Handle("/metrics", promhttp.Handler())
 	server := &http.Server{
 		Handler: handler,
 	}
+	return server
+}
 
-	tcpListener, err := net.Listen("tcp", listenAddress)
-	if err != nil {
-		return err
+func shutdownHttpServer(parentCtx context.Context, svr *http.Server) {
+	ctx, cancel := context.WithTimeout(parentCtx, 5*time.Second)
+	defer cancel()
+	klog.Info("Shutting down metrics server so it can be recreated with updated TLS configuration.")
+	if err := svr.Shutdown(ctx); err != nil {
+		klog.Errorf("Failed to gracefully shut down metrics server during restart: %v", err)
 	}
+}
+
+func startListening(svr *http.Server, tlsConfig *tls.Config, lAddr string, resultChannel chan asyncResult) {
+	tcpListener, err := net.Listen("tcp", lAddr)
+	if err != nil {
+		resultChannel <- asyncResult{name: "HTTPS server", error: err}
+	}
+	tlsListener := tls.NewListener(tcpListener, tlsConfig)
+	klog.Infof("Metrics port listening for HTTPS on %v", lAddr)
+	err = svr.Serve(tlsListener)
+	resultChannel <- asyncResult{name: "HTTPS server", error: err}
+}
+
+func handleServerResult(result asyncResult, lastLoopError error) error {
+	lastError := lastLoopError
+
+	// API states "Serve always returns non-nil error" but check in case API changes.
+	if result.error == nil || result.error == http.ErrServerClosed {
+		klog.Infof("Collected metrics %s goroutine.", result.name)
+	} else {
+		klog.Errorf("Collected metrics %s goroutine: %v", result.name, result.error)
+		lastError = result.error
+	}
+	return lastError
+}
+
+// RunMetrics launches a server bound to listenAddress serving
+// Prometheus metrics at /metrics over HTTPS.  Continues serving
+// until runContext.Done() and then attempts a clean shutdown
+// limited by shutdownContext.Done().  Assumes runContext.Done()
+// occurs before or simultaneously with shutdownContext.Done().
+// Also detects changes to metrics certificate files upon which
+// the metrics HTTP server is shutdown and recreated with a new
+// TLS configuration.
+func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress, certFile, keyFile string) error {
+	var tlsConfig *tls.Config
+	if listenAddress != "" {
+		var err error
+		tlsConfig, err = makeTLSConfig(certFile, keyFile)
+		if err != nil {
+			return fmt.Errorf("Failed to create TLS config: %w", err)
+		}
+	} else {
+		return errors.New("TLS configuration is required to serve metrics")
+	}
+	server := createHttpServer()
 
 	resultChannel := make(chan asyncResult, 1)
 	resultChannelCount := 1
 
-	go func() {
-		tlsListener := tls.NewListener(tcpListener, tlsConfig)
-		klog.Infof("Metrics port listening for HTTPS on %v", listenAddress)
-		err := server.Serve(tlsListener)
-		resultChannel <- asyncResult{name: "HTTPS server", error: err}
-	}()
+	go startListening(server, tlsConfig, listenAddress, resultChannel)
+
+	certDir := filepath.Dir(certFile)
+	keyDir := filepath.Dir(keyFile)
+
+	origCertChecksum, err := checksumFile(certFile)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize certificate file checksum: %w", err)
+	}
+	origKeyChecksum, err := checksumFile(keyFile)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize key file checksum: %w", err)
+	}
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if watcher == nil || err != nil {
+		return fmt.Errorf("Failed to create file watcher for certificate and key rotation: %w", err)
+	} else {
+		defer watcher.Close()
+		if err := watcher.Add(certDir); err != nil {
+			return fmt.Errorf("Failed to add %v to watcher: %w", certDir, err)
+		}
+		if certDir != keyDir {
+			if err := watcher.Add(keyDir); err != nil {
+				return fmt.Errorf("Failed to add %v to watcher: %w", keyDir, err)
+			}
+		}
+	}
 
 	shutdown := false
+	restartServer := false
 	var loopError error
 	for resultChannelCount > 0 {
 		if shutdown {
 			select {
 			case result := <-resultChannel:
 				resultChannelCount--
-				if result.error == nil {
-					klog.Infof("Collected metrics %s goroutine.", result.name)
-				} else {
-					klog.Errorf("Collected metrics %s goroutine: %v", result.name, result.error)
-					loopError = result.error
-				}
+				loopError = handleServerResult(result, loopError)
 			case <-shutdownContext.Done(): // out of time
 				klog.Errorf("Abandoning %d uncollected metrics goroutines", resultChannelCount)
 				return shutdownContext.Err()
@@ -156,21 +227,61 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
 		} else {
 			select {
 			case <-runContext.Done(): // clean shutdown
-			case result := <-resultChannel: // crashed before a shutdown was requested
-				resultChannelCount--
-				if result.error == nil {
-					klog.Infof("Collected metrics %s goroutine.", result.name)
-				} else {
-					klog.Errorf("Collected metrics %s goroutine: %v", result.name, result.error)
-					loopError = result.error
+			case result := <-resultChannel: // crashed before a shutdown was requested or metrics server recreated
+				if restartServer {
+					klog.Info("Creating metrics server with updated TLS configuration.")
+					server = createHttpServer()
+					go startListening(server, tlsConfig, listenAddress, resultChannel)
+					restartServer = false
+					continue
 				}
+				resultChannelCount--
+				loopError = handleServerResult(result, loopError)
+			case event := <-watcher.Events:
+				if event.Op != fsnotify.Chmod && event.Op != fsnotify.Remove {
+					if changed, err := certsChanged(origCertChecksum, origKeyChecksum, certFile, keyFile); changed {
+
+						// Update file checksums with latest files.
+						//
+						if origCertChecksum, err = checksumFile(certFile); err != nil {
+							klog.Errorf("Failed to update certificate file checksum: %v", err)
+							loopError = err
+							break
+						}
+						if origKeyChecksum, err = checksumFile(keyFile); err != nil {
+							klog.Errorf("Failed to update key file checksum: %v", err)
+							loopError = err
+							break
+						}
+
+						tlsConfig, err = makeTLSConfig(certFile, keyFile)
+						if err == nil {
+							restartServer = true
+							shutdownHttpServer(shutdownContext, server)
+							continue
+						} else {
+							klog.Errorf("Failed to create TLS configuration with updated configuration: %v", err)
+							loopError = err
+						}
+					} else if err != nil {
+						klog.Errorf("%v", err)
+						loopError = err
+					} else {
+						continue
+					}
+				} else {
+					continue
+				}
+			case err = <-watcher.Errors:
+				klog.Errorf("Error from metrics server certificate file watcher: %v", err)
+				loopError = err
 			}
 			shutdown = true
 			shutdownError := server.Shutdown(shutdownContext)
 			if loopError == nil {
 				loopError = shutdownError
 			} else if shutdownError != nil { // log the error we are discarding
-				klog.Errorf("Failed to gracefully shut down metrics server: %s", shutdownError)
+				klog.Errorf("Failed to gracefully shut down metrics server: %v", shutdownError)
 			}
 		}
 	}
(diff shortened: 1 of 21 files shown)


pkg/cvo/metrics.go
```diff
@@ -110,7 +117,7 @@ type asyncResult struct {
 // until runContext.Done() and then attempts a clean shutdown
 // limited by shutdownContext.Done().  Assumes runContext.Done()
 // occurs before or simultaneously with shutdownContext.Done().
-func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress string, tlsConfig *tls.Config) error {
+func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress, certFile, keyFile string, tlsConfig *tls.Config) error {
```
wking, line 120: might as well drop `tlsConfig` here and [use the cert and key file to construct it internally][1].

[1]: https://github.com/jottofar/cluster-version-operator/blob/543e841ae60248ceb57a15ba70d23b7a61c14e51/pkg/start/start.go#L218

pkg/cvo/metrics.go
```diff
@@ -136,6 +143,21 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+	origCertChecksum := checksumFile(certFile)
+	origKeyChecksum := checksumFile(keyFile)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if watcher == nil || err != nil {
+		klog.Infof("!!!! Failed to create file watcher, cert/key rotation will be disabled  %v", err)
+	} else {
+		defer watcher.Close()
+		if err := watcher.Add(certDir); err != nil {
```
wking, line 156: Don't we need another of these for key-dir if the `keyFile` directory diverges from the `certFile` directory?

pkg/cvo/metrics.go
```diff
@@ -136,6 +143,21 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
 	}()
 
+	certDir := filepath.Base(certFile)
+	origCertChecksum := checksumFile(certFile)
+	origKeyChecksum := checksumFile(keyFile)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if watcher == nil || err != nil {
+		klog.Infof("!!!! Failed to create file watcher, cert/key rotation will be disabled  %v", err)
```
wking, line 153: is there a reason to log-and-continue here instead of failing out?  I'd expect our metrics serving now to be TLS-or-nothing.

pkg/cvo/metrics.go
```diff
@@ -164,6 +186,20 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+				klog.V(2).Infof("!!!! Event from filewatcher on file: %v, event: %v", event.Name, event.Op)
+
+				if event.Op == fsnotify.Chmod || event.Op == fsnotify.Remove {
+					continue
+				}
+
+				if certsChanged(origCertChecksum, origKeyChecksum, certFile, keyFile) {
+					// Update file checksums with latest files.
+					origCertChecksum = checksumFile(certFile)
+					origKeyChecksum = checksumFile(keyFile)
```
wking, line 199: we're about to fall through into the shutdown loop.  Do we need to update these checksum variables?

pkg/cvo/metrics.go
jottofar: Agree. This was first pass just to stand something up quickly.

pkg/cvo/metrics.go
jottofar: Probably so.. Still in work...

pkg/cvo/metrics.go
jottofar: Agree. Just first pass for testing.

pkg/cvo/metrics.go
```diff
@@ -129,17 +172,35 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+
+	certDir := filepath.Dir(certFile)
+	keyDir := filepath.Dir(keyFile)
+	origCertChecksum := checksumFile(certFile)
+	origKeyChecksum := checksumFile(keyFile)
+
+	// Set up and start the file watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if watcher == nil || err != nil {
+		return fmt.Errorf("Failed to create file watcher for certificate and key rotation: %v", err)
```
wking, line 185: nit: [`%w` is like `%v` but includes `Unwrap` support][1].  Can we pivot these new entries to that?

[1]: https://pkg.go.dev/fmt#Errorf

pkg/cvo/metrics.go
wking: makes sense to me to do this, now that you're just restarting the metrics server, and not restarting the whole container.

jottofar:
/retitle Bug 2027342: Add notification for certificate changes

pkg/cvo/metrics.go
```diff
@@ -105,38 +115,84 @@ type asyncResult struct {
...
+// Also detects changes to metrics certificate files upon which
+// the metrics HTTP server is shutdown and recreated with a new
+// TLS configuration.
+func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress, certFile, keyFile string) error {
+	var tlsConfig *tls.Config
+	if listenAddress != "" {
+		var err error
+		tlsConfig, err = makeTLSConfig(certFile, keyFile)
+		if err != nil {
+			return fmt.Errorf("Failed to create TLS config: %v", err)
```
wking, line 161: I realize this is just moved from existing code, but it's another place we can update from `%v` to `%w`.

```console
$ git show origin/pr/726 | grep '^+.*Errorf' | grep -v '%w'
+                       return fmt.Errorf("Failed to create TLS config: %v", err)
```

pkg/cvo/metrics.go
```diff
@@ -105,38 +115,84 @@ type asyncResult struct {
...
+	return server
+}
 
-	tcpListener, err := net.Listen("tcp", listenAddress)
+func shutdownHttpServer(parentCtx context.Context, svr *http.Server) {
+	ctx, cancel := context.WithTimeout(parentCtx, 5*time.Second)
+	defer cancel()
+	klog.Info("Shutting down metrics server so it can be recreated with updated TLS configuration.")
+	if err := svr.Shutdown(ctx); err != nil {
+		klog.Errorf("Failed to gracefully shut down metrics server during restart: %w", err)
```
wking, line 132: I think `%w` should be for `fmt.Errorf`, but not for `klog.Errorf`.  Checking locally:

```console
$ cat test.go 
package main

import (
        "fmt"
        "k8s.io/klog/v2"
)

func main() {
        err := fmt.Errorf("a")
        err = fmt.Errorf("b: %w", err)
        klog.Errorf("w: %w", err)
        klog.Errorf("v: %v", err)
}
$ go run test.go 
E0114 15:13:06.218632    8539 test.go:11] w: %!w(*fmt.wrapError=&{b: a 0xc000098380})
E0114 15:13:06.218733    8539 test.go:12] v: b: a
```

So `%w` rendering strings, `fmt.Errorf` is identical to `%v`, but `klog.Errorf` is a noisy Go dump business.

pkg/cvo/metrics.go
```diff
@@ -105,38 +115,84 @@ type asyncResult struct {
...
+	}
+}
+
+func startListening(svr *http.Server, tlsConfig *tls.Config, lAddr string, resultChannel chan asyncResult) {
+	tcpListener, err := net.Listen("tcp", lAddr)
 	if err != nil {
-		return err
+		resultChannel <- asyncResult{name: "HTTPS server", error: err}
+	}
+	tlsListener := tls.NewListener(tcpListener, tlsConfig)
```
wking, line 141: I was wondering why we opened a new listener, instead of attaching a reconfigured server to the existing listener.  But docs for [`Server.Serve`][1]:

> Serve always returns a non-nil error and closes `l`.

so makes sense to me to open a new listener like you are doing here.

Although conveniently, opening the docs helped me notice [`Server.ServeTLS(l net.Listener, certFile, keyFile string)`][2], so perhaps we can drop `makeTLSConfig`.  Which maybe doesn't close the passed in listener?  Anyhow, moving in that direction may be more than we want to bite off in this PR.

[1]: https://pkg.go.dev/net/http#Server.Serve
[2]: https://pkg.go.dev/net/http#Server.ServeTLS

pkg/cvo/metrics.go
wking: There are a few more `klog.Errorf("...%w...", ...)` calls in the current tip that I'd prefer used `%v`:

```console
$ git show origin/pr/726 | grep '^+.*klog.Errorf.*%w'
+               klog.Errorf("Failed to gracefully shut down metrics server during restart: %w", err)
+                               klog.Errorf("Failed to gracefully shut down metrics server: %w", shutdownError)
+               klog.Errorf("Failed to open file %v for checksum: %w", fName, err)
+               klog.Errorf("Failed to compute checksum for file %v: %w", fName, err)
```

pkg/cvo/metrics.go
```diff
@@ -156,21 +212,50 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+					origKeyChecksum = checksumFile(keyFile)
+
+					var tlsErr error
+					tlsConfig, tlsErr = makeTLSConfig(certFile, keyFile)
+					if tlsErr == nil {
+						restartServer = true
+						shutdownHttpServer(shutdownContext, server)
+						continue
+					} else {
+						loopError = fmt.Errorf("Failed to create TLS config: %w", err)
```
wking, line 245: "create" -> "reload" or something so the error string makes it more clear that we had been find, but then someone borked the key/cert files?

pkg/cvo/metrics.go
```diff
@@ -156,21 +212,50 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+					}
+				} else {
+					continue
+				}
+			case err = <-watcher.Errors:
+				klog.Warningf("Error from metrics server certificate file watcher: %w", err)
```
wking, line 251: Add:

```go
loopError = err
```

so we return a non-`nil` `error` in this case (after shutting down and collecting the server result)?  We might end up clobbering it with the `Collected metrics %s goroutine...` case above, but in the case where that server shutdown is not failing, we don't want our caller to feel like we've happily completed our work.

pkg/cvo/metrics.go
```diff
@@ -156,21 +212,50 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
+					go startListening(server, tlsConfig, listenAddress, resultChannel)
+					restartServer = false
+					continue
+				}
 				resultChannelCount--
 				if result.error == nil {
 					klog.Infof("Collected metrics %s goroutine.", result.name)
 				} else {
 					klog.Errorf("Collected metrics %s goroutine: %v", result.name, result.error)
 					loopError = result.error
```
wking, line 228: This probably needs to be updated to something like:

```go
if loopError != nil {
  klog.Error(loopError)
}
loopError = result.error
```

so that we don't silently forget about any `loopError` issues we're clobbering, like an earlier `Failed to create TLS config: ...`.

pkg/cvo/metrics.go
```diff
@@ -449,3 +534,87 @@ func mostRecentTimestamp(cv *configv1.ClusterVersion) int64 {
...
 }
+
+// Determine if the certificates have changed and need to be updated.
+// Returns true if both files have changed AND neither is an empty file.
+func certsChanged(origCertChecksum []byte, origKeyChecksum []byte, certFile, keyFile string) bool {
+	// Check if both files exist.
+	certNotEmpty, err := fileExistsAndNotEmpty(certFile)
+	if err != nil {
+		klog.Warningf("Error checking if changed TLS cert file empty/exists: %w", err)
+		return false
```
wking, line 545: We may not try again until the next fsnotify event, so I think we might want `bool, error` out of this function, and then:

```go
return false, err
```

here and:

```go
if event.Op != fsnotify.Chmod && event.Op != fsnotify.Remove {
  if changed, err := certsChanged(origCertChecksum, origKeyChecksum, certFile, keyFile); err != nil {
    return fmt.Errorf("checking TLS configuration: %w", err)
  } else changed {
    ...
  }
}
```

in the caller.  Or something like that...

pkg/cvo/metrics.go
jottofar: > why we opened a new listener

Correct, I tried the former first but `Shutdown` [works by first closing all open listeners](https://cs.opensource.google/go/go/+/go1.17.6:src/net/http/server.go;l=2709).

> more than we want to bite off in this PR.

Agree. It does however appear Shutdown is still used and still closes all open listeners.also close the lis

pkg/cvo/metrics.go
jottofar: I updated the error to clarify it's an updated TLS config. Also, there is this log entry that will always precede: "Certificate and key changed. Will recreate metrics server with updated TLS configuration."

pkg/cvo/metrics.go
jottofar: So I was thinking that having CVO up with no file watcher is better than no CVO? User would just have to manually restart CVO to pick certs like they do now, But, on the other hand, if we can't create a watcher perhaps that's indicative of other issues?

pkg/cvo/metrics.go
jottofar: Here again I wasn't sure we wanted to shutdown CVO for these types of errors. If an admin improperly rotates the certs and ends up with no key file (if that's even possible?) do we want to kill CVO or just let it continue waiting for correct certs? Isn't it just going to restart and do that anyway?

pkg/cvo/metrics.go
wking: Yeah, I think failing to create the watcher would be something where we'd want a crashlooping CVO so the admin could investigate and file a bug with us (and we might even get logs via Insights).  If someone improperly rotates certs and breaks metrics, they're going to at least start getting TargetDown alerts.  And it's not clear when the next fsnotify event is going to come through and trigger another look.

If crashlooping on these errors is too harsh, our usual soft-complaint channel of "fire an alert" is probably not reliable, because the breakage would be to that metrics pipeline.  We might still be able to complain via ClusterVersion conditions, and I'd be ok with that, but it seems like it would be some work to wire up, and I'm optimistically hoping cred-rotation errors are rare enough that crashlooping is a sufficiently elegant UX when they do come up.  I'm fine punting this sort of handling out to follow-up work which we can pick up if Insights and all show us that cred-rotation issues are more common than I expect them to be.

pkg/cvo/metrics.go
wking: I've added crashlooping motivation over [here][1].  If that makes sense, I think we do want the fatal-to-the-CVO `return fmt.Errorf(...)` case here.

[1]: https://github.com/openshift/cluster-version-operator/pull/726#discussion_r786276935

pkg/cvo/metrics.go
jottofar: I created a new method `handleServerResult` which resets `lastLoopError` if necessary. It also checks for and treats as a non-error `ErrServerClosed`, since this is always returned upon successful shutdown, as to not clobber our real shutdown reason which may be stored in `lastLoopError`. 
In the case of a `result.error != ErrServerClosed`, `lastLopError` is set to this error however the previous `lastLoopError` has already been logged so it is not clobbered. I considered constructing and returning a new `lastLoopError` from both and returning it but that seemed like overkill.

pkg/cvo/metrics.go
jottofar: Restructured so this results in a graceful shutdown.

pkg/cvo/metrics.go
jottofar: Restructured so this results in a graceful shutdown.

pkg/cvo/metrics.go
```diff
@@ -424,6 +520,13 @@ func gaugeFromInstallConfigMap(cm *corev1.ConfigMap, gauge *prometheus.GaugeVec,
 	return g
 }
 
+func logCurrentAndReturnNewError(currentError, newError error) error {
```
wking, line 523: [lint][1]:

```
pkg/cvo/metrics.go:523:6  deadcode  `logCurrentAndReturnNewError` is unused
```

Can we drop this function?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/726/pull-ci-openshift-cluster-version-operator-master-golangci-lint/1483827196928724992#1:build-log.txt%3A16

pkg/cvo/metrics.go
```diff
@@ -449,3 +552,85 @@ func mostRecentTimestamp(cv *configv1.ClusterVersion) int64 {
...
+			return &certificate, nil
+		},
+	}), nil
+}
+
+// Compute the sha256 checksum for file 'fName'.
+func checksumFile(fName string) []byte {
+	file, err := os.Open(fName)
+	if err != nil {
+		klog.Errorf("Failed to open file %v for checksum: %v", fName, err)
```
wking, line 612: Per [my pitch for crashlooping when we have trouble with things that should always succeed][1], can we make `Open` and `Copy` failures fatal and return an `error` from this function?

[1]: https://github.com/openshift/cluster-version-operator/pull/726#discussion_r786276935

pkg/cvo/metrics.go
```diff
@@ -105,72 +115,171 @@ type asyncResult struct {
...
+							klog.Errorf("Failed to update certificate file checksum: %v", err)
+							loopError = err
+						}
+						if origKeyChecksum, err = checksumFile(keyFile); err != nil {
+							klog.Errorf("Failed to update key file checksum: %v", err)
+							loopError = err
+						}
+
+						tlsConfig, err = makeTLSConfig(certFile, keyFile)
+						if err == nil {
```
wking, line 256: if we failed either of the `checksumFile` calls above, the `makeTLSConfig` call here will have clobbered `err`, and we might have a successful `makeTLSConfig` call get us through to this restart, when really we want to fall through to shutdown.  Maybe `if loopError == nil`?

jottofar: Oh crap, yeah. Let me re-structure so it just bails as soon as it it's any of those errors rather than continuing. That was unintentional.

jottofar: I just added a `break` to each one.

wking: I don't think `break` works, because the shutdown code is currently inside the `for` loop, and after `Shutdown`, we also have the wait on `resultChannel` and/or a timeout in this same `for` loop, so we can consider that error when deciding which error to complain about when the function returns.  I think dropping the `break` and using `if loopError == nil` here will work.  "If we haven't hit an error yet, try to restart, otherwise, continue to shutdown".

pkg/cvo/metrics.go
wking: As of 0e6136c727b055, `checksumFile` returns an `error` too.

pkg/cvo/metrics.go
wking: As of 0e6136c727b055, `logCurrentAndReturnNewError` is gone.

pkg/cvo/metrics.go
wking: As of 0e6136c727b05509, `certsChanged` and `fileExistsAndNotEmpty` calls both return errors which are treated as fatal.

pkg/cvo/metrics.go
wking: `handleServerResult` handling in 0e6136c727b0 looks good to me.

pkg/cvo/metrics.go
wking: In 0e6136c727b05509c, `watcher.Errors` feeds `loopError` and proceeds into graceful shutdown,

pkg/cvo/metrics.go
wking: In 0e6136c727b05509c, this is `Failed to create TLS configuration with updated configuration`, which looks good to me.

pkg/cvo/metrics.go
wking: As of 0e6136c727b05509c:

```console
$ git show origin/pr/726 | grep -c '^+.*klog.Errorf.*%w'
0
$ git show origin/pr/726 | grep '^+.*fmt.Errorf' | grep -v '%w'
+               return false, fmt.Errorf("Certificate or key is missing or empty, certificates will not be rotated.")
```

So none of the `klog.Errorf` calls use `%w` (which is good).  And all of the `fmt.Errrof` calls use `%w` (which is good), except for a single entry that's not actually formatting strings at all.  That could use [`errors.New`][1] instead, but isn't a big deal, and I'm going to mark this thread resolved without waiting on that last nit.

[1]: https://pkg.go.dev/errors#New

pkg/cvo/metrics.go
wking: Fixed in 0e6136c727b05.

pkg/cvo/metrics.go
jottofar: These `break`s break out of the select not the `for` loop. So it falls through to line 279 and goes through the normal `shutdown` logic.

wking:
/lgtm

[e2e-agnostic][1] passed.

[e2e-agnostic-upgrade][2] had orthogonal install failures, and there's some green deeper in the history.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/726/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1484276895288659968
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/726/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1484276895481597952

PR #728, opened 2022-01-17
Bug 2009845: Respect overrides changes during install (second round)

wking:
We'd taken a pass at this in bee2a5e634 (#713), but subsequently reverted in 11bca782a8 (#727), due to [some hot-looping on overrides-change detection][1].

With this new attempt, we're lifting the state check up out of `equalSyncWork`, where it has lived since the check landed in 213726df85 (#500).  This allows us to guard more specifically in the `Update` method, keeping `spec.desiredUpdate` changes out of the sync worker's work config until we complete installation, while passing `overrides` changes through.

On the sync-worker side, `calculateNext` keeps its traditional behavior of accepting any work that the `Update` method set up for it.  It does grow a new guard to keep it from transitioning from `InitializingPayload` to `UpdatingPayload`.  With the new guard, the only way to transition out of `InitializingPayload` is a complete reconciliation (with the softened per-manifest levels of "is this happy enough?" that go with `InitializingPayload`).

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2009845#c8

PR #733, opened 2022-02-01
Bug 2009845: pkg/cvo/sync_worker: Use current state, not suggested state, for guarding Initializing->Updating

wking:
Fixing a bug from 5c7ed6da32 (#728), where:

1. The bootstrap cluster-version operator does its happy thing.
2. With bootstrap winding down, the bootstrap CVO exits, in favor of a control-plane-hosted CVO.
3. Control-plane CVO acquires the leader lock, and [continues `Initializing`][e].
4. As a final bootstrap action, the installer [patches away its `spec.overrides` additions][1].
5. The control-plane CVO watcher gets notified after the `overrides` change, and [call `SyncWorker.Update`][f].
6. `SyncWorker.Update` notices that `w.work != nil`, so it ignores [its `state` argument][a], and [the newly-created work structure][b]'s `State` defaults to 0, because:

    ```console
    $ grep iota pkg/payload/payload.go
      UpdatingPayload State = iota
    ```

    and [`iota` starts at zero][2].  This is the bit I hadn't understood before.
7. `SyncWorker.calculateNext` sees [`changed` (because `!sameOverrides`)][d] and `w.work.State != payload.InitializingPayload` (because it defaulted to the `UpdatingPayload` zero value), and we'd get logs [like][3]:

    ```console
    $ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-aws-serial/1488223505995534336/artifacts/e2e-aws-serial/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-79759cfdfd-h4fpr_cluster-version-operator.log | grep 'Work changed, transitioning from'
    I0131 19:10:19.526493       1 sync_worker.go:442] Work changed, transitioning from Initializing and the suggested incoming Updating state to Updating
    ```

With this commit, the sync worker respects the initial state passed in via `SyncWorker.Update` (which is [the `w.work == nil` case where `SyncWorker.Update` actually overrides the default state value][c]), but the sync worker ignores further `w.work.State` input, because it already knows its own state.

[1]: https://github.com/openshift/installer/pull/5258
[2]: https://go.dev/ref/spec#Iota
[3]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-aws-serial/1488223505995534336
[a]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/sync_worker.go#L208
[b]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/sync_worker.go#L212-L216
[c]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/sync_worker.go#L238-L239
[d]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/sync_worker.go#L434
[e]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/cvo.go#L568-L569
[f]: https://github.com/openshift/cluster-version-operator/blob/07295564ba659631396b6274171a78132b122bac/pkg/cvo/cvo.go#L577

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -438,8 +438,8 @@ func (w *SyncWorker) calculateNext(work *SyncWork) bool {
 	if work.Empty() {
 		work.State = w.work.State
 		work.Attempt = 0
-	} else if changed && w.work.State != payload.InitializingPayload {
-		klog.V(2).Infof("Work changed, transitioning from %s and the suggested incoming %s state to %s", work.State, w.work.State, payload.UpdatingPayload)
+	} else if changed && work.State != payload.InitializingPayload {
+		klog.V(2).Infof("Work changed, transitioning from %s to %s", work.State, payload.UpdatingPayload)
 		work.State = payload.UpdatingPayload
 		work.Attempt = 0
 	}


wking:
[`etcdHighNumberOfFailedGRPCRequests`][1] and [network connectivity][2] are unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/733/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1488429275827671040
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/733/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1488429275915751424

PR #734, opened 2022-02-02
WIP: Report metrics and alert on verified=false

Davoska:
Work in progress.


Davoska:
/hold
I need to resolve the failing test `disruption_tests: [sig-arch] Check if alerts are firing during or after upgrade success` due to the new alert `CurrentVersionNotVerified` firing.

vrutkovs:
>due to the new alert CurrentVersionNotVerified firing.

These alerts are defined in https://github.com/openshift/origin (i.e. https://github.com/openshift/origin/commit/0a3cbf528f67a7091706b7b4f73f8da35f47cdc4), you may use the same bug for a new PR there

Davoska:
I have created the https://github.com/openshift/origin/pull/26900 pull request to add the new alert `CurrentVersionNotVerified` to the allowed alerts in the appropriate tests (https://github.com/openshift/cluster-version-operator/pull/734#issuecomment-1056861239).

I am removing the `do-not-merge/hold` label, and I will rerun the tests when the https://github.com/openshift/origin/pull/26900 pull request gets merged.

/unhold

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -57,6 +57,14 @@ spec:
         (time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 3600 and ignoring(condition, name, reason) cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"}
       labels:
         severity: warning
+    - alert: CurrentVersionNotVerified
+      annotations:
+        summary: The current version of the cluster is not verified.
+        description: The cluster was forcefully upgraded with a provided release image which verification failed, and therefore this cluster may not be trusted or may experience some issues. The verification process does not cover upgradeable checks that depend on the cluster state at the time when the update target was accepted.
+      expr: |
+        cluster_version{type="current", verified="false"}
```
simonpasquier, line 65: To avoid resetting on failed scrapes, it's best to use `last_over_time()`

```suggestion
        last_over_time(cluster_version{type="current", verified="false"}[5m])
```

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -57,6 +57,14 @@ spec:
         (time()-cluster_version_operator_update_retrieval_timestamp_seconds) >= 3600 and ignoring(condition, name, reason) cluster_operator_conditions{name="version", condition="RetrievedUpdates", endpoint="metrics", reason!="NoChannel"}
       labels:
         severity: warning
+    - alert: CurrentVersionNotVerified
+      annotations:
+        summary: The current version of the cluster is not verified.
+        description: The cluster was forcefully upgraded with a provided release image which verification failed, and therefore this cluster may not be trusted or may experience some issues. The verification process does not cover upgradeable checks that depend on the cluster state at the time when the update target was accepted.
```
simonpasquier, line 63: can we explain here how the user would resolve the alert?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -57,6 +57,14 @@ spec:
...
+      expr: |
+        cluster_version{type="current", verified="false"}
+      labels:
+        severity: warning
```
simonpasquier, line 67: Wondering if it should be `info` but no strong feeling.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
Davoska: Oh, thank you for the tip! I'll modify the code accordingly.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
Davoska: Yes, that is a good idea.

> The cluster was forcefully upgraded with a provided release image which verification failed, and therefore this cluster may not be trusted or may experience some issues. The verification process does not cover upgradeable checks that depend on the cluster state at the time when the update target was accepted. **To resolve this alert, please upgrade the cluster to a verified version, or silence this alert if you are aware of the potential risks.**

What do you think?

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
Davoska: Yes, I understand. There could be an argument for both of the severity labels.

In the said [Jira Card](https://issues.redhat.com/browse/OTA-187) it is suggested to have the `warning` label.

The [Openshift documentation](https://docs.openshift.com/container-platform/4.6/monitoring/managing-alerts.html) defines the labels as:

- **Warning**. The alert provides a warning notification about something that might require attention in order to prevent a problem from occurring. Warnings are typically routed to a ticketing system for non-immediate review.
- **Info**. The alert is provided for informational purposes only.

Resolving the issue of the cluster's version not being verified could prevent a problem from occurring, and it is rather important information.
However, the user can only apply such a version if an update is applied forcefully. And the information about the `--force` flag does state the potential risks. So the user that applied the upgrade knew about the risks, therefore a `warning` label maybe isn't necessary. 

I am leaning more toward the `warning` label, but I'll keep the discussion up and I will ask the team for opinions.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
simonpasquier: I would remove the second part (`, or silence this alert if you are aware of the potential risks.`) since silencing doesn't really resolve the alert (it's still there but hidden).

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
simonpasquier: @openshift/sre-alert-sme experts might have a view on this question :)

mmazur: tl;dr: this is too unactionable for a Warning, however there's definitive value for an Info alert providing additional context when investigating cluster issues.


From a fleet (managed openshift) point of view this alerts isn't really actionable for us SREs. The most we can do is send a servicelog message and silencing the alert on our end, but the msg would just be restating the contents of the alert ("please don't do that"). (For context: managed openshift SREs by default see Critical and Warning alerts from clusters except for the ones that we [explicitly chose not to route](https://github.com/openshift/configure-alertmanager-operator/blob/master/pkg/controller/secret/secret_controller.go#L160-L266).)

Using the [Alerting Consistency Enhancement Proposal](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md) as a guide:
* **Warning** – The vast majority of alerts should use the severity. Issues at the warning level should be addressed in a timely manner, but don't pose an immediate threat to the operation of the cluster as a whole.
* **Info** – Info level alerts represent situations an administrator should be aware of, but that don't necessarily require any action. Use these sparingly, and consider instead reporting this information via Kubernetes events.

I'd say in this context this falls into the category of "Info". This alert on its own doesn't really require any action, as the cluster admin explicitly chose to do an upgrade like this and if there are no issues with a cluster running like this, there's also no need for action to remedy this situation.

This alert's usefulness is in the additional context (information) that it provides – if there __is__ something wrong with the cluster and Critical/Warning alerts are firing, then you as a cluster admin/investigating SRE will be glad that this alert's there to point at the unverified upgrade being a potential culprit.

dofinn: +1 to @mmazur ! execellent dive into this. We should also look to automate messages to customer on this as well :)

Davoska: Thank you for the great insights! I'll change the severity to `info`.

Davoska:
/hold
After the previous commit, there are now multiple alerts `CurrentVersionNotVerified` firing when testing on a cluster. I need to resolve this issue.

Davoska:
> After the previous commit, there are now multiple alerts CurrentVersionNotVerified firing at the beginning of a cluster creation when testing on a cluster. I need to resolve this issue.

/unhold
Fixed using the `count` function in the commit https://github.com/openshift/cluster-version-operator/commit/d979b7db0f188187a96aabeaeffb398cbc78d61b.

Davoska:
> I have created the [openshift/origin#26900](https://github.com/openshift/origin/pull/26900) pull request to add the new alert `CurrentVersionNotVerified` to the allowed alerts in the appropriate tests ([#734 (comment)](https://github.com/openshift/cluster-version-operator/pull/734#issuecomment-1056861239)).

With the severity level set to `info`, the https://github.com/openshift/origin/pull/26900 wouldn't be necessary.

Davoska:
Refactored the code a little bit. Added the namespace `openshift-cluster-version` to the alert to comply with the Alerting rules guidelines. Fixed reporting `verified` in the label `desired` and `failure`.

/hold
Hold while I test the pull request via Cluster Bot.

PR #735, opened 2022-02-04
Add utility to check for deleted resources

jottofar:


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
 	gopkg.in/fsnotify.v1 v1.4.7
+	gopkg.in/yaml.v2 v2.4.0
```
wking, line 23: We can avoid the additional direct dependency by sticking to `github.com/ghodss/yaml` for the new code.

hack/cluster-version-util/releasediff/README.md
```diff
@@ -0,0 +1,36 @@
...
+-r, --releases string   top-level directory containing subdirectories for each OCP release to be checked
+-t, --target string     file containing all resources from a running, newly installed, OCP release
+-v, --verbose           verbose logging
+```
+The **"--target"** file contains all the resources from a running, newly installed, OCP release. This file is created by connecting to an OCP cluster running the OCP release to be compared against. The cluster must have been newly installed with the release, not upgraded, in order to get only the set of resources created by that specific release. Once connected to the cluster the file can be created by running **"tools/create-target-release-file.sh TARGET_FILE_NAME"** where  **TARGET_FILE_NAME** is the path to the file to be created. This file will have 4 columns for each resource: APIVersion, Kind, Name, Namespace. The release-resource-diff utility depends on this file having these 4 columns in the given order. The file will be used for comparison against resources from other OCP releases to verify whether the resource still exists. For an example see **test/4.10.0-0.nightly-2022-01-27-104747-all-resources.txt**.
```
wking, line 23: nit: `--target`, `TARGET_FILE_NAME`, etc. to use backticks where you want something marked up as a literal.  Fewer characters in the Markdown source and more distinctive coloring when rendered ;).

Also:

    For an example, see [`test/4.10.0-0.nightly-2022-01-27-104747-all-resources.txt`](test/4.10.0-0.nightly-2022-01-27-104747-all-resources.txt).

or:

    For an example, see [here](test/4.10.0-0.nightly-2022-01-27-104747-all-resources.txt).

to make these references to local files clickable for folks viewing the README in GitHub or other Markdown-aware viewers.

hack/cluster-version-util/releasediff/release-resource-diff.go
```diff
@@ -0,0 +1,421 @@
...
+		}
+	default:
+		return newId, ""
+	}
+	return newId, newVersion
+}
+
+func checkIfOrphaned(resourceIds map[ResourceId]ResourceSource, currentOrphaned map[ResourceId]ResourceSource) {
+	for k, v := range resourceIds {
+		if _, ok := currentOrphaned[k]; !ok {
```
vrutkovs, line 361: nitpick: I'd rather have `if ok { continue}` instead of nested loops

PR #528, opened 2021-03-01
Bug 1871303: metrics: serve metrics after leader lock acquired

jottofar:
We'll get a CI error if installer metric is generated with either `type` or `invoker` set to "". Fix possible race condition whereby informers are not yet sync'ed before metrics are served which results in an API IsNotFound error.

PR #741, opened 2022-02-14
Get cluster version object earlier in startup

jottofar:


wking:
Looking back at 4d421e9582b91, where you were adjusting the existing `InitializeFromPayload` in `start.go`, that was happening in `Options.Run`.  But `Options.Run` [calls `Options.run` as a final step][1], and [the informers get started in `Options.run`][2], so at that point you'll be waiting on ClusterVersion forever watching an unstarted informer.

In 4611337249a, you've moved the load into `OnStartedLeading`, and that works because it's after we've started the informers.  But we shouldn't need an actual lease to load the data, and there's some latency benefits by working on those in parallel, or at least performing the more fixed-time payload load first, before blocking on a possibly contended lease acquisition.  Can we move the `InitializeFromPayload` call to sit right after the informer starts, and before the lease acquisition?  Possibly with a leading wait for the `controllerCtx.CVInformerFactory` cache to sync?  Or are we ok with the low latency cost of keeping the `InitializeFromPayload` behind `OnStartedLeading`?

[1]: https://github.com/openshift/cluster-version-operator/blob/e03ec756ed2a122709fd8bf63f07d574f0d1e652/pkg/start/start.go#L158
[2]: https://github.com/openshift/cluster-version-operator/blob/e03ec756ed2a122709fd8bf63f07d574f0d1e652/pkg/start/start.go#L193-L196

jottofar:
> Looking back at [4d421e9](https://github.com/openshift/cluster-version-operator/commit/4d421e9582b9142fd5259f1595e7b81cac845a7d), where you were adjusting the existing `InitializeFromPayload` in `start.go`, that was happening in `Options.Run`. But `Options.Run` [calls `Options.run` as a final step](https://github.com/openshift/cluster-version-operator/blob/e03ec756ed2a122709fd8bf63f07d574f0d1e652/pkg/start/start.go#L158), and [the informers get started in `Options.run`](https://github.com/openshift/cluster-version-operator/blob/e03ec756ed2a122709fd8bf63f07d574f0d1e652/pkg/start/start.go#L193-L196), so at that point you'll be waiting on ClusterVersion forever watching an unstarted informer.
> 
> In [4611337](https://github.com/openshift/cluster-version-operator/commit/4611337249aebbdeb907c0ef76a119a5877a07d3), you've moved the load into `OnStartedLeading`, and that works because it's after we've started the informers. But we shouldn't need an actual lease to load the data, and there's some latency benefits by working on those in parallel, or at least performing the more fixed-time payload load first, before blocking on a possibly contended lease acquisition. Can we move the `InitializeFromPayload` call to sit right after the informer starts, and before the lease acquisition? Possibly with a leading wait for the `controllerCtx.CVInformerFactory` cache to sync? Or are we ok with the low latency cost of keeping the `InitializeFromPayload` behind `OnStartedLeading`?

No, I see no reason to wait on the lease acquisition so I'll move things around. I wasn't convinced yet that it was the actual lease acquisition but had not circled back to look further.

pkg/start/start.go
```diff
@@ -195,96 +192,103 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
...
-			Lock:            lock,
-			ReleaseOnCancel: true,
-			LeaseDuration:   o.leaderElection.LeaseDuration.Duration,
-			RenewDeadline:   o.leaderElection.RenewDeadline.Duration,
-			RetryPeriod:     o.leaderElection.RetryPeriod.Duration,
-			Callbacks: leaderelection.LeaderCallbacks{
-				OnStartedLeading: func(_ context.Context) { // no need for this passed-through postMainContext, because goroutines we launch inside will use runContext
-					launchedMain = true
-					if o.ListenAddr != "" {
+	if !cache.WaitForCacheSync(informersDone, controllerCtx.CVO.GetCacheSynced()...) {
```
wking, line 195: nit: Instead of exposing `GetCacheSynced`, maybe the public API should be `controllerCtx.CVO.WaitForCacheSync(postMainContext)`?  Or... where are we're creating these informers?  Maybe we don't need to claw anything back out of the `CVO` property at all?

pkg/start/start.go
```diff
@@ -195,96 +192,103 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
...
+		err := postMainContext.Err()
+		klog.V(2).Infof("Caches never synchronized: %v", err)
+	} else if err := controllerCtx.CVO.InitializeFromPayload(ctx, restConfig, burstRestConfig); err != nil {
+		klog.V(2).Infof("Initialize from payload: err=%v", err)
+	} else {
```
wking, line 200: nit: instead of nesting all of this, can we have the earlier blocks of this `if`/`else` return their errors instead of logging them and falling through?

pkg/start/start.go
```diff
@@ -195,6 +192,17 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 	controllerCtx.OpenshiftConfigManagedInformerFactory.Start(informersDone)
 	controllerCtx.InformerFactory.Start(informersDone)
 
+	klog.Info("!!!! Waiting for cache to sync...")
+	if !cache.WaitForCacheSync(informersDone) {
+		klog.Errorf("Caches never synchronized: %v", postMainContext.Err())
+		return
```
wking, line 198: Hmm.  `klog.Fatalf`, or update `run` and above to allow returning an error we can bubble up the stack?

wking: Done in  bba8ed4eae.

pkg/start/start.go
```diff
@@ -195,6 +192,15 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
 	controllerCtx.OpenshiftConfigManagedInformerFactory.Start(informersDone)
 	controllerCtx.InformerFactory.Start(informersDone)
 
+	if !cache.WaitForCacheSync(informersDone, controllerCtx.CVO.GetClusterVersionCacheSynced()) {
```
wking, line 195: [`CVInformerFactory`][1] is an [`externalversions.SharedInformerFactory`][2].  Maybe we can use:

```go
if !controllerCtx.CVInformerFactory.WaitForCacheSync(informersDone) {
```

here, and not need the new `GetClusterVersionCacheSynced`?

[1]: https://github.com/openshift/cluster-version-operator/blob/b133d822ea2de7ccf7f9b6ee5530da703f3d0b2c/pkg/start/start.go#L414
[2]: https://pkg.go.dev/github.com/openshift/client-go/config/informers/externalversions#SharedInformerFactory

pkg/cvo/cvo.go
```diff
@@ -516,18 +532,19 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
...
 	}
 	if changed {
 		klog.V(2).Infof("Cluster version changed, waiting for newer event")
 		return nil
 	}
+
+	// this should never happen
 	if original == nil {
-		klog.V(2).Infof("No ClusterVersion object and defaulting not enabled, waiting for one")
+		klog.V(2).Infof("No ClusterVersion object, waiting for one")
```
wking, line 547: I'm fine dropping this now-dead code, and just making it part of the `getClusterVersion` contract that `original` must be non-`nil` when `err == nil`.  Good excuse to add `getClusterVersion` godocs explaining the contract ;)

jottofar:
/retitle Get cluster version object earlier in startup

pkg/start/start.go
jottofar: used Fatalf

pkg/cvo/cvo.go
jottofar: re-factored and documented

pkg/start/start.go
jottofar: That works

pkg/cvo/cvo.go
```diff
@@ -225,10 +218,33 @@ func New(
 	return optr
 }
 
-// InitializeFromPayload retrieves the payload contents and verifies the initial state, then configures the
-// controller that loads and applies content to the cluster. It returns an error if the payload appears to
-// be in error rather than continuing.
-func (optr *Operator) InitializeFromPayload(restConfig *rest.Config, burstRestConfig *rest.Config) error {
+func (optr *Operator) GetClusterVersionCacheSynced() cache.InformerSynced {
```
wking, line 221: This method and the `cvCacheSynced` property are no longer needed.

pkg/start/start.go
```diff
@@ -195,6 +191,18 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock *resourc
...
 	controllerCtx.InformerFactory.Start(informersDone)
 
+	allSynced := controllerCtx.CVInformerFactory.WaitForCacheSync(informersDone)
+	for _, synced := range allSynced {
+		if !synced {
+			klog.Fatalf("Caches never synchronized: %v", postMainContext.Err())
+			return
+		}
+	}
+	if err := controllerCtx.CVO.InitializeFromPayload(ctx, restConfig, burstRestConfig); err != nil {
```
wking, line 201: I'd argued earlier for this position, to get this disk-load out of the way before the possibly-slow wait for a lease.  But I'm going to reverse myself now, because I want #746 to go `ClusterOperatorDown{name="version"}` if we're waiting for a ClusterVersion for too long, and, since #528, we don't start  serving metrics until we have the leader lock.  Can we move this call to [right after the metrics launch][1]?  Sorry for waffling!

[1]: https://github.com/openshift/cluster-version-operator/blob/0e9bc4ef03004fcc2bb0c58d39e5b49445a9f8f6/pkg/start/start.go#L217

jottofar: Moved it. Still simply calling `Fatalf` on error to exit.

pkg/cvo/cvo.go
```diff
@@ -213,7 +205,8 @@ func New(
 	optr.cacheSynced = append(optr.cacheSynced, coInformer.Informer().HasSynced)
 
 	optr.cvLister = cvInformer.Lister()
-	optr.cacheSynced = append(optr.cacheSynced, cvInformer.Informer().HasSynced)
+	optr.cvCacheSynced = cvInformer.Informer().HasSynced
+	optr.cacheSynced = append(optr.cacheSynced, optr.cvCacheSynced)
```
wking, line 209: we can drop `cvCacheSynced` and go back to the old:

```go
optr.cacheSynced = append(optr.cacheSynced, cvInformer.Informer().HasSynced)
```

here.

wking:
[e2e-agnostic-operator][1]:

```
F0301 20:39:30.447890   12993 start.go:222] Initialize from payload: err=Error when attempting to get cluster version object: timed out waiting for the condition
```

Checking:

```console
$ git --no-pager log -1 --oneline origin/pr/741
978603ed (origin/pr/741) Get cluster version object earlier in startup
$ git grep -n . origin/pr/741 -- pkg/start/start.go | grep -1 222:
origin/pr/741:pkg/start/start.go:221:                                   if err := controllerCtx.CVO.InitializeFromPayload(ctx, restConfig, burstRestConfig); err != nil {
origin/pr/741:pkg/start/start.go:222:                                           klog.Fatalf("Initialize from payload: err=%v", err)
origin/pr/741:pkg/start/start.go:223:                                   }
```

Maybe we need to raise a timeout somewhere?

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/741/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator/1498750147855650816#1:build-log.txt%3A135

wking:
[Sandbox issue][1] is orthogonal:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/741/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1498652270525419520

wking:
[earlier e2e-agnostic pass][1], and 6164494 -> 978603e is just the `cvCacheSynced` nit, so:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/741/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1498433381367549952

pkg/cvo/cvo.go
wking: fixed in 6164494f -> 978603ed22.

wking:
/hold

Seems like we need a chance for e2e-agnostic-operator.  No need to keep banging away in the meantime.

bootstrap/bootstrap-pod.yaml
```diff
@@ -14,7 +14,6 @@ spec:
       - "start"
       - "--release-image={{.ReleaseImage}}"
       - "--enable-auto-update=false"
-      - "--enable-default-cluster-version=false"
```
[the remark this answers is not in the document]
LalatenduMohanty: I mean along with ``` cmd.PersistentFlags().BoolVar(&opts.EnableDefaultClusterVersion, "enable-default-cluster-version", opts.EnableDefaultClusterVersion, "Allows the operator to create a ClusterVersion object if one does not already exist.")```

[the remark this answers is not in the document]
jottofar: CVO waits "forever" (or until context is cancelled) for creation of ClusterVersion.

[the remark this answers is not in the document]
wking: The commit message (currently 41d36644025) explains why we can drop this.

hack/test-prerequisites.go
```diff
@@ -55,5 +59,29 @@ func main() {
 		if err != nil {
 			log.Fatalf("Could not install %s CRD: %v", name, err)
 		}
+		err = wait.PollImmediate(time.Second, 30*time.Second, func() (bool, error) {
+			_, err = cvClient.ConfigV1().ClusterVersions().Create(ctx,
+				&configv1.ClusterVersion{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "version",
```
wking, line 66: The defaulting code we're removing had been using `Name: optr.name`, so these e2e-operator tests could drop in a ClusterVersion for testing without clobbering the ClusterVersion that the host cluster was using.  I think it works like this:

1. Host cluster comes up with `version` ClusterVersion.
2. `hack/test-integration.sh` called
3. New ClusterVersions created with different names that aren't `version`.
4. Some kind of integration testing on CVOs which consume those per-test ClusterVersions.
5. Testing completes.
6. Non-`version` ClusterVersions removed.

Or something.  I may be completely wrong :p

hack/test-prerequisites.go
```diff
@@ -55,5 +59,29 @@ func main() {
 		if err != nil {
 			log.Fatalf("Could not install %s CRD: %v", name, err)
 		}
+		err = wait.PollImmediate(time.Second, 30*time.Second, func() (bool, error) {
+			_, err = cvClient.ConfigV1().ClusterVersions().Create(ctx,
```
wking, line 63: I think we want to move this creation out of this file, and into a `pkg/start/start_integration_test.go`.  I don't think we need the poll wrapper, unless we feel like we need that for robustness in the face of a flaky Kube-API server connection or some such.  It would slot into each test-case right after namespace creation, around [here][1], and similarly for each of the other `TestIntegrationCVO_...` functions.  Something like:

```go
ns := fmt.Sprintf("e2e-cvo-%s", randutil.String(4))

if _, err := kc.CoreV1().Namespaces().Create(
	ctx,
	&v1.Namespace{
		ObjectMeta: metav1.ObjectMeta{
			Name: ns,
		},
	},
	metav1.CreateOptions{},
); err != nil {
	t.Fatal(err)
}
defer func() {
	if err := kc.CoreV1().Namespaces().Delete(ctx, ns, metav1.DeleteOptions{}); err != nil {
		t.Logf("failed to delete namespace %s: %v", ns, err)
	}
}()

if _, err := client.ConfigV1().ClusterVersions().Create(
	ctx,
	&configv1.ClusterVersion{
		ObjectMeta: metav1.ObjectMeta{
			Name: ns,
		},
		Spec: configv1.ClusterVersionSpec{
			Channel: "fast",
		},
	},
	metav1.CreateOptions{},
); err != nil {
	t.Fatal(err)
}
defer func() {
	if err := client.ConfigV1().ClusterVersions().Delete(ctx, ns, metav1.DeleteOptions{}); err != nil {
		t.Logf("failed to delete cluster version %s: %v", ns, err)
	}
}()
```

It's a bit long-winded, but writing a helper to wrap what's basically two function calls seems fairly complicated too.  And having the two calls inline within `TestIntegrationCVO_...` does make for the convenient `defer` teardown.

[1]: https://github.com/openshift/cluster-version-operator/blob/addb685a2e9f6375ba4ccd3dbfb9f9236475494a/pkg/start/start_integration_test.go#L252

hack/test-prerequisites.go
wking: moving conversation to [here][1].

[1]: https://github.com/openshift/cluster-version-operator/pull/741#discussion_r818372189

wking:
[e2e-agnostic-operator][1] failed in installer Terraform:

```
level=fatal msg=Error: Unable to locate Storage Account "clustermvrlm"!
```

[CI-search][2] suggests this is an isolated flake, so try again:

/test e2e-agnostic-operator

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/741/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator/1499423741107507200#1:build-log.txt%3A52
[2]: https://search.ci.openshift.org/chart?maxAge=24h&type=build-log&name=.&search=Unable%20to%20locate%20Storage%20Account.*clustermvrlm

wking:
Also, because we think we've fixed the earlier issue:

/hold cancel

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -79,20 +79,33 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
...
+		queue:          workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "cvo-loop-test"),
+		client:         client,
+		cvLister:       &clientCVLister{client: client},
+		exclude:        "exclude-test",
+		eventRecorder:  record.NewFakeRecorder(100),
+		clusterProfile: payload.DefaultClusterProfile,
 	}
 
 	dynamicScheme := runtime.NewScheme()
-	//dynamicScheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "test.cvo.io", Version: "v1", Kind: "TestA"}, &unstructured.Unstructured{})
```
wking, line 95: nit: dead code, just delete?

wking (approved):
/lgtm

wking:
Nothing in [the update run][1] sounds like it's this PR, and it's Friday, and build02 and Azure are both struggling, so:

and we'll have lots of cook time in 4.11 release informers by the time we're back next week ;)

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/741/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1502310470520934400

PR #744, opened 2022-02-21
Consume post install static spec capabilities

jottofar:
Implements initial, postinstall, support of static capabilities
per Jira [OTA-567](https://issues.redhat.com/browse/OTA-567).

### DIFF
--- a/lib/capability/capability.go
+++ b/lib/capability/capability.go
@@ -1,0 +1,111 @@
+package capability
+
+import (
+	"fmt"
+	"strings"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+const (
+	CapabilityAnnotation = "capability.openshift.io/name"
+
+	DefaultCapabilitySet = configv1.ClusterVersionCapabilitySetCurrent
+)
+
+type ClusterCapabilities struct {
+	KnownCapabilities   map[configv1.ClusterVersionCapability]struct{}
+	EnabledCapabilities map[configv1.ClusterVersionCapability]struct{}
+}
+
+// SetCapabilities populates and returns cluster capabilities from ClusterVersion capabilities spec.
+func SetCapabilities(config *configv1.ClusterVersion) ClusterCapabilities {
+	var capabilities ClusterCapabilities
+	capabilities.KnownCapabilities = setKnownCapabilities(config)
+	capabilities.EnabledCapabilities = setEnabledCapabilities(config)
+	return capabilities
+}
+
+// GetCapabilitiesStatus populates and returns ClusterVersion capabilities status from given capabilities.
+func GetCapabilitiesStatus(capabilities ClusterCapabilities) configv1.ClusterVersionCapabilitiesStatus {
+	var status configv1.ClusterVersionCapabilitiesStatus
+	for k := range capabilities.EnabledCapabilities {
+		status.EnabledCapabilities = append(status.EnabledCapabilities, k)
+	}
+	for k := range capabilities.KnownCapabilities {
+		status.KnownCapabilities = append(status.KnownCapabilities, k)
+	}
+	return status
+}
+
+// CheckResourceEnablement, given resource annotations and defined cluster capabilities, checks if the capability
+// annotation exists. If so, each capability name is validated against the known set of capabilities. Each valid
+// capability is then checked if it is disabled. If any invalid capabilities are found an error is returned listing
+// all invalid capabilities. Otherwise, if any disabled capabilities are found an error is returned listing all
+// disabled capabilities.
+func CheckResourceEnablement(annotations map[string]string, capabilities ClusterCapabilities) error {
+	val, ok := annotations[CapabilityAnnotation]
+	if !ok {
+		return nil
+	}
+	caps := strings.Split(val, "+")
+	numCaps := len(caps)
+	unknownCaps := make([]string, 0, numCaps)
+	disabledCaps := make([]string, 0, numCaps)
+
+	for _, c := range caps {
+		if _, ok = capabilities.KnownCapabilities[configv1.ClusterVersionCapability(c)]; !ok {
+			unknownCaps = append(unknownCaps, c)
+		} else if _, ok = capabilities.EnabledCapabilities[configv1.ClusterVersionCapability(c)]; !ok {
+			disabledCaps = append(disabledCaps, c)
+		}
+	}
+	if len(unknownCaps) > 0 {
+		return fmt.Errorf("unrecognized capability names: %s", strings.Join(unknownCaps, ", "))
+	}
+	if len(disabledCaps) > 0 {
+		return fmt.Errorf("disabled capabilities: %s", strings.Join(disabledCaps, ", "))
+	}
+	return nil
+}
+
+// setKnownCapabilities populates a map keyed by capability from all known capabilities as defined in ClusterVersion.
+func setKnownCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterVersionCapability]struct{} {
+	known := make(map[configv1.ClusterVersionCapability]struct{})
+
+	for _, v := range configv1.ClusterVersionCapabilitySets {
+		for _, capability := range v {
+			if _, ok := known[capability]; ok {
+				continue
+			}
+			known[capability] = struct{}{}
+		}
+	}
+	return known
+}
+
+// setEnabledCapabilities populates a map keyed by capability from all enabled capabilities as defined in ClusterVersion.
+// DefaultCapabilitySet is used if a baseline capability set is not defined by ClusterVersion.
+func setEnabledCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterVersionCapability]struct{} {
+	enabled := make(map[configv1.ClusterVersionCapability]struct{})
+
+	capSet := DefaultCapabilitySet
+
+	if config.Spec.Capabilities != nil && len(config.Spec.Capabilities.BaselineCapabilitySet) > 0 {
+		capSet = config.Spec.Capabilities.BaselineCapabilitySet
+	}
+
+	for _, v := range configv1.ClusterVersionCapabilitySets[capSet] {
+		enabled[v] = struct{}{}
+	}
+
+	if config.Spec.Capabilities != nil {
+		for _, v := range config.Spec.Capabilities.AdditionalEnabledCapabilities {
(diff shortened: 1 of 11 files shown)


lib/capability/capability.go
```diff
@@ -0,0 +1,94 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+const CapabilityAnnotation = "capability.openshift.io/name"
+
+type ClusterCapabilities struct {
+	KnownCapabilities map[configv1.ClusterVersionCapability]struct{}
+
+	// value indicates whether capability was explicitly included
+	EnabledCapabilities map[configv1.ClusterVersionCapability]bool
```
wking, line 16: I'd have expected vendoring [`ClusterVersionCapabilitiesStatus`][1].  Is the local type with the `map` properties because you want a place to hold local state so you could explain tainting?  If so, possibly `map[configv1.ClusterVersionCapability]error` where `error` is `nil` for "spec asked for this" and non-nil for other enabled capabilities, explaining why the property was tainted in?

[1]: https://github.com/openshift/api/blob/7e3ffb09accd36fb0536fa0e69bed5d70cccd6e5/config/v1/types_cluster_version.go#L292-L305

lib/capability/capability.go
```diff
@@ -0,0 +1,94 @@
...
+			return fmt.Errorf("unrecognized capability names: %s", strings.Join(unknownCaps, ", "))
+		}
+		if len(disabledCaps) > 0 {
+			return fmt.Errorf("disabled capabilities: %s", strings.Join(disabledCaps, ", "))
+		}
+	}
+	return nil
+}
+
+func setKnownCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterVersionCapability]struct{} {
```
wking, line 64: nit: add a godoc like:

```go
// setKnownCapabilities populates status.capabilities.knownCapabilities on the ClusterVersion argument, and returns a map for convenient known-property lookup.
```

or some such.

lib/capability/capability.go
```diff
@@ -0,0 +1,94 @@
...
+	known := make(map[configv1.ClusterVersionCapability]struct{})
+
+	for _, v := range configv1.ClusterVersionCapabilitySets {
+		for _, capability := range v {
+			if _, ok := known[capability]; ok {
+				continue
+			}
+			known[capability] = struct{}{}
+			config.Status.Capabilities.KnownCapabilities = append(config.Status.Capabilities.KnownCapabilities, capability)
```
wking, line 73: Do we have reliable locking on this?  With the status sync'er running in parallel, I'd expect at least something like building a slice in a local variable, and after completing the local array, slotting it in with:

```go
config.Status.Capabilities.KnownCapabilities = knownCapabilities
```

lib/capability/capability.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+	return known
+}
+
+func setEnabledCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterVersionCapability]bool {
+	enabled := make(map[configv1.ClusterVersionCapability]bool)
+
+	for _, v := range configv1.ClusterVersionCapabilitySets[config.Spec.Capabilities.BaselineCapabilitySet] {
```
wking, line 82: [The API][1] docs `vCurrent` as the default cap set, but does not actually declare that default anywhere in an API property, so we'll need to define that defaulting locally here (in a way that we can easily factor it out into library-go later to support `oc adm release extract ...`).

[1]: https://github.com/openshift/api/blob/354aa98a475c1fcd60b41aaee735da95d7318b42/config/v1/types_cluster_version.go#L280-L281

lib/capability/capability.go
```diff
@@ -0,0 +1,94 @@
...
+	return status
+}
+
+func CheckResourceEnablement(annotations map[string]string, capabilities ClusterCapabilities) error {
+	val, ok := annotations[CapabilityAnnotation]
+	if ok {
+		caps := strings.Split(val, "+")
+		numCaps := len(caps)
+		unknownCaps := make([]string, numCaps)
+		disabledCaps := make([]string, numCaps)
```
wking, line 43: I think you want `make([]string, 0, numCaps)` for both of these to say "I want to allocate room for `numCaps` items, but I don't actually have any items for this slice yet".  As it stands, you are initializing `numCaps` empty-string items, and then below you're appending additional caps after those empty-string items.  [Go play][1].

[1]: https://go.dev/play/p/ekvG3pqKMm5

lib/capability/capability.go
jottofar: Originally, yes, it was to be able to distinguish implicitly enabled caps but it's no longer needed since that will come later. And yeah, I see it becoming whatever we need to report.
And use of maps for easy lookup.

lib/capability/capability.go
jottofar: This is old code and will be removed but the question of the need for a lock is valid.

The capabilities reside in get set `SyncWork` and are set in `Update` so are protected by the lock there. Just yesterday I added setting `CapabilitiesStatus` in `SyncWorkerStatus` also in `Update` [here](https://github.com/openshift/cluster-version-operator/blob/abafa6061c661debb0c37b07fab039c827f2df0c/pkg/cvo/sync_worker.go#L436) so it's available to `syncStatus` [here](https://github.com/openshift/cluster-version-operator/blob/abafa6061c661debb0c37b07fab039c827f2df0c/pkg/cvo/status.go#L204).

jottofar:
/retitle Consume post install static spec capabilities

jottofar:
retitle WIP: Consume post install static spec capabilities

[Get cluster version object earlier in startup](https://github.com/openshift/cluster-version-operator/pull/741) needs to land first and unit test fixes in progress.

jottofar:
/retitle WIP: Consume post install static spec capabilities

jottofar:
/retitle Consume post install static spec capabilities

lib/capability/capability.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// CheckResourceEnablement, given resource annotations and defined cluster capabilities, checks if the capability
+// annotation exists. If so, each capability name is validated against the known set of capabilities. Each valid
+// capability is then checked if it is disabled. If any invalid capabilities are found an error is returned listing
+// all invalid capabilities. Otherwise, if any disabled capabilities are found an error is returned listing all
+// disabled capabilities.
+func CheckResourceEnablement(annotations map[string]string, capabilities ClusterCapabilities) error {
+	val, ok := annotations[CapabilityAnnotation]
+	if ok {
```
wking, line 48: nit: save a layer of nesting with:

```go
if !ok {
    return nil
}
caps := ...
```

lib/capability/capability.go
```diff
@@ -0,0 +1,115 @@
...
+		caps := strings.Split(val, "+")
+		numCaps := len(caps)
+		unknownCaps := make([]string, 0, numCaps)
+		disabledCaps := make([]string, 0, numCaps)
+
+		for _, c := range caps {
+			trimmedCap := strings.TrimSpace(c)
+			if len(trimmedCap) == 0 {
+				continue
```
wking, line 57: This opens up space for accepting values like `a+b++c+   +d`.  I don't think we say we'll trim spaces anywhere.  Can we drop `trimmedCap` and use the raw `c` below?

lib/capability/capability_test.go
```diff
@@ -0,0 +1,275 @@
...
+	tests := []struct {
+		name            string
+		config          *configv1.ClusterVersion
+		wantKnownKeys   []string
+		wantEnabledKeys []string
+	}{
+		{name: "capabilities nil",
+			config:          &configv1.ClusterVersion{},
+			wantKnownKeys:   []string{string(configv1.ClusterVersionCapabilityOpenShiftSamples)},
+			wantEnabledKeys: []string{string(configv1.ClusterVersionCapabilityOpenShiftSamples)},
```
wking, line 19: This will break as folks continue to add caps, like openshift/api#1125.  Speaking of which, we will need to bump our vendored API version again, although I don't mind if we delay that until we get closer to feature freeze.  But anyway, I expect we want to use `configv1.ClusterVersionCapabilitySets[configv1.ClusterVersionCapabilitySetCurrent]` as the default value for enabled keys.  And probably for the time being assume that will be the full set of known keys too?  Otherwise this feels a lot like it's re-implementing `setKnownCapabilities`.

lib/capability/capability_test.go
```diff
@@ -0,0 +1,275 @@
...
+			},
+			wantKnownKeys:   []string{string(configv1.ClusterVersionCapabilityOpenShiftSamples)},
+			wantEnabledKeys: []string{string(configv1.ClusterVersionCapabilityOpenShiftSamples)},
+		},
+		{name: "set capabilities None",
+			config: &configv1.ClusterVersion{
+				Spec: configv1.ClusterVersionSpec{
+					Capabilities: &configv1.ClusterVersionCapabilitiesSpec{
+						BaselineCapabilitySet:         configv1.ClusterVersionCapabilitySetNone,
+						AdditionalEnabledCapabilities: []configv1.ClusterVersionCapability{},
```
wking, line 35: can we drop this line?  I expect most folks who use `None` to not also include an explicit empty array for additional caps.

lib/capability/capability.go
jottofar: sure. You guys are really anti-nesting :)

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -2236,7 +2252,7 @@ func TestCVO_RestartAndReconcile(t *testing.T) {
 		t.Fatal(err)
 	}
 	actions = client.Actions()
-	if len(actions) != 1 {
+	if len(actions) != 2 {
```
wking, line 2255: what is the second action here?

lib/capability/capability.go
jottofar: I okay with either but figured since we didn't say they couldn't put spaces I'd check. However, I did that when we were using a coma where a space would be somewhat natural. Not the case with `+` so yeah, I'll just drop it.

wking (approved):
/lgtm

[sandbox and PBD issues][1] are orthogonal.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/744/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1503373665729777664

PR #745, opened 2022-02-21
Bug 1822752: pkg/cvo: Fix ups from separating load from apply #683

jottofar:
Fix remaining comments from https://github.com/openshift/cluster-version-operator/pull/683.

Remove unused parameter lister `configlistersv1.ClusterVersionLister` from `SyncWorker.Update`.

PR #753, opened 2022-03-17
Bug 2064991: pkg/cvo: Separate payload load from payload apply

wking:
Bringing #683 and #745 back to 4.10.


LalatenduMohanty (approved):
/lgtm

PR #754, opened 2022-03-22
capability: disallow disabling, add enabling capabilities

jottofar:
This PR adds support for the following per the. Post-install, an admin can change a capability to enable it but cannot disable a previously enabled capability. This PR changes CVO to override any admin attempt to disable a previously enabled capability and reports any such action through a newly added condition showing these "tainted-in" capabilities. These "tainted-in" capabilities will remain enabled and will be reported as such in the status.

Supporting this new logic requires that capability filtering be moved from payload load time to payload apply time in a similar fashion to how overrides are handled.


lib/capability/capability.go
```diff
@@ -25,12 +25,17 @@ func (caps capabilitiesSort) Len() int           { return len(caps) }
 func (caps capabilitiesSort) Swap(i, j int)      { caps[i], caps[j] = caps[j], caps[i] }
 func (caps capabilitiesSort) Less(i, j int) bool { return string(caps[i]) < string(caps[j]) }
 
-// SetCapabilities populates and returns cluster capabilities from ClusterVersion capabilities spec.
-func SetCapabilities(config *configv1.ClusterVersion) ClusterCapabilities {
+// SetCapabilities populates and returns cluster capabilities from ClusterVersion capabilities spec. This method also
+// ensures that no previousily enabled capability is now disabled and returns any such (tainted-in) capabilities.
+func SetCapabilities(config *configv1.ClusterVersion,
+	existingEnabled map[configv1.ClusterVersionCapability]struct{}) (ClusterCapabilities, []configv1.ClusterVersionCapability) {
```
wking, line 31: Why the separate return field, instead of adding this as a new property in `ClusterCapabilities`, or converting `ClusterCapabilities.EnabledCapabilities` to a `map[configv1.ClusterVersionCapability]bool` for "tainted vs. requested"?

lib/capability/capability.go
```diff
@@ -47,35 +52,43 @@ func GetCapabilitiesStatus(capabilities ClusterCapabilities) configv1.ClusterVer
...
+	for _, c := range resourceCapabilities {
+		if _, ok := enabledCapabilities[c]; !ok {
+			return false
+		}
 	}
-	return nil
+	return true
 }
 
 // setKnownCapabilities populates a map keyed by capability from all known capabilities as defined in ClusterVersion.
```
wking, line 94: orthogonal to this PR, but I'm realizing that `setKnownCapabilities` doesn't need its current `config *configv1.ClusterVersion` argument.

lib/capability/capability.go
```diff
@@ -94,8 +107,13 @@ func setKnownCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterV
 }
 
 // setEnabledCapabilities populates a map keyed by capability from all enabled capabilities as defined in ClusterVersion.
-// DefaultCapabilitySet is used if a baseline capability set is not defined by ClusterVersion.
-func setEnabledCapabilities(config *configv1.ClusterVersion) map[configv1.ClusterVersionCapability]struct{} {
+// DefaultCapabilitySet is used if a baseline capability set is not defined by ClusterVersion. A check is then made to
+// ensure that no previousily enabled capability is now disabled and if any such capabilities are found each is enabled,
+// saved, and returned.
+func setEnabledCapabilities(config *configv1.ClusterVersion,
+	priorEnabled map[configv1.ClusterVersionCapability]struct{}) (map[configv1.ClusterVersionCapability]struct{},
```
wking, line 114: It feels a bit strange to pass through both `priorEnabled` and a `*configv1.ClusterVersion` (whose `status.capabilities` will also include a set of prior, enabled caps.  To make the distinction clear, can we adjust the function signature here to take `*configv1.ClusterVersionCapabilitiesSpec` and `priorEnabled`?

pkg/cvo/status.go
```diff
@@ -354,6 +360,33 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
+			Status:             configv1.ConditionTrue,
+			Reason:             "Attempt to disable previousily enabled capability",
+			Message:            message,
+			LastTransitionTime: now,
+		})
+	} else {
+		resourcemerge.SetOperatorStatusCondition(&config.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:               TaintedCapabilities,
+			Status:             configv1.ConditionFalse,
+			LastTransitionTime: now,
```
wking, line 385: can we get an all-is-well reason/message here, like:

```console
$ git --no-pager grep -A1 AsExpected
pkg/cvo/availableupdates.go:                            Reason:  "AsExpected",
pkg/cvo/availableupdates.go-                            Message: "The update is recommended, because none of the conditional update risks apply to this cluster.",
```

Maybe `AsExpected` and `All enabled capabilities are requested by the configured spec.`, or some such?

pkg/cvo/sync_worker.go
```diff
@@ -220,6 +225,7 @@ func (w *SyncWorker) StatusCh() <-chan SyncWorkerStatus {
 }
 
 func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork, reporter StatusReporter) error {
+
```
wking, line 228: nit: we can probably drop this addition ;)

pkg/cvo/sync_worker.go
```diff
@@ -818,6 +839,10 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 					klog.V(2).Infof("Skipping precreation of %s as unmanaged", task)
 					continue
 				}
+				if !capability.ResourceEnabled(resourceCapabilities[i], work.Capabilities.EnabledCapabilities) {
+					klog.V(2).Infof("Skipping precreation of %s, part of disabled capability", task)
```
wking, line 843: I'd rather be more explicit and have `capability.ResourceEnabled` return an `error` when we decide to skip, explaining why.  Then we can `"Skipping precreation of %s: %s", task, err)` here.

pkg/payload/payload.go
```diff
@@ -100,6 +100,11 @@ const (
 	DefaultClusterProfile = "self-managed-high-availability"
 )
 
+type ManifestWithCapabilities struct {
+	Man          manifest.Manifest
+	Capabilities []configv1.ClusterVersionCapability
```
wking, line 105: Hmm, do we want to defer this until we get openshift/library-go#1333 ?  Because it seems like unnecessary churn to land this shift, then land openshift/library-go#1333 , and then roll back this shift.  But it's not clear to me how invasive this particular pivot is.  If it's fairly small, doing that dance might be worth not blocking on 1333.

lib/capability/capability.go
jottofar: The reason I changed this and removed the `bool` was so I could easily compare for update. I can then use `equalSyncWork` and fold capabilities in just like overrides. The other option of course is to have a capabilities specific equals method which gets called from `equalSyncWork`, which ignores the `bool`.

pkg/cvo/sync_worker.go
jottofar: Why have it differ from the other checks in this same context?  `ResourceEnabled` is simply that, is this resource enabled given it's configured capabilities, with no knowledge of the impact of it being enabled or not. If we really want a method returning an error I would create another but to me it's an invalid logic shift from `SyncWorker.apply` to `capability` package.

pkg/payload/payload.go
jottofar: I didn't want to wait on that. I thought it more important to get CVO working the way we think it should regardless of where some of the code ends up. If I don't do it this way then I'd have to read the manifest annotations again at this point which I know will be dead code and churn anyway.

pkg/cvo/sync_worker.go
jottofar: Okay, so I see now. Have it list exactly which capability(s) is preventing enablement?

lib/capability/capability.go
```diff
@@ -15,8 +16,14 @@ const (
...
-	KnownCapabilities   map[configv1.ClusterVersionCapability]struct{}
-	EnabledCapabilities map[configv1.ClusterVersionCapability]struct{}
+	KnownCapabilities     map[configv1.ClusterVersionCapability]struct{}
+	EnabledCapabilities   map[configv1.ClusterVersionCapability]struct{}
+	TaintedInCapabilities []configv1.ClusterVersionCapability
+}
+
+func (c *ClusterCapabilities) Equal(capabilities *ClusterCapabilities) bool {
+	return reflect.DeepEqual(c.KnownCapabilities, capabilities.KnownCapabilities) &&
+		reflect.DeepEqual(c.EnabledCapabilities, capabilities.EnabledCapabilities)
```
wking, line 26: This is the only bit that matters for `equalSyncWork`, isn't it?  `KnownCapabilities` would only ever change as a result of a target-release change.  And even then, `KnownCapabilities` changes would only affect "why we're skipping this manifest" log messages.  I'm not strongly against having this `Equal` method, but I think all we need is this `reflect.DeepEqual(c.EnabledCapabilities, capabilities.EnabledCapabilities)` inlined in `equalSyncWork`

pkg/payload/payload.go
wking: Oh, hmm, yes, we will need something in the manifest library about scraping caps of sets of manifests...

pkg/cvo/sync_worker.go
```diff
@@ -818,6 +840,10 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 					klog.V(2).Infof("Skipping precreation of %s as unmanaged", task)
 					continue
 				}
+				if err := capability.ResourceEnabled(resourceCapabilities[i], work.Capabilities.EnabledCapabilities); err != nil {
```
wking, line 843: I'm a bit confused by this.  Is it "include the manifest if the capabilities that that manifest needs match the capabilities we pulled off of the manifest when we loaded it"?  Why can't we pass in the cluster's current set of enabled capabilities, without needing to `[i]` index anything?

pkg/payload/payload.go
wking: ok, thinking this through a bit more, I think openshift/library-go#1333`'s `Include` is sufficient to cover your current `capability.ResourceEnabled` and `ManifestWithCapabilities` functionality.  We'll need something new for update-tainting, to compare the incoming manifest set with the outgoing included manifest set and figure out what caps to taint in, but we don't need that for this PR's non-update `spec.capabilities` case.

jottofar:
/retitle capability: disallow disabling, add enabling capabilities

pkg/payload/payload.go
```diff
@@ -134,13 +133,15 @@ type metadata struct {
...
 func LoadUpdate(dir, releaseImage, excludeIdentifier string, includeTechPreview bool, profile string,
-	capabilities capability.ClusterCapabilities) (*Update, error) {
+	knownCapabilities []configv1.ClusterVersionCapability) (*Update, error) {
 
 	payload, tasks, err := loadUpdatePayloadMetadata(dir, releaseImage, profile)
 	if err != nil {
 		return nil, err
 	}
 
+	onlyKnownCaps := configv1.ClusterVersionCapabilitiesStatus{knownCapabilities, knownCapabilities}
```
wking, line 143: Can we use explicit properties here:

```golang
onlyKnownCaps := configv1.ClusterVersionCapabilitiesStatus{
  EnabledCapabilities: knownCapabilities, // enabled-caps filtering happens at apply time, so don't filter on that here
  KnownCapabilities:   knownCapabilities,
}
```

Or some such?  I don't expect `ClusterVersionCapabilitiesStatus` to be all that volatile, but if it does add additional properties in the future, the explicit references here seem like they'd make life easier on the future dev adjusting this.

pkg/payload/payload.go
```diff
@@ -134,13 +133,19 @@ type metadata struct {
...
 	payload, tasks, err := loadUpdatePayloadMetadata(dir, releaseImage, profile)
 	if err != nil {
 		return nil, err
 	}
 
+	// We only pass known capabilities at payload load time to avoid doing any capability
+	// enable filtering which only occurs during apply.
+	onlyKnownCaps := configv1.ClusterVersionCapabilitiesStatus{
+		EnabledCapabilities: knownCapabilities,
+		KnownCapabilities:   knownCapabilities}
```
LalatenduMohanty, line 147: `EnabledCapabilities: knownCapabilities` does not look right. At this point we do not know what is enabled, do we?

LalatenduMohanty, line 147: `EnabledCapabilities: knownCapabilities` does not look right. At this point we do not know what is enabled, do we?

pkg/cvo/sync_worker.go
```diff
@@ -803,6 +823,11 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 
 	// update each object
 	errs := payload.RunGraph(ctx, graph, maxWorkers, func(ctx context.Context, tasks []*payload.Task) error {
+
+		capabilities := configv1.ClusterVersionCapabilitiesStatus{
+			EnabledCapabilities: capability.GetCapabilities(work.Capabilities.EnabledCapabilities),
+			KnownCapabilities:   capability.GetKnownCapabilities()}
```
wking, line 829: Can we replace this with:

```go
capabilties := capability.GetCapabilitiesStatus(work.Capabilities)
```

Then we could drop `GetCapabilities`.

wking: also, I'm not all that concerned about the weight of the `GetCapabilitiesStatus` call, but we could shift it up outside of `RunGraph`, because all the nodes in the graph will want the result, and none of them need to mutate `capabilities`.

LalatenduMohanty (changes requested):

pkg/cvo/status.go
```diff
@@ -354,6 +360,35 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
+		caps := make([]string, len(tainted))
+		for i, c := range tainted {
+			caps[i] = string(c)
+		}
+		message = message + strings.Join([]string(caps), ", ")
+
+		resourcemerge.SetOperatorStatusCondition(&config.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:               TaintedCapabilities,
+			Status:             configv1.ConditionTrue,
+			Reason:             "Attempt to disable previousily enabled capability",
```
wking, line 377: `Reason` should be [a CamelCase slug][1].  Maybe `CapabilitiesImplicitlyEnabled` or whatever, to use @bparees' preferred words (assembled into a camel CaseSlug by me).  Maybe just echo the `Type`?  I don't think we're planning on putting in the effort to distinguish the various cases for implicitly enabling caps.

[1]: https://github.com/openshift/api/blob/67c28690af52a69e0b8fa565916fe1b9b7f52f10/config/v1/types_cluster_operator.go#L131

pkg/payload/payload.go
wking: We do know enabled caps here, but as the comment above points out, we are deferring enabled-cap filtering to apply-time (so we don't have to reload from disk if the admin enables an additional cap).

pkg/cvo/status.go
```diff
@@ -159,6 +159,9 @@ const ClusterVersionInvalid configv1.ClusterStatusConditionType = "Invalid"
 // and no failures occurred during image verification and precondition checking.
 const DesiredReleaseAccepted configv1.ClusterStatusConditionType = "ReleaseAccepted"
 
+// ImplicitlyEnabledCapabilities
```
wking, line 162:
```go
// ImplicitlyEnabledCapabilities is True if there are enabled capabilities which the user is not currently
// requesting via spec.capabilities, because the cluster version operator does not support uninstalling
// capabilities if any associated resources were previously managed by the CVO.
```

or some such?

pkg/cvo/sync_worker.go
```diff
@@ -379,20 +386,26 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
...
 
 	if work.Empty() {
 		klog.V(2).Info("Update work has no release image; ignoring requested change")
 		return w.status.DeepCopy()
 	}
 
-	work.Capabilities = capability.SetCapabilities(config)
-	versionEqual, overridesEqual := equalSyncWork(w.work, work, fmt.Sprintf("considering cluster version generation %d", generation))
+	work.Capabilities = capability.SetCapabilities(config, priorCaps)
+	w.status.CapabilitiesStatus.ImplicitlyEnabledCaps = work.Capabilities.ImplicitlyEnabledCapabilities
```
wking, line 403: don't we want this down under the `equalSyncWork` guard if/when we set `w.work`?

wking (approved):
/lgtm

wking:
[MCD `TargetDown`][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/754/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1509635926538588160

PR #759, opened 2022-04-01
Bug 2070854: syncWorkerStatus: Avoid saving stale status values

jottofar:
which can occur since the splitting of payload load and apply to be called in separate goroutines https://github.com/openshift/cluster-version-operator/pull/683. Each thread will now call it's own `update status` function that only sets status values it manages.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -580,12 +580,15 @@ func (w *statusWrapper) ValidPayloadStatus(release configv1.Release) bool {
 	return equalDigest(w.previousStatus.loadPayloadStatus.Release.Image, release.Image)
 }
 
+// ReportPayload reports payload load status.
 func (w *statusWrapper) ReportPayload(payloadStatus LoadPayloadStatus) {
 	status := w.previousStatus
 	status.loadPayloadStatus = payloadStatus
-	w.w.updateStatus(*status)
+	w.w.updateLoadStatus(*status)
 }
 
+// Report reports payload application status. It does not overwrite payload load status and capabilities status
+// since payload application does not update these statuses they could therefore be out-of-date.
 func (w *statusWrapper) Report(status SyncWorkerStatus) {
 	p := w.previousStatus
 	var fractionComplete float32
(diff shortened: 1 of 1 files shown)


wking:
[Seems unrelated][1]:

```
: operator conditions node-tuning | 0s
-- | --
{Operator progressing (ProfileProgressing): Waiting for 1/6 Profiles to be applied  Operator progressing (ProfileProgressing): Waiting for 1/6 Profiles to be applied}
```

But also means we have no update coverage at all.  I'm pretty confident that this change did not break updates though, so:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/759/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1511442289912713216

pkg/cvo/sync_worker.go
```diff
@@ -713,11 +716,18 @@ func equalSyncWork(a, b *SyncWork, context string) (equalVersion, equalOverrides
 
 // updateStatus records the current status of the sync action for observation
 // by others. It sends a copy of the update to the report channel for improved
-// testability.
-func (w *SyncWorker) updateStatus(update SyncWorkerStatus) {
+// testability. It does not overwrite payload load status and capabilities status
+// when payloadApplication is true since these statuses could be out-of-date.
+func (w *SyncWorker) updateStatus(update SyncWorkerStatus, payloadApplication bool) {
```
wking, line 721: Hmm, if we have payload-loaders and payload-appliers both feeding this method, couldn't one always have stale information about the other?  Can we set this up so they only push the information that they're responsible for, with pointers or other "I don't have an opinion" markers so they can clearly leave the properties they aren't responsible for alone?  Maybe `updateStatus(*loadStatus, *applyStatus)` or something like that?

jottofar: > Can we set this up so they only push the information that they're responsible

That's essentially what I have using the `bool`. But I had also thought about simply making another update method so I'll do that. I think it will be easier to understand than a `bool` or using `nil`s.

jottofar:
/retitle Bug 2070854: syncWorkerStatus: Avoid saving stale status values

wking (approved):
/lgtm

PR #760, opened 2022-04-04
Bug 2070805: pkg/cvo/updatepayload: Prune previous payload downloads

wking:
[Avoid][1]:

    mv: inter-device move failed: '/manifests' to '/etc/cvo/updatepayloads/HbO7IDc7tyIg9utw3sd_tg/manifests/manifests'; unable to remove target: Directory not empty

by pruning all scratch directories before downloading a new payload.  Also shift the job pruning up here too, so it's all collected together.  This avoids sticking subsequent download jobs after a corrupted download attempt (e.g. if something terminates a download job part way through its `mv` call, subsequent download ).  Reproducing this issue locally, with `/tmp` on a different filesystem:

```console
$ mkdir -p a /tmp/a/a
$ /bin/mv a /tmp
/bin/mv: inter-device move failed: 'a' to '/tmp/a'; unable to remove target: Directory not empty
$ /bin/mv --version
mv (GNU coreutils) 8.31
...
```

The directory pruning will also avoid leaking previous sets of manifests.  Before this commit, the `/etc/cvo/updatepayloads` directories on the control plane nodes would gradually accumulate more and more release manifest sets as new targets were downloaded for inspection.  The manifest sets aren't that big:

```console
$ oc adm release extract --to manifests quay.io/openshift-release-dev/ocp-release:4.10.6-x86_64
Extracted release payload from digest sha256:88b394e633e09dc23aa1f1a61ededd8e52478edf34b51a7dbbb21d9abde2511a created at 2022-03-23T07:34:52Z
$ du -hs manifests
6.5M    manifests
```

But still, better to have the pruning than leak forever.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2070805#c0

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -135,14 +135,28 @@ func (r *payloadRetriever) targetUpdatePayloadDir(ctx context.Context, update co
 	hash := md5.New()
 	hash.Write([]byte(update.Image))
 	payloadHash := base64.RawURLEncoding.EncodeToString(hash.Sum(nil))
-
 	tdir := filepath.Join(r.workingDir, payloadHash)
-	err := payload.ValidateDirectory(tdir)
-	if os.IsNotExist(err) {
-		// the dirs don't exist, try fetching the payload to tdir.
-		err = r.fetchUpdatePayloadToDir(ctx, tdir, update)
+
+	// Prune older jobs and directories while gracefully handling errors.
+	if err := r.pruneJobs(ctx, 0); err != nil {
+		klog.Warningf("failed to prune jobs: %v", err)
 	}
-	if err != nil {
+	if files, err := os.ReadDir(r.workingDir); err != nil {
+		klog.Warningf("failed to list update payload directory: %v", err)
+	} else {
+		for _, file := range files {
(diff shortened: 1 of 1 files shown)


wking:
[`pods should successfully create sandboxes by not timing out` is the only failure][1], and that's orthogonal.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/760/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1510874470456233984

pkg/cvo/updatepayload.go
```diff
@@ -248,12 +255,6 @@ func (r *payloadRetriever) pruneJobs(ctx context.Context, retain int) error {
 		// Ignore jobs not beginning with operatorName
 		case !strings.HasPrefix(job.Name, r.operatorName+"-"):
 			break
```
jottofar, line 257: Why is it that we no longer have to ignore jobs that have not yet started or are still active? Will these never show up now because they will have been pruned?

wking: when we are getting ready to launch the new job, we have already given up on the old job, so we are pruning those, even if they were still pending or running.  The new job will succeed, or not, and will remain around until the CVO decides to launch a successor.

PR #761, opened 2022-04-04
Bug 2070887: pkg/cvo/sync_worker.go: set implicitly enabled caps earlier

jottofar:
since implicitly enabled capability changes do not trigger !capabilitiesEqual so we return before setting the status.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -404,11 +404,13 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 	versionEqual, overridesEqual, capabilitiesEqual :=
 		equalSyncWork(w.work, work, fmt.Sprintf("considering cluster version generation %d", generation))
 
+	// needs to be set here since changes in implicitly enabled capabilities are not considered a "capabilities change"
+	w.status.CapabilitiesStatus.ImplicitlyEnabledCaps = work.Capabilities.ImplicitlyEnabledCapabilities
+
 	if versionEqual && overridesEqual && capabilitiesEqual {
 		klog.V(2).Info("Update work is equal to current target; no change required")
 		return w.status.DeepCopy()
 	}
-	w.status.CapabilitiesStatus.ImplicitlyEnabledCaps = work.Capabilities.ImplicitlyEnabledCapabilities
 
 	// initialize the reconciliation flag and the status the first time
 	// update is invoked


pkg/cvo/sync_worker.go
```diff
@@ -401,14 +401,16 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 
 	work.Capabilities = capability.SetCapabilities(config, priorCaps)
 
+	// needs to be set here since changes in implicitly enabled capabilities are not considered a "capabilities change"
```
wking, line 404: But:

 * `equalSyncWork` [uses `ClusterCapabilities.Equal`][1], and
 * `ClusterCapabilities.Equal` [only looks at `EnabledCapabilities`][2].

I think we may want to change [this line][3] to:

```go
capabilitiesError := a.Capabilities.Equal(&b.Capabilities)
...
if capabilitiesError != nil {
  msgs = append(msgs, fmt.Sprintf("capabilities changed (%v)", capabilitiesChanged))
}
```

So we don't have to make local guesses about how the caps changed, but I don't understand where `ImplicitlyEnabledCaps` / `ImplicitlyEnabledCapabilities` is coming into this.

[1]: https://github.com/openshift/cluster-version-operator/blob/779b102a550e7622a6f885e7b41207b12c585704/pkg/cvo/sync_worker.go#L696
[2]: https://github.com/openshift/cluster-version-operator/blob/779b102a550e7622a6f885e7b41207b12c585704/lib/capability/capability.go#L23
[3]: https://github.com/openshift/cluster-version-operator/blob/779b102a550e7622a6f885e7b41207b12c585704/pkg/cvo/sync_worker.go#L706

wking: > I think we may want to change this line to...

I've floated #762 with this tweak.

wking: oh, I think I understand now.  This isn't about adjusting what `equalSyncWork` does, it's about ensuring the new `ImplicitlyEnabledCapabilities` make it through to `w.status.CapabilitiesStatus.ImplicitlyEnabledCaps`, regardless of where this function later returns.  Can we make that distinction a bit more obvious with something like:

```go
versionEqual, overridesEqual, capabilitiesEqual :=
	equalSyncWork(w.work, work, fmt.Sprintf("considering cluster version generation %d", generation))

// Changes to implicitly enabled capabilities are not considered a "capabilities change",
// but still need to be propagated through and reported in status.
w.status.CapabilitiesStatus.ImplicitlyEnabledCaps = work.Capabilities.ImplicitlyEnabledCapabilities

if versionEqual && overridesEqual && capabilitiesEqual {
	...
```

If the positioning you have today makes more sense than what I'm suggesting here, that's fine with me too.

jottofar: Hmm, maybe I'm misunderstanding your comment and suggestion but let me try and explain my change and reasoning again. The `capability.Equal` works fine as is and there's no "guessing" about how caps have changed.

- The only caps change we care about for the purposes of an update are changes to `EnabledCapabilities`. `Equal` works correctly in this regard.
-  We only care about `ImplicitlyEnabledCapabilities` for purposes of capturing them in status.

So for example, post-install C1 and C2 are enabled. Admin then changes C2 to disabled. On next `Update` `SetCapabilities` will set C2 back to enabled and save C2 into `ImplicitlyEnabledCapabilities`. `equalSyncWork` properly returns no change since `EnabledCapabilities` haven't changed. Before my change `Update` would return [here](https://github.com/openshift/cluster-version-operator/blob/033a9d8399163c9ef5d57339eb481ed589281e8c/pkg/cvo/sync_worker.go#L412) before `w.status.CapabilitiesStatus.ImplicitlyEnabledCaps` was updated.

wking (approved):
:lgtm:

PR #765, opened 2022-04-18
Bug 2070805: pkg/cvo/updatepayload: Shift previous-download removal into the job

wking:
I'd attempted to add this cleanup on the CVO side in 507b474632 (#760), but @evakhoni pointed out that that doesn't work because the CVO's volume mount is `readOnly: true`:

    2022-04-11T20:42:32.056428784Z W0411 20:42:32.056387       1 updatepayload.go:149] failed to prune update payload directory: unlinkat /etc/cvo/updatepayloads/brRTeZnZSIkZ2J2YMdQozg: read-only file system

This commit shifts removal into the job.

And because we want that removal to also cover "a previous job run half-populated the target release directory", I've removed the pre-fetch `ValidateDirectory` check.  In cases where we already had valid content locally, that's one extra download, which seems like a lower price than assuming a previous partial download was actually complete.  If we see many extra downloads, we use atomic renames or something so it's easy for an incoming CVO to distinguish between content that was completely populated by a previous job run from content that was incompletely populated.

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -161,15 +161,35 @@ func (r *payloadRetriever) targetUpdatePayloadDir(ctx context.Context, update co
 func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir string, update configv1.Update) error {
 	var (
 		version         = update.Version
-		payload         = update.Image
+		image           = update.Image
 		name            = fmt.Sprintf("%s-%s-%s", r.operatorName, version, randutil.String(5))
 		namespace       = r.namespace
 		deadline        = pointer.Int64Ptr(2 * 60)
 		nodeSelectorKey = "node-role.kubernetes.io/master"
 		nodename        = r.nodeName
-		cmd             = []string{"/bin/sh"}
-		args            = []string{"-c", copyPayloadCmd(dir)}
 	)
+
+	baseDir, targetName := filepath.Split(dir)
+	tmpDir := filepath.Join(baseDir, fmt.Sprintf("%s-%s", targetName, randutil.String(5)))
+
+	setContainerDefaults := func(container corev1.Container) corev1.Container {
+		container.Image = image
+		container.VolumeMounts = []corev1.VolumeMount{{
+			MountPath: targetUpdatePayloadsDir,
+			Name:      "payloads",
+		}}
+		container.SecurityContext = &corev1.SecurityContext{
+			Privileged: pointer.BoolPtr(true),
+		}
+		container.Resources = corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:              resource.MustParse("10m"),
+				corev1.ResourceMemory:           resource.MustParse("50Mi"),
+				corev1.ResourceEphemeralStorage: resource.MustParse("2Mi"),
+			},
+		}
+		return container
+	}
 
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


jottofar:
>  In cases where we already had valid content locally, that's one extra download, which seems like a lower price than assuming a previous partial download was actually complete. If we see many extra download

With https://bugzilla.redhat.com/show_bug.cgi?id=2072389 still open a typical `extra download` scenario is payload fails signature validation and user comes back and uses`force`. However I'm thinking the fix for the aforementioned bug is to continually `syncPayload` (and thereby recheck preconditions) if the reason for payload load failure is `precondition failed` which, with this change, will continually download the payload too if I'm understanding correctly.

pkg/cvo/updatepayload.go
```diff
@@ -295,6 +280,8 @@ func (r *payloadRetriever) pruneJobs(ctx context.Context, retain int) error {
 // `mkdir -p <target dir> && mv <default release manifest dir> <target release manifests dir>`
```
jottofar, line 280: Update, now three commands

wking: Good catch.  No need to echo what folks can read a few lines down in this internal function, so I've just dropped this from the docstring with 624f7fd0 -> d70b3c9e.

wking:
> However I'm thinking the fix for the aforementioned bug is to continually `syncPayload` (and thereby recheck preconditions) if the reason for payload load failure is `precondition failed` which, with this change, will continually download the payload too if I'm understanding correctly.

I've pushed d70b3c9e -> dee4b9be to pivot to populating `{baseDir}/{targetName}-{randomSuffix}` and then atomically rename that populated result to `{baseDir}/{targetName}`.  That way, future `targetUpdatePayloadDir` calls using the `ValidateDirectory` precheck will successfully distinguish "previous job got everything for this release" (where we don't want a new download job) from "previous job got some bits over, but not everything" (where we do want a new download job).

pkg/cvo/updatepayload.go
```diff
@@ -290,20 +280,24 @@ func (r *payloadRetriever) pruneJobs(ctx context.Context, retain int) error {
...
+
+	fromCVOPath := filepath.Join(payload.DefaultPayloadDir, payload.CVOManifestDir)
+	toCVOPath := filepath.Join(tmpDir, payload.CVOManifestDir)
+	cvoCmd := fmt.Sprintf("mv %s %s", fromCVOPath, toCVOPath)
+
+	fromReleasePath := filepath.Join(payload.DefaultPayloadDir, payload.ReleaseManifestDir)
+	toReleasePath := filepath.Join(tmpDir, payload.ReleaseManifestDir)
+	releaseCmd := fmt.Sprintf("mv %s %s", fromReleasePath, toReleasePath)
+
+	moveInPlaceCmd := fmt.Sprintf("mv %s %s", tmpDirCmd, tdir)
```
jottofar, line 299: `tmpDirCmd` -> `tmpDir`

jottofar, line 299: I think you may have to first create `tdir` to make the `rename` atomic, https://man7.org/linux/man-pages/man2/rename.2.html

wking: Thanks; fixed in dee4b9be -> 295db8e9.

wking: From [POSIX][1]:

> This `rename()` function is equivalent for regular files to that defined by the ISO C standard. Its inclusion here expands that definition to include actions on directories and specifies behavior when the new parameter names a file that already exists. That specification requires that the action of the function be atomic.

The "If `newpath` already exists ..." section of the Linux man page is also in the "... and also when the target does exist..." category.  Although it would be nice if someone said "this is atomic when the target does not exist" more explicitly on either side.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/rename.html

wking: [This article][1] talks through the atomic rename approach, and from that, it looks like we might need a pre-rename `fsync` to be very secure about the transaction.  I'm not sure that covering the kernel-crash corner case is worth the overhead of calling `fsync`.

[1]: https://lwn.net/Articles/789600/

wking: I've pushed 4be737e0 -> a5af89d3 with an updated commit message to walk through this.

wking:
> Or better, I think in this case instead of returning commands-as-strings you could return an array of strings and avoid /bin/sh interpreting them at all.

I've added c45a9813 with a pivot to `initContainers` for command chaining, so I can avoid `sh` and `&&`.

wking:
[`Watchdog` gap][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/765/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1518666600574619648

PR #766, opened 2022-04-18
Bug 2072389: Do not save desired update on load failures

jottofar:
since this will cause `equalSyncWork` to continually return `true` and no reattempt to load the payload will occur until the desired update is updated by the user again, e.g. image change or change to `force`. This was accomplished by moving `work` update, including the desired version, after the desired version payload has been successfully loaded.

See commit message for more details on precondition failures in general and the `RecentEtcdBackup` precondition specifically.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -439,11 +439,8 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 		oldDesired = &w.work.Desired
 	}
 
-	w.work = work
-
-	if !versionEqual && oldDesired == nil {
-		klog.Infof("Propagating initial target version %v to sync worker loop in state %s.", desired, state)
-	} else if !versionEqual && state == payload.InitializingPayload {
+	// since oldDesired is not nil this is not the first time update is invoked and therefore w.work is not nil
+	if !versionEqual && oldDesired != nil && state == payload.InitializingPayload {
 		klog.Warningf("Ignoring detected version change from %v to %v during payload initialization", *oldDesired, work.Desired)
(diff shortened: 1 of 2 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -435,11 +435,13 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 				Image:   work.Desired.Image,
 			},
 		}
+		w.work = work
```
wking, line 438: setting this before the `w.loadUpdatedPayload` feels weird.  Why do we need this?

pkg/cvo/sync_worker.go
```diff
@@ -435,11 +435,13 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
...
 	} else {
 		oldDesired = &w.work.Desired
 	}
 
-	w.work = work
+	// save latest capability settings
+	w.work.Capabilities = work.Capabilities
 
 	if !versionEqual && oldDesired == nil {
 		klog.Infof("Propagating initial target version %v to sync worker loop in state %s.", desired, state)
```
wking, line 447: we may want to shift this log block to after the `w.loadUpdatedPayload` too.  We want to keep the `Ignoring detected version change...` bailout block before the `w.loadUpdatedPayload`.

pkg/cvo/sync_worker.go
jottofar: It is weird. Wanted to get `Capabilities` saved since they can change independent of payload. Changed such that I only save capabilities and not all of work.

wking (approved):
/lgtm

PR #768, opened 2022-04-28
Bug 2079789: pkg/cvo/sync_worker.go: Initialize implicitlyEnabledCaps

jottofar:
so `syncPayload` does not return nil value and stomp on valid implicitly enabled capabilities value.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -224,10 +224,17 @@ func (w *SyncWorker) StatusCh() <-chan SyncWorkerStatus {
 	return w.report
 }
 
+// syncPayload retrieves, loads, and verifies the specified payload, aka sync's the payload, whenever there is no current
+// payload or the current payload differs from the desired payload. Whenever a payload is sync'ed a check is made for
+// implicitly enabled capabilities. For the purposes of the check made here, implicitly enabled capabilities are
+// capabilities which are disabled in the desired payload but must remain enabled since the capability defines one or
+// more resources which are enabled in the current payload. All such capabilities are returned along with any previously
+// existing implicitly enabled capabilities. If no new implicitly enabled capabilities are found, just the previously
+// existing implicitly enabled capabilities are returned.
 func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork,
 	reporter StatusReporter) ([]configv1.ClusterVersionCapability, error) {
 
-	var implicitlyEnabledCaps []configv1.ClusterVersionCapability
+	implicitlyEnabledCaps := work.Capabilities.ImplicitlyEnabledCapabilities
 
 	desired := configv1.Release{
 		Version: work.Desired.Version,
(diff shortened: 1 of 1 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -227,7 +227,7 @@ func (w *SyncWorker) StatusCh() <-chan SyncWorkerStatus {
 func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork,
 	reporter StatusReporter) ([]configv1.ClusterVersionCapability, error) {
```
wking, line 228: We should grow godocs here, so the semantics of the returned caps list are explicitly declared.  The returned caps list is from a313095a1c3 (#758), and gets passed up through `loadUpdatedPayload` to `Update` where it is fed into `SetFromImplicitlyEnabledCapabilities`.  But thinking about it now, I'm a bit confused by the semantics.  Wouldn't we actually want to feed the set of enabled capabilities (both implicit and explicit), to be compared with the set of requested caps, to determine the new set of enabled capabilities?  For example:

1. Admin requests cap1 and cap2.  Enabled set is cap1 and cap2.  Implicitly enabled set is empty.
2. Admin stops requesting cap1.  Enabled set is cap1 and cap2.  Implicitly enabled set is cap2.
3. Admin requests cap1 again, but stops requesting cap2.  Enabled set is cap1 and cap2.  Implicitly enabled set is cap1.

I'm worried that feeding just the currently-implicit-set into `SetFromImplicitlyEnabledCapabilities` will flub step 3 and miss the fact that cap1 needs to stay enabled.

jottofar: The logic to figure out "1. 2. 3." is in `GetImplicitlyEnabledCapabilities` which does take the entire Capabilities structure. Per its godoc, `SetFromImplicitlyEnabledCapabilities`, given implicitly enabled capabilities which were already determined by `GetImplicitlyEnabledCapabilities` simply "...updates ... with the given implicitly enabled capabilities and ensures each is in the enabled map".

jottofar: ACK to the godoc for this method and a unit test.

pkg/cvo/sync_worker.go
```diff
@@ -274,7 +281,7 @@ func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork,
 				Release:            desired,
 				LastTransitionTime: time.Now(),
 			})
-			return nil, err
+			return implicitlyEnabledCaps, err
```
wking, line 284: nit: we don't need these or the other `err != nil` touches, because:

* [`loadUpdatedPayload`][1] is going to throw them out and return a `nil` cap set on errors, and
* [`Update`][2] doesn't care what the returned cap set is when there's an error.

We only care about the returned cap set when the error is `nil`.

[1]: https://github.com/openshift/cluster-version-operator/blob/4e074889fc30cef47dae8b1309648be821bd02d8/pkg/cvo/sync_worker.go#L377
[2]: https://github.com/openshift/cluster-version-operator/blob/4e074889fc30cef47dae8b1309648be821bd02d8/pkg/cvo/sync_worker.go#L458

wking (approved):
/lgtm

wking:
[This run][1] had a router issue, which would be orthogonal, but otherwise looks good.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/768/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1520117854203023360

PR #770, opened 2022-05-02
Bug 2080429: pkg/cvo/sync_worker.go: Save overrides and capabilities

jottofar:
if changed before returning due to a desired update payload load failure.

### DIFF
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -209,8 +209,8 @@ func TestCVO_StartupAndSync(t *testing.T) {
 				{State: configv1.PartialUpdate, Image: "image/image:1", Version: "1.0.0-abc", StartedTime: defaultStartedTime},
 			},
 			Capabilities: configv1.ClusterVersionCapabilitiesStatus{
-				EnabledCapabilities: []configv1.ClusterVersionCapability{"baremetal", "marketplace", "openshift-samples"},
-				KnownCapabilities:   []configv1.ClusterVersionCapability{"baremetal", "marketplace", "openshift-samples"},
+				EnabledCapabilities: []configv1.ClusterVersionCapability{configv1.ClusterVersionCapabilityBaremetal, configv1.ClusterVersionCapabilityMarketplace, configv1.ClusterVersionCapabilityOpenShiftSamples},
+				KnownCapabilities:   []configv1.ClusterVersionCapability{configv1.ClusterVersionCapabilityBaremetal, configv1.ClusterVersionCapabilityMarketplace, configv1.ClusterVersionCapabilityOpenShiftSamples},
 			},
 			Conditions: []configv1.ClusterOperatorStatusCondition{
 				{Type: ImplicitlyEnabledCapabilities, Status: "False", Reason: "AsExpected", Message: "Capabilities match configured spec"},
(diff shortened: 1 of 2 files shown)


pkg/cvo/sync_worker.go
```diff
@@ -459,9 +459,14 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
...
 	w.lock.Lock()
 	if err != nil {
-		// save latest capability settings if not first time through
+		// save override and capability changes if not first time through
 		if w.work != nil {
-			w.work.Capabilities = work.Capabilities
+			if !overridesEqual {
+				w.work.Overrides = config.Spec.Overrides
+			}
+			if !capabilitiesEqual {
```
wking, line 467: Do we need these qualifiers?  If they are equal, then `w.work.Capabilities = work.Capabilities` is a no-op, so you get the same behavior by performing the assignment unconditionally, right?  Same for overrides above.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1566,6 +1566,233 @@ func TestCVO_UpgradeUnverifiedPayload(t *testing.T) {
...
+			ResourceVersion: "1",
+			Generation:      1,
+		},
+		Spec: configv1.ClusterVersionSpec{
+			ClusterID:     clusterUID,
+			Channel:       "fast",
+			DesiredUpdate: &configv1.Update{Version: desired.Version, Image: desired.Image},
+			Capabilities: &configv1.ClusterVersionCapabilitiesSpec{
+				BaselineCapabilitySet:         configv1.ClusterVersionCapabilitySetNone,
+				AdditionalEnabledCapabilities: []configv1.ClusterVersionCapability{"baremetal"},
```
wking, line 1591: nit: we have `configv1.ClusterVersionCapabilityBaremetal`, which is a lot longer than `"baremetal"`, but gets us compile-time typo protection.  And maybe helps folks grepping for consumers?  Doesn't seem like a big win, so update or leave your current string as it stands, as you see fit.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1566,6 +1566,233 @@ func TestCVO_UpgradeUnverifiedPayload(t *testing.T) {
...
+			// Prefers the image version over the operator's version (although in general they will remain in sync)
+			ObservedGeneration: 1,
+			Desired:            desired,
+			VersionHash:        "DL-FFQ2Uem8=",
+			History: []configv1.UpdateHistory{
+				{State: configv1.PartialUpdate, Image: "image/image:1", Version: "1.0.1-abc", StartedTime: defaultStartedTime},
+				{State: configv1.CompletedUpdate, Image: "image/image:0", Version: "1.0.0-abc", Verified: true, StartedTime: defaultStartedTime, CompletionTime: &defaultCompletionTime},
+			},
+			Conditions: []configv1.ClusterOperatorStatusCondition{
+				{Type: ImplicitlyEnabledCapabilities, Status: "False", Reason: "AsExpected", Message: "Capabilities match configured spec"},
```
wking, line 1688: I expected `status.Capabilities` to be set at this point.  Perhaps that is not something that `o.sync` syncs?

wking (approved):
I don't entirely understand what's going on in the unit test, but the code change looks good to me.  Feel free to discuss the unit tests more, or just pull the hold, as you see fit.

/lgtm
/hold

pkg/cvo/cvo_scenarios_test.go
jottofar: Right. I changed `Update` to also set CV status before returning on load error.

wking (approved):
/lgtm
/hold cancel

PR #771, opened 2022-05-02
Bug 2081895: lib/resourcebuilder/apiext: Restore check for Established=True CRDs

wking:
I'd dropped this in cc9292a061 (#400), claiming:

> There's no object status for CRDs or DaemonSets that marks "we are really hurting".  The v1.18.0 Kubernetes CRD and DaemonSet controllers do not set any conditions in their operand status (although the API for those conditions exists [2,3]).  With this commit, we have very minimal wait logic for either.  Sufficiently unhealthy DaemonSet should be reported on via their associated ClusterOperator, and sufficiently unhealthy CRD should be reported on when we fail to push any custom resources consuming them (Task.Run retries will give the API server time to ready itself after accepting a CRD update before the CVO fails its sync cycle).

But from [upstream docs][2]:

> It might take a few seconds for the endpoint to be created. You can watch the `Established` condition of your CustomResourceDefinition to be true or watch the discovery information of the API server for your resource to show up.

So I was correct that we will hear about CRD issues when we fail to push a dependent custom resource.  But I was not correct in claiming that the CRD controller set no conditions.  And the code I removed in cc9292a061 was in fact looking at the `Established` condition already.

This commit restores the `Established` check, but without the previous `PollImmediateUntil` wait.  We'll see how it plays in CI.  If we spend too long in between failed sync cycles waiting on new CRDs to show up, we can look at restoring a short wait.

[2]: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition


wking:
[Update][1]:

```
Cluster did not complete upgrade: timed out waiting for the condition: Could not update customresourcedefinition "clusteroperators.config.openshift.io" (4 of 792)
```

From the CVO logs:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521351550553821184/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-5f5f9cd4b4-gwq8w_cluster-version-operator.log | grep customresourcedef | tail -n2
E0503 08:14:05.090838       1 task.go:112] error running apply for customresourcedefinition "clusteroperators.config.openshift.io" (4 of 792): CustomResourceDefinition clusteroperators.config.openshift.io does not declare an Established status condition
E0503 08:14:20.932028       1 task.go:112] error running apply for customresourcedefinition "clusteroperators.config.openshift.io" (4 of 792): CustomResourceDefinition clusteroperators.config.openshift.io does not declare an Established status condition
```

From the must-gather:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521351550553821184/artifacts/e2e-agnostic-upgrade/gather-must-gather/artifacts/must-gather.tar | tar xOz registry-build03-ci-openshift-org-ci-op-sbcr982j-stable-initial-sha256-862bbab55fec772e032e10f2765e9221f8a743e6fbcdd2d376bbe10a9e875c79/cluster-scoped-resources/apiextensions.k8s.io/customresourcedefinitions/clusteroperators.config.openshift.io.yaml | yaml2json | jq -r '.status.conditions[] | .lastTransitionTime + " " + .type + "=" + .status + " " + .reason + ": " + .message' 
2022-05-03T05:07:37Z NamesAccepted=True NoConflicts: no conflicts found
2022-05-03T05:07:37Z Established=True InitialNamesAccepted: the initial names have been accepted
```

Looks like it's declaring `Established` to me.  I've pushed c1328ada -> 4998c6c0 to log the conditions we did get.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521351550553821184

wking:
[upgrade][1] failed to install.  Unrelated to this PR, because the installed version is the base branch.

/test e2e-agnostic-upgrade

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521540867054112768

wking:
[New run][1] failed the same way, and confirms that the CVO code is not seeing the conditions that show up in the must-gather:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521622968680058880/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-644b6dd49d-rkdzj_cluster-version-operator.log | grep 'does not declare' | tail -n2
E0504 03:05:53.570755       1 task.go:112] error running apply for customresourcedefinition "clusteroperators.config.openshift.io" (4 of 791): CustomResourceDefinition clusteroperators.config.openshift.io does not declare an Established status condition: []
I0504 03:06:13.358865       1 sync_worker.go:1098] Update error 4 of 791: UpdatePayloadFailed Could not update customresourcedefinition "clusteroperators.config.openshift.io" (4 of 791) (*errors.errorString: CustomResourceDefinition clusteroperators.config.openshift.io does not declare an Established status condition: [])
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521622968680058880/artifacts/e2e-agnostic-upgrade/gather-must-gather/artifacts/must-gather.tar | tar xOz registry-build03-ci-openshift-org-ci-op-dwkpjrf0-stable-initial-sha256-862bbab55fec772e032e10f2765e9221f8a743e6fbcdd2d376bbe10a9e875c79/cluster-scoped-resources/apiextensions.k8s.io/customresourcedefinitions/clusteroperators.config.openshift.io.yaml | yaml2json | jq -r '.status.conditions[] | .lastTransitionTime + " " + .type + "=" + .status + " " + .reason + ": " + .message'
2022-05-03T23:08:12Z NamesAccepted=True NoConflicts: no conflicts found
2022-05-03T23:08:12Z Established=True InitialNamesAccepted: the initial names have been accepted
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1521622968680058880

lib/resourceapply/apireg.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"context"
+	"fmt"
+
+	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	apiregclientv1 "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
+)
+
+// ApplyAPIServicev1 applies the required API service to the cluster.
+func ApplyAPIServicev1(ctx context.Context, client apiregclientv1.APIServicesGetter, required *apiregv1.APIService, reconciling bool) (*apiregv1.APIService, bool, error) {
```
wking, line 12: Can we just drop `DeleteAPIServicev1`?  Seems like it was added in 0afb8a8658 (#438), but that it has never had any consumers.

jottofar: I'd leave it in so our `resourcebuilder` continues to support apply and delete for all the resources.

jottofar: But looking at this further, this is saying CVO will never apply this resource. Did CVO ever apply the resource such that there could ever be an orphan? If not then yeah, I suppose there's no reason for the delete.

lib/resourcebuilder/resourcebuilder.go
```diff
@@ -130,10 +132,11 @@ func (b *builder) Do(ctx context.Context) error {
 			if err := b.modifyDeployment(ctx, typedObject); err != nil {
 				return err
 			}
-			if _, _, err := resourceapply.ApplyDeploymentv1(ctx, b.appsClientv1, typedObject, reconcilingMode); err != nil {
+			if actual, _, err := resourceapply.ApplyDeploymentv1(ctx, b.appsClientv1, typedObject, reconcilingMode); err != nil {
 				return err
+			} else if actual != nil {
```
jottofar, line 137: So before your change if a resource was `IsCreateOnly` this would have segfault'ed? Seems like a pretty big hole.

Should the `nil` check be in each of the functions instead?

lib/resourceapply/apireg.go
wking: I've pushed ebb12713 -> 6e12b9f2 with a new commit that re-drops APIService handling.

lib/resourcebuilder/resourcebuilder.go
jottofar: ACK. Did not notice that param change.

wking:
[PDB/DNS stuff][1] is unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/771/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1523741507503263744

PR #773, opened 2022-05-05
Bug 2079789: capability: Init prior known from CV status

jottofar:
so newly started CVO pod, including one which results from an upgrade, will be able to determine and set implicitly enabled capabilities.
Also made the following changes:
- Set `w.status` when `w.work == nil` so as not to clobber `CapabilitiesStatus`. `CapabilitiesStatus` would never had been set at that point in the code before this change.
- Duplicate manifests are handled by `GetImplicitlyEnabledCapabilities` by tainting in all applicable capabilities across the duplicated manifests.

### DIFF
--- a/lib/capability/capability.go
+++ b/lib/capability/capability.go
@@ -48,6 +48,15 @@ func SetCapabilities(config *configv1.ClusterVersion,
 	return capabilities
 }
 
+// GetCapabilitiesAsMap returns the slice of capabilities as a map with default values.
+func GetCapabilitiesAsMap(capabilities []configv1.ClusterVersionCapability) map[configv1.ClusterVersionCapability]struct{} {
+	caps := make(map[configv1.ClusterVersionCapability]struct{}, len(capabilities))
+	for _, c := range capabilities {
+		caps[c] = struct{}{}
+	}
+	return caps
+}
+
 // SetFromImplicitlyEnabledCapabilities, given implicitly enabled capabilities and cluster capabilities, updates
 // the latter with the given implicitly enabled capabilities and ensures each is in the enabled map. The updated
 // cluster capabilities are returned.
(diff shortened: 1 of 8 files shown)


lib/capability/capability.go
```diff
@@ -48,6 +48,15 @@ func SetCapabilities(config *configv1.ClusterVersion,
 	return capabilities
 }
 
+// GetCapabilitiesAsMap returns the slice of capabilities as a map with default values.
+func GetCapabilitiesAsMap(capabilities []configv1.ClusterVersionCapability) map[configv1.ClusterVersionCapability]struct{} {
```
wking, line 52: nit, you can use this helper to seed an initial `enabled` set [here][1].

[1]: https://github.com/openshift/cluster-version-operator/blob/910117429c2ceef37ec7cc25b64bfa7a0058b9da/lib/capability/capability.go#L154-L156

pkg/payload/payload.go
```diff
@@ -248,9 +248,14 @@ func GetImplicitlyEnabledCapabilities(updatePayloadManifests []manifest.Manifest
 				continue
```
jottofar, line 248: Prob should be a `break`

pkg/payload/payload.go
```diff
@@ -245,9 +245,11 @@ func GetImplicitlyEnabledCapabilities(updatePayloadManifests []manifest.Manifest
...
 			}
 			if err := currentManifest.Include(nil, nil, nil, &clusterCaps); err != nil {
-				continue
+				break
 			}
 			if err := updateManifest.Include(nil, nil, nil, &clusterCaps); err != nil {
+				klog.V(2).Infof("%s has changed and is now part of one or more disabled capabilities which will be implicitly enabled: %v",
+					getManifestResourceId(updateManifest), err)
```
wking, line 252: I feel like this is asking us to create a `Manifest.String()` method in library go, as follow-up work ;)

pkg/payload/payload.go
```diff
@@ -257,6 +259,7 @@ func GetImplicitlyEnabledCapabilities(updatePayloadManifests []manifest.Manifest
 					}
 				}
 			}
+			break
```
wking, line 262: I don't understand this `break` or the one a few lines up^^.  Can you explain why we want them in the commit message?

jottofar: Removed breaks such that all update manifests are compared against all current manifests, i.e. handles duplicate manifest resources.

wking (approved):
/lgtm

PR #788, opened 2022-06-08
Bug 2094174: pkg/cvo: reset payload load status

jottofar:
when an upgrade in progress is cleared.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -668,15 +668,26 @@ func (optr *Operator) getClusterVersion(ctx context.Context) (*configv1.ClusterV
 	return nil, false, err
 }
 
-// versionString returns a string describing the desired version.
-func versionString(update configv1.Release) string {
-	if len(update.Version) > 0 {
-		return update.Version
+// versionString returns a string describing the version. The given version is used if set otherwise
+// the given image is used if set. If neither of these is set "<unknown>" is returned.
+func versionString(version string, image string) string {
+	if len(version) > 0 {
+		return version
 	}
-	if len(update.Image) > 0 {
-		return update.Image
+	if len(image) > 0 {
+		return image
 	}
 	return "<unknown>"
+}
+
+// versionStringFromUpdate returns a string describing the desired version from Update type.
+func versionStringFromUpdate(update configv1.Update) string {
+	return versionString(update.Version, update.Image)
+}
+
+// versionStringFromRelease returns a string describing the desired version from Release type.
+func versionStringFromRelease(release configv1.Release) string {
+	return versionString(release.Version, release.Image)
 }
 
 // currentVersion returns an update object describing the current
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -232,7 +232,7 @@ func TestCVO_StartupAndSync(t *testing.T) {
 				Step:               "RetrievePayload",
 				Message:            "Retrieving and verifying payload version=\"1.0.0-abc\" image=\"image/image:1\"",
 				LastTransitionTime: time.Unix(1, 0),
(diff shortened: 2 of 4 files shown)


jottofar:
/override ci/prow/golangci-lint

linter is broke. Pushing this past.

pkg/cvo/cvo.go
```diff
@@ -668,13 +668,18 @@ func (optr *Operator) getClusterVersion(ctx context.Context) (*configv1.ClusterV
 	return nil, false, err
 }
 
+// versionStringUpdate returns a string describing the desired version from Update type.
+func versionStringUpdate(update configv1.Update) string {
+	return versionString(configv1.Release{Version: update.Version, Image: update.Image})
```
LalatenduMohanty, line 673: This looks odd. Why versionStringUpdate  need to call versionString(). It can directly reply with the version string. Also we should change the name of functions as versionStringFromUpdate() and versionStringFromRelease().

LalatenduMohanty (changes requested):

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1566,6 +1566,253 @@ func TestCVO_UpgradeUnverifiedPayload(t *testing.T) {
 	})
 }
 
+func TestCVO_ResetPayloadLoadStatus(t *testing.T) {
+	o, cvs, client, _, shutdownFn := setupCVOTest("testdata/payloadtest-2")
+
+	// Setup: a successful sync from a previous run, and the operator at the same image as before
+	//
```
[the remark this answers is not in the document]
jottofar: > This can be removed.

The extra comment line? If so I'm just following the convention already established in the file. Removing these would be extensive.

pkg/cvo/cvo.go
jottofar: Agree with the name changes but don't understand the questioning of why this method calls `versionString`. `versionString` has logic in it which is being reused by `versionStringUpdate`.

LalatenduMohanty: I was suggesting that we can directly do following in versionStringFromUpdate()
```
if len(update.Version) > 0 {
		return update.Version
}
return "unknown"

Comment (Davoska) [reply pkg/cvo/cvo.go]: Regarding the reuse of the logic, keeping the names clearer, and not creating a new data structure. We could have the functions `versionStringFromUpdate()` and `versionStringFromRelease()` call a function like `func versionString(version string, image string) string`. This would keep the logic in one function, and only strings are passed.

```go
func versionString(version string, image string) string {
	if len(version) > 0 {
		return version
	}
	if len(image) > 0 {
		return image
	}
	return "<unknown>"
}

// versionStringFromUpdate returns a string describing the desired version from Update type.
func versionStringFromUpdate(update configv1.Update) string {
	return versionString(update.Version, update.Image)
}

// versionStringFromRelease returns a string describing the desired version.
func versionStringFromRelease(release configv1.Release) string {
	return versionString(release.Version, release.Image)
}
```

Davoska:
Looks good to me :+1:

I tested the pull request on a live cluster, and the fix seems to be working as expected. But keep in mind that I am still learning the codebase and the changes made in the `pkg/cvo/cvo_scenarios_test.go` file are a little bit too advanced for me to properly review at the moment. But still a great source for learning!

I would personally add more information to the commit's description. For someone like me who is new to the codebase, commit descriptions are a great source of information regarding why some changes in the code were made.

LalatenduMohanty (approved):
/lgtm

wking:
I'm getting `Application is not available` out of Prow, but [the e2e-agnostic presubmit][1] does not excercise this new code, and the pull request has been open long enough:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/788/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1552354158739197952

PR #791, opened 2022-06-17
Bug 2097067: pkg/cvo: retain initial completed update history entry

jottofar:
which was the intent of https://github.com/openshift/cluster-version-operator/commit/6971c2bce79327fd51045fe6726c5d9c4524aaed. But that change failed to keep the last completed if last completed is at position maxHistory-1. It also did not work for a "prune as you go" `(len(History) == maxHistory+1)` strategy which is how the pruner is called by mergeOperatorHistory.

### DIFF
--- a/pkg/cvo/status.go
+++ b/pkg/cvo/status.go
@@ -29,6 +29,10 @@ const (
 	// cannot reach the desired state. It is considered more serious than Degraded
 	// and indicates the cluster is not healthy.
 	ClusterStatusFailing = configv1.ClusterStatusConditionType("Failing")
+
+	// MaxHistory is the maximum size of ClusterVersion history. Once exceeded
+	// ClusterVersion history will be pruned.
+	MaxHistory = 50
 )
 
 func mergeEqualVersions(current *configv1.UpdateHistory, desired configv1.Release) bool {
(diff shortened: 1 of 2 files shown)


pkg/cvo/status.go
```diff
@@ -128,26 +132,25 @@ func mergeOperatorHistory(config *configv1.ClusterVersion, desired configv1.Rele
...
-		if item.State != configv1.CompletedUpdate {
-			continue
-		}
-		// guarantee the last position in the history is always a completed item
-		if i >= maxHistory {
-			config.Status.History[maxHistory-1] = item
-		}
-		break
+	if config.Status.History[maxHistory].State == configv1.CompletedUpdate {
+		item := config.Status.History[maxHistory]
```
wking, line 148: I doubt we will drop `MaxHistory` below the current 50, but if we do, is it worth teaching this to prefer the oldest `>=maxHistory` index that's completed?

jottofar: Currently, because `pruneStatusHistory` gets called each time a new entry is added the history index will never exceed `Maxhistory`. This could change in the future but most likely if it does it will be part of a more robust history pruning re-do, e.g. implementing our enhancement, which will change things significantly anyway.

wking (approved):
/lgtm

wking:
[Update pod sandboxes and `Watchdog` continuity][1] are unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/791/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1539263947029352448

PR #796, opened 2022-07-06
Pull arch from payload

jottofar:
and use when querying OSUS. Per https://issues.redhat.com/browse/OTA-655, heterogeneous cluster payloads will contain metadata key "architecture" set to "multi". Homogeneous cluster payloads are unchanged and will not contain the metadata key "architecture" in which case CVO will set the architecture using runtime.GOARCH.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -23,6 +22,7 @@ import (
 	"github.com/openshift/cluster-version-operator/pkg/clusterconditions"
 )
 
+const noArchitecture string = "NoArchitecture"
 const noChannel string = "NoChannel"
 
 // syncAvailableUpdates attempts to retrieve the latest updates and update the status of the ClusterVersion
@@ -35,8 +35,9 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 		usedDefaultUpstream = true
 		upstream = optr.defaultUpstreamServer
 	}
-	arch := runtime.GOARCH
+
 	channel := config.Spec.Channel
+	arch := optr.getArchitecture()
 
 	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
 	u := optr.getAvailableUpdates()
@@ -46,6 +47,9 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 		klog.V(2).Infof("Retrieving available updates again, because more than %s has elapsed since %s", optr.minimumUpdateCheckInterval, u.LastAttempt)
 	} else if channel != u.Channel {
 		klog.V(2).Infof("Retrieving available updates again, because the channel has changed from %q to %q", u.Channel, channel)
+	} else if arch != u.Architecture {
+		klog.V(2).Infof("Retrieving available updates again, because the architecture has changed from %q to %q",
+			u.Architecture, arch)
 	} else if upstream == u.Upstream || (upstream == optr.defaultUpstreamServer && u.Upstream == "") {
 		klog.V(2).Infof("Available updates were recently retrieved, with less than %s elapsed since %s, will try later.", optr.minimumUpdateCheckInterval, u.LastAttempt)
 		return nil
@@ -67,6 +71,7 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 	au := &availableUpdates{
 		Upstream:           upstream,
 		Channel:            config.Spec.Channel,
+		Architecture:       arch,
 		Current:            current,
 		Updates:            updates,
 		ConditionalUpdates: conditionalUpdates,
@@ -82,8 +87,9 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 }
 
 type availableUpdates struct {
-	Upstream string
-	Channel  string
+	Upstream     string
+	Channel      string
+	Architecture string
 
 	// LastAttempt records the time of the most recent attempt at update
 	// retrieval, regardless of whether it was successful.
@@ -115,6 +121,7 @@ func (u *availableUpdates) NeedsUpdate(original *configv1.ClusterVersion) *confi
 	if u == nil {
 		return nil
 	}
+	// Architecture could change but does not reside in ClusterVersion
 	if u.Upstream != string(original.Spec.Upstream) || u.Channel != original.Spec.Channel {
 		return nil
 	}
@@ -160,6 +167,7 @@ func (optr *Operator) setAvailableUpdates(u *availableUpdates) {
 	if u != nil && (optr.availableUpdates == nil ||
 		optr.availableUpdates.Upstream != u.Upstream ||
 		optr.availableUpdates.Channel != u.Channel ||
+		optr.availableUpdates.Architecture != u.Architecture ||
 		success) {
 		u.LastSyncOrConfigChange = u.LastAttempt
 	} else if optr.availableUpdates != nil {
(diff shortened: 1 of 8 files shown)


pkg/cvo/cvo.go
```diff
@@ -93,6 +93,9 @@ type Operator struct {
 	// minimumUpdateCheckInterval is the minimum duration to check for updates from
 	// the upstream.
 	minimumUpdateCheckInterval time.Duration
+	// architecture identifies the architecture of the cluster nodes. The cluster architecture
+	// is heterogeneous if the cluster has nodes running on different architectures.
```
wking, line 97: > ... if the release image declares 'release.openshift.io/architecture: multi'.

right?  Because you can be, and ask for, heterogeneous release images without actually having heterogeneous nodes.  And the homogeneous -> heterogeneous transition will (eventually) be:

1. Running a homogeneous release on a homogeneous cluster, asking for homogeneous update recommendations.
2. Running a homogeneous release on a homogeneous cluster, asking for heterogeneous update recommendations.
3. Updating to a heterogeneous release on a homogeneous cluster, asking for heterogeneous update recommendations.
4. Running a heterogeneous release on a homogeneous cluster, asking for heterogeneous update recommendations.
5. Running a heterogeneous release on a heterogenous cluster, asking for heterogeneous update recommendations.

pkg/payload/payload.go
```diff
@@ -412,3 +437,8 @@ func loadImageReferences(releaseDir string) (*imagev1.ImageStream, error) {
 	}
 	return nil, fmt.Errorf("%s is a %T, not a v1 ImageStream", imageReferencesFile, imageRefObj)
 }
+
+// isHeterogeneousArchitecture returns true if arch is set to heterogeneousArchitectureID indicating a heterogeneous architecture.
+func isHeterogeneousArchitecture(arch string) bool {
+	return arch == heterogeneousArchitectureID
```
wking, line 443: nit: at the moment, the `==` seems straightforward enough that I'd rather inline it and drop `isHeterogeneousArchitecture`.  But I don't feel all that strongly about it, so feel free to ignore me if you prefer it the way you have it.

pkg/cvo/cvo.go
jottofar: Yeah, agree, but that's the "to be" logic. Implementation of this Jira is just a baby step so I made the comment only reflect what this code does now. This value simply gets to set to either "mullti" if the payload contains it or GOARCH. We can expand when we add more logic.

pkg/payload/payload.go
jottofar: Prefer to leave it and just move on since it has to be touched again to fully implement heter.

wking (approved):
/lgtm

LalatenduMohanty:
Still working on the review.

pkg/cvo/cvo.go
LalatenduMohanty, line 97: Can we add about the possible value of architecture when cluster is heterogeneous i.e. `multi` and for others it would be `GOARCH`.

LalatenduMohanty:
Added some comments which are nit picks and does not need to be fixed before the merge.

pkg/cvo/availableupdates.go
```diff
@@ -160,6 +167,7 @@ func (optr *Operator) setAvailableUpdates(u *availableUpdates) {
 	if u != nil && (optr.availableUpdates == nil ||
 		optr.availableUpdates.Upstream != u.Upstream ||
 		optr.availableUpdates.Channel != u.Channel ||
+		optr.availableUpdates.Architecture != u.Architecture ||
```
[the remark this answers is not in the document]
wking: What would we log?  Something like "bumping LastSyncOrConfigChange because..." and then "upstream changed from A to B", "channel changed from A to B", or "we successfully synchronized available updates"?

pkg/cvo/cvo.go
wking: if we want to doc this, I think we should just say "to match the Cincinnati API's 'arch' property" or something, and then update [this][1] in  openshift/cincinnati#711 to cover `multi` as well.

[1]: https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md#request

pkg/cvo/availableupdates.go
```diff
@@ -203,7 +225,7 @@ func calculateAvailableUpdatesStatus(ctx context.Context, clusterID string, tran
 
 	if len(arch) == 0 {
 		return cvoCurrent, nil, nil, configv1.ClusterOperatorStatusCondition{
-			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: "NoArchitecture",
+			Type: configv1.RetrievedUpdates, Status: configv1.ConditionFalse, Reason: noArchitecture,
 			Message: "The set of architectures has not been configured.",
```
wking, line 229: Not sure why yet, but this is [tripping e2e-agnostic like][1]:

```
: [sig-instrumentation][Late] Alerts shouldn't report any unexpected alerts in firing or pending state [Suite:openshift/conformance/parallel]
--
Run #0: Failed 5s
{  fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Jul 28 15:51:34.182: Unexpected alerts fired or pending after the test run:  alert CannotRetrieveUpdates fired for 313 seconds with labels: {endpoint="metrics", instance="10.0.0.8:9099", job="cluster-version-operator", namespace="openshift-cluster-version", pod="cluster-version-operator-7674b7b444-f7rn9", service="cluster-version-operator", severity="warning"} ...
```

and:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/796/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1552651611942162432/artifacts/e2e-agnostic/gather-extra/artifacts/clusterversion.json | jq -r '.items[].status.conditions[] | select(.type == "RetrievedUpdates")'
{
  "lastTransitionTime": "2022-07-28T14:21:43Z",
  "message": "The set of architectures has not been configured.",
  "reason": "NoArchitecture",
  "status": "False",
  "type": "RetrievedUpdates"
}
```

/hold

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/796/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1552651611942162432

wking:
```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/796/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1552651611942162432/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-7674b7b444-f7rn9_cluster-version-operator.log | grep 'Payload loaded from'
I0728 14:46:02.143948       1 sync_worker.go:385] Payload loaded from registry.build02.ci.openshift.org/ci-op-7ipcn54g/release@sha256:70dd7baaa8c8550e19dddbe79d75a9d78bc37855974649b8292da0e1dccb4d56 with hash oiJMeRYBbXY=, architecture 
```

not sure if that's relevant, but sounds suspicious.

pkg/cvo/availableupdates.go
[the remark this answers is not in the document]
LalatenduMohanty: Never mind. I misunderstood the purpose of this

pkg/cvo/cvo.go
LalatenduMohanty: I am fine with that as long as we can find out from the doc what values can be used for it.

jottofar: Updated comment

pkg/payload/payload.go
jottofar: done

pkg/cvo/cvo.go
jottofar: Updated comment

pkg/cvo/availableupdates.go
jottofar: Needed call to `SetArchitecture` in `InitializeFromPayload`

wking (approved):
/lgtm

wking:
[`TestCVO_UpgradeFailedPayloadLoadWithCapsChanges`][1] is a frequent race recently.  We should fix it, but not in this PR.

/override ci/prow/unit

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/796/pull-ci-openshift-cluster-version-operator-master-unit/1552761191300861952

wking (approved):
/lgtm

LalatenduMohanty (approved):
/lgtm

jottofar:
TestCVO_UpgradeFailedPayloadLoadWithCapsChanges race, unrelated to this PR

/override ci/prow/unit

PR #805, opened 2022-07-27
/pkg/cvo: improve CV history pruning

jottofar:
by implementing enhancement https://github.com/openshift/enhancements/blob/master/enhancements/update/clusterversion-history-pruning.md.


pkg/cvo/status_history.go
```diff
@@ -0,0 +1,235 @@
...
+	// partialZStreamWeight is given to any partial z-stream updates between a completed z-stream version transition.
+	partialZStreamWeight = -20.0
+
+	// sliceIndexWeight when applied will favor more recent updates and avoid ties.
+	sliceIndexWeight = -1.01
+)
+
+// maxHistorySize is set by the function prune and used by all the history pruning functions. Setting and using
+// this local package variable eases unit testing by allowing testing against a smaller history slice.
+var maxHistorySize int
```
wking, line 36: I don't understand the need for this package local variable, when `prune` takes an explicit `maxSize` that it can pass through to functions it calls in turn.  Can we drop this?

pkg/cvo/status_history.go
```diff
@@ -0,0 +1,235 @@
...
+
+// prune prunes history when at maxSize by ranking each entry using the above defined weights and removing the entry with
+// the lowest rank. maxSize is passed in to allow the use of smaller values which eases unit test.
+func prune(history []configv1.UpdateHistory, maxSize int) []configv1.UpdateHistory {
+	if len(history) <= maxSize {
+		return history
+	}
+	maxHistorySize = maxSize
+
+	idx := getTheMostRecentCompletedEntryIndex(history)
```
wking, line 46: nit: `idx` for `index` is pretty generic.  Can we use `mostRecentCompletedEntryIndex` or something super-obvious here?  Less room for someone to misunderstand what it's about later, if they come in here and start shuffling code.

pkg/cvo/status_history.go
```diff
@@ -0,0 +1,235 @@
...
+	lowestRank := math.MaxFloat64
+	var lowestRankIdx int
+
+	for i := range history {
+		rank := 0.0
+		thisReason := ""
+		if isTheInitialEntry(i) {
+			rank = mostImportantWeight
+		}
+		if rank == 0.0 && isAFinalEntry(i) {
```
wking, line 58: nit: these `if rank == 0.0 &&` could be collapsed into `else if`.  Or instead of a big `else if` chain, you could use [a `switch`][1]:

```go
switch {
case i <= maxFinalEntryIndex || i == mostRecentCompletedEntryIndex || i == maxSize:
  rank = mostImportantWeight
case ...
```

[1]: https://go.dev/ref/spec#Switch_statements

pkg/cvo/status_history.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		if rank == 0.0 && isPartialPortionOfMinorTransition(i, history) {
+			rank = rank + partialMinorWeight
+			thisReason = "a partial update within a minor transition"
+		}
+		if rank == 0.0 && isPartialWithinAZStream(i, history) {
+			rank = rank + partialZStreamWeight
+			thisReason = "a partial update within a z-stream transition"
+		}
+		rank = rank + (sliceIndexWeight * float64(i))
```
wking, line 76: nit: [`rank += sliceIndexWeight * float64(i)`][1].

[1]: https://go.dev/ref/spec#Assignment_statements

pkg/cvo/status_history.go
```diff
@@ -0,0 +1,235 @@
...
+
+		if rank < lowestRank {
+			lowestRank = rank
+			lowestRankIdx = i
+			reason = thisReason
+		}
+	}
+	if reason == "" {
+		klog.V(2).Infof("Pruning version %s at index %d with rank %f.", history[lowestRankIdx].Version, lowestRankIdx, lowestRank)
```
wking, line 85: nit: maybe we should drop this case, and default `reason := "<bug should never prune without a reason>"` or something?  Because reading the cases above, you're setting `reason` for everything except completed within a z stream, and maybe that deserves an explicit case and a `a completed update within a z-stream transition` `thisReason`?

pkg/cvo/status_history_test.go
```diff
@@ -0,0 +1,819 @@
...
+		{
+			name: "partial update within a z-stream transition",
+			history: []configv1.UpdateHistory{
+				{State: configv1.CompletedUpdate, Version: "4.5.3"},
+				{State: configv1.CompletedUpdate, Version: "4.4.3"},
+				{State: configv1.CompletedUpdate, Version: "4.4.3"},
+				{State: configv1.CompletedUpdate, Version: "4.4.2"},
+				{State: configv1.CompletedUpdate, Version: "4.1.10"},
+				{State: configv1.CompletedUpdate, Version: "4.1.9"},
+				{State: configv1.CompletedUpdate, Version: "4.1.3"},
```
wking, line 58: nit: having a partial update followed by a completed update with the same target version is unlikely.    Can we use 4.1.4 for this line?

pkg/cvo/status_history_test.go
```diff
@@ -0,0 +1,819 @@
...
+				{State: configv1.CompletedUpdate, Version: "4.7.0"},
+				{State: configv1.CompletedUpdate, Version: "4.6.0"},
+				{State: configv1.CompletedUpdate, Version: "4.5.0"},
+				{State: configv1.CompletedUpdate, Version: "4.4.0"},
+				{State: configv1.CompletedUpdate, Version: "4.3.0"},
+				{State: configv1.CompletedUpdate, Version: "4.1.0"},
+			},
+		},
+		{
+			name: "prune oldest not in mostImportantWeight set",
```
wking, line 108: nit: looks like a dup of the `prune oldest that's not initial` `history`?

pkg/cvo/status_history.go
jottofar: I did this at first thinking there'd be many more functions needing it. But there's only four so sure, I can pass it.

pkg/cvo/status_history.go
jottofar: I left it unset because it always seems possible, even if very unlikely, that a generic message may not be accurate. Working with only 10 made up updates may have clouded my vision. I suppose if someone hits 100 entries on a real cluster they would certainly had had to do a z-stream update, well unless they had 95 partials or something crazy. I think some more generic message would do but agree with your general point that we should always have a reason.

pkg/cvo/status_history_test.go
jottofar: It now doesn't matter to the tested code. It did when I was explicitly checking for a "same z-stream" version transition. But I thought this would be fairly common - i.e. my upgrade to 4.1.3 didn't complete (is Partial), oh, let me try again?

pkg/cvo/status_history.go
jottofar: I had originally had different reasons for the first three but then realized they should never be pruned so yeah no reason not to collapse them. I avoided `else if`s because in pass reviews I sensed an aversion to them, not necessarily by you,  but I prefer them here too. Regarding the switch option, I believe keeping the functions makes the code both more readable and more maintainable. Won't be surprised if we go back in here and tweak some things.

pkg/cvo/status_history.go
jottofar: agree

pkg/cvo/status_history_test.go
wking: if you are on A, attempt an update to B, give up (returning to A), and then attempt B again, you get ABAB history, not ABB history.  But :shrug: I don't have strong feelings about what goes into your unit test-case here.

pkg/cvo/status_history.go
wking: yeah, my generic message was just getting at "if you see this, please file a bug, because we always want something more specific"

pkg/cvo/status_history_test.go
jottofar: I did change it as you suggested so no problem there. Yeah, I see what you're saying now about the ABAB history.

wking (approved):
/lgtm

wking:
There's a bugfix in #828.  We're unlikely to backport this particular change, but commenting here in case we change our minds, so we remember to backport both this and the fix.

PR #806, opened 2022-07-28
Bug 2109374: pkg/clusterconditions/promql: Cap PromQL queries at 5 minutes

wking:
In some clusters, [these PromQL queries can hang for hours, possibly forever][1].  I think we have [a 30s default `KeepAlive` timeout][2], but apparently there's enough socket traffic to keep from tripping that.  This adds a 5m cap to the PromQL calls, although I'm not particularly attached to that particular number.  We can always raise it if we start seeing timeouts in Insights for queries where taking that long seems reasonable.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2109374#c12
[2]: https://pkg.go.dev/github.com/prometheus/client_golang/api#pkg-variables

### DIFF
--- a/pkg/clusterconditions/promql/promql.go
+++ b/pkg/clusterconditions/promql/promql.go
@@ -27,6 +27,9 @@ type PromQL struct {
 
 	// HTTPClientConfig holds the client configuration for connecting to the Prometheus service.
 	HTTPClientConfig config.HTTPClientConfig
+
+	// QueryTimeout limits the amount of time we wait before giving up on the Prometheus query.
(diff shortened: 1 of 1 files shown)


jottofar:
Hitting same old `TestCVO_UpgradeFailedPayloadLoadWithCapsChanges` unit test failure

wking:
[Verified via regression testing][1].

/cherrypick release-4.11

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2109374#c16

PR #808, opened 2022-08-01
Bug 2006611: Upgrade takes too much time when upgrading via --to-image

Davoska:
This pull request will synchronize the Upgradeable status more frequently when the precondition checks on the payload start failing and will improve the logging regarding synchronization of the upgradeable status. The more frequent synchronization will happen for a given period after the precondition checks start to fail.

### TL;DR: 
A precondition check originally could be failing for at least `optr.minimumUpdateCheckInterval` time because of `Upgradeable==false`. The `Upgradeable==false` could have been potentially already resolved (cluster operator reporting `Upgradeable==false` just momentarily) during the duration resulting in necessary waiting for up to `optr.minimumUpdateCheckInterval` for the next synchronization. Synchronize the upgradeable status more frequently in case of failing precondition checks to speed up initializing an upgrade stuck on a precondition check that potentially has been solved. We don't want to check this forever in case of precondition checks failing for a long time due to a bigger problem.

The bug was caused by slow syncing of the CVO's Upgradeable status when a precondition check fails and less frequent running of the precondition checks. The frequency of precondition checks was fixed by https://github.com/openshift/cluster-version-operator/pull/683 fixing the Etcd backup halting an upgrade for a prolonged time (https://bugzilla.redhat.com/show_bug.cgi?id=2072348). The problem of the `Upgradeable==false` thanks to the `ErrorCheckingOperatorCompatibility` caused by the OLM operator (https://bugzilla.redhat.com/show_bug.cgi?id=2006611) was fixed by the https://github.com/openshift/cluster-version-operator/pull/766. However, the main root cause of the `ErrorCheckingOperatorCompatibility` error probably remains. This pull request will synchronize the CVO's Upgradeable status more often for `optr.minimumUpdateCheckInterval` after the `ReleaseAccepted` reports false due to the reason `PreconditionChecks`.

The fixes mentioned above exist in releases as follows:
[Bug 1822752: pkg/cvo: Separate payload load from payload apply](https://github.com/openshift/cluster-version-operator/pull/683): `4.8` and newer
[Bug 2072389: Do not save desired update on load failures](https://github.com/openshift/cluster-version-operator/pull/766): `4.9` and newer

#### The behavior of the problem:
Precondition checks are run on a new payload before initializing an upgrade https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/cvo.go#L842-L847.  These checks are categorized into two categories: **EtcdBackup** and **Upgradeable**.

The Upgradeable precondition check is only checking the CVO's Upgradeable status https://github.com/openshift/cluster-version-operator/blob/master/pkg/payload/precondition/clusterversion/upgradeable.go#L66-L74.

The CVO's Upgradeable status originally synchronizes only every `optr.minimumUpdateCheckInterval` https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/upgradeable.go#L38-L50.

A problem can occur when a `Upgradeable==false` is reported for the CVO's status during a synchronization before an upgrade. A precondition check registers the `Upgradeable==false` and halts the upgrade, and then regularly checks if the `Upgradeable==false` still exists. However, the CVO's Upgradeable status synchronizes less often which can result in necessary waiting time for the user when initializing an upgrade.

This is normally an expected behavior. However, when a cluster operator reports `Upgradeable==false` just momentarily (for example, for 10 seconds at the start of the upgrade) and it gets registered by the CVO and thus the CVO's Upgradeable status gets set to false, then the upgrade can be halted for up to 3 minutes waiting for the next synchronization of the CVO's Upgradeable status. Making it seems as like the CVO is slow when applying an upgrade.

#### Bug 2072348 - Upgrade Precondition "EtcdRecentBackup" failed because of "ControllerStarted"
Link: https://bugzilla.redhat.com/show_bug.cgi?id=2072348 (marked as a duplicate of this bug):

Running Etcd backup at the start of an upgrade could hold up the upgrade for up to three minutes (slow checking of the Etcd backup status). This was fixed by https://github.com/openshift/cluster-version-operator/pull/683 allowing to run the precondition checks more frequently when initializing an upgrade (checking the Etcd backup status more frequently). 

#### Bug 2006611 - CVO resolves the version takes a long time sometimes when upgrading via `--to-image`
Link: https://bugzilla.redhat.com/show_bug.cgi?id=2006611 (this bug):

The problem can also arise by the CVO reporting `Upgradeable==false` with the reason `ErrorCheckingOperatorCompatibility` when a cluster operator reports `Upgradeable==false`. Most notably by the OLM operator. In the past, when the `status.desired.version` (when upgrading via `--to-image`) got the value `""` at the start of an upgrade.  If this value was empty for a prolonged time (~30 seconds), the OLM operator reported `Upgradeable==false`. If this got picked up by the CVO, then the next synchronization of the CVO's Upgradeable would happen in 2--3 minutes. Making it seems like the upgrade is slow.
This behavior was fixed by the https://github.com/openshift/cluster-version-operator/pull/766, which ensured the `status.desired.version` doesn't get the empty value and gets the new version's value only after the update is successfully loaded.

### Conclusion:
This behavior was fortunately fixed to an extent during the development. However, the main cause of the issues mentioned above was slow syncing at the start of an upgrade when a precondition check fails. These behaviors were fixed during development, but the root cause remains, to my knowledge.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -121,6 +121,11 @@ type Operator struct {
 	upgradeable           *upgradeable
 	upgradeableChecks     []upgradeableCheck
 
+	// minimumUpgradeableCheckInterval is the minimum duration before another
+	// synchronization of the upgradeable status can happen during failing
+	// precondition checks.
+	minimumUpgradeableCheckInterval time.Duration
+
 	// verifier, if provided, will be used to check an update before it is executed.
 	// Any error will prevent an update payload from being accessed.
 	verifier verify.Interface
@@ -179,10 +184,11 @@ func New(
 			Image: releaseImage,
 		},
 
-		statusInterval:             15 * time.Second,
-		minimumUpdateCheckInterval: minimumInterval,
-		payloadDir:                 overridePayloadDir,
-		defaultUpstreamServer:      "https://api.openshift.com/api/upgrades_info/v1/graph",
+		statusInterval:                  15 * time.Second,
+		minimumUpdateCheckInterval:      minimumInterval,
+		minimumUpgradeableCheckInterval: 15 * time.Second,
+		payloadDir:                      overridePayloadDir,
+		defaultUpstreamServer:           "https://api.openshift.com/api/upgrades_info/v1/graph",
 
 		client:        client,
 		kubeClient:    kubeClient,
@@ -627,7 +633,7 @@ func (optr *Operator) upgradeableSync(ctx context.Context, key string) error {
 		return nil
 	}
 
-	return optr.syncUpgradeable()
+	return optr.syncUpgradeable(config)
 }
 
 // isOlderThanLastUpdate returns true if the cluster version is older than
--- a/pkg/cvo/upgradeable.go
+++ b/pkg/cvo/upgradeable.go
@@ -33,12 +33,13 @@ const (
 
 var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
 
-// syncUpgradeable. The status is only checked if it has been more than
-// the minimumUpdateCheckInterval since the last check.
-func (optr *Operator) syncUpgradeable() error {
-	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
+// syncUpgradeable synchronizes the upgradeable status only if it has been more than
+// the minimumUpdateCheckInterval since the last synchronization or the precondition
+// checks on the payload are failing for less than minimumUpdateCheckInterval, and it has
+// been more than the minimumUpgradeableCheckInterval since the last synchronization.
+func (optr *Operator) syncUpgradeable(config *configv1.ClusterVersion) error {
 	u := optr.getUpgradeable()
-	if u != nil && u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
+	if u != nil && u.RecentlyChanged(optr.minimumUpdateCheckInterval) && !shouldSyncUpgradeableDueToPreconditionChecks(optr, config, u) {
 		klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
 		return nil
 	}
@@ -92,8 +93,14 @@ type upgradeable struct {
 	Conditions []configv1.ClusterOperatorStatusCondition
 }
 
+// hasPassedDurationSinceTime checks if a certain amount of time specified by duration
+// has passed since time. Returns true if the duration has passed, returns false otherwise.
+func hasPassedDurationSinceTime(t time.Time, duration time.Duration) bool {
+	return time.Now().After(t.Add(duration))
+}
+
 func (u *upgradeable) RecentlyChanged(interval time.Duration) bool {
-	return u.At.After(time.Now().Add(-interval))
+	return !hasPassedDurationSinceTime(u.At, interval)
 }
 
 func (u *upgradeable) NeedsUpdate(original *configv1.ClusterVersion) *configv1.ClusterVersion {
@@ -441,4 +448,23 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
 		UpdateFunc: optr.updateFunc,
 		DeleteFunc: optr.deleteFunc,
 	}
+}
+
+// shouldSyncUpgradeableDueToPreconditionChecks checks if the upgradeable status should
+// be synchronized due to the precondition checks. It checks whether the precondition
+// checks on the payload are failing for less than minimumUpdateCheckInterval, and it has
+// been more than the minimumUpgradeableCheckInterval since the last synchronization.
+// This means, upon precondition failure, we will synchronize the upgradeable status
+// more frequently at beginning of an upgrade.
+//
+// shouldSyncUpgradeableDueToPreconditionChecks expects the parameters not to be nil.
+//
+// Function returns true if the synchronization should happen, returns false otherwise.
+func shouldSyncUpgradeableDueToPreconditionChecks(optr *Operator, config *configv1.ClusterVersion, u *upgradeable) bool {
+	cond := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, DesiredReleaseAccepted)
+	if cond != nil && cond.Reason == "PreconditionChecks" && cond.Status == configv1.ConditionFalse &&
+		hasPassedDurationSinceTime(u.At, optr.minimumUpgradeableCheckInterval) && !hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval) {
+		return true
+	}
+	return false
 }


pkg/cvo/upgradeable.go
```diff
@@ -33,13 +33,27 @@ const (
...
-	// updates are only checked at most once per minimumUpdateCheckInterval or if the generation changes
+// syncUpgradeable synchronizes the upgradeable status only if it has been more than
+// the minimumUpdateCheckInterval since the last synchronization or the precondition checks
+// on the payload are failing for less than minimumUpdateCheckInterval.
+func (optr *Operator) syncUpgradeable(config *configv1.ClusterVersion) error {
 	u := optr.getUpgradeable()
-	if u != nil && u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
-		klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
+	if u == nil {
+		klog.V(2).Info("Synchronizing the upgradeable status, because it's the first attempt.")
```
LalatenduMohanty, line 42: When u == nil, it does not mean that it is the first attempt. If you want to check the u == nil, then we should be able to remove the else section of the conditional loop.

pkg/cvo/upgradeable.go
```diff
@@ -33,13 +33,27 @@ const (
...
+	} else if !u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
+		klog.V(2).Infof("Synchronizing the upgradeable status again, because more than %s has elapsed since %s", optr.minimumUpdateCheckInterval, u.At)
+	} else if cond := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, DesiredReleaseAccepted); cond != nil &&
+		cond.Reason == "PreconditionChecks" &&
+		cond.Status == configv1.ConditionFalse &&
+		!hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval) {
```
LalatenduMohanty, line 48: What is the downside of just checking ` else if u != nil && !hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval)`

LalatenduMohanty (changes requested):

pkg/cvo/upgradeable.go
```diff
@@ -129,7 +149,8 @@ func collectUpgradeableConditions(conditions []configv1.ClusterOperatorStatusCon
 	return ret
 }
 
-// setUpgradeable updates the currently calculated status of Upgradeable
+// setUpgradeable updates the operator's upgradeable field using the upgradeable u.
+// The field At of the struct upgradeable gets set by the setUpgradeable method.
```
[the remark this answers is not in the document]
Davoska: It's meant to reference the parameter `u`. However, I can see why it looks like a typo.

I can change the wording to:
```go
// setUpgradeable updates the operator's upgradeable field using the parameter.
```

Or we can drop this change from the commit. Originally when working on the fix, I had some code in this function, and I have modified the description of this function. Change was then dropped, but I have left the more descriptive description.

pkg/cvo/upgradeable.go
Davoska: Yes, I agree that `...because it's the first attempt.` is not correct and if the condition is kept should be changed to something like `Synchronizing the upgradeable status, because it's not set.`.

Regarding `u == nil`:
Since I am using the `u.At` value in messages, it's a little bit trickier to change it properly. Could you post a suggestion? I would appreciate it!

Otherwise, I only came up with:
```go
if cond := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, DesiredReleaseAccepted); cond != nil &&
	cond.Reason == "PreconditionChecks" &&
	cond.Status == configv1.ConditionFalse &&
	!hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval) {
	// The precondition check could have been potentially already resolved
	// (e.g. operator reporting Upgradeable==false just momentarily).
	// Synchronize the upgradeable status again in case of failing precondition checks to speed up initializing
	// an upgrade stuck on a precondition check that potentially has been solved.
	// We don't want to do this forever in case of precondition checks failing for a long time due to a bigger problem.
	klog.V(2).Infof("Synchronizing the upgradeable status again, because precondition checks are failing for less than %s since %s", optr.minimumUpdateCheckInterval, cond.LastTransitionTime)
} else if u == nil || !u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
	klog.V(2).Infof("Synchronizing the upgradeable status again.")
} else {
	klog.V(2).Infof("Upgradeable status was recently synchronized, with less than %s elapsed since %s, will try later.", optr.minimumUpdateCheckInterval, u.At)
	return nil
}
```
But I have lost the more descriptive reason why the synchronization happened in the message.

pkg/cvo/upgradeable.go
Davoska: If we leave the `cond.Reason == "PreconditionChecks"` we could remove the `cond.Status == configv1.ConditionFalse`, because the `cond.Reason == "PreconditionChecks"` [only gets set when the preconditions checks fail](https://github.com/openshift/cluster-version-operator/blob/5ad3c1434f781c68b8fab13b760d4c0c894fd21a/pkg/cvo/sync_worker.go#L349-L360), thus if the reason is `PreconditionChecks` then the status is implicitly `configv1.ConditionFalse`.

Davoska: > Since during the step PreconditionChecks the Etcd Backup is run. And while the Etcd backup is running, the ReleaseAccepted is False

> Thus the cond.LastTransitionTime.Time is probably changed after every upgrade.

Is not 100% correct. The EtcdBackup precondition check fails only in given instances:
https://github.com/openshift/cluster-version-operator/blob/c77c3360b86d8788e94493406a0d23a144e14e43/pkg/payload/precondition/clusterversion/etcdbackup.go#L76-L86

LalatenduMohanty: Ok, so I think the code as it is good.  I believe we want to keep `cond.Status == configv1.ConditionFalse` in the condition. I will suggest that we change 
```
if u == nil {
		klog.V(2).Info("Synchronizing the upgradeable status because is it not set")
```

LalatenduMohanty (changes requested):

Davoska:
I have fixed the pull request's description. There was incorrectly written `spec.desired.version` which should have been `status.desired.version`. I apologize to anyone for any inconvenience.

Davoska:
/hold
There is a work in progress.

pkg/cvo/upgradeable.go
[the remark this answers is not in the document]
Davoska: I have removed the change from the commit.

wking:
See if openshift/release#31580 is working:

/test e2e-agnostic-upgrade-out-of-change

pkg/cvo/upgradeable.go
Davoska: Hi @LalatenduMohanty ,
I have come up with a few possibilities to your suggestions, let me know your opinion. I would like to keep the debug logging. However, I wouldn't like the logic of the logging to make the code more complicated since the logging isn't that necessary.

The `shouldSyncUpgradeableDueToPreconditionChecks()` either returns just the `bool` or also the `DesiredReleaseAccepted` condition in the examples below

#### A) Short and simple, close to the original code, no debug logging (currently proposed in the PR)
```go
	u := optr.getUpgradeable()
	if u != nil && u.RecentlyChanged(optr.minimumUpdateCheckInterval) && !shouldSyncUpgradeableDueToPreconditionChecks(optr, config, u) {
		klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
		return nil
	}
	optr.setUpgradeableConditions()
```

#### B) Debug logging plus now code is similar to its equivelant in `pkg/cvo/availableupdates.go` (https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/availableupdates.go#L43-L58)
I am leaning more towards this option. But if the logging isn't necessary, I would go with the option A.
```go
	u := optr.getUpgradeable()
	if u == nil {
		klog.V(4).Info("Synchronizing the upgradeable status because it's not set.")
	} else if shouldSync, cond := shouldSyncUpgradeableDueToPreconditionChecks(optr, config, u); shouldSync {
		klog.V(4).Infof("Synchronizing the upgradeable status again, because precondition checks are failing for less than %s since %s", optr.minimumUpdateCheckInterval, cond.LastTransitionTime)
	} else if u.RecentlyChanged(optr.minimumUpdateCheckInterval) {
		klog.V(2).Infof("Upgradeable status was recently synchronized, with less than %s elapsed since %s, will try later.", optr.minimumUpdateCheckInterval, u.At)
		return nil
	} else {
		klog.V(4).Infof("Synchronizing the upgradeable status again, because more than %s has elapsed since %s", optr.minimumUpdateCheckInterval, u.At)
	}
	optr.setUpgradeableConditions()
```

#### C) Tried to keep the code regarding the logic close to the original code plus debug logging. A little messy.
```go
	var cond *configv1.ClusterOperatorStatusCondition
	var recentlyChanged bool
	var shouldSyncDueToPreconditionChecks bool

	u := optr.getUpgradeable()
	if u != nil {
		recentlyChanged = u.RecentlyChanged(optr.minimumUpdateCheckInterval)
	}
	if recentlyChanged {
		shouldSyncDueToPreconditionChecks, cond = shouldSyncUpgradeableDueToPreconditionChecks(optr, config, u)
		if !shouldSyncDueToPreconditionChecks {
			klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
			return nil
		}
	}

	if u == nil {
		klog.V(4).Info("Synchronizing the upgradeable status because it's not set.")
	} else if !recentlyChanged {
		klog.V(4).Infof("Synchronizing the upgradeable status again because more than %s has elapsed since %s", optr.minimumUpdateCheckInterval, u.At)
	} else if shouldSyncDueToPreconditionChecks {
		klog.V(4).Infof("Synchronizing the upgradeable status again because precondition checks are failing for less than %s since %s", optr.minimumUpdateCheckInterval, cond.LastTransitionTime)
	}
	optr.setUpgradeableConditions()
```

LalatenduMohanty (approved):
/lgtm

pkg/cvo/upgradeable.go
```diff
@@ -442,3 +449,31 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
...
+// CVO should synchronize the upgradeable status again in case of failing precondition
+// checks to speed up initializing an upgrade stuck on a precondition check that
+// potentially has been solved.
+//
+// We don't want to check this forever in case of precondition checks failing for a long
+// time due to a bigger problem that won't resolve by itself in a timely manner.
+func shouldSyncUpgradeableDueToPreconditionChecks(optr *Operator, config *configv1.ClusterVersion, u *upgradeable) bool {
+	cond := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, DesiredReleaseAccepted)
+	if cond != nil && cond.Reason == "PreconditionChecks" && cond.Status == configv1.ConditionFalse &&
+		hasPassedDurationSinceTime(u.At, optr.statusInterval) && !hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval) {
```
jottofar, line 475: > optr.statusInterval
Update the comment where this is declared now that we're also using it to limit frequency of upgradeable checks here.

pkg/cvo/upgradeable.go
```diff
@@ -442,3 +449,31 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
 		DeleteFunc: optr.deleteFunc,
 	}
 }
+
+// shouldSyncUpgradeableDueToPreconditionChecks checks if the upgradeable status should
+// be synchronized due to the precondition checks. It checks whether the precondition
+// checks on the payload are failing for less than minimumUpdateCheckInterval, and it has
+// been more than the statusInterval since the last synchronization. This means we will
+// synchronize the upgradeable status more frequently at beginning of an upgrade for
+// a while in a case the precondition checks are failing.
```
jottofar, line 458: > // ...This means we will
> // synchronize the upgradeable status more frequently at beginning of an upgrade for
> // a while in a case the precondition checks are failing.
How about just:
`This means, upon precondition failure, we will synchronize the upgradeable status more frequently at beginning of an upgrade.`

pkg/cvo/upgradeable.go
```diff
@@ -442,3 +449,31 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
...
+// Function returns true if the synchronization should happen, returns false otherwise.
+//
+// A precondition check preventing an upgrade could have been potentially already resolved
+// (e.g. operator reporting Upgradeable==false just momentarily).
+// CVO should synchronize the upgradeable status again in case of failing precondition
+// checks to speed up initializing an upgrade stuck on a precondition check that
+// potentially has been solved.
+//
+// We don't want to check this forever in case of precondition checks failing for a long
+// time due to a bigger problem that won't resolve by itself in a timely manner.
```
jottofar, line 471: This was already explained so I would drop.

pkg/cvo/upgradeable.go
```diff
@@ -442,3 +449,31 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
...
+//
+// shouldSyncUpgradeableDueToPreconditionChecks expects the parameters not to be nil.
+//
+// Function returns true if the synchronization should happen, returns false otherwise.
+//
+// A precondition check preventing an upgrade could have been potentially already resolved
+// (e.g. operator reporting Upgradeable==false just momentarily).
+// CVO should synchronize the upgradeable status again in case of failing precondition
+// checks to speed up initializing an upgrade stuck on a precondition check that
+// potentially has been solved.
```
jottofar, line 468: Not sure this is needed for this function's godoc but if so I would try and combine with the above and make it all more concise or it may not be read.

pkg/cvo/cvo.go
```diff
@@ -627,7 +627,7 @@ func (optr *Operator) upgradeableSync(ctx context.Context, key string) error {
 		return nil
 	}
 
-	return optr.syncUpgradeable()
+	return optr.syncUpgradeable(config)
```
wking, line 630: The CVO using the in-cluster ClusterVersion to pass state from one goroutine to another feels sticky... I'm not set against it, but it would be nice if we could find a way to plumb the data we need through without having to pass this thing we just got from the cluster here.  `syncUpgradeable` is already an `Operator` method, so it should have access to a lot of state already...

pkg/cvo/upgradeable.go
```diff
@@ -92,8 +93,14 @@ type upgradeable struct {
 	Conditions []configv1.ClusterOperatorStatusCondition
 }
 
+// hasPassedDurationSinceTime checks if a certain amount of time specified by duration
+// has passed since time. Returns true if the duration has passed, returns false otherwise.
+func hasPassedDurationSinceTime(t time.Time, duration time.Duration) bool {
+	return t.Before(time.Now().Add(-duration))
```
wking, line 99: nit: brains are different, but mine finds it easier to read as:

```go
time.Now().After(t.Add(duration))
```

pkg/cvo/upgradeable.go
wking: this feels like a separate enough thing, let's just declare a new property for it, or hard-code it here.

wking: or maybe we don't even need the `statusInterval` too-fast guard at all.  Looking for `Started syncing upgradeable...` in [a recent CI run][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-aws-sdn-serial/1557180107716562944/artifacts/e2e-aws-sdn-serial/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-69fc9586b6-h8fjp_cluster-version-operator.log | grep 'Started syncing upgradeable' | tail
I0810 04:21:46.656158       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:21:46.656147659 +0000 UTC m=+7759.515637877)
I0810 04:22:37.695673       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:22:37.695665773 +0000 UTC m=+7810.555155995)
I0810 04:26:10.068836       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:26:10.068828676 +0000 UTC m=+8022.928318900)
I0810 04:29:42.441410       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:29:42.44139783 +0000 UTC m=+8235.300888061)
I0810 04:33:14.814429       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:33:14.814418303 +0000 UTC m=+8447.673908518)
I0810 04:34:53.026298       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:34:53.026289695 +0000 UTC m=+8545.885779917)
I0810 04:38:25.399818       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:38:25.399809113 +0000 UTC m=+8758.259299332)
I0810 04:41:57.772956       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:41:57.772946757 +0000 UTC m=+8970.632436978)
I0810 04:45:30.147801       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:45:30.147790368 +0000 UTC m=+9183.007280583)
I0810 04:49:02.520685       1 cvo.go:615] Started syncing upgradeable "openshift-cluster-version/version" (2022-08-10 04:49:02.520678086 +0000 UTC m=+9395.380168305)
```

I haven't dug into code to be clearer on the trigger, but it looks like it can already be ~minutes between triggers.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-aws-sdn-serial/1557180107716562944

wking: Hmm, in fact:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-aws-sdn-serial/1557180107716562944/artifacts/e2e-aws-sdn-serial/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-69fc9586b6-h8fjp_cluster-version-operator.log | grep -o 'Started syncing upgradeable\|Upgradeable conditions were recently checked' | sort | uniq -c
     59 Started syncing upgradeable
     34 Upgradeable conditions were recently checked
```

Even if we completely drop the guards, we're still only ~doubling the number of times we'd sync.  If the idea is to noticeably lower latency, we might need to go figure out that trigger so we can trigger more frequently.

pkg/cvo/upgradeable.go
Davoska: The flow of the code, logic, and debugging was resolved using option A. Marking this discussion as resolved.

pkg/cvo/upgradeable.go
LalatenduMohanty: I am fine with dropping the guards. If it backfires we can comeback and fix it.

pkg/cvo/cvo.go
LalatenduMohanty: I had seen this, but somehow did not spend time figuring out if we can avoid that.

pkg/cvo/upgradeable.go
Davoska: Ah, yes, I agree. I'll try to improve the comments adequately for the new commit.

pkg/cvo/upgradeable.go
Davoska: Thanks, much better! I didn't think of that.

pkg/cvo/cvo.go
Davoska: Yes, I wasn't keen on it either. However, the `syncAvailableUpdates()` in `pkg/cvo/availableupdates.go` uses the cluster version as a parameter also (https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/availableupdates.go#L31), so I ended up using here as well. (Now I see the commit https://github.com/openshift/cluster-version-operator/commit/db150e6db4f6d00c553804f46a531da4ee0f5552 for that change was from 4 years ago).

Another alternative I can think of is calling the 
```go
config, err := optr.cvLister.Get(optr.name)
```
in `syncUpgradeable()` but I'll try to think of a better implementation.

LalatenduMohanty: I do not think it is a bad idea to pass `config` in `optr.syncUpgradeable(config)` because `upgradeableSync()`  is aleady in a go-routine and it is calling the `optr.cvLister.Get(optr.name)` . So I do not think there is mu h difference between  `upgradeableSync()`  or `SyncUpgradeable()` calling `optr.cvLister.Get(optr.name)`

pkg/cvo/upgradeable.go
LalatenduMohanty: I will still prefer to merge this as it is but we can comeback later to ` go figure out that trigger so we can trigger more frequently.` as @wking suggested.

Davoska: I had a bad understanding of the times when the `upgradeableSync()` gets called.

As of this moment the `upgradeableSync()` only gets called by the worker when a key is added to the `optr.upgradeableQueue`.
https://github.com/openshift/cluster-version-operator/blob/2f174bfbf22e88f9636e10f97e6c107c91ba92ae/pkg/cvo/cvo.go#L388-L392

The only time a key is added to the `optr.upgradeableQueue` is by the `eventHandler()` for the cluster version in the `AddFunc` and `UpdateFunc`.
https://github.com/openshift/cluster-version-operator/blob/2f174bfbf22e88f9636e10f97e6c107c91ba92ae/pkg/cvo/cvo.go#L462-L479

Originally the equation for calculating the next time of executing `upgradeableSync()` would be something like: `optr.minimumUpdateCheckInterval + DURATION_TO_NEXT_UPDATE_OF_THE_CLUSTER_VERSION`

A problem could arise in the future again in the case where the cluster version is updated less often and more frequent checks for the Upgradeable are needed. Thus dropping the guard still wouldn't be sufficient as Trevor pointed out.

However, the time-sensitive part of the code remains the same - initializing an upgrade.

A potential idea could be something like running periodically (every ~15 seconds):
```go
optr.upgradeableQueue.Add(workQueueKey)
```
at the beginning of an upgrade in a case, the CVO's `Upgradeable` is `False`.

In this case, it would be **guaranteed** to run the `upgradeableSync()` minimally every 15 seconds.

 But if we presume there is a sufficient number of updates to the cluster version at the beginning of an upgrade so that the keys get added to the `upgradeableQueue` we could potentially ignore this.

pkg/cvo/upgradeable.go
Davoska: Added in https://github.com/openshift/cluster-version-operator/commit/5aac7a3ef2934cbeae09a2632e22fe57eb20da91.

pkg/cvo/upgradeable.go
Davoska: Dropped in the new commit https://github.com/openshift/cluster-version-operator/commit/5aac7a3ef2934cbeae09a2632e22fe57eb20da91.

pkg/cvo/upgradeable.go
Davoska: Thanks for the better wording, Jack! 
Comments replaced in the new commit https://github.com/openshift/cluster-version-operator/commit/5aac7a3ef2934cbeae09a2632e22fe57eb20da91.

Davoska:
/unhold
Comments were addressed.

pkg/cvo/cvo.go
wking: Right, I don't think it matters if we shift the `Get` around.  What I'd like long-term is no get at all in this pathway.  Looking [here][1], [here][2], and [here][3], something like `optr.getUpgradeable().Conditions` will get you the current conditions without having to leave local CVO memory.  Although a lock guarding access to a pointer property seems a bit suspect...

[1]: https://github.com/openshift/cluster-version-operator/blob/f7f9b8d0b34cb6349478869c529ab6cd575a55b2/pkg/cvo/cvo.go#L123-L125
[2]: https://github.com/openshift/cluster-version-operator/blob/f7f9b8d0b34cb6349478869c529ab6cd575a55b2/pkg/cvo/upgradeable.go#L91-L92
[3]: https://github.com/openshift/cluster-version-operator/blob/f7f9b8d0b34cb6349478869c529ab6cd575a55b2/pkg/cvo/upgradeable.go#L143-L149

LalatenduMohanty (approved):
/lgtm

pkg/cvo/upgradeable.go
LalatenduMohanty: @Davoska You suggestion is same as reducing optr.minimumUpdateCheckInterval  to 15 seconds or 30 seconds. I believe in the beginning of the upgrade cycle we see more frequent updates but not sure. We can check logs to confirm this. 

Also there was recent changes to the event handler methods https://github.com/openshift/cluster-version-operator/commit/9222fa9a6616b58a8056c780b9a6252e82a26e37

LalatenduMohanty: @wking and I talked about this and we think the solution in this PR is doing the right thing.

In https://github.com/openshift/cluster-version-operator/blob/master/pkg/cvo/cvo.go#L397 It is calling optr.worker() every second.  If there is any work then it gets processed by processNextWorkItem() but if there is not then we just continue checking every second till we get work. So with this change if there is a change in upgradeable then with in 1 second CVO will start the process of reconciliation.

PR #818, opened 2022-08-09
Bug 2117033: pkg/cvo/sync_worker: Trigger new sync round on ClusterOperator versions[name=operator] changes

wking:
@deads2k and @stbenjam [identified an uneccessary delay][1]:

* 9:42:00, CVO gives up on Kube API server ClusterOperator
* 9:42:47, Kube API server operator achieves 4.12
* 9:46:22, after a cool-off sleep, the CVO starts in on a new manifest graph-walk attempt
* 9:46:34, CVO notices that the Kube API server ClusterOperator is happy

The 3+ minute delay from 9:42:47 to 9:46:22 is not helpful, and we've probably had delays like this since my old e02d1489a5 (#560), which landed in 4.6.

This commit introduces a "ClusterOperator bumped `versions[name=operator]`" trigger to break out of the cool-off sleep.

There's plenty of room to be more precise here.  For example, you could currently have a `versions[name=operator]` bump during the sync loop that the CVO did notice, and that queued notification will break from the sleep and trigger a possible useless reconciliation round while we wait on some other resource.  You could drain the notification queue before the sleep to avoid that, but you wouldn't want to drain new-work notifications, and I haven't done the work required to be able to split those apart.

I'm only looking at ClusterOperator at the moment, because of the many types the CVO manages, ClusterOperator is the one we most frequently wait on, as large cluster components take their time updating.  It's possible but less likely that we'd want similar triggers for additional types in the future (Deployment, etc.), if/when those types develop more elaborate "is the in-cluster resource sufficient happy?" checks.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2117033#c1

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -197,10 +197,11 @@ func New(
 		clusterProfile:     clusterProfile,
 	}
 
-	cvInformer.Informer().AddEventHandler(optr.eventHandler())
+	cvInformer.Informer().AddEventHandler(optr.clusterVersionEventHandler())
 	cmConfigInformer.Informer().AddEventHandler(optr.adminAcksEventHandler())
 	cmConfigManagedInformer.Informer().AddEventHandler(optr.adminGatesEventHandler())
 
+	coInformer.Informer().AddEventHandler(optr.clusterOperatorEventHandler())
 	optr.coLister = coInformer.Lister()
 	optr.cacheSynced = append(optr.cacheSynced, coInformer.Informer().HasSynced)
 
@@ -458,9 +459,9 @@ func (optr *Operator) queueKey() string {
 	return fmt.Sprintf("%s/%s", optr.namespace, optr.name)
 }
 
-// eventHandler queues an update for the cluster version on any change to the given object.
+// clusterVersionEventHandler queues an update for the cluster version on any change to the given object.
 // Callers should use this with a scoped informer.
-func (optr *Operator) eventHandler() cache.ResourceEventHandler {
+func (optr *Operator) clusterVersionEventHandler() cache.ResourceEventHandler {
 	workQueueKey := optr.queueKey()
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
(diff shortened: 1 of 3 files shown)


pkg/cvo/cvo.go
```diff
@@ -479,6 +480,36 @@ func (optr *Operator) eventHandler() cache.ResourceEventHandler {
...
+				optr.ConfigSyncWorker.NotifyAboutManagedResourceActivity(new, msg)
+			}
+		},
+	}
+}
+
+func clusterOperatorInterfaceVersionOrDie(obj interface{}, name string) (*configv1.ClusterOperator, string) {
+	co, ok := obj.(*configv1.ClusterOperator)
+	if !ok {
+		panic(fmt.Sprintf("%v is %T, not a ClusterOperator", obj, obj))
```
wking, line 502: This feels like a hack, but I wasn't able to find a cleaner way to get at the structured information I want.  Improvements welcome :)

wking:
[Update presubmit][1] failed on the likely orthogonal `events should not repeat pathologically` `pod/machine-api-controllers... Readiness probe error: Get "http://10.129.0.14:9442/healthz": dial tcp 10.129.0.14:9442: connect: connection refused`.  But dropping into Loki with:

```
{invoker="openshift-internal-ci/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1557150272407474176"} | unpack | namespace="openshift-cluster-version" |~ "clusteroperator \"kube-apiserver|in state|[Nn]otif"
```

turns up:

```
...
E0810 01:03:39.710537       1 task.go:117] error running apply for clusteroperator "kube-apiserver" (107 of 816): Cluster operator kube-apiserver is updating versions
I0810 01:07:09.150368       1 sync_worker.go:235] Notify the sync worker: Cluster operator kube-apiserver changed versions[name="operator"] from "4.12.0-0.ci.test-2022-08-09-234236-ci-op-dmywi2n7-initial" to "4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest"
I0810 01:07:09.150539       1 sync_worker.go:864] apply: 4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest on generation 3 in state Updating at attempt 5
...
I0810 01:07:16.712837       1 sync_worker.go:987] Done syncing for clusteroperator "kube-apiserver" (107 of 816)
...
```

Comparing with the test-suite monitor:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/818/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1557150272407474176/artifacts/e2e-agnostic-upgrade/openshift-e2e-test/build-log.txt | grep 'clusteroperator/kube-apiserver versions:'
Aug 10 00:58:05.747 I clusteroperator/kube-apiserver versions: raw-internal 4.12.0-0.ci.test-2022-08-09-234236-ci-op-dmywi2n7-initial -> 4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest
Aug 10 01:07:09.155 I clusteroperator/kube-apiserver versions: operator 4.12.0-0.ci.test-2022-08-09-234236-ci-op-dmywi2n7-initial -> 4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest
Aug 10 02:08:47.822 I clusteroperator/kube-apiserver versions: raw-internal 4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest -> 4.12.0-0.ci.test-2022-08-09-234236-ci-op-dmywi2n7-initial
Aug 10 02:17:51.529 I clusteroperator/kube-apiserver versions: operator 4.12.0-0.ci.test-2022-08-09-235304-ci-op-dmywi2n7-latest -> 4.12.0-0.ci.test-2022-08-09-234236-ci-op-dmywi2n7-initial
```

So the CVO tests are doing A->B->A rollbacks, and I'm looking at the A->B leg in Loki (which is the leg managed by the incoming CVO that has my patch), and the test-suite monitor sees the transition at 01:07:09.155, and the new sync round kicks off at 01:07:09.150539, and the CVO gets deep enough into the manifest graph to notice the kube-apiserver's completion at 01:07:16.712837.  So down from possibly-minutes to seconds :tada:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/818/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1557150272407474176

pkg/cvo/cvo.go
LalatenduMohanty: I am not sure if we need a panic here. If clusterOperatorEventHandler() can be called on an operator why do we need to check if the obj is not an operator?

LalatenduMohanty (approved):
/lgtm

wking:
[`pods should successfully create sandboxes by adding pod to network`][1] is orthogonal.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/818/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1557598894374785024

PR #821, opened 2022-08-15
allow more than one featureset

deads2k:
This is a sketch of how we could allow other featuresets like `""` to be specified in teh annotations.

If no annotation is specified, then it's always present. If any `release.openshift.io/feature-gate` is specified then the manifest only used when the value matches the currently specified featureset.  Also, present-me wishes that past-me had distinguished between gates and sets better.  I wonder if I can change the world still. I might be able to transition during 4.12.


pkg/cvo/cvo.go
```diff
@@ -141,9 +141,9 @@ type Operator struct {
 	// via annotation
 	exclude string
 
-	// includeTechPreview is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
+	// reqiredFeatureSet is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
```
haircommander, line 144: nit: missing a u in `requiredFeatureSet` everywhere in this file

deads2k:
> waiting for the deletion of Security Rule

/retest

pkg/cvo/cvo.go
wking: and stepping back a bit further, can multiple feature sets be simultaneously enabled?  Maybe this should be `enabledFeatureSets []string`?

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -117,7 +117,7 @@ func setupCVOTest(payloadDir string) (*Operator, map[string]runtime.Object, *fak
 			Steps: 1,
 		},
 		"exclude-test",
-		false,
+		"Default",
```
wking, line 120: unclear to me if `Default` or the empty-string is the default value.  Whichever way we go, can you doc it in at least the `Operator` property declaration?

pkg/cvo/cvo.go
deads2k: Only one at a time enforced via the API type.   https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L55

pkg/cvo/sync_worker.go
```diff
@@ -183,15 +183,15 @@ type SyncWorker struct {
 	// of the form exclude.release.openshift.io/<identifier>=true
 	exclude string
 
-	// includeTechPreview is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
-	includeTechPreview bool
+	// requiredFeatureSet is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
```
wking, line 186: nit: Replace `to true`, etc. with whatever the `string` semantics are supposed to be.

pkg/start/start.go
```diff
@@ -132,15 +132,21 @@ func (o *Options) Run(ctx context.Context) error {
 	}
 	// check to see if techpreview should be on or off.  If we cannot read the featuregate for any reason, it is assumed
 	// to be off.  If this value changes, the binary will shutdown and expect the pod lifecycle to restart it.
-	includeTechPreview := false
+	startingFeatureSet := ""
 	gate, err := cb.ClientOrDie("feature-gate-getter").ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
 	switch {
 	case apierrors.IsNotFound(err):
-		includeTechPreview = false // if we have no featuregates, then we aren't tech preview
+		startingFeatureSet = "" // if we have no featuregates, then we assume the default featureset, which is "".
```
wking, line 139: it needs to be "then we exclude everything that could possibly depend on the current feature set".

pkg/start/start.go
```diff
@@ -132,15 +132,21 @@ func (o *Options) Run(ctx context.Context) error {
...
 		klog.Warningf("Error getting featuregate value: %v", err)
+
 	default:
-		includeTechPreview = gate.Spec.FeatureSet == configv1.TechPreviewNoUpgrade
+		// otherwise, you're the default
+		startingFeatureSet = string(gate.Spec.FeatureSet)
+	}
+	requiredFeatureSetAnnotationValue := "Default" // When the featureset is "", we require it to say "Default"
+	if len(startingFeatureSet) > 0 {
+		requiredFeatureSetAnnotationValue = startingFeatureSet
```
wking, line 149: do we need to make this call way up the stack in `start.go`?  It feels like something we could push down into library-go.

vendor/github.com/openshift/library-go/pkg/manifest/manifest.go
```diff
@@ -114,11 +116,46 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+	ret := sets.String{}
+	specified := false
+	for _, featureSetAnnotation := range []string{"release.openshift.io/feature-gate", "release.openshift.io/feature-set"} {
+		featureSetAnnotationValue, featureSetAnnotationExists := annotations[featureSetAnnotation]
+		if featureSetAnnotationExists {
+			specified = true
+			featureSetAnnotationValues := strings.Split(featureSetAnnotationValue, ",")
+			for _, manifestFeatureSet := range featureSetAnnotationValues {
+				if !knownFeatureSets.Has(manifestFeatureSet) {
+					// never include the manifest if the feature-set annotation is outside of allowed values (only TechPreviewNoUpgrade and "" are currently allowed)
```
wking, line 143: Can we drop `(only TechPreviewNoUpgrade and "" are currently allowed)`, because I expect it would go stale, and may already be stale with your `configv1.FeatureSets` source for `knownFeatureSets`.

vendor/github.com/openshift/library-go/pkg/manifest/manifest.go
```diff
@@ -114,11 +116,46 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+			continue
+		}
+		knownFeatureSets.Insert(string(featureSet))
+	}
+}
+
+func getFeatureSets(annotations map[string]string) (sets.String, bool, error) {
+	ret := sets.String{}
+	specified := false
+	for _, featureSetAnnotation := range []string{"release.openshift.io/feature-gate", "release.openshift.io/feature-set"} {
```
wking, line 136: nit: use your `featureSetAnnotation` for one of these?

vendor/github.com/openshift/library-go/pkg/manifest/manifest.go
```diff
@@ -132,15 +169,13 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+	if requiredFeatureSet != nil {
+		manifestFeatureSets, manifestSpecifiesFeatureSets, err := getFeatureSets(annotations)
+		if err != nil {
+			return err
 		}
-		// never include the manifest if the feature-gate annotation is outside of allowed values (only TechPreviewNoUpgrade is currently allowed)
-		if featureGateAnnotationExists && featureGateAnnotationValue != string(configv1.TechPreviewNoUpgrade) {
-			return fmt.Errorf("unrecognized value %s=%s", featureGateAnnotation, featureGateAnnotationValue)
+		if manifestSpecifiesFeatureSets && !manifestFeatureSets.Has(*requiredFeatureSet) {
+			return fmt.Errorf("%q is required, and %s=%s", *requiredFeatureSet, featureSetAnnotation, strings.Join(manifestFeatureSets.List(), ","))
```
wking, line 178: nit: this assumes `featureSetAnnotation`, and not the old annotation name.  Is the idea that we drop the old name before we GA 4.12, once existing folks have migrated?

deads2k: > nit: this assumes featureSetAnnotation, and not the old annotation name. Is the idea that we drop the old name before we GA 4.12, once existing folks have migrated?

correct

pkg/cvo/cvo.go
```diff
@@ -145,9 +145,9 @@ type Operator struct {
 	// via annotation
 	exclude string
 
-	// includeTechPreview is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
+	// requiredFeatureSet is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
 	// label set.  This is set based on whether the featuregates.config.openshift.io|.spec.featureSet is set to "TechPreviewNoUpgrade".
```
wking, line 149: `set to true` and the `TechPreviewNoUpgrade` focus are stale.

pkg/cvo/sync_worker.go
```diff
@@ -184,15 +184,18 @@ type SyncWorker struct {
 	// of the form exclude.release.openshift.io/<identifier>=true
 	exclude string
 
-	// includeTechPreview is set to true when the CVO should create resources with the `release.openshift.io/feature-gate=TechPreviewNoUpgrade`
-	includeTechPreview bool
+	// requiredFeatureSet is set to the value of Feature.config.openshift.io|spec.featureSet.
+	// The CVO should create resources with the `annotations[release.openshift.io/feature-set]` unset or if
+	// the annotation is set, it must contain the requiredFeatureSet.
+	// The library called by the CVO translates "" into "Default" to ease usage.
```
wking, line 190: We might not need to get into this level of detail about what's going on inside the vendored library.  Maybe something more generic like:

> requiredFeatureSet is set to the value of Feature.config.openshift.io|spec.featureSet, which contributes to whether or not some manifests are included for reconciliation.

pkg/featurechangestopper/techpreviewchangestopper.go
```diff
@@ -20,7 +20,7 @@ import (
 // TechPreviewChangeStopper calls stop when the value of the featuregate changes from TechPreviewNoUpgrade to anything else
 // or from anything to TechPreviewNoUpgrade.
```
wking, line 21: Too late for us to rename this to `FeatureChangeStopper` to match the directory name you initially selected?  Might be able to do that with:

```console
$ sed -i 's/TechPreviewChangeStopper/FeatureChangeStopper/g' pkg/featurechangestopper/*
$ mv pkg/featurechangestopper/{techpreview,feature}changestopper.go
$ mv pkg/featurechangestopper/{techpreview,feature}changestopper_test.go
```

or similar.

pkg/start/start.go
```diff
@@ -132,15 +132,19 @@ func (o *Options) Run(ctx context.Context) error {
...
 	// to be off.  If this value changes, the binary will shutdown and expect the pod lifecycle to restart it.
-	includeTechPreview := false
+	startingFeatureSet := ""
 	gate, err := cb.ClientOrDie("feature-gate-getter").ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
 	switch {
 	case apierrors.IsNotFound(err):
-		includeTechPreview = false // if we have no featuregates, then we aren't tech preview
+		// if we have no featuregates, then we assume the default featureset, which is "".
+		// This excludes everything that could possibly depend on a different feature set.
+		startingFeatureSet = ""
```
wking, line 141: Do we need to assume here?  If we have no `cluster` FeatureGate, can we be in any other set?  Isn't that a clear (if implicit) declaration that we're in the default set?

pkg/start/start.go
```diff
@@ -132,15 +132,19 @@ func (o *Options) Run(ctx context.Context) error {
...
 	case err != nil:
 		klog.Warningf("Error getting featuregate value: %v", err)
```
wking, line 143: Hmm, we may need a pointer for passing this thing around (luckily, library-go is already using a pointer).  openshift/enhancements#1227 currently includes:

> If a special manifest is required for a FeatureSet that allows changes (LatencySensitive to Default for instance), the person adding the manifest will have to add the appropriate delete-annotated manifest in Default.  This is expected to be a rare occurrence of a rare occurrence.

So there's a risk here of:

1. User configures LatencySensitive.
2. Resource A, that only happens in LatencySensitive, rolls out.
3. Happy days.
4. Something bumps the CVO.
5. New CVO comes up, but network hiccup fails the FeatureGate GET.
6. We enter this `err != nil` case, and warn, but leave `startingFeatureSet == ""`.
7. Reconciling manifests, we exclude the LatencySensitive resource A manifest (meh), but include the deletion manifest (!), and remove resource A from the cluster.
8. Change-detector has a successful GET, notices the mistake, and shuts down the CVO.
9. Replacement CVO has a successful GET, realizes it's LatencySensitive, and rolls resource A back into the cluster.

That's not the end of the world, but I'd feel more comfortable with pointers tracking whether we knew or not about the feature gate, and... hmm, I guess there's not a clear way to tell library-go to exclude everything that declares a feature-set annotation.  Maybe we don't need a pointer, but we do need a dummy value here like:

```go
startingFeatureSet = "not sure, so exclude everything that might care"
```

?

pkg/start/start.go
```diff
@@ -132,15 +132,19 @@ func (o *Options) Run(ctx context.Context) error {
...
+
 	default:
-		includeTechPreview = gate.Spec.FeatureSet == configv1.TechPreviewNoUpgrade
+		// otherwise, you're the default
```
wking, line 146: what is this comment about?  Just reminding us that we're in the `default:` case?

deads2k:
pushed another commit for the comments.

pkg/featurechangestopper/featurechangestopper.go
```diff
@@ -17,10 +17,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-// TechPreviewChangeStopper calls stop when the value of the featuregate changes from TechPreviewNoUpgrade to anything else
+// FeatureChangeStopper calls stop when the value of the featuregate changes from TechPreviewNoUpgrade to anything else
 // or from anything to TechPreviewNoUpgrade.
```
wking, line 21: Stale `TechPreviewNoUpgrade` focus.  Maybe:

> FeatureChangeStopper calls stop when the value of the featuregate changes.

?

wking (approved):
/lgtm

PR #827, opened 2022-08-19
lib/capability: Sort all slices after building them

wking:
We'd already been sorting the two status properties before attaching them to ClusterVersion, but we also use some of the underlying data internally, e.g. in `TestCVO_InitImplicitlyEnabledCaps`.  By sorting after we build each slice, we don't need to worry about slice-consumers being surprised by order instability.

### DIFF
--- a/lib/capability/capability.go
+++ b/lib/capability/capability.go
@@ -79,6 +79,7 @@ func GetKnownCapabilities() []configv1.ClusterVersionCapability {
 	for _, v := range configv1.ClusterVersionCapabilitySets {
 		known = append(known, v...)
 	}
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

wking:
[Registry uptime][1] should be addressed by openshift/release#31518.  Until then:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/827/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade/1560651902012624896

PR #832, opened 2022-09-08
OCPBUGS-1402: pkg/cvo/sync_worker.go: remove Lock/Unlock

jottofar:
from `SyncWorker.updateLoadStatus` and `SyncWorker.Update`. This then keeps SyncWorker locked during the entire `Update` call not allowing occurrences as described by https://bugzilla.redhat.com/show_bug.cgi?id=2118289#c1. With this change `SyncWorker` must already be locked before `SyncWorker.updateLoadStatus` is called. But currently only `SyncWorker.Update` calls `SyncWorker.updateLoadStatus`.

### DIFF
--- a/pkg/cvo/cvo_scenarios_test.go
+++ b/pkg/cvo/cvo_scenarios_test.go
@@ -1822,7 +1822,7 @@ func TestCVO_ResetPayloadLoadStatus(t *testing.T) {
 }
 
 func TestCVO_UpgradeFailedPayloadLoadWithCapsChanges(t *testing.T) {
-	o, cvs, client, _, shutdownFn := setupCVOTest("testdata/payloadtest-2")
+	o, cvs, client, _, shutdownFn := setupCVOTest("testdata/payloadtest-3")
 
 	// Setup: load and apply payload which sets "work" to non-nil
 	//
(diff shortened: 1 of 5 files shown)


jottofar:
/retitle Bug 2118289: pkg/cvo/sync_worker.go: remove Lock/Unlock

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -1974,9 +1975,9 @@ func TestCVO_UpgradeFailedPayloadLoadWithCapsChanges(t *testing.T) {
 				{Type: ImplicitlyEnabledCapabilities, Status: "False", Reason: "AsExpected", Message: "Capabilities match configured spec"},
 				{Type: DesiredReleaseAccepted, Status: configv1.ConditionFalse, Reason: "VerifyPayloadVersion",
 					Message: "Verifying payload failed version=\"1.0.0-abc\" image=\"image/image:0\" failure=release image version 1.0.1-abc does not match the expected upstream version 1.0.0-abc"},
-				{Type: "Available", Status: "True", Message: "Done applying 1.0.1-abc"},
+				{Type: "Available", Status: "False"},
```
wking, line 1978: Do we understand why we're `Available=False` here with no `Reason` or `Message`?

pkg/cvo/sync_worker.go
```diff
@@ -499,9 +499,7 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 		}
 	}
 
-	w.lock.Unlock()
 	implicit, err := w.loadUpdatedPayload(ctx, work, cvoOptrName)
```
wking, line 502: There's another unlock/lock around a different `loadUpdatedPayload` ~30 lines up^^ that you probably want to drop too.

pkg/cvo/cvo_scenarios_test.go
jottofar: This final status reaping has always been very time sensitive. Appears with the lock changes allowing the Update more time the sync doesn't get as far, and does not complete. To get to completion I reduced test payload to one file and put in an additional wait for the completion of sync'ing the original payload.

jottofar:
/retitle OCPBUGS-1402: pkg/cvo/sync_worker.go: remove Lock/Unlock

wking (approved):
/lgtm

PR #840, opened 2022-09-23
OCPBUGS-1636: pkg/cvo/sync_worker: Pre-create ClusterOperator in reconciling-mode too

wking:
Originally, all component operators were responsible for creating their own ClusterOperator, and we'd just watch to make sure we were happy enough with what they did.  However, on install, or when updating to a version that added a new component, we could have timelines like:

1. CVO creates a namespace for an operator.
2. CVO creates ... for the operator.
3. CVO creates the operator Deployment.
4. Operator deployment never comes up, for whatever reason.
5. Admin must-gathers.
6. Must gather uses ClusterOperators for discovering important stuff, and because the ClusterOperator doesn't exist yet, we get no data about why the deployment didn't come up.

So in 2a469e37c1 (#318), we added ClusterOperator pre-creation to get:

1. CVO pre-creates ClusterOperator for an operator.
2. CVO creates the namespace for an operator.
3. CVO creates ... for the operator.
4. CVO creates the operator Deployment.
5. Operator deployment never comes up, for whatever reason.
6. Admin must-gathers.
7. Must gather uses ClusterOperators for discovering important stuff, and finds the one the CVO had pre-created with hard-coded relatedObjects, gathers stuff from the referenced operator namespace, and allows us to trouble-shoot the issue.

However, all existing component operators already knew how to create their own ClusterOperator, because that was the only path before the CVO learned about pre-creation.  And even since then, most new operators come into the cluster on install or on update, when the CVO is pre-creating.  New in 4.12, [the platform-operator is coming in][1], and it has two relevant characteristics:

* It [does not know how][2] to create the platform-operators-aggregated ClusterOperator.
* It is [gated behind TechPreviewNoUpgrade][3].

So we are exposed to:

1. Admin installs a cluster.  No platform-operators-aggregated, because it's not TechPreviewNoUpgrade.
2. Install complete.  CVO transitions to reconciling mode.
3. Admin enables TechPreviewNoUpgrade.
4. CVO notices, and reboots fc00c62199 (#821).
5. Because we decided to not transition into updating mode for feature-set changes, we stay in reconciling mode.
6. Because we're in reconciling mode, we skip the ClusterOperator pre-creation, and get right in to the status check.
7. Because the platform operator didn't create the ClusterOperator either, the CVO's status check [fails with][2]:

        45657:E0923 01:43:25.610286       1 task.go:117] error running apply for clusteroperator "openshift-platform-operators/platform-operators-aggregated" (587 of 960): clusteroperator.config.openshift.io "platform-operators-aggregated" not found

With this commit, I stop making the ClusterOperator pre-creation conditional, so the new flow is:

...
6. Even in reconciling mode, we pre-create the ClusterOperator.
7. Because we pre-created the ClusterOperator, the CVO's status check succeeds (at least, after the operator writes acceptable status to the ClusterOperator we've created for it).

This will also help us recover components where a bunch of in-cluster resources had been deleted, assuming the CVO was still alive.  There may be other component operators who rely on the CVO for ClusterOperator creation, but which we haven't noticed because they aren't also gated behind TechPreviewNoUpgrade.

[1]: https://github.com/openshift/enhancements/blob/6e1697418be807d0ae567a9f83ac654a1fd0ee9a/enhancements/olm/platform-operators.md
[2]: https://issues.redhat.com/browse/OCPBUGS-1636
[3]: https://github.com/openshift/platform-operators/blob/4ecea427cf5302dfcdf4a5af8d28eadebacc2037/manifests/0000_50_cluster-platform-operator-manager_07-aggregated-clusteroperator.yaml#L8

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -908,14 +908,12 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 
 	graph := payload.NewTaskGraph(tasks)
 	graph.Split(payload.SplitOnJobs)
-	var precreateObjects bool
 	switch work.State {
 	case payload.InitializingPayload:
 		// Create every component in parallel to maximize reaching steady
 		// state.
 		graph.Parallelize(payload.FlattenByNumberAndComponent)
 		maxWorkers = len(graph.Nodes)
-		precreateObjects = true
 	case payload.ReconcilingPayload:
 		// Run the graph in random order during reconcile so that we don't
 		// hang on any particular component - we seed from the number of
@@ -933,7 +931,6 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 		// perform an orderly roll out by payload order, using some parallelization
 		// but avoiding out of order creation so components have some base
 		graph.Parallelize(payload.ByNumberAndComponent)
-		precreateObjects = true
 	}
 	capabilities := capability.GetCapabilitiesStatus(work.Capabilities)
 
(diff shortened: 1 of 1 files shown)


jianzhangbjz:
Passed, test steps see comments: https://issues.redhat.com/browse/OCPBUGS-1636
/label qe-approved

LalatenduMohanty (approved):
/lgtm

wking:
[Azure CSI SCC failures][1] are unrelated.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/840/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1576585708724817920

PR #850, opened 2022-10-12
OCPBUGS-2125: Allow unknown capabilities during payload load and implicitl enablement checking

jottofar:
When OCP is being updated, e.g. 4.y -> 4.y+1, the outgoing 4.y CVO loads the 4.y+1 payload and also checks if any capabilities must be implicitly enabled. If the 4.y+1 payload contains new capabilities those capabilities will not exist, i.e. are unknown, to the 4.y CVO. Before this change this would result in the new capabilities being flagged as unknown resulting in any manifest from the updated payload that contains them not being loaded. They would also be flagged as unknown in the `payload.GetImplicitlyEnabledCapabilities` function and therefore never implicitly enabled.


wking (approved):
/lgtm

PR #851, opened 2022-10-13
OCPBUGS-1458: Allow CVO to update `KUBERNETES_SERVICE_HOST` with LB address

petr-muller:
The problem was identified to be a broken substitution of internal load balancer into `KUBERNETES_SERVICE_HOST` by Trevor and David (see my [JIRA comment](https://issues.redhat.com/browse/OCPBUGS-1458?focusedCommentId=21090756&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21090756) and related [Slack thread](https://coreos.slack.com/archives/C011CSSPBLK/p1664925995946479?thread_ts=1661182025.992649&cid=C011CSSPBLK)).

CVO injects the LB hostname in the [`ModifyDeployment`](https://github.com/openshift/cluster-version-operator/blob/dc1ad0aef5f3e1b88074448d21445a5bddb6b05b/lib/resourcebuilder/apps.go#L19) fine, but then the deployment gets applied in [`ApplyDeployment`](https://github.com/openshift/cluster-version-operator/blob/dc1ad0aef5f3e1b88074448d21445a5bddb6b05b/lib/resourceapply/apps.go#L17) and the  `EnsureDeployment`->`ensurePodTemplateSpec`->`ensurePodSpec`->`ensureContainers`->`ensureContainer`->`ensureEnvVar` chain stomps the updated value in `required` by the old value from `existing` and reverts the injection.

This behavior was added intentionally in https://github.com/openshift/cluster-version-operator/pull/559 as a part of a fix for various hot-looping issues. The substitution apparently caused some hot-looping issues in the past ([slack thread](https://coreos.slack.com/archives/CEGKQ43CP/p1620934857402200?thread_ts=1620895567.367100&cid=CEGKQ43CP)). I have tested removing the special handling `KUBERNETES_SERVICE_HOST` thoroughly, and saw no problematic behavior. After fixing other hot-looping problems in https://github.com/openshift/cluster-version-operator/pull/855 to eliminate noise, no new hot-loops occurs with `KUBERNETES_SERVICE_HOST` handling removed.

Even with the substitution fix, I observed further network problems in CVO ([example](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/851/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1584592473953406976/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-dd6bdb766-qdctn_cluster-version-operator_previous.log)):

```
F1024 17:40:24.591942       1 start.go:29] error: Get "https://api-int.ci-op-02b7hikv-3302f.ci.azure.devcluster.openshift.com:6443/apis/config.openshift.io/v1/featuregates/cluster": dial tcp: lookup api-int.ci-op-02b7hikv-3302f.ci.azure.devcluster.openshift.com on 172.30.0.10:53: read udp 169.254.169.2:48475->172.30.0.10:53: i/o timeout
```

To get rid of these, I added a client-side poll to CVO, but I currently fail to reproduce the issue in https://github.com/openshift/cluster-version-operator/pull/859, so maybe the `i/o timeout` issues were just some co-indidental disruption bug that we happened to fix in the meantime? Not sure.

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -16,6 +16,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -133,19 +134,37 @@ func (o *Options) Run(ctx context.Context) error {
 	// check to see if techpreview should be on or off.  If we cannot read the featuregate for any reason, it is assumed
 	// to be off.  If this value changes, the binary will shutdown and expect the pod lifecycle to restart it.
 	startingFeatureSet := ""
-	gate, err := cb.ClientOrDie("feature-gate-getter").ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
-	switch {
-	case apierrors.IsNotFound(err):
-		// if we have no featuregates, then the cluster is using the default featureset, which is "".
-		// This excludes everything that could possibly depend on a different feature set.
-		startingFeatureSet = ""
-	case err != nil:
-		// client-go automatically retries network blip errors on GETs for 30s by default.  If we fail longer than that
-		// the operator won't be able to do work anyway.  Return the error and crashloop.
-		return err
 
-	default:
-		startingFeatureSet = string(gate.Spec.FeatureSet)
+	// client-go automatically retries some network blip errors on GETs for 30s by default, and we want to
+	// retry the remaining ones ourselves. If we fail longer than that, the operator won't be able to do work
+	// anyway. Return the error and crashloop.
+	//
+	// We implement the timeout with a context because the timeout in PollImmediateWithContext does not behave
+	// well when ConditionFunc takes longer time to execute, like here where the GET can be retried by client-go
+	fgCtx, fgCancel := context.WithTimeout(ctx, 25*time.Second)
+	defer fgCancel()
+	var lastError error
+	if err := wait.PollImmediateInfiniteWithContext(fgCtx, 2*time.Second, func(ctx context.Context) (bool, error) {
+		gate, fgErr := cb.ClientOrDie("feature-gate-getter").ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
+		switch {
+		case apierrors.IsNotFound(fgErr):
+			// if we have no featuregates, then the cluster is using the default featureset, which is "".
+			// This excludes everything that could possibly depend on a different feature set.
+			startingFeatureSet = ""
+			return true, nil
+		case fgErr != nil:
+			lastError = fgErr
+			klog.Warningf("Failed to get FeatureGate from cluster: %v", fgErr)
+			return false, nil
+		default:
+			startingFeatureSet = string(gate.Spec.FeatureSet)
+			return true, nil
+		}
+	}); err != nil {
+		if lastError != nil {
+			return lastError
+		}
+		return err
 	}
 
 	lock, err := createResourceLock(cb, o.Namespace, o.Name)
--- a/lib/resourcemerge/core.go
+++ b/lib/resourcemerge/core.go
@@ -112,13 +112,6 @@ func ensureContainer(modified *bool, existing *corev1.Container, required corev1
 
 func ensureEnvVar(modified *bool, existing *[]corev1.EnvVar, required []corev1.EnvVar) {
 	for envidx := range required {
-		// Currently only CVO deployment uses this variable to inject internal LB host.
-		// This may result in an IP address being returned by API so assuming the
-		// returned value is correct.
-		if required[envidx].Name == "KUBERNETES_SERVICE_HOST" {
-			ensureEnvVarKubeService(*existing, &required[envidx])
-		}
-
 		if required[envidx].ValueFrom != nil {
 			ensureEnvVarSourceFieldRefDefault(required[envidx].ValueFrom.FieldRef)
 		}
@@ -126,14 +119,6 @@ func ensureEnvVar(modified *bool, existing *[]corev1.EnvVar, required []corev1.E
 	if !equality.Semantic.DeepEqual(required, *existing) {
 		*existing = required
 		*modified = true
-	}
-}
-
-func ensureEnvVarKubeService(existing []corev1.EnvVar, required *corev1.EnvVar) {
-	for envidx := range existing {
-		if existing[envidx].Name == required.Name {
-			required.Value = existing[envidx].Value
-		}
 	}
 }
 
--- a/lib/resourcemerge/core_test.go
+++ b/lib/resourcemerge/core_test.go
@@ -1640,6 +1640,26 @@ func TestEnsureEnvVar(t *testing.T) {
 			},
 			expectedModified: false,
 		},
+		{
+			name: "CVO can inject LB into ENVVAR",
+			existing: []corev1.EnvVar{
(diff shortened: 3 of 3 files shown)


petr-muller:
Needs https://github.com/openshift/cluster-version-operator/pull/855 first

petr-muller:

Seems unrelated, happens also in other Azure jobs

lib/resourceapply/apps.go
```diff
@@ -29,13 +29,19 @@ func ApplyDeploymentv1(ctx context.Context, client appsclientv1.DeploymentsGette
...
+	var original appsv1.Deployment
+	existing.DeepCopyInto(&original)
 	modified := pointer.BoolPtr(false)
 	resourcemerge.EnsureDeployment(modified, existing, *required)
 	if !*modified {
 		return existing, false, nil
 	}
 	if reconciling {
-		klog.V(2).Infof("Updating Deployment %s/%s due to diff: %v", required.Namespace, required.Name, cmp.Diff(existing, required))
+		if diff := cmp.Diff(&original, existing); diff != "" {
```
jottofar, line 40: Why are we changing this and doing another diff here? This would differ from how it's done everywhere else (which of course doesn't mean it's wrong but do need a good reason for the change).

lib/resourceapply/apps.go
```diff
@@ -29,13 +29,19 @@ func ApplyDeploymentv1(ctx context.Context, client appsclientv1.DeploymentsGette
 		return nil, false, nil
 	}
 
+	var original appsv1.Deployment
```
jottofar, line 32: If we keep your change below move these two lines into `if` where it's used.

lib/resourceapply/apps.go
petr-muller: TL;DR I'm convinced that the existing diff is incorrect and misleading.

I've tried to explain in the commit message https://github.com/openshift/cluster-version-operator/pull/851/commits/6b6412d64d025b883e88fc58a470b4fe70f2ebaf and in more details on [Slack](https://coreos.slack.com/archives/CJ1J9C3V4/p1666285479676509?thread_ts=1666284447.320419&cid=CJ1J9C3V4) when Trevor asked.

Sorry for not making the change description more prominent. Good point about "everywhere else" - I haven't realized we show the diff even for other resources. Can you check my explanation above and if this fix is correct, I can either make a followup PR to make all diffs like this, or I can extract this change to a separate PR entirely - what do you prefer?

lib/resourceapply/apps.go
petr-muller: Can't do that, we need to make the original deepcopy of `existing` before it is modified on L35 by `resourcemerge.EnsureDeployment`

petr-muller:
/test unit
`TestCVO_UpgradeFailedPayloadLoadWithCapsChanges`

petr-muller:
/retest
openshift-marketplace flake du-jour and `TestCVO_UpgradeFailedPayloadLoadWithCapsChanges`

pkg/start/start.go
```diff
@@ -133,19 +135,27 @@ func (o *Options) Run(ctx context.Context) error {
...
-		// This excludes everything that could possibly depend on a different feature set.
-		startingFeatureSet = ""
-	case err != nil:
-		// client-go automatically retries network blip errors on GETs for 30s by default.  If we fail longer than that
-		// the operator won't be able to do work anyway.  Return the error and crashloop.
-		return err
 
-	default:
-		startingFeatureSet = string(gate.Spec.FeatureSet)
+	if err := wait.PollImmediate(2*time.Second, 30*time.Second, func() (done bool, err error) {
```
LalatenduMohanty, line 139: Use of `wait.PollImmediate` looks good to me. @wking WDYT?

pkg/start/start.go
```diff
@@ -133,19 +135,28 @@ func (o *Options) Run(ctx context.Context) error {
...
+			// if we have no featuregates, then the cluster is using the default featureset, which is "".
+			// This excludes everything that could possibly depend on a different feature set.
+			startingFeatureSet = ""
+			return true, nil
+			// retry some network blips that client-go does not, to survive brief API disruptions during install
+		case net.IsTimeout(err) || net.IsConnectionRefused(err):
+			return false, nil
+		case err != nil:
+			// client-go automatically retries some network blip errors on GETs for 30s by default.  If we fail longer than that
+			// the operator won't be able to do work anyway.  Return the error and crashloop.
```
wking, line 152: This comment doesn't make as much sense now.  Abu pointed out [the client-go retry logic][1] here, and maybe that is giving us 30s of retries under connection resets or probable-EOFs.  But with the wrapping `PollImmediate`, we no longer care about what client-go does, because our poller will retry if client-go fails.  Can we drop this comment?

[1]: https://github.com/kubernetes/kubernetes/blob/93d1c43b491a39740297cf1e76782b99d8739865/staging/src/k8s.io/client-go/rest/request.go#L940-L952

pkg/start/start.go
wking: I'd prefer [`PollImmediateWithContext`][1] so we respect anything canceling `Run`'s `ctx` in the poller (e.g. for the 2s waits), even though we're already passing it through to `Get`.

[1]: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediateWithContext

pkg/start/start.go
```diff
@@ -133,19 +135,28 @@ func (o *Options) Run(ctx context.Context) error {
...
+			return false, err
```
wking, line 153: We might want to log `err`, store in a new `lastError`, and then return `nil` here.  Because [`ConditionFunc`][1] has:

> ConditionFunc returns true if the condition is satisfied, or an error if the loop should be aborted.

and I thought the idea was that until we reach 30s, we want to keep trying.

[1]: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ConditionFunc

pkg/start/start.go
petr-muller: I only made it retry on `net.IsTimeout(err) || net.IsConnectionRefused(err)`, see L148. For any other error we propagate it out which *ends* the polling. Do you thing we should just retry everything?

And if yes, maybe the timeout in `PollImmediate` should be lower, like 25s, so that we either retry short blips ourselves, or we hit the client-go retry cases exactly once.

pkg/start/start.go
petr-muller: addressed in 296ae5f8a950fbda3e348dd131bde3d3a95f92ce

pkg/start/start.go
petr-muller: addressed in 296ae5f8a950fbda3e348dd131bde3d3a95f92ce

pkg/start/start.go
petr-muller: addressed in 296ae5f8a950fbda3e348dd131bde3d3a95f92ce

petr-muller:
@wking @LalatenduMohanty I was seeing some surprising behavior of the timeout in `PollImmediate`. That method does not seem to interact well with `ConditionFunc`s that take a long time to execute (as in our case, where the client-go retry logic can make the GET take 30 seconds to execute. `PollImmediate` as implemented will always execute at least three iterations of `ConditionFunc` even if the timeout expires earlier, check out the testcase here: https://go.dev/play/p/frtOTVNGOHk. This means that our failure cases would cause CVO to shutdown in 90s in worst cases.

I solved this by implementing the timeout with a context instead, happy to hear your thoughts.

petr-muller:
Does not seem related:

```
alert TargetDown fired for 990 seconds with labels: {job="prometheus-k8s", namespace="openshift-monitoring", service="prometheus-k8s", severity="warning"}
```

petr-muller:
`TestCVO_UpgradeFailedPayloadLoadWithCapsChanges`
/test unit

wking (approved):
/lgtm

PR #856, opened 2022-10-21
OCPBUGS-2727: Do not fail precondition check for UnknownUpdate

LalatenduMohanty:
With this change we will not fail precondition checks when a version is not in available updates. So users should be able to upgrade to any version with approriate client side overrides. For example using --allow-explicit-upgrade and --to-imge flags in "oc adm upgrade"

### DIFF
--- a/pkg/payload/precondition/clusterversion/recommendedupdate.go
+++ b/pkg/payload/precondition/clusterversion/recommendedupdate.go
@@ -37,10 +37,11 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 	}
 	if err != nil {
 		return &precondition.Error{
-			Nested:  err,
-			Reason:  "UnknownError",
-			Message: err.Error(),
-			Name:    ru.Name(),
+			Nested:             err,
+			Reason:             "UnknownError",
+			Message:            err.Error(),
+			Name:               ru.Name(),
+			NonBlockingWarning: true,
 		}
 	}
 	for _, recommended := range clusterVersion.Status.AvailableUpdates {
@@ -61,7 +62,8 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 							Reason: condition.Reason,
 							Message: fmt.Sprintf("Update from %s to %s is not recommended:\n\n%s",
 								clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion, condition.Message),
-							Name: ru.Name(),
+							Name:               ru.Name(),
+							NonBlockingWarning: true,
 						}
 					default:
 						return &precondition.Error{
@@ -69,7 +71,8 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 							Message: fmt.Sprintf("Update from %s to %s is %s=%s: %s: %s",
 								clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion,
 								condition.Type, condition.Status, condition.Reason, condition.Message),
-							Name: ru.Name(),
+							Name:               ru.Name(),
+							NonBlockingWarning: true,
 						}
 					}
 				}
@@ -78,7 +81,8 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 				Reason: "UnknownConditionType",
 				Message: fmt.Sprintf("Update from %s to %s has a status.conditionalUpdates entry, but no Recommended condition.",
 					clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion),
-				Name: ru.Name(),
+				Name:               ru.Name(),
+				NonBlockingWarning: true,
 			}
 		}
 	}
@@ -88,13 +92,13 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 			Reason: "NoChannel",
 			Message: fmt.Sprintf("Configured channel is unset, so the recommended status of updating from %s to %s is unknown.",
 				clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion),
-			Name: ru.Name(),
+			Name:               ru.Name(),
+			NonBlockingWarning: true,
 		}
 	}
 
 	reason := "UnknownUpdate"
 	msg := ""
-
 	if retrieved := resourcemerge.FindOperatorStatusCondition(clusterVersion.Status.Conditions, configv1.RetrievedUpdates); retrieved == nil {
 		msg = fmt.Sprintf("No %s, so the recommended status of updating from %s to %s is unknown.", configv1.RetrievedUpdates,
 			clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion)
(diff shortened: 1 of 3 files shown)


pkg/payload/precondition/clusterversion/recommendedupdate.go
```diff
@@ -104,13 +104,15 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
 	} else {
 		msg = fmt.Sprintf("%s=%s (%s), so the recommended status of updating from %s to %s is unknown.", retrieved.Type,
 			retrieved.Status, retrieved.Reason, clusterVersion.Status.Desired.Version, releaseContext.DesiredVersion)
+		warning = true
```
wking, line 107: Should any of these `RecommendedUpdate` cases should be blocking?  Because any change we make seems like it could break somebody.

pkg/payload/precondition/precondition.go
```diff
@@ -71,25 +72,37 @@ func Summarize(errs []error, force bool) (bool, error) {
 		return false, nil
 	}
 	var msgs []string
+	var isWarning bool
 	for _, e := range errs {
 		if pferr, ok := e.(*Error); ok {
 			msgs = append(msgs, fmt.Sprintf("Precondition %q failed because of %q: %v", pferr.Name, pferr.Reason, pferr.Error()))
+			if pferr.NonBlockingWarning {
+				isWarning = true
```
wking, line 80: This logic isn't quite right, because it treats the existence of a single warning error should not make the overall effect warning.  Instead, the existence of a single blocking alert should make the overall effect blocking.  And also, `force` means we won't block regardless, although it may still be worth splitting informative warnings from more serious "would have been blocking except for force" warnings in that case.

LalatenduMohanty: The way the current code is written it ignores the warning if there are multiple precondition check failed. Also if there is one error and NonBlockingWarning is not set to true then it will be blocking.

pkg/payload/precondition/precondition.go
```diff
@@ -12,10 +12,11 @@ import (
...
 type Error struct {
-	Nested  error
-	Reason  string
-	Message string
-	Name    string
+	Nested             error
+	Reason             string
+	Message            string
+	Name               string
+	NonBlockingWarning bool // For some errors we do not want to fail the precondition check but we want to commnicate about it
```
petr-muller, line 19: nit
```suggestion
	NonBlockingWarning bool // For some errors we do not want to fail the precondition check but we want to communicate about it
```

pkg/payload/precondition/precondition.go
LalatenduMohanty: https://github.com/openshift/cluster-version-operator/commit/a1e9171fb415e41ff19f32c5741eb3c5e7a16e7d

pkg/payload/precondition/precondition.go
LalatenduMohanty: https://github.com/openshift/cluster-version-operator/commit/a1e9171fb415e41ff19f32c5741eb3c5e7a16e7d

petr-muller (approved):
/lgtm
/hold

"Unqualified" review, feel free to wait for another reviewer or unhold ;)

LalatenduMohanty:
Squashed the commits in https://github.com/openshift/cluster-version-operator/commit/e6ee0b8a52a47cf5b59e779b246d2f4ac0b8621e

pkg/payload/precondition/clusterversion/recommendedupdate.go
```diff
@@ -88,29 +92,34 @@ func (ru *RecommendedUpdate) Run(ctx context.Context, releaseContext preconditio
...
-			Name: ru.Name(),
+			Name:               ru.Name(),
+			NonBlockingWarning: true,
 		}
 	}
 
 	reason := "UnknownUpdate"
 	msg := ""
-
+	warning := false
```
wking, line 102: no need for this variable, you can just hard-code `NonBlockingWarning: true` belown in the `return` line.

pkg/payload/precondition/precondition.go
```diff
@@ -71,25 +72,40 @@ func Summarize(errs []error, force bool) (bool, error) {
...
+
 	}
 	msg := ""
 	if len(msgs) == 1 {
 		msg = msgs[0]
 	} else {
 		msg = fmt.Sprintf("Multiple precondition checks failed:\n* %s", strings.Join(msgs, "\n* "))
+
+		// when there are multiple precondition checks failed we will ignore if one of them is marked warning
+		isWarning = false
```
wking, line 94: I don't understand what this is about.  If there are two warning messages, won't this code make the summarized condition blocking, when we want to keep it non-blocking?

pkg/payload/precondition/precondition.go
```diff
@@ -71,25 +72,40 @@ func Summarize(errs []error, force bool) (bool, error) {
...
 	}
 
 	if force {
 		msg = fmt.Sprintf("Forced through blocking failures: %s", msg)
+		return !force, &payload.UpdateError{
```
wking, line 99: Previously we'd only get to this log line if we had blocking failures.  Now we can have either a warning or blocking issue.  I think we want something like:

```go
if force and !isWarning {
  msg = fmt.Sprintf("Forced through blocking failures: %s", msg)
  isWarning = true
}

return !isWarning, &payload.UpdateError{
  Nested:  nil,
  Reason:  "UpgradePreconditionCheckFailed",
  Message: msg,
  Name:    "PreconditionCheck",
}
```

pkg/payload/precondition/precondition.go
LalatenduMohanty: Good catch, I forgot to remove it.

pkg/payload/precondition/precondition.go
LalatenduMohanty: This works too. Somehow I thought we want to keep the code around force separate than the warning, just for readability

pkg/payload/precondition/precondition.go
LalatenduMohanty: Resolved in https://github.com/openshift/cluster-version-operator/commit/481bcde393f244f4e4245ebec6a43941c87626de

pkg/payload/precondition/clusterversion/recommendedupdate.go
LalatenduMohanty: Resolved in https://github.com/openshift/cluster-version-operator/commit/481bcde393f244f4e4245ebec6a43941c87626de

pkg/payload/precondition/precondition.go
```diff
@@ -12,10 +12,11 @@ import (
...
 type Error struct {
-	Nested  error
-	Reason  string
-	Message string
-	Name    string
+	Nested             error
+	Reason             string
+	Message            string
+	Name               string
+	NonBlockingWarning bool // For some errors we do not want to fail the precondition check but we want to communicate about it
```
wking, line 19: nit, this could be a godoc, instead of a non-structured comment, but :shrug:

wking (approved):
/lgtm

wking:
[`KubePodNotReady` for `prometheus-k8s-1`][1] is orthogonal.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/856/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic/1586006561224921088

PR #857, opened 2022-10-26
pkg/cvo/cvo_scenarios_test.go: Remove status check from TestCVO_UpgradeFailedPayloadLoadWithCapsChanges

jottofar:
This unnecessary final check was originally added to utilize the newly added expectFinalUpdateStatus function to expose it for future use when test cases are modified or added. Multiple statuses are set making hitting the exact one difficult due to timing which can cause failures. TestCVO_InitImplicitlyEnabledCaps was modified to now use expectFinalUpdateStatus.


wking (approved):
/lgtm

PR #858, opened 2022-10-27
resourcemerge: cleanup deprecated function calls

petr-muller:
These warnings are annoying for anyone new (like me) who opens the codebase in GoLand, and prevent new code from being efficiently checked while writing.

- `pointer.BoolPtr` -> `pointer.Bool`
- `pointer.Int*Ptr` -> `pointer.Int*`
- `diff.ObjectReflectDiff` -> `cmp.Diff`

### DIFF
--- a/lib/resourcemerge/apps_test.go
+++ b/lib/resourcemerge/apps_test.go
@@ -107,7 +107,7 @@ func TestEnsureDeployment(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			defaultDeployment(&test.existing, test.existing)
 			defaultDeployment(&test.expected, test.expected)
-			modified := pointer.BoolPtr(false)
+			modified := pointer.Bool(false)
 			EnsureDeployment(modified, &test.existing, test.required)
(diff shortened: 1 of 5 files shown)


LalatenduMohanty:
We use golangci-lint as part of our pre-submits https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-master.yaml#L55 , checking why golanci-lint did not catch these issues.

LalatenduMohanty:
We are running the golangci-lint with default linters which are not catching this, checking if we need some specific linter combination which will catch these for future.

LalatenduMohanty:
Tried few presets e.g. `$ golangci-lint run --verbose --timeout=5m0s --presets=bugs` to include large set of linters but did not find anything that complains about `BoolPtr`.

LalatenduMohanty:
Also BoolPtr is deprecated https://github.com/kubernetes/utils/blob/cfd413dd9e85/pointer/pointer.go#L236

LalatenduMohanty (approved):
/lgtm

PR #865, opened 2022-11-11
pkg/cvo/updatepayload: Restructure desired-image lookup

wking:
The previous logic ignored `version` if `image` was specified.  This allowed admins to request incompatible `(image, version)` tuples, and the cluster-version operator would silently accept their 'image' and not warn the admin that the actual version diverged from their request.

When `desiredUpdates` has any opinions, the new logic shuffles to create a set of targets, including:

* The current release.
* Everything in `availableUpdates`.
* If `desiredUpdate.image` is not empty, and also not a digest-match for the current release or any `availableUpdates`, also a new entry with information from the `desiredUpdate` request.

That set of targets is filtered to find any that match all populated `desiredUpdate` properties.  So if only `desiredUpdate.image` is set, we look for any targets that match that image digest.  If only `version` is set, we look for any targets that match that version.  If both `image` and `version` are set, we look for any targets that match both that image and that image digest.

In the event that we find no matching targets, or that we find multiple matching targets, we complain, and stick with our current release.

For now, to preserve backwards compatibility with the previous implementation, we warn, but still accept the `image` target, if an incompatible `(image, version)` tuple is requested.


wking:
WIP because I still need to plumb the warning through to something we can measure in Telemetry.

pkg/cvo/cvo.go
```diff
@@ -591,18 +591,12 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
...
-		currentVersion := optr.currentVersion()
-		desired = configv1.Update{
-			Version: currentVersion.Version,
-			Image:   currentVersion.Image,
-		}
-		klog.V(2).Infof("Desired version from operator is %#v", desired)
-	}
+	desired, findUpdateWarning := findDesiredUpdate(config, optr.currentVersion())
+	if findUpdateWarning != nil {
+		klog.Warningf("Deprecated desired update matching: %v", findUpdateWarning)
```
petr-muller, line 596: The message here is confusing to me - I'm not sure if I understood what is wrong if I saw that in the log. What is deprecated? Is the problem that something is deprecated, or is the problem that something is matching... what?

Also - is there a good reason why not to simply call the second argument `err` (or `findUpdateErr` or whatever)? It's an `error`, and because of the convention is so prevalent, naming it otherwise indicates it's not an error. If I only read callsite code I'd probably think it's a string.

Also 2) It's also a convention that callers should not rely on actual return values if the method returned a non-nil error. I think that the intent that it is fine in this case should be documented (probably on both sides, but at least in `findDesiredUpdate` godoc - if the contract is "no matter the error, this method always retuns something sane in its first return value", then IMO it should be made explicit.

pkg/cvo/updatepayload.go
```diff
@@ -314,28 +314,102 @@ func (r *payloadRetriever) pruneJobs(ctx context.Context, retain int) error {
...
+// image and an incompatible version.
+func releaseMatchesUpdate(release *configv1.Release, update *configv1.Update) (error, error) {
+	if release == nil || release.Image == "" {
+		return nil, fmt.Errorf("releases without images never match: %#v", release)
+	}
+
+	var mismatchedVersionError error
+	if update.Version != "" && release.Version != update.Version {
+		mismatchedVersionError := fmt.Errorf("release version %q does not match desired update %q", release.Version, update.Version)
+		if mismatchedVersionError != nil && update.Image == "" { // TODO: eventually drop the image-check here, if we can demonstrate that incompatible (image, version) pairs are uncommon in the wild
```
petr-muller, line 405: `mismatchedVersionError != nil` check is weird here, it's never nil, it's always set on the line just above?

pkg/cvo/cvo.go
wking: The full message would look like:

    Deprecated desired update matching: release version "4.13.0" does not match desired update "4.13.3"

for someone who set:

```yaml
desiredUpdate:
  image: ${PULLSPEC_FOR_4_13_3}
  version: 4.13.0
```

> I think that the intent that it is fine in this case should be documented...

Done with 96dfce7c -> 6c2d7e79

pkg/cvo/updatepayload.go
wking: Dropped in 6c2d7e79 -> 1618a424

PR #869, opened 2022-11-30
OTA-560: Modernize README.md to be user-centered

petr-muller:
CVO-specific user-targeting content will always partially overlap with other
documentation, like OCP docs about upgrades, the [`dev-guide`](https://github.com/openshift/enhancements/tree/master/dev-guide/cluster-version-operator) in the openshift/enhancement
repository and perhaps other locations. Ideally we want avoid duplicating other
content here as much as possible: only cover the basics or really CVO-specific
content in this repository, and link elsewhere when needed.

The goals of the README are the following:
1. Inform a technical, knowledgeable user about how CVO operates
2. Inform a technical, knowledgeable user about how to interact with CVO in a cluster
3. Redirect an incoming developer to another suitable location for information
   about development, testing and contributing
4. Present CVO repository as well-maintained and visitor-friendly

This PR only brings the user-facing content. The developer-facing content is moved/updated in a separate https://github.com/openshift/cluster-version-operator/pull/871

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,70 +1,124 @@
-# Cluster Version Operator (CVO)
+# Cluster Version Operator
 
-## Building and Publishing CVO
+The Cluster Version Operator (CVO) is one of the [Cluster Operators][dev-guide-operators] that run in every OpenShift
+cluster. CVO consumes an artifact called a _"release payload image,"_ which represents a specific version of OpenShift.
+The release payload image contains the [resource manifests][kube-glossary-manifest] necessary for the cluster to
+function, like all Cluster Operator ones. CVO reconciles the resources within the cluster to match the manifests in the
+release payload image. As a result, CVO implements _cluster upgrades_. After being provided a release payload image for
+a newer OpenShift version, CVO reconciles all Cluster Operators to their updated versions, and Cluster Operators
+similarly update their operands.
 
-```sh
-./hack/build-image.sh && REPO=<your personal repo (quay.io/ahbinavdahiya | docker.io/abhinavdahiya)> ./hack/push-image.sh
+## OpenShift Upgrades
+
+For information about upgrading OpenShift clusters, please see the respective documentation:
+
+- [OKD Documentation: Updating clusters][okd-updating-clusters]
+- [Red Hat OpenShift Container Platform: Updating clusters][ocp-updating-clusters]
+
+## `ClusterVersion` Resource
+
+Like other Cluster Operators, the Cluster Version Operator is configured by a Config API resource in the cluster:
+a [`ClusterVersion`][ocp-clusterversion]:
+
+```console
+$ oc explain clusterversion
+  KIND:     ClusterVersion
+  VERSION:  config.openshift.io/v1
+
+  DESCRIPTION:
+       ClusterVersion is the configuration for the ClusterVersionOperator. This is
+       where parameters related to automatic updates can be set. Compatibility
+       level 1: Stable within a major release for a minimum of 12 months or 3
+       minor releases (whichever is longer).
+
+  FIELDS:
+     ...
+     spec	<Object> -required-
+       spec is the desired state of the cluster version - the operator will work
+       to ensure that the desired version is applied to the cluster.
+
+     status	<Object>
+       status contains information about the available updates and any in-progress
+       updates.
 ```
 
-1. This builds image locally and then pushes `${VERSION}` and `latest` tags to `${REPO}/origin-cluster-version-operator`.
+`ClusterVersion` resource follows the established [Kubernetes pattern][kube-spec-and-status] where the `spec`
+property describes the desired state that CVO should achieve and maintain, and the `status` property is populated by the
+CVO to describe its status and the observed state.
 
-2. `${VERSION}` encodes the Git commit used to build the images.
+In a typical OpenShift cluster, there will be a cluster-scoped `ClusterVersion` resource called `version`:
 
-## Building release image using local CVO
-
-1. Make sure you have `oc` binary from https://github.com/openshift/origin master as it requires `adm release` subcommand.
-
-2. Run the following command to create release-image at `docker.io/abhinavdahiya/origin-release:latest`:
-
-```sh
-oc adm release new -n origin \
-    --from-image-stream=origin-v4.0 \
-    --to-image-base=docker.io/abhinavdahiya/origin-cluster-version-operator:latest \
-    --to-image docker.io/abhinavdahiya/origin-release:latest
+```console
+$ oc get clusterversion version
+NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
+version   4.11.17   True        False         6d5h    Cluster version is 4.11.17
 ```
 
-## Installing CVO and operators in cluster.
+Note that as a user or a cluster administrator, you usually do not interact with the `ClusterVersion` resource directly
+but via either the [`oc adm upgrade`][ocp-oc-adm-upgrade] CLI or the [web console][ocp-webconsole-upgrades].
 
-1. Use CVO `render` to render all the manifests from release-payload to a directory. Refer the [Using CVO to render the release-image locally](#using-cvo-to-render-the-release-image-locally) section for more information.
+## Understanding Upgrades
 
-2. Create the operators from the manifests by using `oc create -f <directory when CVO rendered manifests>`.
+> :bulb: This section is only a high-level overview. See the [Update Process][dev-guide-upgrade-workflow] and
+> [Reconciliation][dev-guide-reconciliation] documents in the [dev-guide][dev-guide] for more details.
 
-## Using CVO to render the release-image locally
+The Cluster Version Operator continuously fetches information about upgrade paths for the configured channel from the
+OpenShift Update Service (OSUS). It stores the recommended update options in the `status.availableUpdates` field of
+its `ClusterVersion` resource.
 
-1. Run the following command to get render the release-payload contents to `/tmp/cvo`
+The intent to upgrade the cluster to another version is expressed by storing the desired version in
+the `spec.desiredUpdate` field. When `spec.desiredUpdate` does not match the current cluster version, CVO will start
+updating the cluster. It downloads the release payload image, validates it,
+and [systematically reconciles][dev-guide-reconciliation] the Cluster Operator resources to match the updated manifests
+delivered in the release payload image.
 
-```sh
-podman run --rm -ti \
-    -v /tmp/cvo:/tmp/cvo:z \
-    <release image> \
-        render \
-        --output-dir=/tmp/cvo \
-        --release-image="<release image>"
+## Troubleshooting
+
+A typical OpenShift cluster will have a `Deployment` resource called `cluster-version-operator` in
+the `openshift-cluster-version` namespace, configured to run a single CVO replica. Confirm that its `Pod` is up and
+optionally inspect its log:
+
+```console
+$ oc get deployment -n openshift-cluster-version cluster-version-operator
+NAME                       READY   UP-TO-DATE   AVAILABLE   AGE
+cluster-version-operator   1/1     1            1           2y227d
+
+$ oc get pods -n openshift-cluster-version -l k8s-app=cluster-version-operator
+NAME                                       READY   STATUS    RESTARTS   AGE
+cluster-version-operator-6885cc574-674n6   1/1     Running   0          6d5h
+
+$ oc logs -n openshift-cluster-version -l k8s-app=cluster-version-operator
+...
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,70 +1,137 @@
...
-## Building and Publishing CVO
-
-```sh
-./hack/build-image.sh && REPO=<your personal repo (quay.io/ahbinavdahiya | docker.io/abhinavdahiya)> ./hack/push-image.sh
+# Cluster Version Operator
+
+The Cluster Version Operator (CVO) is one of the [Cluster Operators](https://github.com/openshift/enhancements/blob/master/dev-guide/operators.md#what-is-an-openshift-clusteroperator)
+that run in every OpenShift cluster. CVO consumes an artifact called a _"release
+payload image,"_ which represents a given version of OpenShift and contains the
+manifests needed for all Cluster Operators to function. CVO reconciles
```
wking, line 6: nit: we're introducing "manifests" here, and it might be worth adding a bit more context like "resource manifests" or a parenthetical explaining that manifests are requesting particular configuration for particular Kubernetes resources or something like that.

README.md
```diff
@@ -1,70 +1,137 @@
...
+the resources in the cluster to match the manifests in the release payload
+image. This way, CVO also implements _cluster upgrades_; after being provided
+a release payload image for a newer OpenShift version, CVO reconciles all
+Cluster Operators to their updated versions, and in turn, the Cluster Operators
```
wking, line 10: nit: "cluster operators" are our main operands, but we can do other things too, and the CVO is not all that opinionated about what we're managing.  We might want to soften this part a bit like:

> ... the CVO reconciles the incoming manifests, including those for Cluster Operators, bringing the associated Kubernetes resources up to the configuration requested in the incoming release payload.

or some such.

README.md
```diff
@@ -1,70 +1,137 @@
...
+       updates.
 ```
 
-1. This builds image locally and then pushes `${VERSION}` and `latest` tags to `${REPO}/origin-cluster-version-operator`.
-
-2. `${VERSION}` encodes the Git commit used to build the images.
-
-## Building release image using local CVO
+`ClusterVersion` resource follows the established [Kubernetes pattern](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#object-spec-and-status) where
+the `spec` sub-resource describes the desired state that CVO should achieve and
```
wking, line 49: nit: [only `status` is a strict subresource][1].  Can we use "property" or something instead, to avoid having an opinion on subresource-ness?

[1]: https://github.com/openshift/api/blob/94ac871f4a586d03b0ae11344f3604016065a458/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml#L419-L420

README.md
```diff
@@ -1,70 +1,137 @@
...
-```sh
-oc adm release new -n origin \
-    --from-image-stream=origin-v4.0 \
-    --to-image-base=docker.io/abhinavdahiya/origin-cluster-version-operator:latest \
-    --to-image docker.io/abhinavdahiya/origin-release:latest
+```console
+$ oc get clusterversion version
+NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
+version   4.11.3    True        False         85d     Cluster version is 4.11.3
 ```
```
wking, line 60: Around here it is probably worth pointing out that `oc adm upgrade` is a nicer way to get pretty-printed ClusterVersion information, without `get`'s default output format's attempt to pack all of the information into a single line.

README.md
```diff
@@ -1,70 +1,137 @@
...
+The Cluster Version Operator continuously fetches information about upgrade
+paths for the configured channel from the OpenShift Update Service (OSUS). It
+stores the recommended update options in the `status.availableUpdates` field of
+its `ClusterVersion` resource.
 
-1. Run the following command to get render the release-payload contents to `/tmp/cvo`
+The intent to upgrade the cluster to another version is expressed by storing the
+desired version in the `spec.desiredUpdate` field, either manually by the
+administrator through the CLI or web console, or CVO does it automatically when
+automated upgrades are enabled.
```
wking, line 78: Probably drop the "automated upgrades" mention for now, because we don't expose any user-facing knobs for that (openshift/enhancements#124 floated a knob, but was deferred).  And when automatic updates are enabled, there will be a knob to ask the CVO to perform them, but the CVO won't implement it by touching `spec.desiredUpdate` (because the CVO and clusters admins competing for the same `spec` properties would be racy, so the CVO's ClusterVersion `spec` access is read-only).

README.md
```diff
@@ -1,70 +1,137 @@
...
-```sh
-podman run --rm -ti \
-    -v /tmp/cvo:/tmp/cvo:z \
-    <release image> \
-        render \
-        --output-dir=/tmp/cvo \
-        --release-image="<release image>"
-```
+When `spec.desiredUpdate` does not match the current cluster version, CVO will
+start updating the cluster. It downloads the release payload image, validates
```
wking, line 81: nit: I'm fine with this simplification, but when `spec.desiredUpdate` does not match the current cluster version, the CVO will consider updating the cluster.  We might refuse to update (e.g. if we cannot find a signature for the release image), although if  `force: true`, that waives all of our target-acceptance guards.

README.md
```diff
@@ -1,70 +1,137 @@
...
+$ oc get pods -n openshift-cluster-version -l k8s-app=cluster-version-operator
+NAME                                        READY   STATUS    RESTARTS   AGE
+cluster-version-operator-7fdf58f594-ct92x   1/1     Running   1          85d
+
+$ oc logs -n openshift-cluster-version -l k8s-app=cluster-version-operator
+...
 ```
 
-## Contributing
+The CVO follows the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/4c9ef2d/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) and sets `Conditions` in the status of its `ClusterVersion` resource:
```
wking, line 106: nit: probably spell out the full commit hash (4c9ef2d135294355e7ca33ec7a5e01d31438df12) in the URI.  That reduces the risk of a truncated-hash collision as additional commits land upstream, and the user doesn't have to read this URI anyway.

wking: also, possibly drop this section entirely, because both `oc adm upgrade` and the web-console (which you've already plugged) will handle rendering conditions for the user.

README.md
```diff
@@ -1,70 +1,137 @@
...
+    "type": "Progressing"
+  },
+  ...
+```
 
-[Contributions welcome](CONTRIBUTING.md)!
+### Troubleshooting Cluster Upgrades
 
-## Documentation
+TODO: Are there good links for this? Inspecting update pod? Inspecting `ClusterOperator` resources?
```
wking, line 133: Hopefully our existing condition strings are already actionable enough.  If they aren't, we should improve those strings, or the context `oc` and the web-console add to those strings, instead of trying to paper over any gaps in our GitHub README.

README.md
```diff
@@ -1,70 +1,137 @@
...
-```sh
-./hack/build-image.sh && REPO=<your personal repo (quay.io/ahbinavdahiya | docker.io/abhinavdahiya)> ./hack/push-image.sh
+# Cluster Version Operator
+
+The Cluster Version Operator (CVO) is one of the [Cluster Operators](https://github.com/openshift/enhancements/blob/master/dev-guide/operators.md#what-is-an-openshift-clusteroperator)
+that run in every OpenShift cluster. CVO consumes an artifact called a _"release
+payload image,"_ which represents a given version of OpenShift and contains the
+manifests needed for all Cluster Operators to function. CVO reconciles
+the resources in the cluster to match the manifests in the release payload
+image. This way, CVO also implements _cluster upgrades_; after being provided
```
LalatenduMohanty, line 8: s/implements/initiates the/

obrown1205, line 8: The sentence "This way, CVO also implements..." is too long IMO. I suggest rephrasing a bit :) 

Maybe: 
The Cluster Version Operator (CVO) is one of the [Cluster Operators](https://github.com/openshift/enhancements/blob/master/dev-guide/operators.md#what-is-an-openshift-clusteroperator) that run in every OpenShift cluster. CVO consumes an artifact called a "release payload image," which represents a given version of OpenShift. It also contains the manifests needed for all Cluster Operators to function, reconciling the resources within the cluster to match these manifests in the resource payload image. 
As a result, CVO implements cluster upgrades. After being provided a release payload image for a newer OpenShift version, CVO reconciles all Cluster Operators to their updated versions and the Cluster Operators similarly update their operands.

README.md
petr-muller: Cool. I have softened the paragraph in bf6cc77a13f199601f2f472ce21d6aa2af8d9c1c, also taking into account @obrown1205 's comment below. I think it's generally fine to omit nuances and focus on the main purpose in a README, but I agree I sometimes handwave things too much ;)

README.md
petr-muller: addressed in bf6cc77a13f199601f2f472ce21d6aa2af8d9c1c

README.md
petr-muller: I think it's fine to focus on the happy path. I mention the validation in the next sentence which IMO implies that it can fail.

README.md
petr-muller: yes and yes, done in bf6cc77a13f199601f2f472ce21d6aa2af8d9c1c

README.md
petr-muller: Yeah, I think this is "troubleshooting upgrades", not "troubleshooting" CVO, so I'd like this section to be super short and ideally just lead elsewhere :+1:

README.md
petr-muller: Thanks! I tried to apply your rephrasing in the current version while merging it with Trevor's suggestion above for the same paragraph.

README.md
```diff
@@ -1,70 +1,127 @@
...
+
+- [OKD Documentation: Updating clusters][okd-updating-clusters]
+- [Red Hat OpenShift Container Platform: Updating clusters][ocp-updating-clusters]
 
-1. This builds image locally and then pushes `${VERSION}` and `latest` tags to `${REPO}/origin-cluster-version-operator`.
+## `ClusterVersion` Resource
 
-2. `${VERSION}` encodes the Git commit used to build the images.
+Like other Cluster Operators, the Cluster Version Operator is configured by a Config API resource in the cluster:
+a [`ClusterVersion`][[ocp-clusterversion]]:
```
wking, line 21: nit: looks like doubled brackets here?  I'd have expected:

    [`ClusterVersion`][ocp-clusterversion]

README.md
```diff
@@ -1,70 +1,127 @@
...
+> [Reconciliation][dev-guide-reconciliation] documents in the [dev-guide][dev-guide] for more details.
 
-`<release image>` can be personal release image generated using [this](#building-release-image-using-local-cvo) or Origin's release image like `registry.ci.openshift.org/openshift/origin-release:v4.0`.
+The Cluster Version Operator continuously fetches information about upgrade paths for the configured channel from the
+OpenShift Update Service (OSUS). It stores the recommended update options in the `status.availableUpdates` field of
+its `ClusterVersion` resource.
 
-## Running CVO tests
+The intent to upgrade the cluster to another version is expressed by storing the desired version in
+the `spec. desiredUpdate` field. When `spec.desiredUpdate` does not match the current cluster version, CVO will start
```
wking, line 70: nit: `spec. desiredUpdate` -> `spec.desiredUpdate` (removing a space)

README.md
```diff
@@ -1,70 +1,127 @@
...
+updating the cluster. It downloads the release payload image, validates it, and [systematically reconciles]
+[dev-guide-reconciliation] the Cluster Operator resources to match the updated manifests delivered in the release
```
wking, line 72: nit: GitHub doesn't like breaking the text and anchor-ref across a newline:

image

README.md
```diff
@@ -1,70 +1,127 @@
...
+NAME                                        READY   STATUS    RESTARTS   AGE
+cluster-version-operator-7fdf58f594-ct92x   1/1     Running   1          85d
+
+$ oc logs -n openshift-cluster-version -l k8s-app=cluster-version-operator
+...
 ```
 
-## Contributing
+The CVO follows the [Kubernetes API conventions][kube-api-conventions-status] and sets `Conditions` in the status of
+its `ClusterVersion` resource. These conditions are surfaced by both the OpenShift web console and the `oc adm-cli`.
```
wking, line 95: nit: `oc adm-cli` reads like it's suggesting a specific invocation, when I think you mean either `oc adm upgrade` or "the `oc adm ...` CLI" or some such.

petr-muller:
4d4d4b4aed8f5ce9487049fb12e14d3a5b7fa483 contains all fixes for the formatting artifacts pointed out by @wking 

I have moved this PR from WIP, @LalatenduMohanty pointed out that it's better to submit the "move developer-oriented content to new location" part of OTA-560 as a separate PR.

wking (approved):
/lgtm

PR #870, opened 2022-11-30
OCPBUGS-702: Fix removing `caBundle` field of CRDs when `...inject-cabundle=true`

Davoska:
This pull request will fix CVO's behavior of removing the `caBundle` field of CRDs when reconciling manifests.

This pull request references the bug OCPBUGS-702 [1].

---

*   lib/resourcemerge: Fix removing `caBundle` field of CRDs when `...inject-cabundle=true`

Currently, the users can omit the `caBundle` field in the CRDs and set the `service.beta.openshift.io/inject-cabundle=true` annotation so that the given `caBundle` field is managed by the `service-ca` controller (this functionality is not limited only to CRDs) [2].

However, the CVO is not aware of this feature. CVO, when reconciling such CRD, sees an empty `caBundle` field in the given release manifest, and CVO then removes any value in the existing applied CRD so it matches the release manifest semantically.

This commit makes the CVO look into the existing resource and check for the `service.beta.openshift.io/inject-cabundle=true` annotation. If it's set then the `service-ca` controller is responsible for the field, and CVO will ignore any differences between the existing `caBundle` field and the `caBundle` field in the given release manifest of the CRD.

This bug also caused the CVO to hotloop on CRDs as mentioned in the bug [3].

---

*   lib/resourceapply/apiext.go: Improve diff logging for CRDs

This pull request also improves the logging of handling the CRDs. Previously, the logged `diff` wasn't helpful and it didn't show the reason why the update of the resource happened. Now, CVO logs the different fields that caused the update. This change is similar to the change in the commit [4].

---

[1] https://issues.redhat.com/browse/OCPBUGS-702
[2] https://docs.openshift.com/container-platform/4.6/security/certificates/service-serving-certificate.html
[3] https://bugzilla.redhat.com/show_bug.cgi?id=2089097
[4] https://github.com/openshift/cluster-version-operator/commit/c58385eea897b5af790e4dd36103e978e7e4c597

### DIFF
--- a/lib/resourcemerge/apiext.go
+++ b/lib/resourcemerge/apiext.go
@@ -12,18 +12,45 @@ import (
 // modified is set to true when existing had to be updated with required.
 func EnsureCustomResourceDefinitionV1(modified *bool, existing *apiextv1.CustomResourceDefinition, required apiextv1.CustomResourceDefinition) {
 	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	ensureCustomResourceDefinitionV1Defaults(&required)
+	ensureCustomResourceDefinitionV1CaBundle(&required, *existing)
+	if !equality.Semantic.DeepEqual(existing.Spec, required.Spec) {
+		*modified = true
+		existing.Spec = required.Spec
+	}
+}
 
-	// apply defaults to blue print
+func ensureCustomResourceDefinitionV1Defaults(required *apiextv1.CustomResourceDefinition) {
 	if len(required.Spec.Names.Singular) == 0 {
 		required.Spec.Names.Singular = strings.ToLower(required.Spec.Names.Kind)
 	}
 	if len(required.Spec.Names.ListKind) == 0 {
 		required.Spec.Names.ListKind = required.Spec.Names.Kind + "List"
 	}
+}
 
-	// we stomp everything
-	if !equality.Semantic.DeepEqual(existing.Spec, required.Spec) {
-		*modified = true
-		existing.Spec = required.Spec
+// ensureCustomResourceDefinitionV1CaBundle ensures that the field
+// spec.Conversion.Webhook.ClientConfig.CABundle of a CRD is not managed by the CVO when
+// the service-ca controller is responsible for the field.
+func ensureCustomResourceDefinitionV1CaBundle(required *apiextv1.CustomResourceDefinition, existing apiextv1.CustomResourceDefinition) {
+	if val, ok := existing.ObjectMeta.Annotations[injectCABundleAnnotation]; !ok || val != "true" {
+		return
+	}
+	req := required.Spec.Conversion
+	if req == nil ||
+		req.Webhook == nil ||
+		req.Webhook.ClientConfig == nil {
+		return
+	}
+	if req.Strategy != apiextv1.WebhookConverter {
+		// The service CA bundle is only injected by the service-ca controller into
+		// the CRD if the CRD is configured to use a webhook for conversion
+		return
+	}
+	exc := existing.Spec.Conversion
+	if exc != nil &&
+		exc.Webhook != nil &&
+		exc.Webhook.ClientConfig != nil {
+		req.Webhook.ClientConfig.CABundle = exc.Webhook.ClientConfig.CABundle
 	}
 }
--- a/lib/resourcemerge/apiext_test.go
+++ b/lib/resourcemerge/apiext_test.go
@@ -1,0 +1,191 @@
+package resourcemerge
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+func TestEnsureCustomResourceDefinitionV1(t *testing.T) {
+	tests := []struct {
+		name     string
+		existing apiextv1.CustomResourceDefinition
+		required apiextv1.CustomResourceDefinition
+
+		expectedModified bool
+		expected         apiextv1.CustomResourceDefinition
+	}{
+		{
+			name: "respect injected caBundle when the annotation `...inject-cabundle=true` is set",
+			existing: apiextv1.CustomResourceDefinition{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						injectCABundleAnnotation: "true",
+					},
+				},
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: []byte("CA bundle added by the ca operator"),
+							},
+						},
+					},
+				},
+			},
+			required: apiextv1.CustomResourceDefinition{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						injectCABundleAnnotation: "true",
+					},
+				},
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: nil,
+							},
+						},
+					},
+				},
+			},
+
+			expectedModified: false,
+			expected: apiextv1.CustomResourceDefinition{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						injectCABundleAnnotation: "true",
+					},
+				},
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: []byte("CA bundle added by the ca operator"),
+							},
+						},
+					},
+				}},
+		},
+		{
+			name: "respect injected caBundle when the annotation `...inject-cabundle=true` is set by the user",
+			existing: apiextv1.CustomResourceDefinition{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						injectCABundleAnnotation: "true",
+					},
+				},
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: []byte("CA bundle added by the ca operator"),
+							},
+						},
+					},
+				},
+			},
+			required: apiextv1.CustomResourceDefinition{
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: nil,
+							},
+						},
+					},
+				},
+			},
+
+			expectedModified: false,
+			expected: apiextv1.CustomResourceDefinition{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						injectCABundleAnnotation: "true",
+					},
+				},
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: []byte("CA bundle added by the ca operator"),
+							},
+						},
+					},
+				}},
+		},
+		{
+			name: "remove injected caBundle when the annotation `...inject-cabundle=true` is not set",
+			existing: apiextv1.CustomResourceDefinition{
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: []byte("CA bundle added by the user"),
+							},
+						},
+					},
+				},
+			},
+			required: apiextv1.CustomResourceDefinition{
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: nil,
+							},
+						},
+					},
+				},
+			},
+
+			expectedModified: true,
+			expected: apiextv1.CustomResourceDefinition{
+				Spec: apiextv1.CustomResourceDefinitionSpec{
+					Conversion: &apiextv1.CustomResourceConversion{
+						Strategy: apiextv1.WebhookConverter,
+						Webhook: &apiextv1.WebhookConversion{
+							ClientConfig: &apiextv1.WebhookClientConfig{
+								CABundle: nil,
+							},
+						},
+					},
+				}},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			defaultCustomResourceDefinitionV1(&test.existing, test.existing)
+			defaultCustomResourceDefinitionV1(&test.expected, test.expected)
+			modified := pointer.Bool(false)
+			EnsureCustomResourceDefinitionV1(modified, &test.existing, test.required)
+			if *modified != test.expectedModified {
+				t.Errorf("mismatch modified got: %v want: %v", *modified, test.expectedModified)
+			}
+
+			if !equality.Semantic.DeepEqual(test.existing, test.expected) {
+				t.Errorf("unexpected: %s", cmp.Diff(test.expected, test.existing))
+			}
+		})
+	}
+}
+
+// Ensures the structure contains any defaults not explicitly set by the test
+func defaultCustomResourceDefinitionV1(in *apiextv1.CustomResourceDefinition, from apiextv1.CustomResourceDefinition) {
+	modified := pointer.Bool(false)
+	EnsureCustomResourceDefinitionV1(modified, in, from)
+}
--- a/lib/resourceapply/apiext.go
+++ b/lib/resourceapply/apiext.go
@@ -28,14 +28,20 @@ func ApplyCustomResourceDefinitionv1(ctx context.Context, client apiextclientv1.
 		return nil, false, nil
 	}
 
-	modified := pointer.BoolPtr(false)
+	var original apiextv1.CustomResourceDefinition
+	existing.DeepCopyInto(&original)
+
+	modified := pointer.Bool(false)
 	resourcemerge.EnsureCustomResourceDefinitionV1(modified, existing, *required)
 	if !*modified {
 		return existing, false, nil
 	}
-
 	if reconciling {
-		klog.V(2).Infof("Updating CRD %s due to diff: %v", required.Name, cmp.Diff(existing, required))
+		if diff := cmp.Diff(&original, existing); diff != "" {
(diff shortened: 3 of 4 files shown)


lib/resourcemerge/apiext_test.go
```diff
@@ -0,0 +1,190 @@
...
+			defaultCustomResourceDefinitionV1(&test.existing, test.existing)
+			defaultCustomResourceDefinitionV1(&test.expected, test.expected)
+			modified := pointer.Bool(false)
+			EnsureCustomResourceDefinitionV1(modified, &test.existing, test.required)
+			if *modified != test.expectedModified {
+				t.Errorf("mismatch modified got: %v want: %v", *modified, test.expectedModified)
+			}
+
+			if !equality.Semantic.DeepEqual(test.existing, test.expected) {
+				t.Errorf("mismatch CRD got: %v want: %v", test.existing, test.expected)
```
petr-muller, line 180: can we log `cmp.Diff(expected, actual)` so the test output shows the difference on failure?

lib/resourcemerge/apiext.go
```diff
@@ -12,18 +12,36 @@ import (
...
 
-	// we stomp everything
-	if !equality.Semantic.DeepEqual(existing.Spec, required.Spec) {
-		*modified = true
-		existing.Spec = required.Spec
+// ensureCustomResourceDefinitionV1CaBundle ensures that the field
+// .Spec.Conversion.Webhook.ClientConfig.CABundle is not managed by the CVO when
+// "service.beta.openshift.io/inject-cabundle=true" annotation is set in existing.
+func ensureCustomResourceDefinitionV1CaBundle(required *apiextv1.CustomResourceDefinition, existing apiextv1.CustomResourceDefinition) {
+	if val, ok := existing.ObjectMeta.Annotations["service.beta.openshift.io/inject-cabundle"]; !ok || val != "true" {
```
petr-muller, line 36: I think we should const the `"service.beta.openshift.io/inject-cabundle"` instead of inlining it and use the const also in tests etc.

petr-muller (commented):
LGTM in general, minor feedback inline

lib/resourcemerge/apiext_test.go
Davoska: Most definitely! We should maybe think about unifying the output of tests where it's appropriate.

Currently, some of the tests have the output for `!equality.Semantic.DeepEqual(test.existing, test.expected)` as:

```go
if !equality.Semantic.DeepEqual(test.existing, test.expected) {
	t.Errorf("mismatch <KIND> got: %v want: %v", test.existing, test.expected)
}
```

and some as:

```go
if !equality.Semantic.DeepEqual(test.existing, test.expected) {
	t.Errorf("unexpected: %s", cmp.Diff(test.expected, test.existing))
}
```

But that's just a minor thought.

lib/resourcemerge/apiext.go
Davoska: I agree, thanks for catching this!

I have created a new file `lib/resourcemerge/helper.go` where I have put the constant. I couldn't think of a more appropriate existing file or a better name for the file (maybe `lib/resourcemerge/constants.go`). If anybody has a better idea, please let me know.

lib/resourcemerge/apiext.go
```diff
@@ -12,18 +12,36 @@ import (
...
-		*modified = true
-		existing.Spec = required.Spec
+// ensureCustomResourceDefinitionV1CaBundle ensures that the field
+// .Spec.Conversion.Webhook.ClientConfig.CABundle is not managed by the CVO when
+// the service-ca controller is responsible for the field.
+func ensureCustomResourceDefinitionV1CaBundle(required *apiextv1.CustomResourceDefinition, existing apiextv1.CustomResourceDefinition) {
+	if val, ok := existing.ObjectMeta.Annotations[InjectCABundleAnnotation]; !ok || val != "true" {
+		return
+	}
+	if required.Spec.Conversion != nil && required.Spec.Conversion.Strategy == apiextv1.WebhookConverter {
```
LalatenduMohanty, line 39: This is something new to me. So will dig more and get back.

lib/resourcemerge/apiext_test.go
petr-muller: Yeah... I think that `cmp.Diff` is always better in these compare-output-against-expected unless you have some really specific use-case. Oftentimes you don't even need `equality.Semantic.DeepEqual` and can do `if diff := cmp.Diff(); diff != "" { //print diff }}`. `cmp.Diff` is not *that* old, I imagine some CVO code predates it. I'd consider these as tech dept and would welcome PRs that clean them up.

lib/resourcemerge/apiext.go
petr-muller: TBH I'd just place that in `apiext.go`, it's not used elsewhere. I'd only introduce a new file once we have more of these.

lib/resourcemerge/apiext.go
petr-muller: Looking at https://docs.openshift.com/container-platform/4.11/security/certificates/service-serving-certificate.html#add-service-certificate-crd_service-serving-certificate this PR looks legit to me. Many CRD manifests in the payload have that annotation so I'd expect this documented behavior to happen on them.

lib/resourcemerge/apiext_test.go
Davoska: Yes, you are right. Thank you for the helpful comments.

lib/resourcemerge/apiext.go
petr-muller: Cool, if see a non-hypothetical use case for the const in the observable future, then let's keep the separate file :+1:

petr-muller:
/retest

Azure is saaad today :/

Davoska:
The last two force pushes resolved the discussion https://github.com/openshift/cluster-version-operator/pull/870#discussion_r1043412314.
https://github.com/openshift/cluster-version-operator/commit/90dcc7808ac54367539959390c0dcf09f0f926a7: Remove exporting of the constant `injectCABundleAnnotation`
https://github.com/openshift/cluster-version-operator/commit/60a7cad03c5212af8cb7884ffc5467ba2adcd819: Update commit message

petr-muller (approved):

petr-muller:
/hold

Allow @LalatenduMohanty to [get back](https://github.com/openshift/cluster-version-operator/pull/870#discussion_r1042760739)

lib/resourcemerge/apiext.go
```diff
@@ -12,18 +12,36 @@ import (
...
+// the service-ca controller is responsible for the field.
+func ensureCustomResourceDefinitionV1CaBundle(required *apiextv1.CustomResourceDefinition, existing apiextv1.CustomResourceDefinition) {
+	if val, ok := existing.ObjectMeta.Annotations[injectCABundleAnnotation]; !ok || val != "true" {
+		return
+	}
+	if required.Spec.Conversion != nil && required.Spec.Conversion.Strategy == apiextv1.WebhookConverter {
+		// The service CA bundle is only injected by the service-ca controller into
+		// the CRD if the CRD is configured to use a webhook for conversion
+		if required.Spec.Conversion.Webhook != nil && required.Spec.Conversion.Webhook.ClientConfig != nil &&
+			existing.Spec.Conversion != nil && existing.Spec.Conversion.Webhook != nil && existing.Spec.Conversion.Webhook.ClientConfig != nil {
```
LalatenduMohanty, line 43: Putting all 5 conditions in one if conditional statement is a good practice. This reduces the readability of the code. Is there anyway we can break it down to nested if statements or someway make the code simpler?

LalatenduMohanty (changes requested):

lib/resourcemerge/apiext.go
LalatenduMohanty: Yeah, I agree.

lib/resourcemerge/apiext.go
LalatenduMohanty: Do we need to really check `existing.Spec.Conversion != nil` and   `existing.Spec.Conversion.Webhook` if we are checking for `existing.Spec.Conversion.Webhook.ClientConfig != nil`.

LalatenduMohanty (changes requested):

lib/resourcemerge/apiext.go
Davoska: > Is there anyway we can break it down to nested if statements or someway make the code simpler?

Yes, we can definitely try. I have restructured the conditions and added an `if` statement as to differentiate the checks for `required` and `existing`. What do you think about something like this?

```go
if required.Spec.Conversion != nil && required.Spec.Conversion.Strategy == apiextv1.WebhookConverter {
	// The service CA bundle is only injected by the service-ca controller into
	// the CRD if the CRD is configured to use a webhook for conversion
	if required.Spec.Conversion.Webhook != nil &&
		required.Spec.Conversion.Webhook.ClientConfig != nil {
		if existing.Spec.Conversion != nil &&
			existing.Spec.Conversion.Webhook != nil &&
			existing.Spec.Conversion.Webhook.ClientConfig != nil {
			required.Spec.Conversion.Webhook.ClientConfig.CABundle = existing.Spec.Conversion.Webhook.ClientConfig.CABundle
		}
	}
}
```

---

> Do we need to really check existing.Spec.Conversion != nil and existing.Spec.Conversion.Webhook if we are checking for existing.Spec.Conversion.Webhook.ClientConfig != nil.

Yes, I do believe so. All these fields are pointers and need to be checked before dereferencing, so we don't dereference a `nil` by any chance.

For example, in the [openshift/library-go](https://github.com/openshift/library-go) we can find an example of accessing the `<apiextensionsv1.CustomResourceDefinition>.Spec.Conversion.Webhook.ClientConfig` field like this ([pkg/operator/resource/resourcemerge/apiextensions.go#L41-L46](https://github.com/openshift/library-go/blob/e555322cb70844f90b003cbdfe7ce002d7c61810/pkg/operator/resource/resourcemerge/apiextensions.go#L41-L46)):

```go
func mimicCRDV1Defaulting(required *apiextensionsv1.CustomResourceDefinition) {
	crd_SetDefaults_CustomResourceDefinitionSpec(&required.Spec)

	if required.Spec.Conversion != nil &&
		required.Spec.Conversion.Webhook != nil &&
		required.Spec.Conversion.Webhook.ClientConfig != nil &&
		required.Spec.Conversion.Webhook.ClientConfig.Service != nil {
		crd_SetDefaults_ServiceReference(required.Spec.Conversion.Webhook.ClientConfig.Service)
	}
}
```

A similar thing at [openshift/operator-framework-olm/staging/operator-lifecycle-manager/pkg/controller/operators/olm/apiservices.go#L298-L302](https://github.com/openshift/operator-framework-olm/blob/55503f8eaa40ec8aa02d0b1310609c7551988554/staging/operator-lifecycle-manager/pkg/controller/operators/olm/apiservices.go#L298-L302) as well:

```go
if crd.Spec.Conversion == nil || crd.Spec.Conversion.Webhook == nil || crd.Spec.Conversion.Webhook.ClientConfig == nil && crd.Spec.Conversion.Webhook.ClientConfig.CABundle == nil {
	continue
}

return crd.Spec.Conversion.Webhook.ClientConfig.CABundle, nil
```

LalatenduMohanty: image
This format looks ok to me. Thanks

Davoska:
I have reformatted the code to improve the readability in the last commit [c2791e6](https://github.com/openshift/cluster-version-operator/commit/c2791e664edf9fa5cc5a430b67f88c5716a309ee) (based on the recommendations stated in the [#870 (comment)](https://github.com/openshift/cluster-version-operator/pull/870#discussion_r1052621839)).

lib/resourcemerge/apiext.go
```diff
@@ -12,18 +12,40 @@ import (
...
+		// The service CA bundle is only injected by the service-ca controller into
+		// the CRD if the CRD is configured to use a webhook for conversion
+		if required.Spec.Conversion.Webhook != nil &&
+			required.Spec.Conversion.Webhook.ClientConfig != nil {
+			if existing.Spec.Conversion != nil &&
+				existing.Spec.Conversion.Webhook != nil &&
+				existing.Spec.Conversion.Webhook.ClientConfig != nil {
+				required.Spec.Conversion.Webhook.ClientConfig.CABundle = existing.Spec.Conversion.Webhook.ClientConfig.CABundle
+			}
+		}
```
petr-muller, line 49:
```suggestion
	req := required.Spec.Conversion
	if req == nil ||
		req.Strategy != apiextv1.WebhookConverter ||
		req.Webhook == nil ||
		req.Webhook.ClientConfig == nil {
		return
	}
	// The service CA bundle is only injected by the service-ca controller into
	// the CRD if the CRD is configured to use a webhook for conversion
	if exc := existing.Spec.Conversion; exc != nil && exc.Webhook != nil && exc.Webhook.ClientConfig != nil {
		required.Spec.Conversion.Webhook.ClientConfig.CABundle = exc.Webhook.ClientConfig.CABundle
	}
```

I'm sorry for bikeshedding style this late in the PR but I think we can simplify for readability with guardclauses and some variable extraction... I don't insist, just consider.

Davoska: Don't be sorry! Thank you for the suggestions. I appreciate them.

What do you say about something like this? I tried to make it even more consistent (added a new line for `exc := existing.Spec.Conversion`) and I tried to separate the logic even more (added a new condition only including the `WebhookConverter` check).

```go
	req := required.Spec.Conversion
	if req == nil ||
		req.Webhook == nil ||
		req.Webhook.ClientConfig == nil {
		return
	}
	if req.Strategy != apiextv1.WebhookConverter {
		// The service CA bundle is only injected by the service-ca controller into
		// the CRD if the CRD is configured to use a webhook for conversion
		return
	}
	exc := existing.Spec.Conversion
	if exc != nil &&
		exc.Webhook != nil &&
		exc.Webhook.ClientConfig != nil {
		required.Spec.Conversion.Webhook.ClientConfig.CABundle = exc.Webhook.ClientConfig.CABundle
	}
```

petr-muller: that's good. personally I wouldn't linebreak the conditions in the `if` statements, the resulting lines are short. but really that's just nitpicking. I think the main benefit comes from extracting the long dot-chains to variables :+1:

LalatenduMohanty: The new structure looks good either way.

LalatenduMohanty: I can add a lgtm once the changes are done.

Davoska:
/hold 

I need to check the CVO logs in the CI to confirm there wasn't any regression in the new changes.

Davoska:
The caused hotlooping seems to be gone in the log file from the [e2e-agnostic-upgrade-into-change](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/870/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1616379079022350336/artifacts/e2e-agnostic-upgrade-into-change/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-58cf567b5c-hprmg_cluster-version-operator.log) job:
```bash
$ grep -io 'updating.*due to diff.*' openshift-cluster-version_cluster-version-operator-58cf567b5c-hprmg_cluster-version-operator.log | sort | uniq -c
      6 Updating CronJob openshift-operator-lifecycle-manager/collect-profiles due to diff:   &unstructured.Unstructured{
      6 Updating Namespace cli-artifacts due to diff:   &v1.ImageStream{
      6 Updating Namespace cli due to diff:   &v1.ImageStream{
      6 Updating Namespace driver-toolkit due to diff:   &v1.ImageStream{
      6 Updating Namespace installer-artifacts due to diff:   &v1.ImageStream{
      6 Updating Namespace installer due to diff:   &v1.ImageStream{
      6 Updating Namespace must-gather due to diff:   &v1.ImageStream{
      6 Updating Namespace oauth-proxy due to diff:   &v1.ImageStream{
      6 Updating Namespace tests due to diff:   &v1.ImageStream{
      6 Updating Namespace tools due to diff:   &v1.ImageStream{
      6 Updating ValidatingWebhookConfiguration /controlplanemachineset.machine.openshift.io due to diff:   &unstructured.Unstructured{
      6 Updating ValidatingWebhookConfiguration /performance-addon-operator due to diff:   &unstructured.Unstructured{
```

As opposed to the log file from the `4.13.0-0.ci-2023-01-19-175836` version. As we can see no new hotlooping was introduced in the above output and only the CRD hotloopings was removed as expected.

```bash
[dhurta@fedora Downloads]$ grep -io 'updating.*due to diff.*' cluster-version-operator-c98796f6b-cll9c-cluster-version-operator.log | sort | uniq -c
      2 Updating CRD alertmanagerconfigs.monitoring.coreos.com due to diff:   &v1.CustomResourceDefinition{
      2 Updating CRD consoleplugins.console.openshift.io due to diff:   &v1.CustomResourceDefinition{
      2 Updating CRD performanceprofiles.performance.openshift.io due to diff:   &v1.CustomResourceDefinition{
      2 Updating CronJob openshift-operator-lifecycle-manager/collect-profiles due to diff:   &unstructured.Unstructured{
      2 Updating Namespace cli-artifacts due to diff:   &v1.ImageStream{
      2 Updating Namespace cli due to diff:   &v1.ImageStream{
      2 Updating Namespace driver-toolkit due to diff:   &v1.ImageStream{
      2 Updating Namespace installer-artifacts due to diff:   &v1.ImageStream{
      2 Updating Namespace installer due to diff:   &v1.ImageStream{
      2 Updating Namespace must-gather due to diff:   &v1.ImageStream{
      2 Updating Namespace oauth-proxy due to diff:   &v1.ImageStream{
      2 Updating Namespace tests due to diff:   &v1.ImageStream{
      2 Updating Namespace tools due to diff:   &v1.ImageStream{
      2 Updating ValidatingWebhookConfiguration /controlplanemachineset.machine.openshift.io due to diff:   &unstructured.Unstructured{
      2 Updating ValidatingWebhookConfiguration /performance-addon-operator due to diff:   &unstructured.Unstructured{
```

_I am going to investigate the hotlooping on the namespaces which are `ImageStream`. But this issue is unrelated to this PR._

---

Looking for any mentions of CRDs returns the expected messages. Meaning no `error running...` were reported.

Trimmed output of `grep`:
```bash
$ grep -io '].*CRD.*\|].*customresourcedefinition.*' openshift-cluster-version_cluster-version-operator-58cf567b5c-hprmg_cluster-version-operator.log | sort | uniq -c
      1 ] customresourcedefinition "machinedisruptionbudgets.healthchecking.openshift.io" not found. It either has already been removed or it has never been installed on this cluster.
      1 ] customresourcedefinition "machinehealthchecks.healthchecking.openshift.io" not found. It either has already been removed or it has never been installed on this cluster.
     12 ] Done syncing for customresourcedefinition "alertmanagerconfigs.monitoring.coreos.com" (390 of 832)
...
     12 ] Done syncing for customresourcedefinition "tuneds.tuned.openshift.io" (422 of 832)
      2 ] excluding 0000_10_config-operator_01_infrastructure-TechPreviewNoUpgrade.crd.yaml group=apiextensions.k8s.io kind=CustomResourceDefinition namespace= name=infrastructures.config.openshift.io: "Default" is required, and release.openshift.io/feature-set=TechPreviewNoUpgrade
...
      2 ] excluding 0000_80_machine-config-operator_02_containerruntimeconfig.crd.yaml group=apiextensions.k8s.io kind=CustomResourceDefinition namespace= name=containerruntimeconfigs.machineconfiguration.openshift.io: "Default" is required, and release.openshift.io/feature-set=TechPreviewNoUpgrade
     12 ] Running sync for customresourcedefinition "alertmanagerconfigs.monitoring.coreos.com" (390 of 832)
...
     12 ] Running sync for customresourcedefinition "tuneds.tuned.openshift.io" (422 of 832)
```

Davoska:
/unhold

The new changes are working as expected.

Davoska:
/retest

The failing jobs seems to be unrelated to the introduced changes.

lib/resourcemerge/apiext.go
Davoska: I have pushed the above mentioned solution by me (https://github.com/openshift/cluster-version-operator/pull/870#discussion_r1081545493) and changed the line:

```go
		required.Spec.Conversion.Webhook.ClientConfig.CABundle = exc.Webhook.ClientConfig.CABundle
```

to:

```go
		req.Webhook.ClientConfig.CABundle = exc.Webhook.ClientConfig.CABundle
```

Because the variable `req` is a pointer `*apiextv1.CustomResourceConversion` and assigning the `CABundle` does change the value in the original variable.

LalatenduMohanty (approved):
/lgtm

petr-muller:
Looking at the offending job history, there does not seem to be any pattern or suspicious failure, it's a combination of generic disruption errors and failure-to-installs.

Davoska:
/unhold
/retest

The CI was already passing and the failing jobs seems to be unrelated to these changes.

petr-muller:
Tested a million times ;)

Davoska:

This job [was already overridden](https://github.com/openshift/cluster-version-operator/pull/870#issuecomment-1401942444) but due to another PR merging, the job needs to be rerun and potentially overridden again. The job itself [has passed in the past](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change).

---

The last failed run doesn't seem to be caused by these changes as well. 

`Cluster did not complete upgrade: timed out waiting for the condition: deployment openshift-dns-operator/dns-operator is Progressing=False: ProgressDeadlineExceeded: ReplicaSet "dns-operator-6c6df4c7db" has timed out progressing.`

`ProgressDeadlineExceeded: ReplicaSet "dns-operator-6c6df4c7db" has timed out progressing.`

And multiple events caused by `reason/FailedCreatePodSandBox Failed to create pod sandbox`

PR #871, opened 2022-12-06
OTA-560: Improve developer-oriented docs

petr-muller:
- Update `CONTRIBUTING.md`
- Provide more convenience `Makefile` targets
- Cleanup codegen `Makefile` targets
- Move developer-facing content from `README.md` to `docs/dev/README.md`

This is a companion PR to https://github.com/openshift/cluster-version-operator/pull/869

I have omitted the `Installing CVO and operators in cluster` and `Using CVO to render the release-image locally` sections from the [original README](https://github.com/openshift/cluster-version-operator/blob/e6aeb8dd171b9e442767cf8e88b034c4844de749/README.md) b/c I don't think these use-cases are too interesting when developing CVO. I have updated a section on building a custom release payload to build from an existing payload (instead if imagestream) and added a section that instructs how to inject such release payload to the cluster


Makefile
```diff
@@ -11,16 +11,24 @@ test:
...
-verify-codegen: verify-codegen-crds
+update-codegen:
+	./hack/update-codegen.sh
+
+verify-codegen:
+	./hack/verify-codegen.sh
+
 verify: verify-codegen
-.PHONY: update-codegen-crds update-codegen verify-codegen-crds verify-codegen verify
+.PHONY: update-codegen verify-codegen verify
```
wking, line 34: nit: dup `verify-codegen` entries?  Also, the pattern in the rest of the `Makefile` appears to be a new `.PHONY` entry after each target.  I'm agnostic about whether we use that or a single large `.PHONY` entry, but I'd rather not split between both patterns.

docs/dev/README.md
```diff
@@ -1,4 +1,90 @@
...
+This pushes `${VERSION}` and `latest` tags to `${REPO}/origin-cluster-version-operator`, where `${VERSION}` encodes the
+Git commit used to build the image.
+
+### Building and publishing a release payload image with development CVO
+
+After publishing a [development CVO image](#publishing-a-development-cvo-image), you can build a release payload image
+that will contain all manifests from an existing release payload and the development CVO binary:
+
+```console
+$ REPO=<your personal repo, such as quay.io/login> ./hack/push-image.sh
```
wking, line 49: down in _Replace cluster's CVO with development CVO release payload image_, you list publishing this image as a prereq, and do not repeat the steps to publish.  Here, you list publishing this image as a prereq and are repeating the steps to publish.  I'm fine with both inline copies of the command, or expecting folks to go read the CVO-image-publishing section to get the command, but it seems like both CVO-image-dependent sections should use the same approach.

petr-muller: Thanks, addressed in 7971a6a96def102c258193b3a768ab075a33d62f by removing the repeated publishing steps. I also noticed the crosslink in _Replace cluster's CVO with development CVO release payload image_ was leading to a wrong section, so fixed that too.

Makefile
petr-muller: I've just removed the codegen stuff entirely: the `update-codegen` was just a tombstone and we do not use it in our CI jobs: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-master.yaml

petr-muller:
I'm not ignoring the flake, I'm looking if I can fix it in a separate PR

petr-muller:

Azure flakes during cluster installation:

```
 level=error msg=Error: creating/updating Load Balancer Backend Address Pool "Load Balancer Backend Address Pool: (Backend Address Pool Name \"ci-op-2psbn19r-b1493-qqklb\" / Load Balancer Name \"ci-op-2psbn19r-b1493-qqklb\" / Resource Group \"ci-op-2psbn19r-b1493-qqklb-rg\")": network.LoadBalancerBackendAddressPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=500 -- Original Error: Code="MultipleErrorsOccurred" Message="Multiple error occurred: . Please see details." Details=[{"code":"InternalServerError","message":"Encountered internal server error. Diagnostic information: timestamp '20221213T161143Z', subscription id 'd38f1e38-4bed-438e-b227-833f997adf6a', tracking id 'a6b364d1-60d9-4263-892a-317b8fb28b8d', request correlation id '2e2361f4-15bd-b6bf-c55d-4b480921a88e'."},{"code":"InternalServerError","message":"Encountered internal server error. Diagnostic information: timestamp '20221213T161143Z', subscription id 'd38f1e38-4bed-438e-b227-833f997adf6a', tracking id 'a6b364d1-60d9-4263-892a-317b8fb28b8d', request correlation id '2e2361f4-15bd-b6bf-c55d-4b480921a88e'."}] 
```

CONTRIBUTING.md
```diff
@@ -22,20 +22,18 @@ For contributors who want to work up pull requests, the workflow is roughly:
 1. Create a topic branch from where you want to base your work (usually master).
 2. Make commits of logical units.
 3. Make sure your commit messages are in the proper format (see [below](#commit-message-format)).
-4. If necessary, you may use `make update-codegen` to update generated code.
-5. Make sure the tests pass (`go test ./...`), and add any new tests as appropriate.
+4. Make sure your [code builds][build-cvo] and follows the [coding style](#coding-style)
```
LalatenduMohanty, line 25: Any reason we not suggesting `make update-codegen` ?

LalatenduMohanty: Do we expect that users will notice the test failures in the PR and then run it?

LalatenduMohanty (changes requested):

CONTRIBUTING.md
petr-muller: update-codegen.sh is a tombstone and does nothing: https://github.com/openshift/cluster-version-operator/blob/master/hack/update-codegen.sh

@wking [suggested](https://github.com/openshift/cluster-version-operator/pull/871#discussion_r1047353017) that I fix some style in the related makefile targets so I ended up removing all that to avoid maintenance of code with no value

petr-muller: Leaving a note here for our future selves after a Slack convo. The answer is that the code generation happens in openshift/api repository and then we vendor it.

PR #874, opened 2022-12-08
OCPBUGS-1443: Avoid using a lister before client caches sync

petr-muller:
Previously, CVO could use a `ClusterVersion` lister before its informer's cache was synced, leading to the following noise in a CVO log, that could also propagate into an `Upgradeable=False` condition and cause user-observable noise:

```
Unable to get ClusterVersion, err=clusterversion.config.openshift.io "version" not found
```

This is coming from one of the upgradeable checks called inside `setUpgradeableCondition`. This method is called in two scenarios: 
- in `ConfigMap` informer create/update/delete handlers
- in `syncUpgradeable->upgradeableSync` path that handles work items  from the `upgradeableQueue` queue. This path throttles the sync and only sets the conditions if sufficient time passed since the last sync

The failure case is that while caches of all informers are synced, the `ConfigMap` informer handler can be called while the `ClusterVersion` informer cache is not synced yet (so this is a race condition bug).

My original solution was to stop calling `setUpgradeableCondition` from the handlers entirely, and queue the full upgradeable sync instead. This fixes the issue at startup, but then it causes CVO to handle admin-ack CM changes by going through the potentially throttled full sync path, which (I think) could mean that CVO would react on manual admin action with a non-trivial delay, which is poor UX (people expect their actions to have observable effects). So instead, I made the handler checks whether the informers are synced, and if they are, then sync the conditions immediately, but if they are not, queue a full upgradeable sync.

I'm not 100% convinced the increased complexity is worth it for a mostly cosmetic fix, so I'm open to opinions.

### DIFF
--- a/pkg/cvo/upgradeable.go
+++ b/pkg/cvo/upgradeable.go
@@ -403,11 +403,27 @@ func (optr *Operator) defaultUpgradeableChecks() []upgradeableCheck {
 	}
 }
 
+// setUpgradableConditionsIfSynced calls setUpgradableConditions if all informers were synced at least once, otherwise
+// it queues a full upgradable sync that will eventually execute
+//
+// This method is needed because it is called in an informer handler, but setUpgradeableConditions uses a lister
+// itself, so it races on startup (it can get triggered while one informer's cache is being synced, but depends
+// on another informer's cache already synced
+func (optr *Operator) setUpgradableConditionsIfSynced() {
+	for _, synced := range optr.cacheSynced {
+		if !synced() {
+			optr.upgradeableQueue.Add(optr.queueKey())
+			return
+		}
+	}
+	optr.setUpgradeableConditions()
+}
+
 func (optr *Operator) addFunc(obj interface{}) {
 	cm := obj.(*corev1.ConfigMap)
 	if cm.Name == internal.AdminGatesConfigMap || cm.Name == internal.AdminAcksConfigMap {
 		klog.V(2).Infof("ConfigMap %s/%s added.", cm.Namespace, cm.Name)
-		optr.setUpgradeableConditions()
+		optr.setUpgradableConditionsIfSynced()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


wking (approved):
Waiting until all the caches are sync'ed seemed like it might be too conservative, delaying updates to the `Upgradeable` condition when we had sync'ed sufficiently many caches, but maybe not some peripheral caches.  However, inspecting today, our only caches for ClusterVersion and ClusterOperator:

```console
$ git --no-pager grep 'cacheSynced.*append'
pkg/autoupdate/autoupdate.go:   ctrl.cacheSynced = append(ctrl.cacheSynced, cvInformer.Informer().HasSynced)
pkg/autoupdate/autoupdate.go:   ctrl.cacheSynced = append(ctrl.cacheSynced, coInformer.Informer().HasSynced)
pkg/cvo/cvo.go: optr.cacheSynced = append(optr.cacheSynced, coInformer.Informer().HasSynced)
pkg/cvo/cvo.go: optr.cacheSynced = append(optr.cacheSynced, cvInformer.Informer().HasSynced)
```

And we need both of those to perform the `Upgradeable` checks.  And also, if we grow a new peripheral informer in the future, we're likely to crash-loop if we fail to sync that new cache, and the crash-looping CVO will summon an admin to help us recover, regardless of whether ClusterVersion's `Upgradeable` goes stale or not.  So:

/lgtm

evakhoni:
pre-merge verified in https://issues.redhat.com/browse/OCPBUGS-1443?#comment-21404909
/label qe-approved

petr-muller:
Closing was a mistake, sorry for the noise :/

wking:
[`nodes should not go unready after being upgraded and go unready only once`][1] is unrelated:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/874/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1602621433580621824

PR #875, opened 2022-12-13
tests: fix rare nil dereference in `TestCache`

petr-muller:
Looking at [`Cache.Match()`](https://github.com/openshift/cluster-version-operator/blob/master/pkg/clusterconditions/cache/cache.go#L63-L148) implementation, the `c.Condition.Match()` on [L133](https://github.com/openshift/cluster-version-operator/blob/master/pkg/clusterconditions/cache/cache.go#L133) can be called with `targetCondition=nil` and that seems to be intentional. In tests, `c.Condition` is a mock, which did not expect that pointer to be `nil` and dereferenced it, resulting in an occassional panic in tests like https://github.com/openshift/cluster-version-operator/pull/871#issuecomment-1348927131. The fix stores the full pointer (its deepcopy to avoid sharing that memory) in the mock instead.

I needed to restructure the tests a little to allow adding a nil parameter testcase

### DIFF
--- a/pkg/clusterconditions/cache/cache_test.go
+++ b/pkg/clusterconditions/cache/cache_test.go
@@ -94,10 +94,12 @@ func TestCache(t *testing.T) {
 
 	for i := 0; i < maxConcurrence; i++ {
 		name := fmt.Sprintf("condition %d", i)
-		condition := configv1.ClusterCondition{Type: name}
+		condition := &configv1.ClusterCondition{Type: name}
 		var previousCall *mock.Call
+		found := false
 		for j, call := range m.Calls {
 			if reflect.DeepEqual(call.Condition, condition) {
+				found = true
 				if previousCall != nil {
 					spaceBetweenCalls := call.When.Sub(previousCall.When)
 					if spaceBetweenCalls < c.MinForCondition {
@@ -106,6 +108,9 @@ func TestCache(t *testing.T) {
 				}
 				previousCall = &m.Calls[j]
 			}
+		}
+		if !found {
+			t.Errorf("condition %v not found in recorded calls", condition)
 		}
 	}
 }
@@ -161,7 +166,7 @@ func Test_calculateMostStale(t *testing.T) {
 			expectedKey: `{"type": "a"}`,
 		},
 		{
-			name: "two entries, both old evaluatations, clear evaluation winner",
+			name: "two entries, both old evaluations, clear evaluation winner",
 			cache: map[string]*MatchResult{
 				`{"type": "a"}`: {When: now.Add(-2 * time.Hour), Access: now, Match: true, Error: nil},
 				`{"type": "b"}`: {When: now.Add(-time.Hour), Access: now.Add(-time.Minute), Match: true, Error: nil},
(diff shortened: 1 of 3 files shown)


petr-muller:

```
Node ci-op-ss56w10k-b1493-6h6wr-worker-eastus21-6w9b7 went unready multiple times: 2022-12-15T00:54:13Z, 2022-12-15T00:58:34Z
Node ci-op-ss56w10k-b1493-6h6wr-worker-eastus21-6w9b7 went ready multiple times: 2022-12-15T00:57:01Z, 2022-12-15T00:58:39Z
```

pkg/clusterconditions/mock/mock_test.go
```diff
@@ -68,39 +80,58 @@ func TestMock(t *testing.T) {
...
+		condition     *configv1.ClusterCondition
+		pushMatches   []mock.MatchResult
 		expectedMatch bool
 		expectedError *regexp.Regexp
 	}{
 		{
+			condition:     nil,
+			pushMatches:   []mock.MatchResult{{Match: true, Error: nil}},
 			expectedMatch: true,
-			expectedError: nil,
```
Davoska, line 83: Possible nitpick:

I would maybe leave these explicit `nil` declarations (this line and previous line [91](https://github.com/openshift/cluster-version-operator/blob/fa8004488fbdc02dc58f004db09c7597c832b2bc/pkg/clusterconditions/mock/mock_test.go#L91)). They make it explicit that the expected value for the error in the respective test is `nil` as in the original code (since we are checking their values). 

But I don't have a strong feeling against removing them. Don't feel blocked by this.

Davoska (changes requested):
One minor requested change and one even smaller possible nitpick :+1: 

Other than that, the code looks good to me! I really like the new restructured code. At least for me, it's more readable.

pkg/clusterconditions/cache/cache_test.go
```diff
@@ -161,7 +161,7 @@ func Test_calculateMostStale(t *testing.T) {
 			expectedKey: `{"type": "a"}`,
 		},
 		{
-			name: "two entries, both old evaluatations, clear evaluation winner",
+			name: "two entries, both old evaluations, clear evaluation winner",
```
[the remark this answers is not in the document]
petr-muller: Great catch :+1: 

I'll also check whether the test could be made more robust against such silent failure, it seems that it's quite easy to unwittingly make this test silently pass by never passing that `DeepEqual`.

pkg/clusterconditions/mock/mock_test.go
petr-muller: I have a slight preference for omitting these and relying on golang zero value defaults, it's idiomatic in go.

Davoska: Roger that, feel free to close this thread :+1: 

Meanwhile, I'll go re-read idioms in Go :books:

pkg/clusterconditions/cache/cache_test.go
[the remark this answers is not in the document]
petr-muller: Looks like this was the only other place where `.Condition`  was used, so we should be safe now. I also improved the test to fail when the checked condition is not found in the recorded calls (it fails before the fix you suggested).

Davoska:
[Override](https://github.com/openshift/cluster-version-operator/pull/875#issuecomment-1357630250) seems to be justified, this PR only affects unit tests and shouldn't result in the given error.

New changes looks good to me :+1:

/lgtm

PR #876, opened 2022-12-16
OCPBUGS-4986: pkg/payload/precondition: Do not claim warnings would have blocked

wking:
Avoid errors like:

```console
$ oc get -o json clusterversion version | jq -r '.status.history[0].acceptedRisks'
Forced through blocking failures: Precondition "ClusterVersionRecommendedUpdate" failed because of "UnknownUpdate": RetrievedUpdates=True (), so the update from 4.13.0-0.okd-2022-12-11-064650 to 4.13.0-0.okd-2022-12-13-052859 is probably neither recommended nor supported.
```

Instead, tweak the logic from 481bcde393 (#856), and only append the `Forced through blocking failures:` prefix when the forcing was required.

### DIFF
--- a/pkg/payload/precondition/precondition.go
+++ b/pkg/payload/precondition/precondition.go
@@ -81,6 +81,7 @@ func Summarize(errs []error, force bool) (bool, error) {
 			}
 			continue
 		}
+		isWarning = false
 		msgs = append(msgs, e.Error())
 
 	}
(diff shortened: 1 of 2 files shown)


bshephar:
Nice clean solution. fwiw, looks good to me.

LalatenduMohanty (approved):
/lgtm

wking:
[Install failed][1].

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/876/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1603824085119275008

PR #882, opened 2023-01-09
OCPBUGS-5505: Set upgradeability check throttling period to 2m

petr-muller:
Previously, the throttling reused the `minimumUpdateCheckInterval` value which is derived from the full CVO minimum sync period. This value is set between 2m and 4m at CVO startup. This period is unecessarily long and bad for UX, things happen with a delay and our own testcase expects upgradeability to be propagated in 3 minutes at worst.

Hardcode the throttling to 2m to prevent flapping on flurries but allow changes to propagate deterministically faster. We will still get a bit of non-determinisim from sync periods and requeueing, so this change should not cause any periodic API-hammering.

I'd like to backport this change to 4.11 where it causes CI flakes in `[bz-Cluster Version Operator] Verify presence of admin ack gate blocks upgrade until acknowledged` test.

### DIFF
--- a/pkg/cvo/upgradeable.go
+++ b/pkg/cvo/upgradeable.go
@@ -29,20 +29,45 @@ import (
 const (
 	adminAckGateFmt             string = "^ack-[4-5][.]([0-9]{1,})-[^-]"
 	upgradeableAdminAckRequired        = configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired")
-	checkThrottlePeriod                = 2 * time.Minute
 )
 
 var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
 
-// syncUpgradeable synchronizes the upgradeable status only if it has been more than
-// the checkThrottlePeriod since the last synchronization or the precondition
-// checks on the payload are failing for less than minimumUpdateCheckInterval, and it has
-// been more than the minimumUpgradeableCheckInterval since the last synchronization.
-func (optr *Operator) syncUpgradeable(config *configv1.ClusterVersion) error {
-	u := optr.getUpgradeable()
-	if u != nil && u.RecentlyChanged(checkThrottlePeriod) && !shouldSyncUpgradeableDueToPreconditionChecks(optr, config, u) {
-		klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
-		return nil
+// upgradeableCheckIntervals holds the time intervals that drive how often CVO checks for upgradeability
+type upgradeableCheckIntervals struct {
+	// min is the base minimum interval between upgradeability checks, applied under normal circumstances
+	min time.Duration
+
+	// minOnFailedPreconditions is the minimum interval between upgradeability checks when precondition checks are
+	// failing and were recently (see afterPreconditionsFailed) changed. This should be lower than min because we want CVO
+	// to check upgradeability more often
+	minOnFailedPreconditions time.Duration
+
+	// afterFailingPreconditions is the period of time after preconditions failed when minOnFailedPreconditions is
+	// applied instead of min
+	afterPreconditionsFailed time.Duration
+}
+
+func defaultUpgradeableCheckIntervals() upgradeableCheckIntervals {
+	return upgradeableCheckIntervals{
+		// 2 minutes are here because it is a lower bound of previously nondeterministicly chosen interval
+		// TODO (OTA-860): Investigate our options of reducing this interval. We will need to investigate
+		// the API usage patterns of the underlying checks, there is anecdotal evidence that they hit
+		// apiserver instead of using local informer cache
+		min:                      2 * time.Minute,
+		minOnFailedPreconditions: 15 * time.Second,
+		afterPreconditionsFailed: 2 * time.Minute,
+	}
+}
+
+// syncUpgradeable synchronizes the upgradeable status only if the sufficient time passed since its last update. This
+// throttling period is dynamic and is driven by upgradeableCheckIntervals.
+func (optr *Operator) syncUpgradeable(cv *configv1.ClusterVersion) error {
+	if u := optr.getUpgradeable(); u != nil {
+		if u.RecentlyChanged(optr.upgradeableCheckIntervals.throttlePeriod(cv)) {
+			klog.V(2).Infof("Upgradeable conditions were recently checked, will try later.")
+			return nil
+		}
 	}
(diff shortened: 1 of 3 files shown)


petr-muller:
/test e2e-agnostic-upgrade-out-of-change
^^^ failed to install

/override e2e-agnostic-upgrade-into-change
^^^ passed everything except the following, which is unrelated
```
: [sig-api-machinery] disruption/cache-kube-api connection/reused should be available throughout the test
{
  cache-kube-api-reused-connections was unreachable during disruption testing for at least 2s of 1h35m24s (maxAllowed=1s):  
}
```

pkg/cvo/upgradeable.go
```diff
@@ -29,17 +29,18 @@ import (
...
+	checkThrottlePeriod                = time.Minute
 )
 
 var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
 
 // syncUpgradeable synchronizes the upgradeable status only if it has been more than
-// the minimumUpdateCheckInterval since the last synchronization or the precondition
+// the checkThrottlePeriod since the last synchronization or the precondition
 // checks on the payload are failing for less than minimumUpdateCheckInterval, and it has
 // been more than the minimumUpgradeableCheckInterval since the last synchronization.
```
wking, line 40: Do these two `minimumUpdateCheckInterval` refs need to be updated too?  I see `shouldSyncUpgradeableDueToPreconditionChecks` later in this file still using `minimumUpdateCheckInterval`, but if `minimumUpdateCheckInterval` is going to be scoped to `upstream` Cincinnati-side stuff, I'd expect us to be using an `Upgradeable`-specific knob for `Upgradeable`-side stuff.  Maybe `checkThrottlePeriod` should be specific to `Upgradeable` throttling, and we want a third variable for whatever `shouldSyncUpgradeableDueToPreconditionChecks` is doing?  Or maybe they can both share `checkThrottlePeriod`?  Or something?

pkg/cvo/upgradeable.go
```diff
@@ -29,17 +29,18 @@ import (
 const (
 	adminAckGateFmt             string = "^ack-[4-5][.]([0-9]{1,})-[^-]"
 	upgradeableAdminAckRequired        = configv1.ClusterStatusConditionType("UpgradeableAdminAckRequired")
+	checkThrottlePeriod                = time.Minute
```
wking, line 32: nit: some day I'd like to break `pkg/cvo` up into more, smaller sub-packages.  But at the moment it's a big bucket doing lots of things, and we might want to use `checkUpgradeableThrottlePeriod` for clearer scoping here.  And add godocs explaining its intended semantics (not strictly required for these internal variables, but still potentially helpful to avoid accidental developer misuse)?

petr-muller: Good point. I'm still pondering making it a member to allow unit-testing, then this const would be called `default...`.

petr-muller:
@wking PTAL. Encouraged by your _"some day I'd like to break pkg/cvo up into more, smaller sub-packages"_ comment, I went for a slightly larger refactor of the related code and decoupled the logic from `cvo.go` better. I extracted the code that drives the "is our `Upgradeable` recent enough?" question to a dedicated method that is now tested and bundled the upgradeable-related intervals into a helper struct.

petr-muller:
```
openshift-api-reused-connections was unreachable during disruption testing for at least 4s of 1h39m12s (maxAllowed=2s):
oauth-api-reused-connections was unreachable during disruption testing for at least 4s of 1h39m12s (maxAllowed=2s):
```
Unrelated
/retest

petr-muller:
3rd attempt failed to install, previous two attempts did not indicate a problem with CVO

pkg/cvo/upgradeable.go
LalatenduMohanty: What is the downside of reducing the checkThrottlePeriod further?

petr-muller: Good question, I'm not entirely sure. Abhinav's message from https://github.com/openshift/cluster-version-operator/commit/04528144feb7a8141801bce591fda43d65acc48a that introduced the throttling does not help too much about why exactly `minUpdateCheckInterval` was chosen:

> Adds another long running sync like the availableUpdates that updates the operator with upgradeable conditions every minUpdateChekInterval.

`minUpdateCheckInterval` itself happened in this beast https://github.com/openshift/cluster-version-operator/pull/45 and there's too much content there to be useful, the commit message says just:

> * Avoid reconciling too often (exit before applying) when no spec change recorded

I'm not sure about how expensive the upgradeability checks are - that's one of the two reasons I can imagine throttling is useful (the other is suppressing user-observable flapping in busy periods).

petr-muller: But the refactoring on this PR should make experimenting with this easier ;)

wking (approved):
/lgtm

pkg/cvo/upgradeable.go
```diff
@@ -466,21 +492,20 @@ func (optr *Operator) adminGatesEventHandler() cache.ResourceEventHandler {
...
-	cond := resourcemerge.FindOperatorStatusCondition(config.Status.Conditions, DesiredReleaseAccepted)
-	if cond != nil && cond.Reason == "PreconditionChecks" && cond.Status == configv1.ConditionFalse &&
-		hasPassedDurationSinceTime(u.At, optr.minimumUpgradeableCheckInterval) && !hasPassedDurationSinceTime(cond.LastTransitionTime.Time, optr.minimumUpdateCheckInterval) {
-		return true
-	}
-	return false
+// The cv parameter is expected to be non-nil.
+func (intervals *upgradeableCheckIntervals) throttlePeriod(cv *configv1.ClusterVersion) time.Duration {
+	if cond := resourcemerge.FindOperatorStatusCondition(cv.Status.Conditions, DesiredReleaseAccepted); cond != nil {
+		// Function returns true if the synchronization should happen, returns false otherwise.
```
evakhoni, line 504: nit: this comment probably belongs to the previous `func shouldSyncUpgradeableDueToPreconditionChecks`

petr-muller: good catch, I want to submit a little code cleanup PR for this area that wasnt appropriate for a to-be-backported bugfix, I'll make sure I fix this there

petr-muller: Included in https://github.com/openshift/cluster-version-operator/pull/886

PR #887, opened 2023-01-19
start.go: simplify code for computing sync period

petr-muller:
Previously, `resyncPeriod` returned a `func() time.Duration` that did the actual computation, but there is no place where the function is stored and used later as a callback. It was always immediately called after creation.

Remove one layer of unnecessary indirection and make `resyncPeriod` returns the computed sync period directly.

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -354,11 +354,9 @@ func createResourceLock(cb *ClientBuilder, namespace, name string) (resourcelock
 	})
 }
 
-func resyncPeriod(minResyncPeriod time.Duration) func() time.Duration {
-	return func() time.Duration {
-		factor := rand.Float64() + 1
-		return time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

[These][1] are orthogonal, and everything in this run suggests the CVO is fine:

And [these][2] are orthogonal, and while they blocked us from measuring CVO behavior, I'm confident enough in the other runs and the clarity of the code change to skip retesting there too:

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/887/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-out-of-change/1616099183746355200
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/887/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator/1616099183654080512

PR #888, opened 2023-01-19
lib/validation: Drop "a previous version" from desiredUpdate validation

wking:
Narrowly catching up with c80bd024a8 (#684, [4.10][1]), which removed that option from the implementation.  Having the validation code living so far from the implementation code makes us more vulnerable to this kind of divergence, but I'll leave larger refactorings to follow-up work.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1878925


petr-muller (approved):

petr-muller:

Both results only failed on disruption tests and the OLM breakage-du-jour

PR #862, opened 2022-11-09
Bug 2089093: CVO hotloops on OperatorGroup due to the diff of "upgradeStrategy"

Davoska:
This pull request will add `/apis/operators.coreos.com/v1/operatorgroups` handling that enables a custom way of processing manifests representing the `OperatorGroup` resource. Currently, the `OperatorGroup` resources are reconciled as generic resources. This doesn't provide defaulting of the fields.

This pull request will provide more precise processing of the `OperatorGroup` resources and will provide defaulting of the resources, which will fix the current hotlooping of the CVO.

Currently, the CVO hotloops on the `OperatorGroup` resources due to the missing defaulting of the resource (Bug 2089093 [1]).

The [2] commit introduced the `UpgradeStrategy` field to the `OperatorGroup` type. This field has a default value of `"Default"`. The CVO currently doesn't know of this default value, and thus the CVO hotloops when handling the `OperatorGroup` resource as a generic resource because some manifests don't specify the value of the field and rely on defaulting [3].

This pull request will provide a more precise way of processing the `OperatorGroup` type and will also fix the mentioned bug by providing the defaulting for the resource.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2089093
[2] openshift/operator-framework-olm@d895574
[3] https://github.com/openshift/operator-framework-olm/blob/ab5ecfd7f235142f191d91d103afabccdfd6ff04/manifests/0000_50_olm_13-operatorgroup-default.yaml

PR #893, opened 2023-01-26
Bug 2089138: CVO hotloops on ValidatingWebhookConfiguration

Davoska:
This pull request will fix hotlooping on `ValidatingWebhookConfiguration`.

The hotlooping was caused by missing CVO-side defaulting of the `ValidatingWebhookConfiguration` resources and by missing CVO awareness of the `service-ca` operator who is responsible for injecting the CA bundle into resources annotated by `service.beta.openshift.io/inject-cabundle=true` [1].

This pull request adds the defaulting and ensures that the CVO ignores the respective `CABundle` fields inside the webhooks when the annotation is set.

This pull request references bug 2089138 [2].

[1] https://docs.openshift.com/container-platform/4.12/security/certificates/service-serving-certificate.html
[2] https://bugzilla.redhat.com/show_bug.cgi?id=2089138

### DIFF
--- a/lib/resourcebuilder/resourcebuilder.go
+++ b/lib/resourcebuilder/resourcebuilder.go
@@ -45,9 +45,9 @@ type builder struct {
 	configClientv1          *configclientv1.ConfigV1Client
 	coreClientv1            *coreclientv1.CoreV1Client
 	imageClientv1           *imageclientv1.ImageV1Client
+	operatorsClientv1       *operatorsclientv1.OperatorsV1Client
 	rbacClientv1            *rbacclientv1.RbacV1Client
 	securityClientv1        *securityclientv1.SecurityV1Client
-	operatorsClientv1       *operatorsclientv1.OperatorsV1Client
 }
 
 func newBuilder(config *rest.Config, m manifest.Manifest) Interface {
@@ -61,9 +61,9 @@ func newBuilder(config *rest.Config, m manifest.Manifest) Interface {
 		configClientv1:          configclientv1.NewForConfigOrDie(config),
 		coreClientv1:            coreclientv1.NewForConfigOrDie(withProtobuf(config)),
 		imageClientv1:           imageclientv1.NewForConfigOrDie(config),
+		operatorsClientv1:       operatorsclientv1.NewForConfigOrDie(config),
 		rbacClientv1:            rbacclientv1.NewForConfigOrDie(withProtobuf(config)),
 		securityClientv1:        securityclientv1.NewForConfigOrDie(config),
-		operatorsClientv1:       operatorsclientv1.NewForConfigOrDie(config),
 	}
 }
 
@@ -104,6 +104,18 @@ func (b *builder) Do(ctx context.Context) error {
 			return err
 		} else if !deleteReq {
 			if _, _, err := resourceapply.ApplySecurityContextConstraintsv1(ctx, b.securityClientv1, typedObject, reconcilingMode); err != nil {
+				return err
+			}
+		}
+	case *operatorsv1.OperatorGroup:
+		if b.modifier != nil {
+			b.modifier(typedObject)
+		}
+		if deleteReq, err := resourcedelete.DeleteOperatorGroupv1(ctx, b.operatorsClientv1, typedObject,
+			updatingMode); err != nil {
+			return err
+		} else if !deleteReq {
+			if _, _, err := resourceapply.ApplyOperatorGroupv1(ctx, b.operatorsClientv1, typedObject, reconcilingMode); err != nil {
 				return err
 			}
 		}
@@ -265,18 +277,6 @@ func (b *builder) Do(ctx context.Context) error {
 				return b.checkCustomResourceDefinitionHealth(ctx, actual)
 			}
 		}
-	case *operatorsv1.OperatorGroup:
-		if b.modifier != nil {
-			b.modifier(typedObject)
-		}
-		if deleteReq, err := resourcedelete.DeleteOperatorGroupv1(ctx, b.operatorsClientv1, typedObject,
-			updatingMode); err != nil {
-			return err
-		} else if !deleteReq {
-			if _, _, err := resourceapply.ApplyOperatorGroupv1(ctx, b.operatorsClientv1, typedObject, reconcilingMode); err != nil {
-				return err
-			}
-		}
 	default:
 		return fmt.Errorf("unrecognized manifest type: %T", obj)
 	}
@@ -295,11 +295,11 @@ func init() {
 	rm.RegisterGVK(corev1.SchemeGroupVersion.WithKind("Service"), newBuilder)
 	rm.RegisterGVK(corev1.SchemeGroupVersion.WithKind("ServiceAccount"), newBuilder)
 	rm.RegisterGVK(imagev1.SchemeGroupVersion.WithKind("ImageStream"), newBuilder)
+	rm.RegisterGVK(operatorsv1.SchemeGroupVersion.WithKind("OperatorGroup"), newBuilder)
 	rm.RegisterGVK(rbacv1.SchemeGroupVersion.WithKind("ClusterRole"), newBuilder)
 	rm.RegisterGVK(rbacv1.SchemeGroupVersion.WithKind("ClusterRoleBinding"), newBuilder)
 	rm.RegisterGVK(rbacv1.SchemeGroupVersion.WithKind("Role"), newBuilder)
 	rm.RegisterGVK(rbacv1.SchemeGroupVersion.WithKind("RoleBinding"), newBuilder)
 	rm.RegisterGVK(securityv1.SchemeGroupVersion.WithKind("SecurityContextConstraints"), newBuilder)
-	rm.RegisterGVK(operatorsv1.SchemeGroupVersion.WithKind("OperatorGroup"), newBuilder)
 	rm.AddToMap(Mapper)
 }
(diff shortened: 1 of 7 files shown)


lib/resourcebuilder/resourcebuilder.go
```diff
@@ -107,6 +111,18 @@ func (b *builder) Do(ctx context.Context) error {
 				return err
 			}
 		}
+	case *admissionregistrationv1.ValidatingWebhookConfiguration:
```
wking, line 114: There's [a Python builder][1] to help with these that I'd like to keep up to date (sorry I didn't notice you were touching these soon enough for the generator to help you create this content).  Something like:

```console
$ git --no-pager diff -U1 hack/
diff --git a/hack/generate-lib-resources.py b/hack/generate-lib-resources.py
index 2a45a2ac..a7fa16e8 100755
--- a/hack/generate-lib-resources.py
+++ b/hack/generate-lib-resources.py
@@ -284,2 +284,4 @@ if __name__ == '__main__':
         'github.com/openshift/api/security/v1': {'SecurityContextConstraints'},
+        'github.com/operator-framework/api/pkg/operators/v1': {'OperatorGroup'},
+        'k8s.io/api/admissionregistration/v1': {'ValidatingWebhookConfiguration'},
         'k8s.io/api/apps/v1': {'DaemonSet', 'Deployment'},
@@ -294,2 +296,4 @@ if __name__ == '__main__':
         'github.com/openshift/api/image/v1': {'package': 'github.com/openshift/client-go/image/clientset/versioned/typed/image/v1', 'type': 'ImageV1Client'},
+        'github.com/operator-framework/api/pkg/operators/v1': {'package': 'github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/typed/operators/v1', 'type': 'OperatorGroup'},
+        'k8s.io/api/admissionregistration/v1': {'package': 'k8s.io/client-go/kubernetes/typed/admissionregistration/v1', 'type': 'AdmissionregistrationV1Client'},
         'k8s.io/api/apps/v1': {'package': 'k8s.io/client-go/kubernetes/typed/apps/v1', 'type': 'AppsV1Client'},
```

with the OperatorGroup stuff to catch up with #862.

[1]: https://github.com/openshift/cluster-version-operator/blob/964e193242a5a89e9f52ae42278927051e491fee/hack/generate-lib-resources.py#L282-L301

lib/resourcemerge/admissionregistration.go
```diff
@@ -0,0 +1,72 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+// EnsureValidatingWebhookConfiguration ensures that the existing matches the required.
+// modified is set to true when existing had to be updated with required.
+func EnsureValidatingWebhookConfiguration(modified *bool, existing *admissionregv1.ValidatingWebhookConfiguration, required admissionregv1.ValidatingWebhookConfiguration) {
+	EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	ensureValidatingWebhookConfigurationDefaults(&required)
+	ensureValidatingWebhookConfigurationCaBundle(&required, *existing)
```
petr-muller, line 15: nit: let's keep the `existing, required` ordering in the parameters here, for consistency. it's error-prone when two arguments with identical type have different ordering in similar-purpose methods

also I'd prefer the shorter `existing, required *admissionregv1.ValidatingWebhookConfiguration` instead of repeating the full type everywhere

lib/resourcemerge/admissionregistration_test.go
```diff
@@ -0,0 +1,644 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/utils/pointer"
```
petr-muller, line 11: nit: no reason to keep these separate?

lib/resourcedelete/admissionregistration.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	admissionregclientv1 "k8s.io/client-go/kubernetes/typed/admissionregistration/v1"
+)
+
+// DeleteValidatingWebhookConfigurationv1 checks the given resource for a valid delete annotation. If found
+// it checks the status of a previously issued delete request. If delete has not been
+// requested and in UpdatingMode it will issue a delete request.
```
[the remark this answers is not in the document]
Davoska: You are right. I will align the comments :+1:

lib/resourcemerge/admissionregistration_test.go
Davoska: Nice catch! I must have deleted the line and forgotten about it.

lib/resourcemerge/admissionregistration.go
Davoska: > also I'd prefer the shorter existing, required *admissionregv1.ValidatingWebhookConfiguration instead of repeating the full type everywhere

Thanks for the tip!

---

> let's keep the existing, required ordering in the parameters here, for consistency. it's error-prone when two arguments with identical type have different ordering in similar-purpose methods

The  two arguments of the function **don't** have identical types because one is a pointer `*admissionregv1.ValidatingWebhookConfiguration`, and the second is a structure `admissionregv1.ValidatingWebhookConfiguration`.

I suppose the name of the function `ensureValidatingWebhookConfigurationCaBundle` causes the problem.

It contains the word `ensure` which is used in a little bit different context than the other `ensure<Resource>` functions that contain the argument `modified *bool` to signal if the `existing` has been modified.

Here, the CVO would set the fields of `required` so as to not trigger an unnecessary update. It's more similar to `ensure<Resource>Defaults` functions but here we need the information from `existing` (the CA bundle) thus I put it as a second argument (I guess I was prioritizing the importance of arguments in the ordering).

But I understand the problem with the consistency and the error-prone use. Do you think swapping the parameters, as you mentioned, would be enough? Now that I am thinking about it, the current name of the function could cause confusion like "_It's ensuring that the CABundle fields of webhooks are the same like the other functions do_"

lib/resourcebuilder/resourcebuilder.go
Davoska: Thanks, Trevor! I have added the change for the `OperatorGroup` type in the new commit bdac5ff491a5839370afb2cca793a1e767ab9b4b. And I have updated the `lib: Add ValidatingWebhookConfiguration defaulting` commit (now 58dceccc858962d6a64025d5ffcbe788a55e176c) to include the changes in the helper script, and I have committed the changes generated from the script.

lib/resourcemerge/admissionregistration_test.go
Davoska: Fixed the redundant line breaking in `lib/resourcemerge/admissionregistration_test.go` and `lib/resourceapply/admissionregistration.go`.

lib/resourcemerge/admissionregistration.go
petr-muller: > The two arguments of the function don't have identical types because one is a pointer 

doh, right, I overlooked that. That makes sense, because one is to be modified and one isn't. And the one being modified is always first, so it is consistent enough. The confusing part is that sometimes we modify `existing` and sometimes we modify `required`. The latter is defaulting. I agree that the names contribute to confusion, but they are not *obviously* wrong, so I'd perhaps just proceed.

If you feel like trying to improve it, then I  feel that doing all defaulting inside `ensureValidatingWebhookConfigurationDefaults` could be a step in a good direction. That method also iterates `required.Webhooks` so maybe it could do the `caBundle` determination, something like:

```go
func ensureValidatingWebhookConfigurationDefaults(required *admissionregv1.VWC, existing VWC) {
        var caBundle string
        if val, ok := <get annotation>; ok && val = "true" && len(existing.Webhooks) == 0 {
		caBundle = existing.Webhooks[0].ClientConfig.CABundle
	}
	for i := range required.Webhooks {
		ensureValidatingWebhookDefaults(&required.Webhooks[i])
		if caBundle != "" {
			required.Webhooks[i].ClientConfig.CABundle = caBundle
                }
	}
}
```

/shrug
But it's not that much better so I guess we may as well proceed with the PR. I'll /hold this if you want to explore some options, but feel free to unhold if you don't want to bother.

/hold
/lgtm
/approve

petr-muller (approved):
/meow

lib/resourcemerge/admissionregistration.go
Davoska: That's also a possibility that ensures the function wouldn't be misused. But then the function name would be semantically incorrect since setting the CA bundles is not technically defaulting. I guess the best option would be to introduce new functions like `initiliazeRequired<Resource>` that would set the default values and would set things like these for the CVO to ignore. But that's reformatting on another level that should ideally be done for all the resources, and it's not that necessary.

I'll keep the PR on `hold` for some time to see if anybody has any suggestions or opinions.

Any suggestions are welcome! :heart: 

If there are no suggestions after ~24 hours, I'll `unhold` the PR.

petr-muller: > should ideally be done for all the resources, and it's not that necessary.

Yeah that was my thinking as well: there are refactoring opportunities that would streamline the code and eliminate these issues but that's likely a separate effort and definitely not a must for a simple hotlooping bugfix.

PR #894, opened 2023-01-26
OCPBUGS-6097: CVO hotloops on ImageStream and logs the information incorrectly

Davoska:
This pull request will fix hotlooping on `ImageStream` and fixes the logging happening during updating `ImageStream` resources.

The hotlooping was caused by missing CVO-side defaulting of the `ImageStream` resources. In `4.12`, a new import value, `importMode`, has been added to the `importPolicy` parameter of image streams [1]. Previously, the CVO wasn't aware of its default value `Legacy` [1].

This pull request also improves the logging when handling the `ImageStream` resources. Previously, the logged diff wasn't helpful in  discovering the issue that caused the update of the resource (this change is similar to the change introduced for improving logging for deployments in commit https://github.com/openshift/cluster-version-operator/commit/c58385eea897b5af790e4dd36103e978e7e4c597). And previously, the logging contained a typo `Updating Namespace` instead of `Updating ImageStream`.

This pull request references the bug OCPBUGS-6097 [2].

[1] https://docs.openshift.com/container-platform/4.12/release_notes/ocp-4-12-release-notes.html#ocp-4-12-images
[2] https://issues.redhat.com/browse/OCPBUGS-6097

### DIFF
--- a/lib/resourcemerge/imagestream.go
+++ b/lib/resourcemerge/imagestream.go
@@ -30,6 +30,7 @@ func EnsureImagestreamv1(modified *bool, existing *imagev1.ImageStream, required
 }
 
 func ensureTagReferencev1(modified *bool, existing *imagev1.TagReference, required imagev1.TagReference) {
+	ensureTagReferencev1Defaults(&required)
 	if !equality.Semantic.DeepEqual(existing.ImportPolicy, required.ImportPolicy) {
 		*modified = true
 		existing.ImportPolicy = required.ImportPolicy
@@ -37,5 +38,14 @@ func ensureTagReferencev1(modified *bool, existing *imagev1.TagReference, requir
 	if !equality.Semantic.DeepEqual(existing.From, required.From) {
 		*modified = true
 		existing.From = required.From
+	}
+}
+
+func ensureTagReferencev1Defaults(required *imagev1.TagReference) {
+	if required.ImportPolicy.ImportMode == "" {
+		required.ImportPolicy.ImportMode = imagev1.ImportModeLegacy
+	}
+	if required.ReferencePolicy.Type == "" {
+		required.ReferencePolicy.Type = imagev1.SourceTagReferencePolicy
 	}
 }
--- a/lib/resourcemerge/imagestream_test.go
+++ b/lib/resourcemerge/imagestream_test.go
@@ -1,0 +1,696 @@
+package resourcemerge
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	imagev1 "github.com/openshift/api/image/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+func TestEnsureImageStreamv1_ImageStreamStatus(t *testing.T) {
+	tests := []struct {
+		name     string
+		existing imagev1.ImageStreamStatus
+		required imagev1.ImageStreamStatus
+	}{
+		{
+			name: "cvo should ignore same status field",
+			existing: imagev1.ImageStreamStatus{
+				DockerImageRepository:       "registry",
+				PublicDockerImageRepository: "public_registry",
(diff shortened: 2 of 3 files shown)


lib/resourcemerge/imagestream.go
```diff
@@ -30,6 +30,7 @@ func EnsureImagestreamv1(modified *bool, existing *imagev1.ImageStream, required
 }
 
 func ensureTagReferencev1(modified *bool, existing *imagev1.TagReference, required imagev1.TagReference) {
+	ensureTagReferencev1Defaults(&required)
```
petr-muller, line 33: Can I extort you into writing a unit test for `EnsureImagestreamv1` that would cover the basic cases (and the newly added one)? No unit tests so far are not your fault, but we can use the opportunity of changing the code to pay the debt.

Davoska: You got it :muscle:

Davoska:
/hold

Adding tests.

Davoska:
Still not ready. I have spotted a potential small issue.

lib/resourcemerge/imagestream.go
Davoska: Tests added :wrench:

lib/resourcemerge/imagestream_test.go
```diff
@@ -0,0 +1,818 @@
...
+		existing imagev1.ImageStream
+		required imagev1.ImageStream
+
+		expectedModified bool
+		expected         imagev1.ImageStream
+	}{
+		{
+			name: "same lookup policy",
+			existing: imagev1.ImageStream{
+				Spec: imagev1.ImageStreamSpec{
```
petr-muller, line 25: nit: looks like we could reduce a lot of visual clutter if our existing/required/actual data was just `ImageStreamSpec` (it seems that we always work with just this field) and we would build the `ImageStream` instances dynamically:

```go
existing := imagev1.ImageStream{Spec: tc.existing}
```

Alternatively, we could have testcases that show that `Status` part is not reconciled.

petr-muller (approved):
/hold

Definitely good enough for merge, thanks for the effort! I've left two comments but they are not required - feel free to ignore and `/unhold`

lib/resourcemerge/imagestream_test.go
Davoska: Both nitpicks seems like good suggestions! I am working on the changes :+1:

lib/resourcemerge/imagestream_test.go
```diff
@@ -0,0 +1,818 @@
...
+	"k8s.io/utils/pointer"
+)
+
+func TestEnsureImageStreamv1(t *testing.T) {
+	tests := []struct {
+		name     string
+		existing imagev1.ImageStream
+		required imagev1.ImageStream
+
+		expectedModified bool
```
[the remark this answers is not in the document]
Davoska: Yes, that is true if you meant "...we expect modified==**false**...". And that is a good suggestion to reduce a lot of duplicated code :+1:

lib/resourcemerge/imagestream_test.go
Davoska: It does bring additional logic to the testing but I think the removal of the visual clutter is worth it. And the developer can focus only on the `ImageStreamSpec` when testing CVO behaviour on `ImageStreamSpec`. And `ImageStreamStatus` respectively.

---

> Alternatively, we could have testcases that show that Status part is not reconciled.

Yes, this test should be present! I have broken up the previous function `TestEnsureImageStreamv1` into `TestEnsureImageStreamv1_ImageStreamStatus` and `TestEnsureImageStreamv1_ImageStreamSpec` while keeping your suggestion to reduce the visual clutter in mind.

Let me know what you think of it!

Davoska:
/cc LalatenduMohanty jiajliu
/retest

Oops, I didn't want to remove anyone's review request. Sorry about the noise.

evakhoni:
/uncc jiajliu
/cc evakhoni
Jia is on leave

petr-muller (approved):
Thanks for putting the effort into the tests :+1:

petr-muller:
/retest
TRT-860 was biting us but should be good now

petr-muller:

We only fail on disruption tests, and it is unlikely that CVO would ever caused that

PR #881, opened 2023-01-06
Revert "pkg/cvo/updatepayload.go: timeout payload retrieval"

petr-muller:
Reverts openshift/cluster-version-operator#846

Apparently introduced a regression that blocks some upgrades, opening to allow `/payload` testing of `periodic-ci-openshift-release-master-nightly-4.13-e2e-metal-ipi-upgrade-ovn-ipv6`. If it fixes the job we should consider merging the revert and work on the fix with clean main branch.

PR #846, opened 2022-10-06
Bug 2090680: pkg/cvo/updatepayload.go: timeout payload retrieval

jottofar:
When we separated payload load from payload apply (https://github.com/openshift/cluster-version-operator/pull/683) the context used for the retrieval changed as well. It went from one that was constrained by syncTimeout (2 -4 minutes) [1] to being the unconstrained shutdownContext [2]. However if "force" is specified we explicitly set a 2 minute timeout in RetrievePayload. This commit creates a new context with a reasonable timeout for RetrievePayload regardless of "force".

[1] https://github.com/openshift/cluster-version-operator/blob/57ffa7c610fb92ef4ccd9e9c49e75915e86e9296/pkg/cvo/sync_worker.go#L605
[2] https://github.com/openshift/cluster-version-operator/blob/57ffa7c610fb92ef4ccd9e9c49e75915e86e9296/pkg/cvo/cvo.go#L413

PR #896, opened 2023-02-03
Bug 2090680: RetrievePayload: Improve timeouts and cover behavior with tests

petr-muller:
The `RetrievePayload` performs two operations: verification and download. Both can take a non-trivial amount of time to terminate, up to "hanging" where CVO needs to abort the operation. The verification result can be ignored when upgrade is forced. The CVO calls `RetrievePayload` with a context that does not set a deadline, so `RetrievePayload` previously set its own internal deadline, common for both operations. This led to a suboptimal behavior on forced upgrades, where "hanging" verification could eat the whole timeout budget, got cancelled but its result was ignored (because of force). The code tried to proceed with download but that immediately aborts because of the expired context.

Improve timeouts in `RetrievePayload` for both input context states: with and without deadline. If the input context sets a deadline, it is respected. If it does not, the default, separate deadlines are applied for both operations. In both cases, the code makes sure the hanging verification never spends the whole budget. When verification terminates fast, the rest of its allotted time is given to the download operation.

To keep context about the change, I have kept separate commits where the intermediate ones show the newly-added tests failing for both the original bug and the regression that caused us to revert in https://github.com/openshift/cluster-version-operator/pull/881.

The tests are a little clunky, being time-based and all. The happy case takes ~4s to execute (the parallelism helps), failures should be capped to ~5s. In theory the tests could be flakey (there's no strong guarantee that a `select` will always see a after-4s channel operation before the after-5s one) but I'm not seeing any problems with several tries of high-count `--count 30` runs :
```
$ go test --count 30 ./pkg/cvo/... -run TestPayloadRetrieverRetrievePayload
ok  	github.com/openshift/cluster-version-operator/pkg/cvo	120.096s
```

---

Original, pre-https://github.com/openshift/cluster-version-operator/pull/846 CVO fails the newly added test for RHBZ#2090680:
```
 --- FAIL: TestPayloadRetrieverRetrievePayload (0.00s)
    --- FAIL: TestPayloadRetrieverRetrievePayload/when_sha_digest_pullspec_image_passes_and_download_hangs_then_it_is_terminated_and_returns_error_(RHBZ#2090680) (5.00s)
        updatepayload_test.go:105: downloader: test backstop hit (expected cancel via ctx)
        updatepayload_test.go:339: Returned error differs from expected:
              interface{}(
            - 	e"Unable to download and prepare the update: download was canceled",
            + 	e"Unable to download and prepare the update: downloader: test backstop hit (expected cancel via ctx)",
              ) 
```
---
Reinstated https://github.com/openshift/cluster-version-operator/pull/846 code fails two new tests (simulating the scenario for which we needed to revert it in https://github.com/openshift/cluster-version-operator/pull/881):

```
 --- FAIL: TestPayloadRetrieverRetrievePayload (0.00s)
    --- FAIL: TestPayloadRetrieverRetrievePayload/when_sha_digest_pullspec_image_is_timing_out_verification_with_unlimited_context_and_update_is_forced_then_verification_times_out_promptly_and_retrieval_proceeds_but_download_fails_then_return_error (2.00s)
        updatepayload_test.go:118: downloader: unexpected ctx cancel (expected download to finish)
        updatepayload_test.go:339: Returned error differs from expected:
              interface{}(
            - 	e"Unable to download and prepare the update: fails to download",
            + 	e"Unable to download and prepare the update: downloader: unexpected ctx cancel (expected download to finish)",
              )
    --- FAIL: TestPayloadRetrieverRetrievePayload/when_sha_digest_pullspec_image_fails_to_verify_until_timeout_but_is_forced_then_it_allows_enough_time_for_download_and_it_returns_successfully (2.00s)
        updatepayload_test.go:118: downloader: unexpected ctx cancel (expected download to finish)
        updatepayload_test.go:339: Returned error differs from expected:
              interface{}(
            + 	e"Unable to download and prepare the update: downloader: unexpected ctx cancel (expected download to finish)",
              ) 
```
---
Last commit implements the separate timeouts and makes all tests pass.

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -38,20 +38,24 @@ func (optr *Operator) defaultPayloadDir() string {
 
 func (optr *Operator) defaultPayloadRetriever() PayloadRetriever {
 	return &payloadRetriever{
-		kubeClient:   optr.kubeClient,
-		operatorName: optr.name,
-		releaseImage: optr.release.Image,
-		namespace:    optr.namespace,
-		nodeName:     optr.nodename,
-		payloadDir:   optr.defaultPayloadDir(),
-		workingDir:   targetUpdatePayloadsDir,
-		verifier:     optr.verifier,
+		kubeClient:           optr.kubeClient,
+		operatorName:         optr.name,
+		releaseImage:         optr.release.Image,
+		namespace:            optr.namespace,
+		nodeName:             optr.nodename,
+		payloadDir:           optr.defaultPayloadDir(),
+		workingDir:           targetUpdatePayloadsDir,
+		verifier:             optr.verifier,
+		verifyTimeoutOnForce: 2 * time.Minute,
+		downloadTimeout:      2 * time.Minute,
 	}
 }
 
 const (
 	targetUpdatePayloadsDir = "/etc/cvo/updatepayloads"
 )
+
+type downloadFunc func(context.Context, configv1.Update) (string, error)
 
 type payloadRetriever struct {
 	// releaseImage and payloadDir are the default payload identifiers - updates that point
@@ -67,7 +71,11 @@ type payloadRetriever struct {
 	operatorName string
 
 	// verifier guards against invalid remote data being accessed
-	verifier verify.Interface
+	verifier             verify.Interface
+	verifyTimeoutOnForce time.Duration
+
+	downloader      downloadFunc
+	downloadTimeout time.Duration
 }
 
 func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.Update) (PayloadInfo, error) {
@@ -94,7 +102,7 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
 	// if 'force' specified, ensure call to verify payload signature times out well before parent context
 	// to allow time to perform forced update
 	if update.Force {
-		timeout := time.Minute * 2
+		timeout := r.verifyTimeoutOnForce
 		if deadline, deadlineSet := ctx.Deadline(); deadlineSet {
 			timeout = time.Until(deadline) / 2
 		}
@@ -120,9 +128,13 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
 		info.Verified = true
 	}
 
+	if r.downloader == nil {
+		r.downloader = r.targetUpdatePayloadDir
+	}
+
 	// download the payload to the directory
 	var err error
-	info.Directory, err = r.targetUpdatePayloadDir(ctx, update)
+	info.Directory, err = r.downloader(ctx, update)
 	if err != nil {
 		return PayloadInfo{}, &payload.UpdateError{
 			Reason:  "UpdatePayloadRetrievalFailed",
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -90,19 +90,29 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
 		return PayloadInfo{}, fmt.Errorf("no payload image has been specified and the contents of the payload cannot be retrieved")
 	}
 
-	var info PayloadInfo
-
 	// verify the provided payload
 	var releaseDigest string
 	if index := strings.LastIndex(update.Image, "@"); index != -1 {
 		releaseDigest = update.Image[index+1:]
 	}
 
-	// set up a new context with reasonable timeout for signature and payload retrieval
-	retrieveCtx, cancel := context.WithTimeout(ctx, r.verifyTimeoutOnForce+r.downloadTimeout)
-	defer cancel()
+	var downloadCtx context.Context
+	var downloadCtxCancel context.CancelFunc
+	var verifyTimeout time.Duration
 
-	if err := r.verifier.Verify(retrieveCtx, releaseDigest); err != nil {
+	if deadline, ok := ctx.Deadline(); ok {
+		verifyTimeout = time.Until(deadline) / 2
+		downloadCtx = ctx
+	} else {
+		verifyTimeout = r.verifyTimeoutOnForce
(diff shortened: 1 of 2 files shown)


petr-muller:
```
[sig-api-machinery] disruption/kube-api connection/new should be available throughout the test 
[sig-api-machinery] disruption/kube-api connection/reused should be available throughout the test
[sig-api-machinery] disruption/openshift-api connection/new should be available throughout the test
[sig-api-machinery] disruption/openshift-api connection/reused should be available throughout the test
```
Only disruption tests, unlikely to be caused by CVO

petr-muller:
```
: [sig-arch][Feature:ClusterUpgrade] Cluster should remain functional during upgrade [Disruptive] [Serial]
: [sig-network] pods should successfully create sandboxes by other
: [sig-api-machinery] disruption/kube-api connection/reused should be available throughout the test
: [sig-api-machinery] disruption/oauth-api connection/reused should be available throughout the test
: [sig-api-machinery] disruption/openshift-api connection/new should be available throughout the test
: [sig-api-machinery] disruption/openshift-api connection/reused should be available throughout the test
upgrade: [sig-cluster-lifecycle] ClusterOperators are available and not degraded after upgrade
```
Nothing pointing at CVO, one more try

petr-muller:
/hold

We want to investigate findings from https://bugzilla.redhat.com/show_bug.cgi?id=2090680#c22

pkg/cvo/updatepayload.go
```diff
@@ -38,21 +38,25 @@ func (optr *Operator) defaultPayloadDir() string {
...
+		verifyTimeoutOnForce: 2 * time.Minute,
+		downloadTimeout:      2 * time.Minute,
 	}
 }
 
 const (
 	targetUpdatePayloadsDir = "/etc/cvo/updatepayloads"
 )
 
+type downloadFunc func(context.Context, configv1.Update) (string, error)
```
wking, line 58: Possibly grow some godocs here, like:

> downloadFunc downloads the requested update, and returns either a path on the local filesystem containing extracted manifests or an error.

or some such, so folks don't wonder "what's that `string` return about?".

pkg/cvo/updatepayload.go
```diff
@@ -82,28 +90,28 @@ func (r *payloadRetriever) RetrievePayload(ctx context.Context, update configv1.
...
+	var downloadCtx context.Context
+	var downloadCtxCancel context.CancelFunc
+	var verifyTimeout time.Duration
+
+	if deadline, ok := ctx.Deadline(); ok {
+		verifyTimeout = time.Until(deadline) / 2
+		downloadCtx = ctx
+	} else {
+		verifyTimeout = r.verifyTimeoutOnForce
+		downloadCtx, downloadCtxCancel = context.WithTimeout(ctx, r.verifyTimeoutOnForce+r.downloadTimeout)
```
wking, line 108: this looks like "if we don't use up our entire block of allocated time on verification, include any saved time in the allowed download window".  But I don't understand why having quicker verification would make us more generous in accepting slower downloads.  On the other hand, I expect that almost all clusters with faster verification will also have faster downloads, so there may be no clusters in the "fast verification, slow download" bucket to worry about.

pkg/cvo/updatepayload_test.go
```diff
@@ -0,0 +1,348 @@
...
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/verify/store"
+
+	//nolint:staticcheck // verify,Verifier from openshift/library-go uses a type from this deprecated package (needs to be addressed there)
```
wking, line 14: do we have a Jira ticket tracking this?

pkg/cvo/updatepayload.go
petr-muller: :+1: good point

pkg/cvo/updatepayload_test.go
petr-muller: good catch, I'll file one

pkg/cvo/updatepayload.go
petr-muller: I'd ask the opposite, why shouldn't we be generous? Networks suck :) If we tolerate the whole `RetrievePayload` to take 4 minutes then we can occasionally soak a weird corner case this way and succeed.

But I'm not too strong on this, I can remove it if you worry about it making the code more complicated.

pkg/cvo/updatepayload.go
petr-muller: I've sprinkled several more godocs when at it

pkg/cvo/updatepayload.go
petr-muller: When adding the godocs I have slightly refactored the code to better correspond to `RetrievePayload` carrying the contract: set just a single timeout on the whole method, and consider the division between verification/download an internal detail. 

It was hard to describe the semantics of `downloadTimeout` (it this plus sometimes a bit more but always bounded) and when things are hard to document it usually means something smells.

pkg/cvo/updatepayload_test.go
petr-muller: https://issues.redhat.com/browse/OCPBUGS-7479

petr-muller:
oh nice, all-pass on the first try ❤️

wking (approved):
/lgtm

petr-muller:
I have investigated the concerns from https://bugzilla.redhat.com/show_bug.cgi?id=2090680#c22 and filed https://issues.redhat.com/browse/OCPBUGS-7714 for the problem. That behavior is just "uncovered" by this bugfix, not caused by it. The fix actually improves the state for the clusters that hit it - I believe that previously the manifest reconciliation was completely broken while the unbounded retrieval was timing out for hours, we just did not notice.

petr-muller:

DNS operator issue, unrelated to this PR

PR #898, opened 2023-02-14
pkg/cvo/sync_worker: Fix "The sync worker already has a pending notification" formatting

wking:
Cleaning up a log line I'd fumbled in 9222fa9a66 (#818).  Changes:

* Remove redundant `so do not notify about`, because `no need to inform about` covers that with more clarity (we're not informing, and even if we did, it would be useless, because there's already a pending notification in place).
* Use `Infof` to get the `%s` formatted in, avoiding:

        ...no need to inform about: %sCluster operator...

   `


LalatenduMohanty (approved):
/lgtm

PR #900, opened 2023-02-16
pkg/cvo: Set 'controller' in our ownerReferences

wking:
We've been setting `ownerReferences` on resources we create or manage since 4dbb32741a (#617).  But when we are actively controlling a resource, we should be setting `controller` to conform with:

```console
$ grep -A100 'type OwnerReference struct' vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go | grep -B3 Controller
          UID types.UID `json:"uid" protobuf:"bytes,4,opt,name=uid,casttype=k8s.io/apimachinery/pkg/types.UID"`
          // If true, this reference points to the managing controller.
          // +optional
          Controller *bool `json:"controller,omitempty" protobuf:"varint,6,opt,name=controller"`
```

With this commit, I'm moving the ClusterVersion owner to `controller: true` for resources we could possibly be comfortable replacing from scratch.  Although things like "I've deleted the `openshift-etcd` namespace" might cause some disruption today, possibly including bricking the cluster, and are not recommended.  They could still theoretically be hardened in the future to support clean replacement.

I considered using `controller: true` for `create-only` resources too, because:

* We'll still re-create `create-only` resources, and
* We don't include the ClusterVersion owner at all when there is a competing owner.

But `create-only` resources are create-only because we expect the admin or other actor to manage them after creation, so the cluster-version operator cannot commit to replacing all of that possible post-creation configuration state.  Actors who try and use `controller: true` to judge "I may be able to delete this and let the controller replace it" should not try to delete `create-only` resources unless they understand the resource well enough to replace any desired configuration state on their own.


LalatenduMohanty (approved):
/lgtm

wking:
[upgrade-into-change][1] looks good:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/900/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1626379647581163520/artifacts/e2e-agnostic-upgrade-into-change/gather-extra/artifacts/namespaces.json | jq -r '.items[].metadata | select(.name == "openshift-cluster-version").ownerReferences'
[
  {
    "apiVersion": "config.openshift.io/v1",
    "controller": true,
    "kind": "ClusterVersion",
    "name": "version",
    "uid": "aa25e5a7-28b5-4bcf-8bd9-f1b14272cf4f"
  }
]
```

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/900/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1626379647581163520

wking:
And here's a create-only resource, also looking good:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/900/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-upgrade-into-change/1626379647581163520/artifacts/e2e-agnostic-upgrade-into-change/gather-must-gather/artifacts/must-gather.tar | tar xOz registry-build05-ci-openshift-org-ci-op-mw5yg4py-stable-sha256-0e9d2765b097e16288212a4d64d74441a84e57d094b322d39b2546bf9ebb64a1/cluster-scoped-resources/config.openshift.io/authentications/cluster.yaml | yaml2json | jq '.metadata | {annotations, ownerReferences}'
{
  "annotations": {
    "include.release.openshift.io/ibm-cloud-managed": "true",
    "include.release.openshift.io/self-managed-high-availability": "true",
    "include.release.openshift.io/single-node-developer": "true",
    "release.openshift.io/create-only": "true"
  },
  "ownerReferences": [
    {
      "apiVersion": "config.openshift.io/v1",
      "kind": "ClusterVersion",
      "name": "version",
      "uid": "aa25e5a7-28b5-4bcf-8bd9-f1b14272cf4f"
    }
  ]
}
```

PR #902, opened 2023-02-18
pkg/cvo: code cleanups

petr-muller:
Fix a bunch of smells I encountered while working on https://github.com/openshift/cluster-version-operator/pull/896. Mostly simplify method signatures, but also remove two unused methods.

PR #904, opened 2023-02-23
OCPBUGS-7419: Trigger new sync round on ClusterOperator Available changes

jeckersb:


pkg/cvo/cvo.go
```diff
@@ -499,28 +499,54 @@ func (optr *Operator) clusterOperatorEventHandler() cache.ResourceEventHandler {
...
 			if optr.configSync != nil && oldVersion != newVersion {
 				msg := fmt.Sprintf("Cluster operator %s changed versions[name=%q] from %q to %q", newStruct.ObjectMeta.Name, versionName, oldVersion, newVersion)
 				optr.configSync.NotifyAboutManagedResourceActivity(new, msg)
+				return
+			}
+
+			oldAvailable := clusterOperatorConditionStatus(oldStruct, configv1.OperatorAvailable)
+			newAvailable := clusterOperatorConditionStatus(newStruct, configv1.OperatorAvailable)
+			if optr.configSync != nil && oldAvailable != newAvailable {
+				msg := fmt.Sprintf("Cluster operator %s changed Available from %q to %q", newStruct.ObjectMeta.Name, oldAvailable, newAvailable)
```
wking, line 516: nit: you could use `%s` and `configv1.OperatorAvailable` here to avoid hard-coding the `type` string, although it seems unlikely that it will ever diverge from `Available`.

pkg/cvo/cvo.go
```diff
@@ -499,28 +499,54 @@ func (optr *Operator) clusterOperatorEventHandler() cache.ResourceEventHandler {
...
+			newVersion := clusterOperatorVersion(newStruct, versionName)
 			if optr.configSync != nil && oldVersion != newVersion {
 				msg := fmt.Sprintf("Cluster operator %s changed versions[name=%q] from %q to %q", newStruct.ObjectMeta.Name, versionName, oldVersion, newVersion)
 				optr.configSync.NotifyAboutManagedResourceActivity(new, msg)
+				return
+			}
+
+			oldAvailable := clusterOperatorConditionStatus(oldStruct, configv1.OperatorAvailable)
+			newAvailable := clusterOperatorConditionStatus(newStruct, configv1.OperatorAvailable)
+			if optr.configSync != nil && oldAvailable != newAvailable {
```
wking, line 515: We might want to do this same sort of thing for `Degraded` too while we're at it.

pkg/cvo/cvo.go
jeckersb: Done

petr-muller (approved):
LGTM
I'll need to adjust https://github.com/openshift/cluster-version-operator/pull/902 after this merges but /shrug

Just isolated disruption test failures -> overriding

PR #910, opened 2023-03-08
OCPBUGS-9070: Fix hotlooping on Cronjob resources

Davoska:
Currently the CVO hotloops on `CronJob` resources.

The issue was caused by the missing defaulting of these resources.

To fix this issue I have added the handling for `CronJob` resources so they are not reconciled as `generic` resources. This has then allowed me to introduce the defaulting.

To fix this issue I have also updated the applying for `Job` resources because its logic can be then reused as part of the fix for the hotlooping on `Cronjob` resources.  While updating the applying for `Job` I have added missing defaulting,  and I have added reconciling of fields that were added later in the development, and the CVO was not reconciling them.

This pull request references https://issues.redhat.com/browse/OCPBUGS-9070

### DIFF
--- a/lib/resourceapply/batch.go
+++ b/lib/resourceapply/batch.go
@@ -47,3 +47,38 @@ func ApplyJobv1(ctx context.Context, client batchclientv1.JobsGetter, required *
 	actual, err := client.Jobs(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
 	return actual, true, err
 }
+
+// ApplyCronJobv1 applies the required CronJob to the cluster.
+func ApplyCronJobv1(ctx context.Context, client batchclientv1.CronJobsGetter, required *batchv1.CronJob, reconciling bool) (*batchv1.CronJob, bool, error) {
+	existing, err := client.CronJobs(required.Namespace).Get(ctx, required.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.V(2).Infof("CronJob %s/%s not found, creating", required.Namespace, required.Name)
+		actual, err := client.CronJobs(required.Namespace).Create(ctx, required, metav1.CreateOptions{})
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+	// if we only create this resource, we have no need to continue further
+	if IsCreateOnly(required) {
+		return nil, false, nil
+	}
+
+	var original batchv1.CronJob
+	existing.DeepCopyInto(&original)
+	modified := pointer.BoolPtr(false)
+	resourcemerge.EnsureCronJob(modified, existing, *required)
+	if !*modified {
+		return existing, false, nil
+	}
+	if reconciling {
+		if diff := cmp.Diff(&original, existing); diff != "" {
+			klog.V(2).Infof("Updating CronJob %s/%s due to diff: %v", required.Namespace, required.Name, diff)
+		} else {
+			klog.V(2).Infof("Updating CronJob %s/%s with empty diff: possible hotloop after wrong comparison", required.Namespace, required.Name)
+		}
+	}
+
+	actual, err := client.CronJobs(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
+	return actual, true, err
+}
--- a/lib/resourceapply/batch.go
+++ b/lib/resourceapply/batch.go
@@ -29,14 +29,19 @@ func ApplyJobv1(ctx context.Context, client batchclientv1.JobsGetter, required *
 		return nil, false, nil
 	}
 
+	var original batchv1.Job
+	existing.DeepCopyInto(&original)
 	modified := pointer.BoolPtr(false)
 	resourcemerge.EnsureJob(modified, existing, *required)
 	if !*modified {
 		return existing, false, nil
 	}
-
 	if reconciling {
-		klog.V(2).Infof("Updating Job %s/%s due to diff: %v", required.Namespace, required.Name, cmp.Diff(existing, required))
+		if diff := cmp.Diff(&original, existing); diff != "" {
+			klog.V(2).Infof("Updating Job %s/%s due to diff: %v", required.Namespace, required.Name, diff)
+		} else {
+			klog.V(2).Infof("Updating Job %s/%s with empty diff: possible hotloop after wrong comparison", required.Namespace, required.Name)
+		}
 	}
 
 	actual, err := client.Jobs(required.Namespace).Update(ctx, existing, metav1.UpdateOptions{})
--- a/lib/resourcebuilder/resourcebuilder.go
+++ b/lib/resourcebuilder/resourcebuilder.go
@@ -169,6 +169,18 @@ func (b *builder) Do(ctx context.Context) error {
 				return b.checkDeploymentHealth(ctx, actual)
 			}
 		}
+	case *batchv1.CronJob:
+		if b.modifier != nil {
+			b.modifier(typedObject)
+		}
+		if deleteReq, err := resourcedelete.DeleteCronJobv1(ctx, b.batchClientv1, typedObject,
+			updatingMode); err != nil {
+			return err
+		} else if !deleteReq {
+			if _, _, err := resourceapply.ApplyCronJobv1(ctx, b.batchClientv1, typedObject, reconcilingMode); err != nil {
+				return err
+			}
+		}
 	case *batchv1.Job:
 		if b.modifier != nil {
 			b.modifier(typedObject)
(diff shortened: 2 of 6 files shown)


Davoska:
/hold
I want to verify in the CI logs that the fix is working as expected.

Davoska:
Regarding `bugzilla/invalid-bug`:

Grr, I don't see the `4.14.0` version in the list. I will look into it later.

lib/resourcebuilder/resourcebuilder.go
```diff
@@ -169,6 +169,18 @@ func (b *builder) Do(ctx context.Context) error {
...
 		}
+	case *batchv1.CronJob:
+		if b.modifier != nil {
+			b.modifier(typedObject)
+		}
+		if deleteReq, err := resourcedelete.DeleteCronJobv1(ctx, b.batchClientv1, typedObject,
+			updatingMode); err != nil {
+			return err
+		} else if !deleteReq {
+			if _, _, err := resourceapply.ApplyCronJobv1(ctx, b.batchClientv1, typedObject, reconcilingMode); err != nil {
```
wking, line 180: One wrinkle with adding official CronJob support is deciding if we want a `return b.checkCronJobHealth(ctx, actual)` like we see below for Jobs.  If we do want that, we can have a second discussion about whether we want it in this pull request, or if we'd rather file a ticket to track it as future work.

lib/resourcemerge/batch.go
```diff
@@ -2,21 +2,97 @@ package resourcemerge
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/utils/pointer"
 )
 
 // EnsureJob ensures that the existing matches the required.
 // modified is set to true when existing had to be updated with required.
 func EnsureJob(modified *bool, existing *batchv1.Job, required batchv1.Job) {
-
 	if required.Spec.Selector != nil {
 		panic("spec.selector is not supported in Job manifests")
```
wking, line 13: Why doesn't this move down into `ensureJobSpec`?

Davoska: Nice catch, I didn't notice that. I will move it to the `ensureJobSpec`.

lib/resourcebuilder/resourcebuilder.go
petr-muller: Feels like we *should* have it but not necessarily in this PR, just making a note/card should be fine, IMO

lib/resourceapply/batch.go
```diff
@@ -29,16 +29,56 @@ func ApplyJobv1(ctx context.Context, client batchclientv1.JobsGetter, required *
...
+	if !*modified {
+		return existing, false, nil
+	}
+	if reconciling {
+		if diff := cmp.Diff(&original, existing); diff != "" {
+			klog.V(2).Infof("Updating CronJob %s/%s due to diff: %v", required.Namespace, required.Name, diff)
+		} else {
+			klog.V(2).Infof("Updating CronJob %s/%s with empty diff: possible hotloop after wrong comparison", required.Namespace, required.Name)
+		}
+	}
```
petr-muller, line 80: Reading the copy of this code for fifth time makes me think whether we can somehow DRY these `Apply*` methods out :/ Not an actual suggestion for you to fix, just thinking loud.

lib/resourcedelete/batch.go
```diff
@@ -37,3 +37,32 @@ func DeleteJobv1(ctx context.Context, client batchclientv1.JobsGetter, required
...
+// DeleteCronJobv1 checks the given resource for a valid delete annotation. If found
+// it checks the status of a previously issued delete request. If delete has not been
+// requested and in UpdatingMode it will issue a delete request.
+func DeleteCronJobv1(ctx context.Context, client batchclientv1.CronJobsGetter, required *batchv1.CronJob,
+	updateMode bool) (bool, error) {
+
+	if delAnnoFound, err := ValidDeleteAnnotation(required.Annotations); !delAnnoFound || err != nil {
+		return delAnnoFound, err
+	}
+	existing, err := client.CronJobs(required.Namespace).Get(ctx, required.Name, metav1.GetOptions{})
```
petr-muller, line 50: Hmmm is it right to entirely ignore errors here? We immediately overwrite it on L56 and we end up using `existing` on L62. Either we should not be ignoring the error, or we should do so explicitly with a comment why (`existing, _ := ...`)

lib/resourcedelete/batch.go
```diff
@@ -37,3 +37,32 @@ func DeleteJobv1(ctx context.Context, client batchclientv1.JobsGetter, required
...
+	resource := Resource{
+		Kind:      "cronjob",
+		Namespace: required.Namespace,
+		Name:      required.Name,
+	}
+	if deleteRequested, err := GetDeleteProgress(resource, err); err == nil {
```
petr-muller, line 56: It'd read a bit better if the condition was inverse:

```go
deleteRequested, err := GetDeleteProgress(resource, err)
if err != nil {
   return true, fmt.Errorf("Error running delete for %s, err=%v", resource, err)
}
if ... {
  if ... {
    return ...
  }
  SetDeleteRequested(...
}
return true, nil
```

lib/resourcebuilder/resourcebuilder.go
Davoska: Good point. I will create a card regarding this issue so we can address it in the future. This pull request already improves the CVO's CronJob support and fixes the addressed issue.

Personally, at first glance, I am not sure how would the `checkCronJobHealth` method look like. And thus it would require additional effort. But as Petr said it does feel like we should have it.

lib/resourceapply/batch.go
Davoska: Yes, that would be great :( I will create a card so we can keep track of these potential places to be refactored. This is a good idea for a shift-week activity or for anyone interested in helping to develop the CVO.

lib/resourcedelete/batch.go
Davoska: It took me few a seconds but the initial `err` does get passed to the function `GetDeleteProgress()` on the L56 and thus is not immediately overwritten:

```go
	if deleteRequested, err := GetDeleteProgress(resource, err); err == nil {
```

And there it is used quite a few times:

https://github.com/openshift/cluster-version-operator/blob/master/lib/resourcedelete/helper.go#L100-L132

lib/resourcedelete/batch.go
Davoska: Me too! We would return early rather than nest deeper.

I'll change the condition in this current PR, and in the future, we can refactor the other conditions in the `resourcedelete` package (or using the DRY principle we will leave reuse this condition).

Davoska:
/hold
Oops, wrong push.

Davoska:
/unhold

I have addressed the comments in https://github.com/openshift/cluster-version-operator/compare/659889f559b396b5ecb7c37d48aac98557ff0563..f665f3c39136b7e55ee92cb741f9b2d0b4e18a6e.

However, I had to rebase the PR in the latest push.

lib/resourcedelete/batch.go
Davoska: Sorry, I forgot about your new comment when pushing the new changes. However, the new code structure simplifies the logic in my opinion. 

https://github.com/Davoska/cluster-version-operator/blob/bug-2039228-cvo-hotloops-on-cronjob/lib/resourcedelete/batch.go#L43-L66

```go
	if delAnnoFound, err := ValidDeleteAnnotation(required.Annotations); !delAnnoFound || err != nil {
		return delAnnoFound, err
	}
	existing, err := client.CronJobs(required.Namespace).Get(ctx, required.Name, metav1.GetOptions{})
	resource := Resource{
		Kind:      "cronjob",
		Namespace: required.Namespace,
		Name:      required.Name,
	}
	deleteRequested, err := GetDeleteProgress(resource, err)
	if err != nil {
		return true, fmt.Errorf("Error running delete for %s, err=%v", resource, err)
	}
	if !deleteRequested && updateMode {
		if err := client.CronJobs(required.Namespace).Delete(ctx, required.Name, metav1.DeleteOptions{}); err != nil {
			return true, fmt.Errorf("Delete request for %s failed, err=%v", resource, err)
		}
		SetDeleteRequested(existing, resource)
	}
```

Now it's clearer than before, and something like `getErr` is no longer needed IMO. However, if you feel like it should be there even now, I am happy to add it.

petr-muller: It's fine, no more modifications needed, thanks!

Davoska:
/hold

Non-critical merge

petr-muller (approved):
:shipit: :tada:

/retest

evakhoni:
@Davoska @petr-muller is this one ready for pre-merge verification? any major changes expected?

Davoska:
Hello @evakhoni :wave: 

The PR is ready. No new changes are expected.

Davoska:
Oh no, it seems I have incorrectly broken up the commits the last week. Now, the refactoring commit https://github.com/openshift/cluster-version-operator/pull/910/commits/2f4e5508c2b103e5381b79d72b1cf128f5a6d54b contains changes that were inside a now non-existing commit that explained the added defaulting for the `Job` resources. Working on it :wrench:

Davoska:
Sorry everyone, for the noise! I have fixed the commits, and in the most recent force-push comparison, we can see that no new changes were added to the code.

petr-muller (approved):

evakhoni:
/hold
appears to be an issue with `concurrencyPolicy` default
commented in [Jira](https://issues.redhat.com/browse/OCPBUGS-9070?focusedId=22132293&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22132293)

lib/resourcebuilder/resourcebuilder.go
Davoska: I have created [OTA-957: CVO should check the health of CronJob resources](https://issues.redhat.com/browse/OTA-957).

evakhoni:
follow up the bug ticket, since results are expected, lets consider this pre-merge verified
/unhold
/label qe-approved

petr-muller:
Looks safe to override to me

Davoska:

The `ci/prow/e2e-agnostic-ovn-upgrade-into-change` job has already passed two times for this code change.

The currently failing run's cause doesn't seem to be related to the PR.

PR #920, opened 2023-03-29
Update dnsPolicy to allow consistent resolution of the internal LB

deads2k:
The kubelet consistently resolves the name.  This change allows the CVO to use the kubelet's DNS configuration.

### DIFF
--- a/pkg/cincinnati/cincinnati.go
+++ b/pkg/cincinnati/cincinnati.go
@@ -37,14 +37,17 @@ type Client struct {
 	// requests.  If empty, the User-Agent header will not be
 	// populated.
 	userAgent string
+
+	conditionRegistry clusterconditions.ConditionRegistry
 }
 
 // NewClient creates a new Cincinnati client with the given client identifier.
-func NewClient(id uuid.UUID, transport *http.Transport, userAgent string) Client {
+func NewClient(id uuid.UUID, transport *http.Transport, userAgent string, conditionRegistry clusterconditions.ConditionRegistry) Client {
 	return Client{
-		id:        id,
-		transport: transport,
-		userAgent: userAgent,
+		id:                id,
+		transport:         transport,
+		userAgent:         userAgent,
+		conditionRegistry: conditionRegistry,
 	}
 }
 
@@ -246,7 +249,7 @@ func (c Client) GetUpdates(ctx context.Context, uri *url.URL, desiredArch, curre
 
 	for i := len(conditionalUpdates) - 1; i >= 0; i-- {
 		for j, risk := range conditionalUpdates[i].Risks {
-			conditionalUpdates[i].Risks[j].MatchingRules, err = clusterconditions.PruneInvalid(ctx, risk.MatchingRules)
+			conditionalUpdates[i].Risks[j].MatchingRules, err = c.conditionRegistry.PruneInvalid(ctx, risk.MatchingRules)
 			if len(conditionalUpdates[i].Risks[j].MatchingRules) == 0 {
 				klog.Warningf("Conditional update to %s, risk %q, has empty pruned matchingRules; dropping this target to avoid rejections when pushing to the Kubernetes API server. Pruning results: %s", conditionalUpdates[i].Release.Version, risk.Name, err)
 				conditionalUpdates = append(conditionalUpdates[:i], conditionalUpdates[i+1:]...)
--- a/pkg/cincinnati/cincinnati_test.go
+++ b/pkg/cincinnati/cincinnati_test.go
@@ -10,11 +10,11 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/openshift/cluster-version-operator/pkg/clusterconditions/standard"
+
 	"github.com/blang/semver/v4"
 	"github.com/google/uuid"
 	configv1 "github.com/openshift/api/config/v1"
-	_ "github.com/openshift/cluster-version-operator/pkg/clusterconditions/always"
-	_ "github.com/openshift/cluster-version-operator/pkg/clusterconditions/promql"
 	_ "k8s.io/klog/v2" // integration tests set glog flags.
 )
 
@@ -745,7 +745,7 @@ func TestGetUpdates(t *testing.T) {
 			ts := httptest.NewServer(http.HandlerFunc(handler))
 			defer ts.Close()
 
-			c := NewClient(clientID, nil, "")
+			c := NewClient(clientID, nil, "", standard.NewConditionRegistry(nil))
 
 			uri, err := url.Parse(ts.URL)
 			if err != nil {
--- a/pkg/clusterconditions/clusterconditions.go
+++ b/pkg/clusterconditions/clusterconditions.go
@@ -25,28 +25,51 @@ type Condition interface {
 	Match(ctx context.Context, condition *configv1.ClusterCondition) (bool, error)
 }
 
-// Registry is a registry of implemented condition types.
-var Registry map[string]Condition
+type ConditionRegistry interface {
+	// Register registers a condition type, and panics on any name collisions.
+	Register(conditionType string, condition Condition)
+
+	// PruneInvalid returns a new slice with recognized, valid conditions.
+	// The error complains about any unrecognized or invalid conditions.
+	PruneInvalid(ctx context.Context, matchingRules []configv1.ClusterCondition) ([]configv1.ClusterCondition, error)
+
+	// Match returns whether the cluster matches the given rules (true),
+	// does not match (false), or the rules fail to evaluate (error).
+	Match(ctx context.Context, matchingRules []configv1.ClusterCondition) (bool, error)
+}
+
+type conditionRegistry struct {
+	// registry is a registry of implemented condition types.
+	registry map[string]Condition
+}
+
+func NewConditionRegistry() ConditionRegistry {
+	ret := &conditionRegistry{
+		registry: map[string]Condition{},
+	}
+
+	return ret
+}
 
 // Register registers a condition type, and panics on any name collisions.
-func Register(conditionType string, condition Condition) {
-	if Registry == nil {
-		Registry = make(map[string]Condition, 1)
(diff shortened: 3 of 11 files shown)


deads2k:
/hold

will cause https://github.com/openshift/cluster-version-operator/blob/d6bb4ddd15e2ff903a1de593a52dfb3717508eb6/pkg/clusterconditions/promql/promql.go#L37 to fail

also, that appears to be missing an e2e test.

pkg/clusterconditions/promql/promql.go
```diff
@@ -32,23 +35,35 @@ type PromQL struct {
...
+		Condition: &PromQL{
+			kubeClient: kubeClient,
+			HTTPClientConfig: config.HTTPClientConfig{
+				Authorization: &config.Authorization{
+					Type:            "Bearer",
+					CredentialsFile: "/var/run/secrets/kubernetes.io/serviceaccount/token",
+				},
+				TLSConfig: config.TLSConfig{
+					CAFile:     "/etc/tls/service-ca/service-ca.crt",
+					ServerName: "thanos-querier.openshift-monitoring.svc.cluster.local",
```
deads2k, line 49: this is what makes sure TLS functions properly

pkg/clusterconditions/promql/promql.go
```diff
@@ -32,23 +35,35 @@ type PromQL struct {
...
-	MinBetweenMatches: 10 * time.Minute,
-	MinForCondition:   time.Hour,
-	Expiration:        24 * time.Hour,
+		MinBetweenMatches: 10 * time.Minute,
+		MinForCondition:   time.Hour,
+		Expiration:        24 * time.Hour,
+	}
+}
+
+func (p *PromQL) Address(ctx context.Context) (string, error) {
```
deads2k, line 60: this avoids the DNS lookup

pkg/clusterconditions/promql/promql.go
```diff
@@ -69,7 +84,11 @@ func (p *PromQL) Valid(ctx context.Context, condition *configv1.ClusterCondition
 // false when the PromQL evaluates to 0, and an error if the PromQL
 // returns no time series or returns a value besides 0 or 1.
 func (p *PromQL) Match(ctx context.Context, condition *configv1.ClusterCondition) (bool, error) {
-	clientConfig := api.Config{Address: p.Address}
+	address, err := p.Address(ctx)
```
deads2k, line 87: lookup is done every match attempt in case of delete and recreate to get different IP.

pkg/clusterconditions/promql/promql.go
```diff
@@ -36,23 +35,26 @@ type PromQL struct {
...
-				Type:            "Bearer",
-				CredentialsFile: "/var/run/secrets/kubernetes.io/serviceaccount/token",
-			},
-			TLSConfig: config.TLSConfig{
-				CAFile:     "/etc/tls/service-ca/service-ca.crt",
-				ServerName: "thanos-querier.openshift-monitoring.svc.cluster.local",
+func NewPromQL(kubeClient kubernetes.Interface) *cache.Cache {
+	return &cache.Cache{
+		Condition: &PromQL{
+			kubeClient: kubeClient,
```
wking, line 41: What's this addition about?

wking: Ah, that's from 2de7afff20250f1eae026df8011c954fa096047a .  Ideally the commits would be distinct pivots, with the `kubeClient` stuff all collected together in one, but :shrug:, not a big deal either way.

deads2k: commit history fixed.

wking (approved):
/lgtm

deads2k:
/hold

upgrade out of change is failing consistently.  It's upgrading just fine, but at the start of the upgrade there are dial failures to routes.  I don't know how this change causes that, but it apparently does.  Even stranger, the installation itself is working (see the agnostic ovn job)

deads2k:
actually this is a general azure regression.  I've scheduled a slack message for TRT in the morning and left details in the thread.

/hold cancel

pkg/cincinnati/cincinnati.go
```diff
@@ -37,14 +37,17 @@ type Client struct {
 	// requests.  If empty, the User-Agent header will not be
 	// populated.
 	userAgent string
+
+	conditionRegistry clusterconditions.ConditionRegistry
 }
 
 // NewClient creates a new Cincinnati client with the given client identifier.
-func NewClient(id uuid.UUID, transport *http.Transport, userAgent string) Client {
+func NewClient(id uuid.UUID, transport *http.Transport, userAgent string, conditionRegistry clusterconditions.ConditionRegistry) Client {
```
petr-muller, line 45: Just a note: this is a really good example why the global sucked. I would never expect an osus client code to depend on the condition registry data, this way it is at least explicit (although it's still fishy and seems wrong to me)

pkg/cincinnati/cincinnati_test.go
```diff
@@ -10,11 +10,11 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/openshift/cluster-version-operator/pkg/clusterconditions/standard"
```
petr-muller, line 13: Can we come up with some better names here? Both `clusterconditions` and `clusterconditions/standard` have a `ConditionRegistry` type and `NewConditionRegistry` constructor, and at callsites they will look like `clusterconditions.NewConditionRegistry` and `standard.NewConditionRegistry` and it is not clear how they are different and when to use which.

pkg/cincinnati/cincinnati_test.go
```diff
@@ -745,7 +745,7 @@ func TestGetUpdates(t *testing.T) {
 			ts := httptest.NewServer(http.HandlerFunc(handler))
 			defer ts.Close()
 
-			c := NewClient(clientID, nil, "")
+			c := NewClient(clientID, nil, "", standard.NewConditionRegistry(nil))
```
petr-muller, line 748: nit: I'd prefer if the registry had a `WithClient` variant or a fluent method but I guess it's just a cosmetic thing

pkg/clusterconditions/clusterconditions.go
```diff
@@ -25,28 +25,51 @@ type Condition interface {
...
 // PruneInvalid returns a new slice with recognized, valid conditions.
 // The error complains about any unrecognized or invalid conditions.
-func PruneInvalid(ctx context.Context, matchingRules []configv1.ClusterCondition) ([]configv1.ClusterCondition, error) {
+func (r *conditionRegistry) PruneInvalid(ctx context.Context, matchingRules []configv1.ClusterCondition) ([]configv1.ClusterCondition, error) {
 	var valid []configv1.ClusterCondition
 	var errs []error
 
 	for _, config := range matchingRules {
-		condition, ok := Registry[config.Type]
+		condition, ok := r.registry[config.Type]
```
petr-muller, line 72: ok this is a code I dislike (not your fault, another artifact of the global), the actual registry is needed here only to validate types, and the remaining code IMO does not belong to a registry. I believe this code should live elsewhere, and the registry should only provide a `IsValid(type)` method or maybe a method that returns a validator func to be used here (probably the latter because the registration can allow dynamic behavior, even if it is not dynamically changed right now)

I don't think this needs to be addressed in this PR, I'll probably want to clean this up as a followup

pkg/clusterconditions/clusterconditions.go
```diff
@@ -63,11 +86,11 @@ func PruneInvalid(ctx context.Context, matchingRules []configv1.ClusterCondition
 
 // Match returns whether the cluster matches the given rules (true),
 // does not match (false), or the rules fail to evaluate (error).
-func Match(ctx context.Context, matchingRules []configv1.ClusterCondition) (bool, error) {
+func (r *conditionRegistry) Match(ctx context.Context, matchingRules []configv1.ClusterCondition) (bool, error) {
 	var errs []error
 
 	for _, config := range matchingRules {
-		condition, ok := Registry[config.Type]
+		condition, ok := r.registry[config.Type]
```
petr-muller, line 93: ditto

pkg/clusterconditions/clusterconditions_test.go
```diff
@@ -194,6 +193,4 @@ func TestMatch(t *testing.T) {
 			}
 		})
 	}
-
-	delete(clusterconditions.Registry, "Error")
```
petr-muller, line 198: :heart:

pkg/clusterconditions/promql/promql.go
```diff
@@ -32,23 +33,40 @@ type PromQL struct {
...
+
+// Address determines the address of the thanos-querier to avoid requiring service DNS resolution.
+// We do this so that our host-network pod can use the node's resolv.conf to resolve the internal load balancer name
+// on the pod before DNS pods are available and before the service network is available.  The side effect is that
+// the CVO cannot resolve service DNS names.
+func (p *PromQL) Address(ctx context.Context) (string, error) {
+	svc, err := p.kubeClient.CoreV1().Services("openshift-monitoring").Get(ctx, "thanos-querier", metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
```
petr-muller, line 67: Do we have any expectations about how reliable this is? Rock-solid, flakey, no idea? I'm just curious.

petr-muller (commented):
Big fan of getting rid of the global registry :vomiting_face: . I am not sure whether the name resolving commit is an ugly hack or not: it looks to me that way but /shrug.

pkg/clusterconditions/promql/promql.go
deads2k: > Do we have any expectations about how reliable this is? Rock-solid, flakey, no idea? I'm just curious.

very reliable.  It's what drives DNS

deads2k:
I'm going for it.  It's https://issues.redhat.com/browse/OCPBUGS-11308

petr-muller:
/retitle OCPBUGS-3783: Update dnsPolicy to allow consistent resolution of the internal LB

PR #936, opened 2023-05-04
Bump go.mod to 1.18

petr-muller:
ART are using 1.18 to build CVO (see Dockerfile.rhel) so this change only makes things more consistent.

Subsequent `go mod tidy` and `go mod vendor` only perform some cleanup in `go.sum` and `vendor` directory, but notably do not result in any changes in actually vendored code.


petr-muller:
/test e2e-agnostic-upgrade

Unlikely to be related:
```
[bz-monitoring][invariant] alert/Watchdog must have no gaps or changes 
{  Watchdog alert not found}
```

PR #944, opened 2023-06-22
Dockerfile: bump to 1.19 and use public UBI pullspec

petr-muller:
This Dockerfile is not used in CI [1] nor in product builds [2], so it may as
well be external contributor friendly and do not contain private CI
registry pullspecs.

We should use 1.20 but that image does not exist yet [3]. I do not see
any reason why to keep using UBI8 in this Dockerfile, so I bumped it to
UBI9.

[1] https://github.com/openshift/release/blob/7d59d083a7aa5f2aec28e807681a4b4e9ccdd088/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-master.yaml#L17-L25
[2] https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.14/images/cluster-version-operator.yml#L3
[3] https://github.com/openshift/release/pull/40567

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.18 AS builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-version-operator
 COPY . .
 RUN hack/build-go.sh; \
     mkdir -p /tmp/build; \
     cp _output/linux/$(go env GOARCH)/cluster-version-operator /tmp/build/cluster-version-operator
 
-FROM registry.ci.openshift.org/ocp/ubi:8
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 COPY --from=builder /tmp/build/cluster-version-operator /usr/bin/
 COPY install /manifests
 COPY vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml /manifests/


petr-muller:
$WORKSONMYMACHINE:

```console
$ podman build .
[1/2] STEP 1/4: FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
[1/2] STEP 2/4: WORKDIR /go/src/github.com/openshift/cluster-version-operator
--> 10fa2b0ca246
[1/2] STEP 3/4: COPY . .
--> f0d4320aaac1
[1/2] STEP 4/4: RUN hack/build-go.sh;     mkdir -p /tmp/build;     cp _output/linux/$(go env GOARCH)/cluster-version-operator /tmp/build/cluster-version-operator
Using version from git...
Building github.com/openshift/cluster-version-operator (v1.0.0-1050-gcdce77d7-dirty)
--> 7e73d1b4757d
[2/2] STEP 1/7: FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
Trying to pull registry.access.redhat.com/ubi9/ubi-minimal:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob 62742f27dce5 done  
Copying config d979387907 done  
Writing manifest to image destination
Storing signatures
[2/2] STEP 2/7: COPY --from=builder /tmp/build/cluster-version-operator /usr/bin/
--> cb85bcb44ded
[2/2] STEP 3/7: COPY install /manifests
--> 1c5b945d6730
[2/2] STEP 4/7: COPY vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml /manifests/
--> 35670a50c8e1
[2/2] STEP 5/7: COPY vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml /manifests/
--> d238df6d64c5
[2/2] STEP 6/7: COPY bootstrap /bootstrap
--> 02377c6ca76b
[2/2] STEP 7/7: ENTRYPOINT ["/usr/bin/cluster-version-operator"]
[2/2] COMMIT
--> 9b3200ab21e6
9b3200ab21e63c1a0b572c02f05f060aeab8b3c64415a245be15a4e0f81567f6
```

Davoska (commented):
I have successfully pulled `registry.ci.openshift.org/openshift/release:golang-1.19` and `registry.access.redhat.com/ubi9/ubi-minimal:latest` images with an empty pull spec file while running `podman build --authfile ~/path/pull-secret-empty.txt .`, and the image was built successfully.

I do have one question out of curiosity. Why do we switch from `ubi` to `ubi-minimal` base image in this PR? I understand that if we do not utilize fully the `ubi` and `ubi-minimal` is enough that we will save space. However, is there any chance something would break in this instance for the CVO? I can't think of anything but I suppose to be fully sure one would need to run e2e tests with the `ubi-minimal` base image?

petr-muller:
> I do have one question out of curiosity. Why do we switch from ubi to ubi-minimal base image in this PR? I understand that if we do not utilize fully the ubi and ubi-minimal is enough that we will save space. However, is there any chance something would break in this instance for the CVO? I can't think of anything but I suppose to be fully sure one would need to run e2e tests with the ubi-minimal base image?

It's more like "why not"... I do not see a reason why we'd need full UBI. CVO is a statically built binary and that Dockerfile just copies a bunch of static content into the final image. CVO does not do any subprocess calls (it spawns `version` Job but for that it would need a release payload image anyway) so it does not need any utils installed (like git or whatever). We *could* exercise the image better in CI if we wanted but I don't think it's worth the resources.

Davoska:
Thank you, @petr-muller, for the explanation! :heart:

LalatenduMohanty (approved):
/lgtm

PR #946, opened 2023-07-07
[WIP] NE-1318: Add required capabilities flag and set Ingress as required

alebedev87:
EP: https://github.com/openshift/enhancements/pull/1415


alebedev87:
/retest e2e-agnostic-ovn

Failures around `unsupported.do-not-use.openshift.io/override-liveness-grace-period-seconds` annotation behavior (`sig-node`). Let's retry to see if it's constant.

alebedev87:
`e2e-agnostic-ovn-upgrade-out-of-change` fails because `Ingress` capability implicitly enabled by this change stays in the status of `ClusterVersion` and is unknown to the CVO version without this change:

> I0725 12:24:29.339077       1 cvo.go:601] Error handling openshift-cluster-version/version: ClusterVersion.config.openshift.io "version" is invalid: status.capabilities.enabledCapabilities[2]: Unsupported value: "Ingress": supported values: "openshift-samples", "baremetal", "marketplace", "Console", "Insights", "Storage", "CSISnapshot", "NodeTuning", "MachineAPI"

I think it's the same case when any new capability is implicitly enabled, for instance while upgrading from 4.12 to 4.13 with `None` capability set.

alebedev87:
/hold

The ingress capability feature has been deferred to OCP 4.15. Reviews are still welcome though.

LalatenduMohanty:
@alebedev87 Can you please link the jira for this feature work? It will help with understanding the context behind the change.

alebedev87:
/unhold

The ingress capability targets 4.16.

alebedev87:
Rebased from `master` to resolve conflicts.

alebedev87:
/unhold

The temporary commit is removed, the API is bumped to the latest version.

alebedev87:
@wking , @petr-muller : would you mind to have a look at the PR? [The API PR](https://github.com/openshift/api/pull/1724) has been merged, I bumped the api to the latest version.

lib/capability/capability_test.go
```diff
@@ -236,10 +243,33 @@ func TestSetCapabilitiesWithImplicitlyEnabled(t *testing.T) {
 			priorEnabled: map[configv1.ClusterVersionCapability]struct{}{"cap1": {}, "cap2": {}, "cap3": {}},
 			wantImplicit: []string{"cap1", "cap2", "cap3"},
 		},
+		{name: "set capabilities 4_11 with always enabled",
```
candita, line 246: nit, this feels more informative:
```suggestion
		{name: "set baseline capabilities with always enabled",
```

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -217,8 +218,8 @@ func TestCVO_StartupAndSync(t *testing.T) {
...
+				EnabledCapabilities: []configv1.ClusterVersionCapability{configv1.ClusterVersionCapabilityBuild, configv1.ClusterVersionCapabilityCSISnapshot, configv1.ClusterVersionCapabilityCloudCredential, configv1.ClusterVersionCapabilityConsole, configv1.ClusterVersionCapabilityDeploymentConfig, configv1.ClusterVersionCapabilityImageRegistry, configv1.ClusterVersionCapabilityIngress, configv1.ClusterVersionCapabilityInsights, configv1.ClusterVersionCapabilityMachineAPI, configv1.ClusterVersionCapabilityNodeTuning, configv1.ClusterVersionCapabilityOperatorLifecycleManager, configv1.ClusterVersionCapabilityStorage, configv1.ClusterVersionCapabilityBaremetal, configv1.ClusterVersionCapabilityMarketplace, configv1.ClusterVersionCapabilityOpenShiftSamples},
```
candita, line 221: meganit:  the capabilities are no longer alphabetically ordered, so you could just add the new one to the end.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -3730,8 +3731,8 @@ func TestCVO_ParallelError(t *testing.T) {
...
+				KnownCapabilities:   []configv1.ClusterVersionCapability{configv1.ClusterVersionCapabilityBuild, configv1.ClusterVersionCapabilityCSISnapshot, configv1.ClusterVersionCapabilityCloudCredential, configv1.ClusterVersionCapabilityConsole, configv1.ClusterVersionCapabilityDeploymentConfig, configv1.ClusterVersionCapabilityImageRegistry, configv1.ClusterVersionCapabilityIngress, configv1.ClusterVersionCapabilityInsights, configv1.ClusterVersionCapabilityMachineAPI, configv1.ClusterVersionCapabilityNodeTuning, configv1.ClusterVersionCapabilityOperatorLifecycleManager, configv1.ClusterVersionCapabilityStorage, configv1.ClusterVersionCapabilityBaremetal, configv1.ClusterVersionCapabilityMarketplace, configv1.ClusterVersionCapabilityOpenShiftSamples},
```
candita, line 3735: Is there any need for a test of what happens when the new capability is missing for some reason?

pkg/cvo/sync_worker.go
```diff
@@ -462,7 +467,7 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 		return w.status.DeepCopy()
 	}
 
-	work.Capabilities = capability.SetCapabilities(config, priorCaps)
+	work.Capabilities = capability.SetCapabilities(config, priorCaps, capability.GetCapabilitiesAsMap(w.alwaysEnableCapabilities))
```
candita, line 470: I think this belongs up where we check `if w.work != nil` in L457

candita: But I also wonder why you didn't add `alwaysEnableCapabilities` to the `Capabilities` in `w.work` instead of keeping it separate from the other capabilities maps?

pkg/cvo/sync_worker.go
```diff
@@ -181,11 +181,16 @@ type SyncWorker struct {
 	requiredFeatureSet string
 
 	clusterProfile string
+
+	// alwaysEnableCapabilities is a list of cluster capabilities which should
```
candita, line 185: Why not add these capabilities to SyncWork.Capabilities?  Then you don't have to pass the variable around in the signatures of other functions.

pkg/start/start.go
```diff
@@ -542,3 +551,26 @@ func (o *Options) NewControllerContext(cb *ClientBuilder, startingFeatureSet str
...
+// parseAlwaysEnableCapabilities parses the string list of capabilities
+// into two lists of configv1.ClusterVersionCapability: known and unknown.
+func parseAlwaysEnableCapabilities(caps []string) ([]configv1.ClusterVersionCapability, []configv1.ClusterVersionCapability) {
+	var (
+		knownCaps   []configv1.ClusterVersionCapability
+		unknownCaps []configv1.ClusterVersionCapability
+	)
+	for _, c := range caps {
+		known := false
+		for _, kc := range configv1.KnownClusterVersionCapabilities {
```
candita, line 564: Is `KnownClusterVersionCapabilites` a map?  If so, you don't need to use range to check every entry.

LalatenduMohanty:
@alebedev87 can you please move the API vendoring to a separate commit. We always commit the file changes related to vendor in a separate commit. It also makes review easier. Here is an example https://github.com/openshift/cluster-version-operator/commit/a98232fcca84b8bc829a63d81dcbcc4242d60c4b

alebedev87:
@LalatenduMohanty : the latest openshift api has been bumped by [this commit](https://github.com/openshift/cluster-version-operator/commit/4561760b5abd85489293dbd147cf41059a57d663). So after the rebase, all vendor updates disappeared from the PR.

pkg/cvo/cvo_scenarios_test.go
alebedev87: This part of the tests [has been changed](https://github.com/openshift/cluster-version-operator/commit/8c2e78ecfebbd44f630575b731dc829b30cfb893) to a fixed (sorted) set of capabilities on `master`. After I rebased my PR, many of my unit test updates regarding the addition of the `Ingress` capability became unnecessary.

pkg/start/start.go
alebedev87: No, it's [a slice](https://github.com/openshift/cluster-version-operator/blob/6a76ba95ed441893e1bdf6616c47701c0464b7f4/vendor/github.com/openshift/api/config/v1/types_cluster_version.go#L401).

lib/capability/capability_test.go
alebedev87: I copied the description from the existing test. But I think you are right, `baseline` would be a neutral term. Also, I modified the baseline parameter to be the current set to avoid mentioning the EOL release (4.11).

pkg/cvo/sync_worker.go
alebedev87: > I think you need to set a variable from w.alwaysEnableCapabilities up where we check if w.work != nil in L457

[I wanted](https://github.com/openshift/cluster-version-operator/pull/946#discussion_r1512788624) to consolidate the process of determining the list of implicitly enabled capabilities within one function: `SetCapabilities`. This way, the logic of compiling the list of  the prior capabilities should occur before the `SetCapabilities` call.

alebedev87:
Seems like a problem with MCO, [e2e-agnostic-ovn](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn) and [e2e-agnostic-operator](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-operator) fail with this error since 3MAR:

> Cluster operator machine-config Degraded is True

> Node ci-op-icwwxmsi-31db5-c66dt-master-0 is reporting: "command \"/usr/bin/rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules\" returned with unexpected error: error: file /etc/audit/rules.d/mco-audit-quiet-containers.rules: Permission denied

lihongan:
cc @melvinjoseph86

pkg/cvo/sync_worker.go
candita: Yeah, I was mixing up `w.work` with `work`.

candita:
@alebedev87 this needs a rebase.  Otherwise
/lgtm

wking (approved):
/lgtm

jiajliu:
Hi @alebedev87 

Thank you for notifying us for a test from cvo side. I just had a pre-merge test to check the new flag in cvo deployment. Following is my test:

1. a regression test with default caps config, cvo works well with new deployment including always-enable-capabilities flag. 
```
// launch 4.16,openshift/cluster-version-operator#946 aws
# ./oc -n openshift-cluster-version get deployment cluster-version-operator -oyaml|grep Ingress
        - --always-enable-capabilities=Ingress
# ./oc get clusterversion version -ojson| jq .status.capabilities
{
  "enabledCapabilities": [
    "Build",
    "CSISnapshot",
    "CloudControllerManager",
    "CloudCredential",
    "Console",
    "DeploymentConfig",
    "ImageRegistry",
    "Ingress",
    "Insights",
    "MachineAPI",
    "NodeTuning",
    "OperatorLifecycleManager",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ],
  "knownCapabilities": [
    "Build",
    "CSISnapshot",
    "CloudControllerManager",
    "CloudCredential",
    "Console",
    "DeploymentConfig",
    "ImageRegistry",
    "Ingress",
    "Insights",
    "MachineAPI",
    "NodeTuning",
    "OperatorLifecycleManager",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ]
}
```
2. a test with none cap config, cvo also works well with the flag, and ingress operator is enabled as an implicitly enabled cap
```
//launch 4.16,openshift/cluster-version-operator#946 aws,no-capabilities
# ./oc get clusterversion version -ojson| jq -r '.status.conditions[]|select(.type=="ImplicitlyEnabledCapabilities")'
{
  "lastTransitionTime": "2024-04-19T09:58:36Z",
  "message": "The following capabilities could not be disabled: Ingress",
  "reason": "CapabilitiesImplicitlyEnabled",
  "status": "True",
  "type": "ImplicitlyEnabledCapabilities"
}

# ./oc get clusterversion version -ojson| jq .status.capabilities
{
  "enabledCapabilities": [
    "CloudControllerManager",
    "CloudCredential",
    "Ingress",
    "MachineAPI"
  ],
  "knownCapabilities": [
    "Build",
    "CSISnapshot",
    "CloudControllerManager",
    "CloudCredential",
    "Console",
    "DeploymentConfig",
    "ImageRegistry",
    "Ingress",
    "Insights",
    "MachineAPI",
    "NodeTuning",
    "OperatorLifecycleManager",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ]
}
# ./oc get co ingress
NAME      VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.16.0-0.test-2024-04-19-093919-ci-ln-w4l7bm2-latest   True        False         False      23m
```
cc @evakhoni for some upgrade test from cvo side. 
cc @jinyunma @shellyyang1989 for awareness

evakhoni:
@jiajliu as far as upgrade is concerned, I expect no much difference between `Ingress`, and any other previously introduced capability, since the standard behavior for any previously non-optional cluster resource becoming an optional capability on the version we're upgrading to, is ImplicitlyEnabled in any case where it is not explicitly enabled via either baseline, additional, or defaulted to enabled by the lack of thereof. in this case a presence of  `--always-enable-capabilities` is expected to manifest itself with the same effect, thus being identical in terms of expected observed behavior.

the only thing comes to my mind, is a basic sanity test verifying that the behavior explained above experienced no difference compared to standard capability introduction during upgrade, in which case any deviation from the standard behavior could be considered a regression. perhaps as well as to verify a presence of the aforementioned container argument following an upgrade. correct me if you think I missed something.

shellyyang1989:
cc @sunzhaohua2 for removing the workaround in upgrade ci after it's merged.

jiajliu:
pre-merge test pass.

/label qe-approved

PR #830, opened 2022-08-31
OCPBUGS-575: lib/resourcemerge/core: Reconcile seccompProfile in ensurePodSecurityContext

wking:
This recently started breaking [4.10-to-4.11-to-4.12 updates][1] with [the first failing run][2] like:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.12-upgrade-from-stable-4.11-from-stable-4.10-e2e-aws-sdn-upgrade/1562409104817786880/artifacts/e2e-aws-sdn-upgrade/gather-extra/artifacts/clusterversion.json | jq -r '.items[].status.conditions[] | .lastTransitionTime + " " + .type + "=" + .status + " " + .reason + ": " + .message'
2022-08-24T12:12:13Z RetrievedUpdates=False NoChannel: The update channel has not been configured.
2022-08-24T12:12:13Z ReleaseAccepted=True PayloadLoaded: Payload loaded version="4.12.0-0.ci-2022-08-24-061508" image="registry.build03.ci.openshift.org/ci-op-2v4w1wbk/release@sha256:37a2aa2a46ed42eedf73cf0ee1ec0f9a37b822823518172c01ef6ad2ba3ffa12" architecture="amd64"
2022-08-24T12:31:04Z Available=True : Done applying 4.11.1
2022-08-24T14:07:56Z Failing=True WorkloadNotProgressing: deployment openshift-operator-lifecycle-manager/package-server-manager has a replica failure FailedCreate: pods "package-server-manager-fbb7cb8b9-5zmwg" is forbidden: violates PodSecurity "restricted:v1.24": seccompProfile (pod or container "package-server-manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
2022-08-24T13:40:48Z Progressing=True WorkloadNotProgressing: Unable to apply 4.12.0-0.ci-2022-08-24-061508: the workload openshift-operator-lifecycle-manager/package-server-manager cannot roll out
2022-08-24T12:36:55Z ImplicitlyEnabledCapabilities=False AsExpected: Capabilities match configured spec
```

while [the previous run][3] had:

```console
[...]

PR #804, opened 2022-07-22
Bug 2108858: lib/resourcemerge: change SecurityContext reconcile

jottofar:
to handle securityContext changes differently. Since https://github.com/openshift/cluster-version-operator/commit/d9f6718, if a securityContext is not explicitly specified in the manifest the resource's securityContext will remain unchanged and it will continue to use the securityContext setting of the currently running resource (if there is one). We're not sure of the exact reason the logic was originally developed in this manner but this change joins a series of similar previous tightenings, including https://github.com/openshift/cluster-version-operator/commit/02bb9ba829174d799e418aeedfbd88bf572a35a0 (lib/resourcemerge/core: Clear env and envFrom if unset in
manifest, 2021-04-20, https://github.com/openshift/cluster-version-operator/pull/549) and https://github.com/openshift/cluster-version-operator/commit/ca299b8871e11ef1bd231e7b011be6920541e7ec (lib/resourcemerge: remove ports which are no longer required, 2020-02-13, https://github.com/openshift/cluster-version-operator/pull/322).

Reconciliation has been changed such that the entire securityContext structure, or any sub field of it, will be cleared if not specified in the manifest. This change affects Deployments, Jobs, and DaemonSets. It affects the securityContext found in both a PodSpec and a Container. Since the functions setInt64Ptr and setBoolPtr have been changed the impact is wide affecting ServiceAccounts, the PodSpec fields ShareProcessNamespace and TerminationGracePeriodSeconds, and the Job fields
ActiveDeadlineSeconds and ManualSelector.

For example, prior to this change assume Deployment machine-api-operator is running on the cluster with the following:

securityContext:
    runAsNonRoot: true
    runAsUser: 65534

and during an upgrade the Deployment machine-api-operator no longer specifies a securityContext. The resulting upgraded Deployment machine-api-operator will still have the original securityContext:

securityContext:
    runAsNonRoot: true
    runAsUser: 65534

Similarly, there is no way to remove, or clear, a securityContext field such as `runAsUser`. You can only modify it.

After this change the above scenario will correctly result in the Deployment machine-api-operator not specifying securityContext upon upgrade completion.

The changes apply to both the `SecurityContext` [1] within a `Container` and the `PodSecurityContext` [2] within a `PodSpec`.

[1] https://github.com/openshift/cluster-version-operator/blob/e0c92036a1939705638e2c0d5daf798a5ef966e9/vendor/k8s.io/api/core/v1/types.go#L2429
[...]

PR #947, opened 2023-07-10
OCPBUGS-5011: Backport `seccompProfile` reconciliation behavior

petr-muller:
> ⚠️ **this PR is mutually exclusive with https://github.com/openshift/cluster-version-operator/pull/940** ⚠️

---

From OCPBUGS-5011:

> After upgrading the cluster from 4.10 to 4.11, openshift-apiserver fails to match the new Pod Security Admission criteria.

Apiserver folks pinned this on CVO reconciliation behavior:

> - the openshift-apiserver-operator deployment/pod on 4.10 doesn't contain any definition of `seccompProfile` in the security context (it does on 4.11)
> - the Cluster Version Operator actually skips `seccompProfile` reconciliation when upgrading to 4.11
> - therefore the resulting deployment/pod after the upgrade will be missing `seccompProfile`, and this will cause the pod to fire PS violation alerts

This behavior in CVO was introduced in https://github.com/openshift/cluster-version-operator/pull/830 in 4.12 which was not backported to 4.11. Cherry-picking aae14d2a0467dbea6c92ed5a30b77f853cd74bb4 does not apply cleanly on release-4.11 without changes from https://github.com/openshift/cluster-version-operator/pull/804 (another 4.12 PR that changes `SecurityContext` reconciliation), so this PR adds necessary adjustments, including changing the behavior expected in the backported tests so it matches the CVO behavior after backporting only https://github.com/openshift/cluster-version-operator/pull/830 behavior without the https://github.com/openshift/cluster-version-operator/pull/804 one. The alternative is to backport both, proposed in https://github.com/openshift/cluster-version-operator/pull/940.

This PR resolves the problem identified by apiserver, but it does not *fully* resolve OCPBUGS-5011 for these two reasons:
- Even with the fixed CVO, there is a period during the update where we already run the 4.11 kube-apiserver (which brings the alert reporting the PSA violations) while we still have the 4.10 openshift-apiserver-operator `ReplicaSet` (which violates the PSA, it is only fixed in 4.11). So even with fixed CVO, we will still trip the alert during the upgrade.
- Because OCPBUGS-5011 was around for a long time, it became a honeypot for any report where the PodSecurityViolation alert fired, even ones unrelated to openshift-apiserver. The CVO change does not fix all culprits that trip this alert. I have not investigated the culprits unrelated to openshift-apiserver. It is possible that some of them are caused by the same gotcha like above, but it is also possible that they are separate problems.

So even if we merge this, the observable behavior of "we trip an alert during the update" will not disappear, disappointing all people who linked a "my update tripped the alert" report to OCPBUGS-5011.

### DIFF
--- a/lib/resourcemerge/core.go
+++ b/lib/resourcemerge/core.go
@@ -487,6 +487,7 @@ func ensureSecurityContext(modified *bool, existing *corev1.SecurityContext, req
 	setBoolPtr(modified, &existing.RunAsNonRoot, required.RunAsNonRoot)
 	setBoolPtr(modified, &existing.ReadOnlyRootFilesystem, required.ReadOnlyRootFilesystem)
 	setBoolPtr(modified, &existing.AllowPrivilegeEscalation, required.AllowPrivilegeEscalation)
+	ensureSeccompProfilePtr(modified, &existing.SeccompProfile, required.SeccompProfile)
 }
 
 func ensureCapabilitiesPtr(modified *bool, existing **corev1.Capabilities, required *corev1.Capabilities) {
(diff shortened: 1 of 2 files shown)


petr-muller:
Based on yesterday's discussion with @LalatenduMohanty we prefer https://github.com/openshift/cluster-version-operator/pull/940

/close

PR #950, opened 2023-08-02
CNF-9385: add ImageRegistry capability

qJkee:
Bump openshift/api to bring new cluster capability ImageRegistry


wking (approved):
The `upgrade-out-of-change` job will never pass, as the incoming old CRD's restrictive `status.capabilities.knownCapabilities` enum rejects the CVO's attempt to write `MachineAPI` and the other new caps.  Which blocks the CVO from setting other `status` properties, including the ones that point out that the release transition has completed.

/lgtm

PR #952, opened 2023-08-07
OCPBUGS-17418: Handle cache.DeletedFinalStateUnknown

ncdc:


pkg/cvo/upgradeable.go
```diff
@@ -460,7 +460,14 @@ func (optr *Operator) updateFunc(oldObj, newObj interface{}) {
...
 func (optr *Operator) deleteFunc(obj interface{}) {
-	cm := obj.(*corev1.ConfigMap)
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone
+	}
+	cm, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		klog.Errorf("Unexpected type %T", obj)
+		return
+	}
```
petr-muller, line 470: This looks good to me. 

I have looked into our other `DeleteFunc` handler and it seems like none of them need a similar treatment because they do not use the object from the informer but just queue something else for processing.

- https://github.com/openshift/cluster-version-operator/blob/9c0a4acf1afd6bd282ab451d2ebcd68c6dbc37f1/pkg/autoupdate/autoupdate.go#L119
- https://github.com/openshift/cluster-version-operator/blob/9c0a4acf1afd6bd282ab451d2ebcd68c6dbc37f1/pkg/cvo/cvo.go#L504-L506
- https://github.com/openshift/cluster-version-operator/blob/9c0a4acf1afd6bd282ab451d2ebcd68c6dbc37f1/pkg/featurechangestopper/featurechangestopper.go#L51-L53

petr-muller (approved):
/hold

LGTM, holding because Lala also said he's gonna take a look

pkg/cvo/upgradeable.go
wking: And just added 2763543246681d6a4d0a38afa68f2b771314af9c to make that more obvious.

wking:
Run the suite that panicked, to ensure we excercise the adjusted ConfigMap delete handler (even if we probably miss the well-timed watch-interruption that triggered [this panic](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.14-e2e-aws-sdn-serial/1687876857824808960)):

/payload-job periodic-ci-openshift-release-master-ci-4.14-e2e-aws-sdn-serial

LalatenduMohanty (changes requested):
New commit message suggestion from @wking which looks good to me.
```
This is to hint to any future developers that there is no existing operand handling in the functions, and they need to add all of that themselves, including (hopefully they realize) DeletedFinalStateUnknown handling.  That's a fair bit of hope to get DeletedFinalStateUnknown in place, but calling that specific aspect out in a comment or something would risk them reading the comment and assuming we were giving them an exhaustive list of guards to consider, when they may need additional guards to match distant-future informer changes (they hopefully read [1]).  And the amount of churn expected for our other delete handlers is low, so investing even more time in trying to protect those future devs seems uneccessary.

[1]: https://pkg.go.dev/k8s.io/client-go/tools/cache#ResourceEventHandler
```

LalatenduMohanty (approved):
/lgtm

PR #954, opened 2023-08-15
OCPBUGS-17418: Really handle DeletedFinalStateUnknown correctly

ncdc:


wking:
Extremely unlikely that this interacts with anything from #950.

PR #956, opened 2023-08-16
Revert "CNF-9385: add ImageRegistry capability"

dgoodwin:
Reverts openshift/cluster-version-operator#950

Testing, hypershift appears very broken now pulling openshift/tools in-cluster. We have no idea why yet but this PR is in the diff so we're black box testing if this is our culprit.


dgoodwin:
/payload-job periodic-ci-openshift-hypershift-release-4.14-periodics-e2e-aws-ovn-conformance

wking:
Dropping capabilities from the ClusterVersion CRD breaks updates (see CI on this pull), and we're discussing internally whether we want this, #957, or some way to save `ImageRegistry` for 4.14.

wking:
openshift/machine-config-operator#3876 and similar efforts have `ImageRegistry` back on track for 4.14, so perhaps we can keep the capability.  Closing for now.

PR #957, opened 2023-08-16
OCPBUGS-17806: Remove ImageRegistry capability

qJkee:


wking:
openshift/machine-config-operator#3876 and similar efforts have `ImageRegistry` back on track for 4.14, so perhaps we can keep the capability.  Closing for now.

PR #961, opened 2023-08-31
OCPBUGS-9050: Alert on failing conditional update risk evaluation

petr-muller:
Add a new `cluster_version_conditional_updates_recommended_conditions_seconds` which, for each conditional update known to CVO, reports how long (in seconds) is the `Recommended` condition on that update in the current state (this is the time since the conditions' `lastTransitionTime`). The metric is labelled with version, reason and status. Note that `lastTransitionTime` was not correctly maintained by CVO until fixed in #964.

Using this metric, we create an alert that fires when there is an update that is in an `Unknown` status for more than 50 minutes and this state is maintained for 10 minutes.

The metric was created slightly more generic than the alert would have needed. Technically the CVO could compute the "bad" state right away and just flap a 0/1 metric, but I believe the metrics are more useful when slightly generic.

PR #939, opened 2023-06-07
OCPBUGS-19512: pkg/clusterconditions/promql: Warm cache with 1s delay

wking:
530a5092a6 (#909) prioritized the order in which risks were evaluated.  But we were still waiting 10 minutes between different PromQL evaluations while evaluating conditional update risks.  This pull request sets up faster initial cache-warming while maintaining the same long-term load on the backing match engine, so it doesn't take 10 minutes to get that initial "I've figured it out now!" result.  ~I'm preserving 10-minute throttling between match calls when the cache is already warm.  But when the cache is warming up, we allow a match call every second, for much faster cache warming.~ [edit: [we dropped the warming-vs-resyncing distinction][1]]

[1]: https://github.com/openshift/cluster-version-operator/pull/939#discussion_r1332205435

PR #962, opened 2023-09-01
OCPBUGS-18454: Avoid using risk names as condition reasons when invalid

petr-muller:
Until / unless we merge https://github.com/openshift/api/pull/1577, conditional risk names may not be valid condition reason strings, so we need to check them and use a generic (new `RiskAppliesToCluster`) reason instead.

I took the opportunity and wrote tests and refactored the code slightly (make it easily testable from a single place and get rid of extensive branching logic that was hard to read). I suggest reviewing individual commits.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -443,7 +444,15 @@ const (
 	recommendedReasonAsExpected       = "AsExpected"
 	recommendedReasonEvaluationFailed = "EvaluationFailed"
 	recommendedReasonMultiple         = "MultipleReasons"
+
+	// recommendedReasonRiskApplies is used instead of the original name if it
+	// does not match the pattern for a valid k8s condition reason.
+	recommendedReasonRiskApplies = "RiskAppliesToCluster"
 )
+
+// Reasons follow same pattern as k8s Condition Reasons
+// https://github.com/openshift/api/blob/59fa376de7cb668ddb95a7ee4e9879d7f6ca2767/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1535-L1536
+var reasonPattern = regexp.MustCompile(`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`)
 
 func newRecommendedReason(now, want string) string {
 	switch {
@@ -473,7 +482,11 @@ func evaluateConditionalUpdate(ctx context.Context, risks []configv1.Conditional
 			errorMessages = append(errorMessages, unknownExposureMessage(risk, err))
 		} else if match {
 			recommended.Status = newRecommendedStatus(recommended.Status, metav1.ConditionFalse)
-			recommended.Reason = newRecommendedReason(recommended.Reason, risk.Name)
+			wantReason := recommendedReasonRiskApplies
+			if reasonPattern.MatchString(risk.Name) {
+				wantReason = risk.Name
+			}
+			recommended.Reason = newRecommendedReason(recommended.Reason, wantReason)
 			errorMessages = append(errorMessages, fmt.Sprintf("%s %s", risk.Message, risk.URL))
 		}
 	}
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -389,19 +389,15 @@ func (u *availableUpdates) evaluateConditionalUpdates(ctx context.Context) {
 		return vi.GTE(vj)
 	})
 	for i, conditionalUpdate := range u.ConditionalUpdates {
-		if errorCondition := evaluateConditionalUpdate(ctx, conditionalUpdate.Risks, u.ConditionRegistry); errorCondition != nil {
-			meta.SetStatusCondition(&conditionalUpdate.Conditions, *errorCondition)
-			u.removeUpdate(conditionalUpdate.Release.Image)
-		} else {
-			meta.SetStatusCondition(&conditionalUpdate.Conditions, metav1.Condition{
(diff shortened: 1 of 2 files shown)


petr-muller:
/hold

Reviews welcome but this will need to be adapted to #964 which is more important to merge.

petr-muller:
#964 merged but not this PR needs to wait for #939 and #961 to avoid conflicts

pkg/cvo/availableupdates.go
```diff
@@ -432,36 +429,69 @@ func unknownExposureMessage(risk configv1.ConditionalUpdateRisk, err error) stri
...
+	recommendedReasonAsExpected       = "AsExpected"
+	recommendedReasonEvaluationFailed = "EvaluationFailed"
+	recommendedReasonMultiple         = "MultipleReasons"
+
+	// recommendedReasonRiskApplies is used instead of the original name if it
+	// does not match the pattern for a valid k8s condition reason.
+	recommendedReasonRiskApplies = "RiskAppliesToCluster"
+)
+
+var reasonPattern = regexp.MustCompile(`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`)
```
wking, line 453: This uses https://github.com/openshift/api/blob/59fa376de7cb668ddb95a7ee4e9879d7f6ca2767/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1535 , to ensure we'll fit into `conditionalUpdates[].conditions`.  Possibly worth calling out that sourcing in a Go comment or commit message?

wking (approved):
/lgtm

pkg/cvo/availableupdates.go
petr-muller: sounds good, will amend

/hold

pkg/cvo/availableupdates.go
```diff
@@ -428,34 +428,52 @@ func unknownExposureMessage(risk configv1.ConditionalUpdateRisk, err error) stri
...
+		return metav1.ConditionFalse
+	case now == metav1.ConditionUnknown || want == metav1.ConditionUnknown:
+		return metav1.ConditionUnknown
+	default:
+		return want
+	}
+}
+
+const (
+	recommendedReasonAsExpected       = "AsExpected"
```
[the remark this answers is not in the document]
LalatenduMohanty: What do you think of `AvailableUpdate`

LalatenduMohanty (changes requested):

pkg/cvo/availableupdates.go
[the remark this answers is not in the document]
LalatenduMohanty: Or RecommendedUpdate instead of AsExpected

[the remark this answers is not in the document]
petr-muller: `AsExpected` is just an existing string literal extracted to a const, but I guess we can improve it in this PR.

How about `RisksDoNotApply`? Update is recommended *because* risks do not apply?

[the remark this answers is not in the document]
LalatenduMohanty: `RisksDoNotApply` sounds better for sure.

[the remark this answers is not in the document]
petr-muller: Perfect, added a commit

pkg/cvo/availableupdates.go
petr-muller: Added in an amend

petr-muller:
@LalatenduMohanty @wking addressed reviews, PTAL

pkg/cvo/availableupdates.go
```diff
@@ -432,36 +429,71 @@ func unknownExposureMessage(risk configv1.ConditionalUpdateRisk, err error) stri
...
+}
+
+const (
+	recommendedReasonRisksDoNotApply  = "RisksDoNotApply"
+	recommendedReasonEvaluationFailed = "EvaluationFailed"
+	recommendedReasonMultiple         = "MultipleReasons"
+
+	// recommendedReasonRiskApplies is used instead of the original name if it
+	// does not match the pattern for a valid k8s condition reason.
+	recommendedReasonRisksApply = "RisksApplyToCluster"
```
LalatenduMohanty, line 450: Nitpick: The reason `RisksApplyToCluster` sounds fine , but it is not consistent with the other strings i.e. `RisksDoNotApply` , `MultipleReasons` , `MultipleReasons` . Do `ReleventRisks` or `ApplicableRisk` sounds consistent or better?

LalatenduMohanty: IMO the wording is important because it is customer facing as console or OC CLI might show these reasons directly to users.

petr-muller: I am happy to change `RisksDoNotApply` to `RisksDoNotApplyToCluster` for consistency. 

I don't think there needs to be any special consistency with other reasons, I think both `EvaluationFailed` and `MultipleReasons` work well in the contexts when we set them?

petr-muller: What about `ExposedToRisks` and `NotExposedToRisks`?

LalatenduMohanty: ExposedRisks and NoExposedRisks sounds good. If not then ExposedToRisks and NotExposedToRisks would ok too.

petr-muller: I did the `ExposedToRisks` pair

LalatenduMohanty (approved):
/lgtm

petr-muller:
```
: [sig-cluster-lifecycle] pathological event should not see excessive Back-off restarting failed containers
{  event [namespace/openshift-console node/ci-op-icws625r-deec6-675nt-master-2 pod/console-6c8659cb4-5ffv7 hmsg/385654efc2 - Back-off restarting failed container console in pod console-6c8659cb4-5ffv7_openshift-console(f816a379-3140-4d58-9975-1a21d6954f0b)] happened 53 times}
```

Does not seem related

PR #966, opened 2023-09-06
OCPBUGS-18386: Properly reconcile SCC resources

petr-muller:
Previously CVO only created SCC resources when they did not exist, but did not enforce any further state from the manifests afterwards, so e.g. SCC changes on updates were never propagated to the cluster.

There is an existing `EnsureSecurityContextConstraint` method in the code, but it was not used anywhere. The method has a "merge" semantics on various lists though (so the resulting state is the union of what is in the cluster and in the manifest), which plays nice and tolerates user changes. Reconciling SCCs this way is not desired nor expected by the Auth team, so the merge semantics is removed in a followup commit because we want CVO to strictly reconcile to the state desired in the manifest. 

This work is done in separate commits to make backports easier and decisions clear. We will want to backport the merge semantics behavior into 4.13 and earlier, and this work will be less risky this way - the backport will simply omit one commit (the one that removes the merge semantics).


lib/resourceapply/security.go
```diff
@@ -29,14 +29,20 @@ func ApplySecurityContextConstraintsv1(ctx context.Context, client securityclien
 		return nil, false, nil
 	}
 
+	var original securityv1.SecurityContextConstraints
+	existing.DeepCopyInto(&original)
 	modified := pointer.Bool(false)
-	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	resourcemerge.EnsureSecurityContextConstraints(modified, existing, *required)
```
stlaz, line 35: `EnsureSecurityContextConstraints` will break when a new field is added into SCCs. Do we need any additional logic that makes this better than applying a generic unstructured object?

lib/resourceapply/security_test.go
```diff
@@ -0,0 +1,122 @@
...
+	AllowedCapabilities:      []corev1.Capability{"NET_BIND_SERVICE"},
+	Volumes:                  []securityv1.FSType{"configMap", "downwardAPI", "emptyDir", "persistentVolumeClaim", "projected", "secret"},
+	SELinuxContext:           securityv1.SELinuxContextStrategyOptions{Type: securityv1.SELinuxStrategyMustRunAs},
+	RunAsUser:                securityv1.RunAsUserStrategyOptions{Type: securityv1.RunAsUserStrategyMustRunAsRange},
+	SupplementalGroups:       securityv1.SupplementalGroupsStrategyOptions{Type: securityv1.SupplementalGroupsStrategyRunAsAny},
+	FSGroup:                  securityv1.FSGroupStrategyOptions{Type: securityv1.FSGroupStrategyMustRunAs},
+	SeccompProfiles:          []string{"runtime/default"},
+}
+
+func TestApplySecurityContextConstraintsv1(t *testing.T) {
```
stlaz, line 35: Add tests for scenarios with "create-only" annotation on the SCC. If you accidentally reconciled those, very bad things would start to happen.

lib/resourceapply/security.go
stlaz: As discussed with @vrutkovs, there might be some advantages to keeping this typed. In particular, Vadim named cases of slice order changes and merging map keys.
The SCCs unfortunately don't have `Spec` that would define their contents. I would suggest combining `DeepCopy` of the whole SCC into the final object while preserving the logic around `ObjectMeta`.

petr-muller: The current treatment follows what CVO does even for other types so I'm not inclined to change that in a bugfix (consistency has value). What you say is right (xref e.g. https://github.com/openshift/cluster-version-operator/pull/965) but it will need to be fixed separately.

lib/resourceapply/security_test.go
petr-muller: added a test

petr-muller:
I have added a commit that makes SCC reconciliation strictly follow the manifest (stomp any user modification) for non-`create-only` manifests. I think that's what we want to do for 4.15 (maybe we can still make it to 4.14), we may discuss whether we want to backport this in the strict form or whether we want to keep the merge semantics to avoid breaking people who patch-update.

petr-muller:
@evakhoni I know Auth's team opinion but I'll need to reconcile it with OTA team's opinion which will likely happen during code review - I'd suggest to wait with testing until that. I'm happy to hold the PR until you have a chance to test it afterwards.

/hold

petr-muller:
```
level=error msg=failed to initialize the cluster: Cluster operator cluster-autoscaler is not available
```
Looks unrelated
/retest

LalatenduMohanty:
@petr-muller can we move the vendor changes to a separate commit?

petr-muller:
@LalatenduMohanty I'd prefer not to. There are no actual changes in dependencies, this is not a bump (note no changes in `go.mod` / `go.sum`). We just vendor more code from our existing deps. In this case `go mod vendor` only vendors the files once the code that uses it actually exists, so splitting it to two commits would mean the first commit is incomplete.

petr-muller:
plenty of that in the CI seems like....
https://search.ci.openshift.org/?search=operator+conditions+control-plane-machine-set&maxAge=24h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

/retest

LalatenduMohanty (approved):
/lgtm

petr-muller:
@LalatenduMohanty ETA tomorrow or monday

lib/resourcemerge/security.go
```diff
@@ -11,78 +12,36 @@ func EnsureSecurityContextConstraints(modified *bool, existing *securityv1.Secur
...
-	for _, required := range required.AllowedCapabilities {
-		found := false
-		for _, curr := range existing.AllowedCapabilities {
-			if equality.Semantic.DeepEqual(required, curr) {
-				found = true
-				break
-			}
-		}
-		if !found {
+	for _, capacityLists := range [][]*[]corev1.Capability{
```
petr-muller, line 15:
```suggestion
	for _, capabilityLists := range [][]*[]corev1.Capability{
```

petr-muller: /lgtm cancel

petr-muller:
/hold

I'd like to get this tested before merge

petr-muller:
This is nice...

```
Suite run returned error: failed due to a MonitorTest failure
error running options: failed due to a MonitorTest failureerror: failed due to a MonitorTest failure
...
response Body: {"ProwJobName":"pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn-upgrade-into-change","ProwJobRunID":1704169262312067072,"Release":"Presubmits","CompareRelease":"4.15","Tests":[],"OverallRisk":{"Level":{"Name":"IncompleteTests","Level":75},"Reasons":["Tests for this run (80) are below the historical average (578): IncompleteTests"]},"OpenBugs":[]}
```

/test e2e-agnostic-ovn-upgrade-into-change

lib/resourceapply/security_test.go
```diff
@@ -0,0 +1,160 @@
...
+			existing: restrictedv2.DeepCopy,
+			required: restrictedv2.DeepCopy,
+			expected: restrictedv2.DeepCopy,
+			expectedAPICalls: []kubetesting.Action{
+				getSCCAction("restricted-v2"),
+			},
+			expectedModified: false,
+		},
+		{
+			name:     "no modified when existing differs but there's release.openshift.io/create-only: true annotation",
```
LalatenduMohanty, line 82: Suggested test names "no modification when existing differs but there's release.openshift.io/create-only: true annotation"

LalatenduMohanty (approved):
/lgtm

lib/resourcemerge/security.go
```diff
@@ -44,7 +45,14 @@ func EnsureSecurityContextConstraints(existing securityv1.SecurityContextConstra
...
 	setBoolPtr(&modified, &result.DefaultAllowPrivilegeEscalation, required.DefaultAllowPrivilegeEscalation)
-	setBoolPtr(&modified, &result.AllowPrivilegeEscalation, required.AllowPrivilegeEscalation)
+
+	// AllowPrivilegeEscalation is optional and defaults to true if not specified,
+	// so we enforce default if manifest does not specify it.
+	if required.AllowPrivilegeEscalation == nil {
+		setBoolPtr(&modified, &result.AllowPrivilegeEscalation, pointer.Bool(true))
+	} else {
+		setBoolPtr(&modified, &result.AllowPrivilegeEscalation, required.AllowPrivilegeEscalation)
+	}
```
LalatenduMohanty, line 55: I do not see other ways to fix the issue. Hopefully this doe not break anything.

lib/resourceapply/security_test.go
```diff
@@ -35,6 +36,16 @@ var restrictedv2 = securityv1.SecurityContextConstraints{
...
+func defaulted(get func() *securityv1.SecurityContextConstraints) func() *securityv1.SecurityContextConstraints {
+	return func() *securityv1.SecurityContextConstraints {
+		scc := get()
+		if scc.AllowPrivilegeEscalation == nil {
+			scc.AllowPrivilegeEscalation = pointer.Bool(true)
+		}
+		return scc
+	}
+}
+
```
LalatenduMohanty, line 48: Any reason you want to use a separate function to set `scc.AllowPrivilegeEscalation = pointer.Bool(true)` vs you use set scc.AllowPrivilegeEscalation = true in `var restrictedv2` and only for test case `name: "no modification when existing differs but there's release.openshift.io/create-only: true annotation",` set it to nil.

LalatenduMohanty (changes requested):

lib/resourceapply/security_test.go
petr-muller: Two reasons. I wanted to keep restrictedv2 as a true version of the restricted-v2 SCC from 4.14 payload, but more importantly, I prefer surprising behavior (like defaulting) to be as explicit as possible. I could set up `var restrictedv2` in a way that works but it woudl be suprising if you e.g. removed that member from test data from whatever reason and all tests started to fail. Done this way, testcases hint about the case that the existing state will be defaulted, the manifest will not, and the expected state is defaulted. E.g. this case shows it nicely:

```
name:     "no modification when existing is equal to required",
existing: defaulted(restrictedv2.DeepCopy),
required: restrictedv2.DeepCopy,
expected: defaulted(restrictedv2.DeepCopy),
...
```

LalatenduMohanty (approved):
/lgtm

petr-muller:
I'm checking whether @evakhoni need to flag this somehow before we let it merge. I'm planning to let it merge at the end of my day at latest.

PR #969, opened 2023-09-15
OCPBUGS-18386: Merge SCCs and block upgrades when modified

petr-muller:
- pkg/cvo/availableupdates: Prioritize conditional risks for largest target version
- Update dnsPolicy to allow consistent resolution of the internal LB
- replace global clustercondition registry with instance based approach
- connect to thanos using IP
- OTA-995: Adding admin-gate ack-4.13-kube-1.27-api-removals-in-4.14
- OCPBUGS-18386: Properly reconcile SCC resources
- review: add create-only test and test API interactions
- resourcemerge: stricty enforce required SCCs
- resourcemerge: restore merging semantics for SCCs
- resourcemerge: detect modified SCCs and set Upgradeable=False on them


petr-muller:
/hold

Needs to be polished and tested to be eligible for merge

LalatenduMohanty (changes requested):

lib/resourceapply/security.go
```diff
@@ -28,7 +81,12 @@ func ApplySecurityContextConstraintsv1(ctx context.Context, client securityclien
 		return nil, false, nil
 	}
 
-	reconcile := resourcemerge.EnsureSecurityContextConstraints(*existing, *required)
+	reconcile, requiredMerges := resourcemerge.EnsureSecurityContextConstraints(*existing, *required)
+
+	if reconciling {
+		trackModifiedScc(required.Name, requiredMerges)
```
wking, line 87: I think we can do this regardless of whether we're in `reconciling` mode, but :shrug: also not all that likely that folks turn straight from an in-progress *-to-4.13 update and target 4.14.

wking (approved):
/lgtm

LalatenduMohanty (approved):

JianLi-RH:
from QA, I give /qe-approved to the PR. @jiajliu @shellyyang1989 @jianlinliu please take a look

PR #972, opened 2023-09-20
OCPBUGS-19465: Properly reconcile SCC resources

petr-muller:
Previously CVO only created SCC resources when they did not exist, but did not enforce any further state from the manifests afterwards, so e.g. SCC changes on updates were never propagated to the cluster.

There is an existing EnsureSecurityContextConstraint method in the code, but it was not used anywhere. The method has a "merge" semantics on various lists though (so the resulting state is the union of what is in the cluster and in the manifest), which plays nice and tolerates user changes. Reconciling SCCs this way is not desired nor expected by the Auth team, so the merge semantics is removed in a followup commit because we want CVO to strictly reconcile to the state desired in the manifest.

This work is done in separate commits to make backports easier and decisions clear. We will want to backport the merge semantics behavior into 4.13 and earlier, and this work will be less risky this way - the backport will simply omit one commit (the one that removes the merge semantics).


wking (approved):
/lgtm

evakhoni:
pre-merge verified successfully.
/label qe-approved

LalatenduMohanty (approved):
/label backport-risk-assessed

PR #964, opened 2023-09-05
availableupdates: do not reset lastTransitionTime on unchanged status

petr-muller:
The code in `evaluateConditionalUpdates` correctly uses `SetStatusCondition` to set conditions, which only updates the
`LastTransitionTime` field when `Status` differs between the original and updated state. Previously though, the original state always contained empty conditions, because conditional updates are always obtained from OSUS and the fresh structure was never updated with existing conditions from the in-cluster status.

PR #926, opened 2023-04-17
OTA-854: Add configurable CVO knobs for risk-evaluation PromQL target

Davoska:
This pull request will add new flags and some minor logic to enable the CVO to evaluate conditional updates in HyperShift.

This pull request references the https://issues.redhat.com/browse/OTA-854

PR #973, opened 2023-09-25
OCPBUGS-19737: pkg/clusterconditions/promql: Warm cache with 1s delay

wking:
Picking #939 back to 4.14, with some minor manual edits because I'm not bringing back #926 (a feature adding support for new environments).  I am bringing back #964, because that was fixing existing behavior (so not a feature) and having it made the rebase slightly easier.


LalatenduMohanty (approved):
/lgtm

PR #982, opened 2023-10-17
OCPBUGS-20711: pkg/cvo/metrics: Disable HTTP/2

wking:
HTTP/2 is a good fit for some use-cases, but is unneccessary complication for our metrics server, which is only expected to have a single client in the local cluster's Prometheus scraper polling every 30 seconds or so.  Disabling HTTP/2 simplifies the server stack, and prevents exposure to CVE-2023-39325 and CVE-2023-44487, which [may not have been fully mitigated by recent Go-side fixes][1].  This also protects us against any future HTTP/2 vulnerabilities.

From [Go docs][2]:

> Starting with Go 1.6, the http package has transparent support for the HTTP/2 protocol when using HTTPS. Programs that must disable HTTP/2 can do so by setting `Transport.TLSNextProto` (for clients) or `Server.TLSNextProto` (for servers) to a non-nil, empty map.

[1]: https://github.com/golang/go/issues/63417#issuecomment-1758858612
[2]: https://pkg.go.dev/net/http#hdr-HTTP_2

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -126,7 +126,8 @@ func createHttpServer() *http.Server {
 	handler := http.NewServeMux()
 	handler.Handle("/metrics", promhttp.Handler())
 	server := &http.Server{
-		Handler: handler,
+		Handler:      handler,
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){}, // disable HTTP/2
 	}
 	return server
 }


wking:
Actually, I'm having trouble getting the existing CVO to talk over HTTP/2, even without this patch.  It seems to not understand ALPN, and default to HTTP/1.1:

```console
$ oc get -o json clusterversion version | jq -r .status.desired.version
4.14.0-rc.6
$ for PROTO in h3 h2 http/1.1 http/1.0; do oc --as=system:admin -n openshift-cluster-version exec deployment/cluster-version-operator -- openssl s_client -connect localhost:9099 -alpn "${PROTO}" 2>/dev/null | grep -i alpn; done
No ALPN negotiated
No ALPN negotiated
No ALPN negotiated
No ALPN negotiated
$ oc --as=system:admin -n openshift-cluster-version exec deployment/cluster-version-operator -- curl -Iksv https://localhost:9099/metrics 2>&1 | grep -i 'alpn\|http'
* ALPN, offering h2
* ALPN, offering http/1.1
* ALPN, server did not agree to a protocol
HTTP/1.1 200 OK
> HEAD /metrics HTTP/1.1
< HTTP/1.1 200 OK
```

Until we understand this:

/hold

shellyyang1989:
Double check on a 4.14 cluster,

# oc adm upgrade 
  Reason: VersionNotFound
Cluster version is 4.14.0-0.nightly-2023-10-17-113123

# oc get pod -n openshift-cluster-version
NAME                                        READY   STATUS    RESTARTS   AGE
cluster-version-operator-64985bc89c-659xz   1/1     Running   0          44m

# oc project openshift-cluster-version
Now using project "openshift-cluster-version" on server "https://api.huliu-nu18e.qe.devcluster.openshift.com:6443".

# oc rsh cluster-version-operator-64985bc89c-659xz

sh-4.4$ curl -Iks https://localhost:9099/metrics | grep "HTTP/"
HTTP/1.1 200 OK
sh-4.4$ 

HTTP/1.1 is being used in CVO.

wking:
We figured out why the CVO is not supporting HTTP/2 today, and it's [because](https://issues.redhat.com/browse/OCPBUGS-20711?focusedId=23302814&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23302814) we [have](https://github.com/openshift/cluster-version-operator/blob/80e7bbda9dadd86c067e09458a9a828d0cc3a1c2/pkg/cvo/metrics.go#L148-L150):

```go
	tlsListener := tls.NewListener(tcpListener, tlsConfig)
	klog.Infof("Metrics port listening for HTTPS on %v", lAddr)
	err = svr.Serve(tlsListener)
```

That means that by the time Go's `http` package gets a hand on anything, we've already configured our TLS listener, with a TLS config that does not include any of `http`'s "maybe we should stuff in HTTP/2 support by default?" logic.

PR #984, opened 2023-10-19
OCPVE-740: vendor: bump openshift/api

qJkee:
Bump openshift/api version with new capability set


petr-muller (approved):
LGTM. My understanding is that we should be able to pass the out-of-change test for this kind of change, the only failures is:

```
: [sig-network] pods should successfully create sandboxes by adding pod to network 
```

/test e2e-agnostic-ovn-upgrade-out-of-change

PR #987, opened 2023-10-25
pkg/clusterconditions/promql: Warm cache with 1s delay

wking:
Picking #974 and #980 back to 4.12.


wking:
oops, wrong base branch 🤦

wking:
replaced with #988.

PR #994, opened 2023-11-11
OTA-917: pkg/customsignaturestore: Implement signatureStores customization

wking:
[The verifier Interface's `AddStore` has always been][1]:

```go
// AddStore adds additional stores for signature verification.
func (v *releaseVerifier) AddStore(additionalStore store.Store) {
  v.store = &serial.Store{Stores: []store.Store{additionalStore, v.store}}
}
```

since it landed in 2020.  [We discussed then][2] whether we were comfortable with "always insert the new addition serially in front of the existing stores", and decided that it was good enough then (although we neglected to commit to that implementation in the `Interface`'s Go docs).

That history sets up my approach in this commit where I add support for ClusterVersion's new `spec.signatureStores`.  The logic is:

1. Try and find the signature in the local ConfigMaps.
    * If we error, e.g. by failing to list ConfigMaps, [fail out of the ConfigMap store][3], which will [fail out of the serial store][4], which will [fail verification][5] without further signature traversal.
    * If we find a valid signature, hooray, [we're done][6].
   * If we run out of ConfigMap entries to check without finding a valid signature, *fall back to...*
2. ClusterVersion's spec.signatureStores.
    * If this property is set to an empty array, fail out.  Admins can configure this if they only want to support ConfigMap lookup.
    * If this property is set to a non-empty array, build a parallel store out of its current contents, and search the new store:
        * If we error, check in with the call-back function about whether the error should be fatal, e.g. [here][7].  But [the verifier callback is very generous about allowing failures][5], so we'll keep going looking at any later signatures in that sigstore and at the other parallel sigstores.
        * If we find a valid signature, hooray, [we're][8] [done][9].
        * If we run out of sigstore entries to check without finding a valid signature, return an error, without wrapping with the callback, because we want to [fail][4] [out][5] without further signature traversal.  Admins who want to check default signature stores as well can configure them explicitly in `signatureStores`.
        * If this property is unset, *fall back to...*
3. [The default signature stores][10].
    * Same handling as the parallel store in `spec.signatureStores`, but with the default signature stores.

We only fall through to the default stores when ConfigMap lookup is exhausted without finding a valid signature and `signatureStores` is unset, to support admins who would rather not have the cluster-version operator reach out to one or more of the default stores, whether or not they have local stores they would rather use.  This allows admins to construct the set of stores they would like to use, while keeping the API surface down at a single property.

Without the `AddStore` history, it would have been possible to construct simplified stores, like a ConfigMaps store alone in the case where `signatureStores` is an empty array.  We could still pivot to that kind of approach with library-go changes, but for the initial implementation, I'm doing the above dance to work with `AddStore` as it stands.

[1]: https://github.com/openshift/library-go/blame/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/verify.go#L247-L250
[2]: https://github.com/openshift/library-go/pull/910#discussion_r500478890
[3]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/configmap/configmap.go#L100
[4]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/serial/serial.go#L32-L38
[5]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/verify.go#L189-L215
[6]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/serial/serial.go#L24-L34
[7]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/sigstore/sigstore.go#L119-L121
[8]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/sigstore/sigstore.go#L134-L136
[9]: https://github.com/openshift/library-go/blob/08d73a9c798b5698d88efd52232127e443bef9b6/pkg/verify/store/parallel/parallel.go#L61-L66
[10]: https://github.com/openshift/cluster-update-keys/blob/804dfeceb7caa41a6212fb405bbcd3f3139b8a6f/manifests.rhel/0000_90_cluster-update-keys_configmap.yaml#L4-L5


pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -0,0 +1,110 @@
...
+		stores = append(stores, &sigstore.Store{
+			URI:        &uri,
+			HTTPClient: s.HTTPClient,
+		})
+	}
+	store := &parallel.Store{Stores: stores}
+	if err := store.Signatures(ctx, name, digest, fn); err != nil {
+		return err
+	}
+	return errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature.")
```
petr-muller, line 63: This bit reads weird to me - at this point we have no chance to *not* return an error? We get to the last line if `store.Signatures` returns `nil` as error, how does that mean that we haven't found a valid signature?

wking:
ClusterBot testing looks good with `launch 4.15,openshift/cluster-version-operator#994 gcp` and:

```console
$ oc patch clusterversion version --type json -p '[{"op": "add", "path": "/spec/upstream", "value": "https://raw.githubusercontent.com/wking/cincinnati-graph-data/signatureStores-demo/cincinnati-graph.json"}]'
$ oc adm upgrade channel demo
$ oc adm upgrade --to 4.15.1
$ oc adm upgrade
Cluster version is 4.15.0-0.test-2023-11-17-182139-ci-ln-wbgj66k-latest

ReleaseAccepted=False

  Reason: VerifyPayloadVersion
  Message: Verifying payload failed version="4.15.1" image="quay.io/openshift-release-dev/ocp-release@sha256:beda83fb057e328d6f94f8415382350ca3ddf99bb9094e262184e0f127810ce0" failure=release image version 4.13.0 does not match the expected upstream version 4.15.1

Upstream: https://raw.githubusercontent.com/wking/cincinnati-graph-data/signatureStores-demo/cincinnati-graph.json
Channel: demo

Recommended updates:

  VERSION     IMAGE
  4.15.1      quay.io/openshift-release-dev/ocp-release@sha256:beda83fb057e328d6f94f8415382350ca3ddf99bb9094e262184e0f127810ce0
```

That isn't initiating the update, because the 4.13.0 signature I stuck in my dummy branch's signature store doesn't match the 4.15.1 version name I stuck in my dummy-branch's graph.  But it is getting past the "is there a signature that I trust" stage.  And shifting the signature in my dummy branch out of the correct spot, I was getting:

```
Retrieving payload failed version="4.15.1" image="quay.io/openshift-release-dev/ocp-release@sha256:beda83fb057e328d6f94f8415382350ca3ddf99bb9094e262184e0f127810ce0" failure=The update cannot be verified: unable to verify sha256:beda83fb057e328d6f94f8415382350ca3ddf99bb9094e262184e0f127810ce0 against keyrings: verifier-public-key-redhat
```

which is my custom store config successfully blocking access to the default stores, which contain a signature for that digest.

pkg/customsignaturestore/customsignaturestore.go
wking: Nice catch :bow:.  I've added a `wrapper` around the callback function so I can track `allDone` about whether we should successfully stop or not.

wking:
[Azure capacity shortage?][1]

```
level=error msg=Error: waiting for creation of Linux Virtual Machine: (Name "ci-op-4jfi2pl2-31db5-wlzzz-master-2" / Resource Group "ci-op-4jfi2pl2-31db5-wlzzz-rg"): Code="OSProvisioningTimedOut" Message="OS Provisioning for VM 'ci-op-4jfi2pl2-31db5-wlzzz-master-2' did not finish in the allotted time. The VM may still finish provisioning successfully. Please check provisioning state later. Also, make sure the image has been properly prepared (generalized).\r\n * Instructions for Windows: https://azure.microsoft.com/documentation/articles/virtual-machines-windows-upload-image/ \r\n * Instructions for Linux: https://azure.microsoft.com/documentation/articles/virtual-machines-linux-capture-image/ \r\n * If you are deploying more than 20 Virtual Machines concurrently, consider moving your custom image to shared image gallery. Please refer to https://aka.ms/movetosig for the same."
```

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/994/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn/1731719610153046016#1:build-log.txt%3A177

wking:
/payload-job periodic-ci-openshift-release-master-ci-4.15-e2e-aws-sdn-techpreview

wking:
openshift/api#1678 merged 🎉 .  I've rebased this pull to use that landed change instead of Pratik's branch, and dropped the WIP.

jiajliu:
/hold

pre-merge test hit a block issue, refer to https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp-modern/1732573738026668032

```
# launch 4.15,openshift/cluster-version-operator#994 gcp,techpreview

your cluster failed to launch: job failed, see logs (logs)
...
level=info msg=Waiting up to 20m0s (until 2:07AM UTC) for the Kubernetes API at https://api.ci-ln-ijdbqvb-72292.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX:6443...
level=info msg=Pulling VM console logs
level=info msg=Pulling debug logs from the bootstrap machine
level=error msg=Attempted to gather ClusterOperator status after installation failure: listing ClusterOperator objects: Get "https://api.ci-ln-ijdbqvb-72292.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX:6443/apis/config.openshift.io/v1/clusteroperators": dial tcp 34.66.53.18:6443: connect: connection refused
level=error msg=Bootstrap failed to complete: Get "https://api.ci-ln-ijdbqvb-72292.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX:6443/version": dial tcp 34.66.53.18:6443: i/o timeout
level=error msg=Failed waiting for Kubernetes API. This error usually happens when there is a problem on the bootstrap host that prevents creating a temporary control plane.
level=warning msg=The bootstrap machine is unable to resolve API and/or API-Int Server URLs
level=info msg=Flag --icsp-file has been deprecated, support for it will be removed in a future release. Use --idms-file instead.
level=info msg=Rendering Kubernetes API server core manifests...
level=info msg=F1207 02:07:47.808587       1 render.go:74] --rendered-manifest-files, are not consistent so results would be unpredictable depending on apply order: "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-Default.crd.yaml" and "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-CustomNoUpgrade.crd.yaml" both set CustomResourceDefinition.apiextensions.k8s.io/clusterversions.config.openshift.io in ns/, but have different values
level=info msg=Bootstrap gather logs captured here "/tmp/installer/log-bundle-20231207020748.tar.gz"
...
```

wking:
```
level=info msg=F1207 02:07:47.808587       1 render.go:74] --rendered-manifest-files, are not consistent so results would be unpredictable depending on apply order: "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-Default.crd.yaml" and "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-CustomNoUpgrade.crd.yaml" both set CustomResourceDefinition.apiextensions.k8s.io/clusterversions.config.openshift.io in ns/, but have different values
level=info msg=Bootstrap gather logs captured here "/tmp/installer/log-bundle-20231207020748.tar.gz"
```

Should be fixed with:

```console
$ git --no-pager diff bff08e52..d862ad20
diff --git a/pkg/payload/render.go b/pkg/payload/render.go
index b54849b4..25122bbe 100644
--- a/pkg/payload/render.go
+++ b/pkg/payload/render.go
@@ -76,8 +76,8 @@ func renderDir(renderConfig manifestRenderConfig, idir, odir string, skipFiles s
                if skipFiles.Has(file.Name()) {
                        continue
                }
-               if strings.Contains(file.Name(), "TechPreviewNoUpgrade") {
-                       continue // tech-preview features can come later, with the production CVO, and not at bootstrap-render time
+               if strings.Contains(file.Name(), "CustomNoUpgrade") || strings.Contains(file.Name(), "TechPreviewNoUpgrade") {
+                       continue // custom/tech-preview features can come later, with the production CVO, and not at bootstrap-render time
                }
 
                ipath := filepath.Join(idir, file.Name())
```

/hold cancel

LalatenduMohanty:
> level=info msg=F1207 02:07:47.808587       1 render.go:74] --rendered-manifest-files, are not consistent so results would be unpredictable depending on apply order: "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-Default.crd.yaml" and "/assets/manifests/0000_00_cluster-version-operator_01_clusterversion-CustomNoUpgrade.crd.yaml" both set CustomResourceDefinition.apiextensions.k8s.io/clusterversions.config.openshift.io in ns/, but have different values

@wking We should change the log type info to warning or error.  Is this coming from CVO?

petr-muller (approved):
Code LGTM to me. I'll hold b/c I was not that involved in this work and I think others may want to look, but feel free to unhold if you feel so. I think Joel had some concerns about different interpretation of a nil slice vs an empty slice but I think this can be sorted out in a followup.

/hold

LalatenduMohanty:
> Code LGTM to me. I'll hold b/c I was not that involved in this work and I think others may want to look, but feel free to unhold if you feel so. I think Joel had some concerns about different interpretation of a nil slice vs an empty slice but I think this can be sorted out in a followup.

We are waiting to see if we can add the CA support code for the signature stores i.e. `CA ConfigMapNameReference `json:"ca"`` before we merge this PR

wking:
`--rendered-manifest-files, are not consistent so results would be unpredictable depending on apply order` is originally from [this `error` from `GenericOptions.Validate`][1].  But the CVO doesn't vendor `github.com/openshift/library-go/pkg/operator/render/options`, so I'm not sure what code path is leading to that error being logged with `info` level.  The installer doesn't seem to vendor it either:

```console
openshift-installer $ git --no-pager grep library-go/pkg/operator/render/options origin/master
...no hits...
```

Ah,  it's probably some later operator's own `render` command checking for inputs to feed their own rendering, and voicing concern about the content they find in the manifest directory,  without knowing that it was the CVO that put the information there.  I'm not sure which operator it was though, although a closer look at bootstrap logs might say. 

[1]: https://github.com/openshift/library-go/blob/653f82d961a1d0cb42c5c86d34840491a8d975d7/pkg/operator/render/options/generic.go#L95

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (s *Store) refreshConfiguration(ctx context.Context) ([]*url.URL, error) {
+	var uris []*url.URL
+	config, err := s.Lister.Get(s.Name)
+	if err != nil {
+		return uris, err
```
LalatenduMohanty, line 80: We should return nil here as `uris` does not have any value in it.  Also if we sending err then the value of `uris` does not matter.

LalatenduMohanty (changes requested):

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			uris = append(uris, uri)
+		}
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.customURIs = uris
+	return uris, err
```
LalatenduMohanty, line 98: Shouldn't we return uris,nil here?

LalatenduMohanty (changes requested):

pkg/customsignaturestore/customsignaturestore.go
wking: `nil`  vs. empty array doesn't seem like it matters here, but sure, I've pivoted to `nil` with  d862ad20 -> a9f10b62.

pkg/customsignaturestore/customsignaturestore.go
wking: the only way to get here is if `err` is `nil`, and if future code changes allow us to fall through to this section with a non-nil `err`, that would be a bug (we'd be storing broken data in our local cache).  But sure, I can pivot to returning `nil`; done with a9f10b62 -> 02957bb4.

wking:
[build02 capacity issues][1].

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/994/pull-ci-openshift-cluster-version-operator-master-images/1732868984258170880

wking:
so close to branch-cut, I'm adding `no-qe` to lock in what we have, and we can patch what we're missing post-merge with bugs :)

wking:
[Disruption errors][1] and  [HyperShift CI account OIDC quotas][2] are both unrelated, because there isn't any CI logic yet to excercise signature retrieval (because we don't sign CI release builds).

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/994/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn/1732946933250527232
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/994/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift/1732947264424382464

PR #997, opened 2023-11-16
OCPBUGS-23377: to reduce log noise, only log sync_worker messages on retry or payload change

deads2k:
initial reduction for https://issues.redhat.com/browse/OTA-923

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -945,6 +945,15 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 
 	var reportEffectErrors []error
 
+	manifestVerbosity := klog.Level(4)
+	switch {
+	case work.Attempt > 0:
+		manifestVerbosity = 2
+	case work.State == payload.UpdatingPayload ||
+		work.State == payload.InitializingPayload:
+		manifestVerbosity = 2
+	}
+
 	// update each object
 	errs := payload.RunGraph(ctx, graph, maxWorkers, func(ctx context.Context, tasks []*payload.Task) error {
 		// attempt to precreate a set of known types (currently ClusterOperator) without retries
@@ -956,14 +965,14 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 				continue
 			}
 			if err := task.Manifest.Include(nil, nil, nil, &capabilities, work.Overrides); err != nil {
-				klog.V(2).Infof("Skipping precreation of %s: %s", task, err)
+				klog.V(manifestVerbosity).Infof("Skipping precreation of %s: %s", task, err)
 				continue
 			}
 			if err := w.builder.Apply(ctx, task.Manifest, payload.PrecreatingPayload); err != nil {
 				klog.V(2).Infof("Unable to precreate resource %s: %v", task, err)
 				continue
 			}
-			klog.V(2).Infof("Precreated resource %s", task)
+			klog.V(manifestVerbosity).Infof("Precreated resource %s", task)
 		}
 
 		for _, task := range tasks {
@@ -972,10 +981,10 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 			}
 			cr.Update()
 
-			klog.V(2).Infof("Running sync for %s", task)
+			klog.V(manifestVerbosity).Infof("Running sync for %s", task)
 
 			if err := task.Manifest.Include(nil, nil, nil, &capabilities, work.Overrides); err != nil {
-				klog.V(2).Infof("Skipping %s: %s", task, err)
+				klog.V(manifestVerbosity).Infof("Skipping %s: %s", task, err)
 				continue
 			}
 			if err := task.Run(ctx, payloadUpdate.Release.Version, w.builder, work.State); err != nil {
@@ -987,7 +996,7 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int, 
 				}
 			}
 			cr.Inc()
-			klog.V(2).Infof("Done syncing for %s", task)
+			klog.V(manifestVerbosity).Infof("Done syncing for %s", task)
 		}
 		return nil
 	})


pkg/cvo/sync_worker.go
```diff
@@ -956,14 +966,14 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 				continue
 			}
 			if err := task.Manifest.Include(nil, nil, nil, &capabilities, work.Overrides); err != nil {
-				klog.V(2).Infof("Skipping precreation of %s: %s", task, err)
+				klog.V(manifestVerbosity).Infof("Skipping precreation of %s: %s", task, err)
 				continue
 			}
 			if err := w.builder.Apply(ctx, task.Manifest, payload.PrecreatingPayload); err != nil {
-				klog.V(2).Infof("Unable to precreate resource %s: %v", task, err)
+				klog.V(manifestVerbosity).Infof("Unable to precreate resource %s: %v", task, err)
```
wking, line 973: I'd rather keep this at two, so we log the first error.  As you have it, we'll presumably soon be in a new sync round with `Attempt > 0`, and we'll log any second precreation error then.  But it still seems like it might be useful for incident timinling to have a timestamp logged for the first round's error.

pkg/cvo/sync_worker.go
```diff
@@ -972,10 +982,10 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 			}
 			cr.Update()
 
-			klog.V(2).Infof("Running sync for %s", task)
+			klog.V(manifestVerbosity).Infof("Running sync for %s", task)
 
 			if err := task.Manifest.Include(nil, nil, nil, &capabilities, work.Overrides); err != nil {
-				klog.V(2).Infof("Skipping %s: %s", task, err)
+				klog.V(manifestVerbosity).Infof("Skipping %s: %s", task, err)
```
wking, line 988: I have no problems being noisy when `overrides` is set.  I don't think we run any `overrides`-setting CI, except during installs where the installer fiddles with `overrides` to keep us from outracing cluster-bootstrap for resources the installer cares about.

Checking the no-capabilities job to see how noisy we are for capabilities filtering, [TestGrid][1] -> [run][2], where it's not crazy-noisy:

```console
$ $ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.15-e2e-aws-ovn-no-capabilities/1724760694437449728/artifacts/e2e-aws-ovn-no-capabilities/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-78698c8956-4wjjc_cluster-version-operator.log | grep -o '[.]go.*Skipping .*' | sort | uniq -c | sort -n | tail
     48 .go:978] Skipping service "openshift-image-registry/image-registry-operator" (347 of 865): disabled capabilities: ImageRegistry
     48 .go:978] Skipping service "openshift-insights/metrics" (629 of 865): disabled capabilities: Insights
     48 .go:978] Skipping service "openshift-machine-api/cluster-baremetal-operator-service" (237 of 865): disabled capabilities: baremetal
     48 .go:978] Skipping service "openshift-machine-api/cluster-baremetal-webhook-service" (238 of 865): disabled capabilities: baremetal
     48 .go:978] Skipping service "openshift-marketplace/marketplace-operator-metrics" (682 of 865): disabled capabilities: marketplace
     48 .go:978] Skipping service "openshift-operator-lifecycle-manager/catalog-operator-metrics" (655 of 865): disabled capabilities: OperatorLifecycleManager
     48 .go:978] Skipping service "openshift-operator-lifecycle-manager/olm-operator-metrics" (654 of 865): disabled capabilities: OperatorLifecycleManager
     48 .go:978] Skipping service "openshift-operator-lifecycle-manager/package-server-manager-metrics" (657 of 865): disabled capabilities: OperatorLifecycleManager
     48 .go:978] Skipping subscription "openshift-operator-lifecycle-manager/packageserver" (668 of 865): disabled capabilities: OperatorLifecycleManager
     48 .go:978] Skipping validatingwebhookconfiguration "performance-addon-operator" (442 of 865): disabled capabilities: NodeTuning
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.15-e2e-aws-ovn-no-capabilities/1724760694437449728/artifacts/e2e-aws-ovn-no-capabilities/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-78698c8956-4wjjc_cluster-version-operator.log | sed -n 's/^.* \([^ ]*[.]go:[0-9]*\).*/\1/p' | sort | uniq -c | sort -n | tail
    389 cvo.go:634
    424 payload.go:210
    576 sync_worker.go:959
   1007 sync_worker.go:966
   2046 task_graph.go:478
   3696 task_graph.go:481
  14602 sync_worker.go:978
  20305 sync_worker.go:701
  26118 sync_worker.go:990
  40833 sync_worker.go:975
```

But it is still the 4th most commonly hit log.  Maybe we can build something that trottles here to log once an hour per manifest or something or when the `err` string changes (e.g. because some capability was enabled, but not all the caps the resource needs).

[1]: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.15-informing#periodic-ci-openshift-release-master-ci-4.15-e2e-aws-ovn-no-capabilities
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.15-e2e-aws-ovn-no-capabilities/1724760694437449728

pkg/cvo/sync_worker.go
LalatenduMohanty: I agree we should not change the logging level for the precreate resource error.

pkg/cvo/sync_worker.go
LalatenduMohanty: I am fine with not printing this as we do not recommend customers using override. This is not something going to be impact our customers in a negative way. If we find something which convince us otherwise we can comeback and fix it.

LalatenduMohanty (changes requested):

pkg/cvo/sync_worker.go
```diff
@@ -945,6 +945,16 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 
 	var reportEffectErrors []error
 
+	manifestVerbosity := klog.Level(4)
+	switch {
+	case work.Attempt > 0:
+		manifestVerbosity = 2
+	case work.State == payload.UpdatingPayload ||
+		work.State == payload.InitializingPayload ||
+		work.State == payload.PrecreatingPayload:
```
wking, line 954: Our typing does not clarify this, but `PrecreatingPayload` is only used at a manifest level, and not on `SyncWorker` state as a whole:

```console
$ git --no-pager grep PrecreatingPayload
pkg/cvo/cvo.go: case payload.PrecreatingPayload:
pkg/cvo/sync_worker.go:                 if err := w.builder.Apply(ctx, task.Manifest, payload.PrecreatingPayload); err != nil {
pkg/payload/payload.go: // PrecreatingPayload indicates we are selectively creating
pkg/payload/payload.go: PrecreatingPayload
```

It doesn't hurt to have this entry here, outside of a chance of confusing future devs about the possible worker-level states.  I'm agnostic about whether we weight that possibility as riskier or not vs. the possibility that future code changes make `PrecreatingPayload` something that could apply to worker-level state.

pkg/cvo/sync_worker.go
wking: it's not just override, it's also excluding capabilities (the `no-capabilities` jobs I poked at above does not set `overrides`, does exclude capabilities, and sees a fair amount of volume from this line).  We do expect some customers to exclude capabilities (that's why we built that feature).  The resulting log volume may or may not be something those customers care about.  The presence of the lines every few minutes per manifest is more than anyone needs.  The presence of the lines at all is occasionally useful, e.g. I used it while internally triaging [OCPBUGS-22266](https://issues.redhat.com/browse/OCPBUGS-22266).  But I haven't needed the line while triaging issues with very old releases, so if folks want it off most of the time by default, I don't expect that to cause me problems.  And hopefully soon we'll have a tunable knob to bring it back in when folks want it to help debugging.

deads2k:
tweaked per comments.

pkg/cvo/sync_worker.go
LalatenduMohanty: I would suggest we go with current PR and if during the testing of 4.15 we feel that we need the log line back then we can create a bug and add it back.

LalatenduMohanty (approved):
/lgtm

wking:
[Pathological `cluster-baremetal-operator` container restarts][1] are unrelated to this change.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/997/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn-upgrade-out-of-change/1725518303201333249

petr-muller:
> [Pathological cluster-baremetal-operator container restarts](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-version-operator/997/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn-upgrade-out-of-change/1725518303201333249) are unrelated to this change.

Otherwise, we just need to wait for the `acknowledge...` gate to be lifted.

petr-muller:
This should fix Tide: https://github.com/kubernetes/test-infra/pull/31323

petr-muller:
We got the fix in https://github.com/openshift/release/pull/46195 and Tide merged this, looks good :crossed_fingers:

PR #998, opened 2023-11-16
Dropping these log levels to reduce the amounts of logs CVO creates

LalatenduMohanty:
With [OTA-923](https://issues.redhat.com//browse/OTA-923) with we will fix these with operator log level


LalatenduMohanty:
/closing this as #997 is a better PR

PR #1005, opened 2023-12-11
OCPBUGS-7714: yield lock while retrieving payload

petr-muller:
Retrieving payload is a potentially long-running operation (download can be slow or even timing out), and `RetrievePayload` is called while holding the `SyncWorker` lock. Holding the lock for extensive amount
of time prohibits the manifest applying look from proceeding, resulting in failures to reconcile state in the cluster.

`RetrievePayload` does not have access to anything protected by the lock, so I believe we can safely yield the lock for its duration and reacquire it afterwards.

~In addition to the bugfix, I refactored some lock-related code in `SyncWorker` code. I recommend reviewing per-commit, individual changes are fairly isolated but squashed together they could be confusing.~ moved to https://github.com/openshift/cluster-version-operator/pull/1020

---

This is why I think simply yielding the lock is sufficient and we do not need to add additional locks. At the moment, the lock is acquired for the whole duration of `SyncWorker.Update` method:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/sync_worker.go#L432-L436

The lock is documented as "guarding changes to `work`, `cancelFn` and `status` fields:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/sync_worker.go#L172-L176

There are two concerns  about yielding and reacquiring the lock somewhere deep in the middle of the execution chain of that method.

**Concern 1:** Do we need to protect against multiple `SyncWorker.Update` executions in parallel? This is the concern Trevor raised in https://github.com/openshift/cluster-version-operator/pull/1005#discussion_r1433130609. If we did then relying on *this* lock may be a mistake and using a second one could be a fix, but I believe this should never happen, its call chain is the following:

* A goroutine spawned by `Operator.Run()` calls `Operator.sync` as a handler for `Operator.worker` executed in a loop by `wait.UntilWithContext`.

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L444-L454

`wait.UntilWithContext` does not create parallelism and neither does `Operator.worker`:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L595-L612

* `Operator.sync` does not parallize anything either and calls `SyncWorker.Update`:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L699

This means CVO should never execute two `SyncWorker.Update` calls in parallel and therefore `RetrievePayload` not being re-entrant is not a concern. We *can* be cautious about this and add a second lock guarding `Update`, and panic when trying to acquire it.

---

**Concern 2:** Does `SyncWorker.Update` assume it "owns" the three fields for the whole duration, IOW can it get broken by other goroutines (only `SyncWorker.Start` does interact with the lock and locked fields) writing into the fields while it yields the lock for the duration of `RetrievePayload`? Again, I believe this is not the case. `SyncWorker.Start` does not write to `work`. It writes to `cancelFn`, but `SyncWorker.Update` does not care about this value before the (potential) yield, just after the reacquire, so there is no possible inconsistency. It also writes to `status`, and here it is a little muddy, but the two methods write into status through `updateApplyStatus` (under `Start`) and `updateLoadStatus` (under `Update`), and do not overwrite each other's parts of the status. So again, this concern should not cause issues.

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -240,7 +240,8 @@ func (w *SyncWorker) NotifyAboutManagedResourceActivity(message string) {
 // existing implicitly enabled capabilities. If no new implicitly enabled capabilities are found, just the previously
 // existing implicitly enabled capabilities are returned.
 //
-// Assumes SyncWorker is locked before syncPayload is called
+// Assumes SyncWorker is locked before syncPayload is called. It yields the lock while retrieving the payload image
+// (potentially long-running IO operation) but acquires it immediately after.
 func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork) ([]configv1.ClusterVersionCapability, error) {
 	// reporter hides status updates that occur earlier than the previous failure,
 	// so that we don't fail, then immediately start reporting an earlier status
@@ -283,7 +284,13 @@ func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork) ([]configv
 			Update:             desired,
 			LastTransitionTime: time.Now(),
 		})
+
+		// syncPayload executes while locked, but RetrievePayload is a potentially long-running operation
+		// which does not need the lock, so holding it may block other loops (mainly the apply loop) from
+		// execution
+		w.lock.Unlock()
 		info, err := w.retriever.RetrievePayload(ctx, work.Desired)
+		w.lock.Lock()
 		if err != nil {
 			msg := fmt.Sprintf("Retrieving payload failed version=%q image=%q failure=%s", desired.Version, desired.Image, strings.ReplaceAll(unwrappedErrorAggregate(err), "\n", " // "))
 			w.eventRecorder.Eventf(cvoObjectRef, corev1.EventTypeWarning, "RetrievePayloadFailed", msg)


pkg/cvo/sync_worker.go
```diff
@@ -714,35 +722,32 @@ func (w *statusWrapper) Report(status SyncWorkerStatus) {
...
 	// the state Update() calculated (to allow us to start in reconciling)
-	if work.Empty() {
-		work.State = w.work.State
-		work.Attempt = 0
-	} else if changed && work.State != payload.InitializingPayload {
-		klog.V(2).Infof("Work changed, transitioning from %s to %s", work.State, payload.UpdatingPayload)
-		work.State = payload.UpdatingPayload
-		work.Attempt = 0
+	if w.Empty() {
+		w.State = desired.State
```
petr-muller, line 736: This is suspicious because the check on L744 indicates `desired` can be nil

petr-muller:
I tested multiple iterations of `go test --race --count=10 ./...` and haven't encountered any failures.

pkg/cvo/sync_worker.go
```diff
@@ -286,7 +282,13 @@ func (w *SyncWorker) syncPayload(ctx context.Context, work *SyncWork,
 			Update:             desired,
 			LastTransitionTime: time.Now(),
 		})
+
+		// syncPayload executes while locked, but RetrievePayload is a potentially long-running operation
+		// which does not need the lock, so holding it may block other loops (mainly the apply loop) from
+		// execution
+		w.lock.Unlock()
 		info, err := w.retriever.RetrievePayload(ctx, work.Desired)
```
wking, line 290: This might be a bit risky, because `RetrievePayload` isn't currently re-entrant.  For example, [this cleanup line][1] assumes it's the only currently-running retrieval attempt.  We could add a lock to [the `payloadRetriever` struct][2] or otherwise find a way to deconflict multiple parallel calls to `RetrievePayload`.  Or we could make some kind of argument about how other loops that might be competing for the lock will never be other `syncPayload` rounds that could be trying to call `RetrievePayload`.  I guess we could add a lock to `payloadRetriever` and panic or otherwise complain loudly if we fail to immediately acquire the lock?

We can also slightly reduce the risk of cross-talk here by replacing `work.Desired` (relying on a pointer passed into this function) with `desired` (which we populated from `work.Desired` shorly after entering this function).  If this sounds useful, we could also use `desired.Force` in the klog line a few dozen lines up^^

[1]: https://github.com/openshift/cluster-version-operator/blob/baef28d57e26b3b6a476ef348d7f3274331907be/pkg/cvo/updatepayload.go#L232
[2]: https://github.com/openshift/cluster-version-operator/blob/baef28d57e26b3b6a476ef348d7f3274331907be/pkg/cvo/updatepayload.go#L63-L88

pkg/cvo/sync_worker.go
```diff
@@ -550,12 +558,14 @@ func (w *SyncWorker) Update(ctx context.Context, generation int64, desired confi
 		klog.V(2).Info("The sync worker has already been notified that new work is available")
 	}
 
-	return w.status.DeepCopy()
+	return w.locked.status.DeepCopy()
```
wking: This is a live-code change tucked into the comment-focused 0153577d3c1c48.  Possibly got squashed into the wrong commit while rebasing?  And then you revert this change in 4f7b4812a6f41?  Maybe you can leave it untouched in both commits?

pkg/cvo/sync_worker.go
```diff
@@ -1050,6 +1076,9 @@ func (r *consistentReporter) Update() {
 	r.reporter.Report(copied)
 }
 
+// Errors updates the status based on the current state of the graph runner.
+//
+// SyncWorker r.reporter.w must not be locked before Update is called.
```
wking, line 1081: "before Update is called" looks like a copy/paste issue?  I'd expect "before Errors is called" here, and similarly for `Completed` below.

pkg/cvo/sync_worker.go
```diff
@@ -893,6 +893,7 @@ func (w *SyncWorker) apply(ctx context.Context, work *SyncWork, maxWorkers int,
 	}
 	payloadUpdate := w.payload
 
+	w.lock.Lock() // lock to access w.status
```
wking, line 889: Is the only `status` access in this block `w.status.Architecture`?  Maybe we can:

```go
w.lock.Lock() // lock to access w.status
architecture := w.status.Architecture
w.lock.Unlock()
```

to make it clear that neither `Initializing()` nor `Reconciling()` care about the lock?  Those functions are just enum->bool conversion at the moment, but that wasn't clear (to me anyway) until I popped over to `payload.go` to check, and just because they're quick, isolated checks today doesn't mean they will stay that way forever.

pkg/cvo/sync_worker.go
petr-muller: Great observation, thanks! I missed this angle. I think locking `payloadRetriever` is a good solution on a first glance but I'll look into it in a bit more detail.

pkg/cvo/sync_worker.go
petr-muller: Yup, bad `fixup` in a rebase, I'll clean it up.

pkg/cvo/sync_worker.go
petr-muller: Great suggestion, will do!

petr-muller:
Addressed the "easy" feedback in 338dc8a , still need to look better into https://github.com/openshift/cluster-version-operator/pull/1005#discussion_r1433130609

petr-muller:
I decided to extract the refactors to a separate PR https://github.com/openshift/cluster-version-operator/pull/1020.

petr-muller:
This is why I think simply yielding the lock is sufficient and we do not need to add additional locks. At the moment, the lock is acquired for the whole duration of `SyncWorker.Update` method:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/sync_worker.go#L432-L436

The lock is documented as "guarding changes to `work`, `cancelFn` and `status` fields:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/sync_worker.go#L172-L176

There are two concerns  about yielding and reacquiring the lock somewhere deep in the middle of the execution chain of that method.

**Concern 1:** Do we need to protect against multiple `SyncWorker.Update` executions in parallel? This is the concern Trevor raised in https://github.com/openshift/cluster-version-operator/pull/1005#discussion_r1433130609. If we did then relying on *this* lock may be a mistake and using a second one could be a fix, but I believe this should never happen, its call chain is the following:

* A goroutine spawned by `Operator.Run()` calls `Operator.sync` as a handler for `Operator.worker` executed in a loop by `wait.UntilWithContext`.

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L444-L454

`wait.UntilWithContext` does not create parallelism and neither does `Operator.worker`:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L595-L612

* `Operator.sync` does not parallize anything either and calls `SyncWorker.Update`:

https://github.com/openshift/cluster-version-operator/blob/7c22805ee23be586dcff97c673ca648f1810770b/pkg/cvo/cvo.go#L699

This means CVO should never execute two `SyncWorker.Update` calls in parallel and therefore `RetrievePayload` not being re-entrant is not a concern. We *can* be cautious about this and add a second lock guarding `Update`, and panic when trying to acquire it.

---

2. **Concern 2:** Does `SyncWorker.Update` assume it "owns" the three fields for the whole duration, IOW can it get broken by other goroutines (only `SyncWorker.Start` does interact with the lock and locked fields) writing into the fields while it yields the lock for the duration of `RetrievePayload`? Again, I believe this is not the case. `SyncWorker.Start` does not write to `work`. It writes to `cancelFn`, but `SyncWorker.Update` does not care about this value before the (potential) yield, just after the reacquire, so there is no possible inconsistency. It also writes to `status`, and here it is a little muddy, but the two methods write into status through `updateApplyStatus` (under `Start`) and `updateLoadStatus` (under `Update`), and do not overwrite each other's parts of the status. So again, this concern should not cause issues.

pkg/cvo/sync_worker.go
petr-muller: I looked into the lock and locked field interactions in CVO, and I believe neither of these two risks applies:

1. CVO does not run `SyncWorker.Update` (and therefore `RetrievePayload`) in parallel
2. `SyncWorker.Update` does not rely on other writers (there is just one, `SyncWorker.Start`) not writing into the locked fields while it executes

Details about my investigation: https://github.com/openshift/cluster-version-operator/pull/1005#issuecomment-1889677070

petr-muller:
Now I am happy I filled https://issues.redhat.com/browse/OCPBUGS-22442 back in October because I now have a proof that this is a flake that existed even before me messing with locks :P

```
    --- FAIL: TestRunGraph/mid-task_cancellation_with_work_in_queue_does_not_deadlock (0.01s) 
```

/test unit

wking (approved):
/lgtm

PR #1027, opened 2024-01-23
NO-JIRA: manifests: fix service manifest filename

petr-muller:
Noticed while reading CVO logs:

```
I0123 17:12:33.973415  127980 task_graph.go:42] Unable to parse run level and component for "0001_00_cluster-version-operator_03_service.yaml"
```

Placed the Service into the same runlevel like CVO Deployment,
immediately after it.


LalatenduMohanty:
> Placed the Service into the same runlevel like CVO Deployment,
immediately after it.

@petr-muller Wondering did it ever change in the past because of which we are seeing this? or it has been this way from the beginning?

petr-muller:
@LalatenduMohanty seems like it was born this way:

https://github.com/openshift/cluster-version-operator/commits/master/install/0001_00_cluster-version-operator_03_service.yaml

Trevor even noticed it in 2019 :rofl: : https://github.com/openshift/cluster-version-operator/pull/49#discussion_r250558797

LalatenduMohanty (approved):
/lgtm

petr-muller:
/label no-qe

No specific testing needed

PR #1029, opened 2024-01-24
install/0000_00_cluster-version-operator_04_testing: Reproduce OCPBUGS-27822

wking:
Effectively reverting openshift/platform-operators#100 to reproduce the ClusterOperator-patching behavior that seems to trigger client-cache corruption in OCPBUGS-27822.

/hold

### DIFF
--- a/install/0000_00_cluster-version-operator_04_testing.yaml
+++ b/install/0000_00_cluster-version-operator_04_testing.yaml
@@ -1,0 +1,12 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: etcd
+  namespace: openshift-etcd
+spec: {}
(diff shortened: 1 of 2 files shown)


petr-muller:
IIUC the Slack thread correctly then this is a hit and we're vulnerable in master. We did not get burned simply because we do not (correctly) have COs with namespace set that would make CVO go into the hotloop and give the client cache the opportunity to corrupt:

```
 [sig-instrumentation] Prometheus [apigroup:image.openshift.io] when installed on the cluster should have important platform topology metrics [apigroup:config.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/parallel] expand_more
: [sig-instrumentation] Prometheus [apigroup:image.openshift.io] when installed on the cluster should start and expose a secured proxy and unsecured metrics [apigroup:config.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/parallel] expand_more
```

PR #1030, opened 2024-01-26
OCPBUGS-24700: Add support for Custom Certificate Authorities for custom signature stores

LalatenduMohanty:
…or custom signature stores

### DIFF
--- a/pkg/customsignaturestore/customsignaturestore.go
+++ b/pkg/customsignaturestore/customsignaturestore.go
@@ -5,47 +5,46 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 	"sync"
 
+	configv1 "github.com/openshift/api/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/verify/store"
 	"github.com/openshift/library-go/pkg/verify/store/parallel"
 	"github.com/openshift/library-go/pkg/verify/store/sigstore"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
 )
 
 type Store struct {
 	// Name is the name of the ClusterVersion object that configures this store.
 	Name string
 
-	// Lister allows the store to fetch the current ClusterVersion configuration.
-	Lister configv1listers.ClusterVersionLister
+	// ClusterVersionLister allows the store to fetch the current ClusterVersion configuration.
+	ClusterVersionLister configv1listers.ClusterVersionLister
 
-	// HTTPClient is called once for each Signatures call to ensure
-	// requests are made with the currently-recommended parameters.
-	HTTPClient sigstore.HTTPClient
+	// HTTPClient construct which respects the customstore CA certs and cluster proxy configuration
+	HTTPClient func(string) (*http.Client, error)
 
 	// lock allows the store to be locked while mutating or accessing internal state.
 	lock sync.Mutex
 
-	// customURIs tracks the most-recently retrieved ClusterVersion configuration.
-	customURIs []*url.URL
+	// customStores tracks the most-recently retrieved ClusterVersion configuration.
+	customStores []configv1.SignatureStore
 }
 
 // Signatures fetches signatures for the provided digest.
 func (s *Store) Signatures(ctx context.Context, name string, digest string, fn store.Callback) error {
-	uris, err := s.refreshConfiguration(ctx)
+	customStores, err := s.refreshConfiguration()
 	if err != nil {
 		return err
-	}
-
-	if uris == nil {
+	} else if customStores == nil {
 		return nil
-	}
-
-	if len(uris) == 0 {
-		return errors.New("ClusterVersion spec.signatureStores is an empty array.  Unset signatureStores entirely if you want to to enable the default signature stores.")
+	} else if len(customStores) == 0 {
+		return errors.New("ClusterVersion spec.signatureStores is an empty array. Unset signatureStores entirely if you want to enable the default signature stores")
 	}
 
 	allDone := false
@@ -58,44 +57,66 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
 		return done, err
 	}
 
-	stores := make([]store.Store, 0, len(uris))
-	for i := range uris {
-		uri := *uris[i]
+	var errs []error
+	stores := make([]store.Store, 0, len(customStores))
+	for _, customStore := range customStores {
+		uri, err := url.Parse(customStore.URL)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to parse the ClusterVersion spec.signatureStores %w", err))
+			continue
+		}
+		newHttpClient, err := s.HTTPClient(customStore.CA.Name)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to process the ClusterVersion spec.signatureStores %w", err))
+			continue
+		}
+
 		stores = append(stores, &sigstore.Store{
-			URI:        &uri,
-			HTTPClient: s.HTTPClient,
-		})
+			URI:        uri,
+			HTTPClient: func() (*http.Client, error) { return newHttpClient, nil }})
+	}
+
+	if len(stores) == 0 {
+		return utilerrors.NewAggregate(errs)
 	}
 	store := &parallel.Store{Stores: stores}
-	if err := store.Signatures(ctx, name, digest, wrapper); err != nil || allDone {
-		return err
+	if err := store.Signatures(ctx, name, digest, wrapper); allDone {
+		if len(errs) > 0 {
+			klog.V(2).Infof("%s", utilerrors.NewAggregate(errs))
+		}
+		return nil
+	} else if err != nil {
+		errs = append(errs, err)
+		return utilerrors.NewAggregate(errs)
 	}
-	return errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature.")
+
+	errs = append(errs, errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature"))
+	return utilerrors.NewAggregate(errs)
 }
 
-func (s *Store) refreshConfiguration(ctx context.Context) ([]*url.URL, error) {
-	config, err := s.Lister.Get(s.Name)
+// refreshConfiguration retrieves the latest configuration from the ClusterVersionLister
+// and updates the customStores with the URL and CA information from the retrieved configuration.
+// It returns the updated customStores slice and any error encountered during the retrieval process.
+func (s *Store) refreshConfiguration() ([]configv1.SignatureStore, error) {
+
+	config, err := s.ClusterVersionLister.Get(s.Name)
 	if err != nil {
 		return nil, err
 	}
 
-	var uris []*url.URL
+	var customStores = make([]configv1.SignatureStore, 0, len(config.Spec.SignatureStores))
 	if config.Spec.SignatureStores != nil {
-		uris = make([]*url.URL, 0, len(config.Spec.SignatureStores))
 		for _, store := range config.Spec.SignatureStores {
-			uri, err := url.Parse(store.URL)
-			if err != nil {
-				return uris, err
-			}
-
-			uris = append(uris, uri)
+			url := store.URL
+			caCert := store.CA
+			customStores = append(customStores, configv1.SignatureStore{URL: url, CA: caCert})
 		}
 	}
 
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.customURIs = uris
-	return uris, nil
+	s.customStores = customStores
+	return customStores, nil
 }
 
 // String returns a description of where this store finds
@@ -104,14 +125,14 @@ func (s *Store) String() string {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	if s.customURIs == nil {
-		return "ClusterVersion signatureStores unset, falling back to default stores"
-	} else if len(s.customURIs) == 0 {
+	if s.customStores == nil {
+		return "ClusterVersion signatureStores not set, falling back to default stores"
+	} else if len(s.customStores) == 0 {
 		return "0 ClusterVersion signatureStores"
 	}
-	uris := make([]string, 0, len(s.customURIs))
-	for _, uri := range s.customURIs {
-		uris = append(uris, uri.String())
+	customStores := make([]string, 0, len(s.customStores))
+	for _, customStore := range s.customStores {
+		customStores = append(customStores, customStore.URL)
 	}
-	return fmt.Sprintf("ClusterVersion signatureStores: %s", strings.Join(uris, ", "))
+	return fmt.Sprintf("ClusterVersion signatureStores: %s", strings.Join(customStores, ", "))
 }
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -294,16 +294,16 @@ func (optr *Operator) LoadInitialPayload(ctx context.Context, startingRequiredFe
 	if err != nil {
 		return nil, fmt.Errorf("the local release contents are invalid - no current version can be determined from disk: %v", err)
 	}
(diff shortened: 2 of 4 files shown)


pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+		}
+		if customStore.CA.Name != "" {
+			stores = append(stores, &sigstore.Store{
+				URI:        uri,
+				HTTPClient: func() (*http.Client, error) { return newHttpClient, nil }})
+		}
+
+	}
+	if len(stores) == 0 {
+		return errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature")
```
petr-muller, line 86: I don't think the message is valid at this point - we have not even started to look for signatures (that's done by `Signatures`  call on L89). We just filtered out CAs without a name and were left at zero

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
 	}
 	store := &parallel.Store{Stores: stores}
 	if err := store.Signatures(ctx, name, digest, wrapper); err != nil || allDone {
 		return err
 	}
-	return errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature.")
+	return nil
```
petr-muller, line 92: This change does not seem correct, getting here means we haven't found a signature after exhausting all locations, I believe. The case when we find something is handled by the `allDone` condition on L89.

It is a bit troubling that this does not make any tests fail. Perhaps we are missing coverage.

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+// refreshConfiguration fetches the current ClusterVersion configuration
+// and saves the information to a local copy
+func (s *Store) refreshConfiguration(ctx context.Context) ([]configv1.SignatureStore, error) {
+	config, err := s.ClusterVersionLister.Get(s.Name)
 	if err != nil {
 		return nil, err
 	}
 
-	var uris []*url.URL
+	//Save configv1.SignatureStore.URL and configv1.SignatureStore.CA information
```
petr-muller, line 103: nit:
```suggestion
	// Save configv1.SignatureStore.URL and configv1.SignatureStore.CA information
```

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+	// from configv1.SignatureStore to customStores variable
```
petr-muller, line 104: Also I think the comment is a bit redundant with the code, I'd consider leaving it out

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -104,14 +123,106 @@ func (s *Store) String() string {
...
+		}
+	}
+
+	tlsConfig, err := s.getTLSConfig(certAuthName)
+	if err != nil {
+		return transport, err
+	} else if tlsConfig != nil {
+		transport.TLSClientConfig = tlsConfig
+	}
+	return transport, err
```
petr-muller, line 182: can be simplified for easier reading

```suggestion
	if err != nil {
		return transport, err
	}
	
	transport.TLSClientConfig = tlsConfig
	return transport, nil
```

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -104,14 +123,106 @@ func (s *Store) String() string {
...
+
+// getTransport constructs an HTTP transport configuration, including
+// any custom proxy configuration.
+func (s *Store) getTransport(certAuthName string) (*http.Transport, error) {
+	transport := &http.Transport{}
+
+	proxyConfig, err := s.getProxyConfig()
+	if err != nil {
+		return transport, err
+	} else if proxyConfig != nil {
```
petr-muller, line 166: these two `if`s are unrelated so lets keep them separate?

```suggestion
	} 
	if proxyConfig != nil {
```

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+		}
+		newHttpClient, err := s.HTTPClient(customStore.CA.Name)
+		if err != nil {
+			return err
+		}
+		if customStore.CA.Name != "" {
+			stores = append(stores, &sigstore.Store{
+				URI:        uri,
+				HTTPClient: func() (*http.Client, error) { return newHttpClient, nil }})
+		}
```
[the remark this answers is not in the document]
LalatenduMohanty: I agree. Updated the code.

pkg/customsignaturestore/customsignaturestore.go
LalatenduMohanty: Updated the message as `ClusterVersion spec.signatureStores exhausted without finding a valid signature store`

LalatenduMohanty: Actually removed the error message from here and moved this to the end of the method.

pkg/customsignaturestore/customsignaturestore.go
LalatenduMohanty: I will change this to following
```
tlsConfig, err := s.getTLSConfig(certAuthName)
	if err != nil {
		return transport, err
	}

	if tlsConfig != nil {
		transport.TLSClientConfig = tlsConfig
	}
	return transport, nil
```

wking:
Getting the whole sentence into the title ;)

/retitle OCPBUGS-24700: Add support for Custom Certificate Authorities for custom signature stores

pkg/customsignaturestore/customsignaturestore.go
wking: Done in 006fd24d4332 -> 4ead92af5e0927; I'm marking this thread resolved.

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,60 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+	var errs error
+	stores := make([]store.Store, 0, len(customStores))
+	for _, customStore := range customStores {
+		uri, err := url.Parse(customStore.URL)
+		if err != nil {
+			errs = fmt.Errorf("%w; %w;", errs, err)
+		}
+		newHttpClient, err := s.HTTPClient(customStore.CA.Name)
+		if err != nil {
+			errs = fmt.Errorf("%w; %w;", errs, err)
```
wking, line 77: This aggregation style could lead to `Unwrap` nesting like:

* Error C, wrapping
  * Error B, wrapping
    * Error A

But there's [no hierarchy to custom signature stores][1], so I'd rather have something like:

```go
var errs []error
stores := make([]store.Store, 0, len(customStores))
for _, customStore := range customStores {
  ...
  if err != nil {
    errs = append(errs, err)
    continue
  }
  ...
}

if len(stores) == 0 {
  return apierrors.NewAggregate(errs)
}

store := &parallel.Store{Stores: stores}
if err := store.Signatures(ctx, name, digest, wrapper); err != nil || allDone {
  errs = append(errs, err)
  return apierrors.NewAggregate(errs)
}

errs = append(errs, errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature."))
return apierrors.NewAggregate(errs)
```

That keeps the "I couldn't create a store for this custom-store entry" errors parallel siblings, and also makes all of those errors siblings with the "...and of the stores that I could create, I had some trouble finding acceptable signatures" errors, all so users get access to debugging content for every branch in the tree that dead-ended for this attempt without arriving at an acceptable signature.

[1]: https://github.com/openshift/api/blob/e2b0b690b6384745331b5e342dad3ed6c2cf4908/config/v1/types_cluster_version.go#L106-L107

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +65,56 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+	stores := make([]store.Store, 0, len(customStores))
+	for _, customStore := range customStores {
+		uri, err := url.Parse(customStore.URL)
+		if err != nil {
+			return err
+		}
+		newHttpClient, err := s.HTTPClient(customStore.CA.Name)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
wking: 4ead92af5e092 adds `errs = fmt.Errorf("%w; %w;", errs, err)` aggregation, so I'm marking this thread resolved.  I've opened [a new thread][1] suggesting some further fine-tuning, and we can continue to hash out how we want to handle error aggregation over there.

[1]: https://github.com/openshift/cluster-version-operator/pull/1030#discussion_r1563333651

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -3,48 +3,55 @@ package customsignaturestore
...
+	ClusterVersionLister configv1listers.ClusterVersionLister
 
-	// HTTPClient is called once for each Signatures call to ensure
-	// requests are made with the currently-recommended parameters.
-	HTTPClient sigstore.HTTPClient
+	// ConfigMapLister allows the store to fetch certificate
+	// authority ConfigMaps for per-store trust.
+	ConfigMapLister corev1listers.ConfigMapNamespaceLister
+
+	// ProxyLister allows the store to fetch the current ClusterVersion proxy configuration.
```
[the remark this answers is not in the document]
wking: fixed in 4ead92af5e0927; I'm marking this thread resolved.

pkg/customsignaturestore/customsignaturestore.go
wking: fixed in 4ead92af5e; I'm marking this thread resolved.  But feel free to spin out the test-coverage point to a new thread if you like.

pkg/customsignaturestore/customsignaturestore.go
wking: fixed in 4ead92af5e; I'm marking this thread resolved.

pkg/customsignaturestore/customsignaturestore.go
wking: related error-handling discussion [in this new thread][1] that addresses Petr's concern, and as of 4ead92af5e09, this error message is back at the end of the `Store.Signatures` function where it belongs, so I'm marking this thread resolved.

[1]: https://github.com/openshift/cluster-version-operator/pull/1030#discussion_r1563333651

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -3,49 +3,56 @@ package customsignaturestore
...
 
-	// HTTPClient is called once for each Signatures call to ensure
-	// requests are made with the currently-recommended parameters.
-	HTTPClient sigstore.HTTPClient
+	// ConfigMapLister allows the store to fetch certificate
+	// authority ConfigMaps for per-store trust.
+	ConfigMapLister corev1listers.ConfigMapNamespaceLister
+
+	// proxyLister allows the store to fetch the current proxy configuration.
+	proxyLister configv1listers.ProxyLister
```
wking, line 38: You cannot get Proxy information out of `ClusterVersionLister` or `ConfigMapLister`.  I'd expect us to need this to be a public `ProxyLister` so we could fall back to the Proxy-configured CA set when the user didn't specify a custom `CA`.  Like:

* If the store declares a custom CA set, use that.  Otherwise fall back to...
* If Proxy declares a system-scoped CA set, use that.  Otherwise fall back to...
* The container image's default CA set.

Regardless of CAs, we'll also still need to consume Proxy to set up the HTTP(S) proxy portions.  That's happening today inside the:

```go
httpClientConstructor := sigstore.NewCachedHTTPClientConstructor(optr.HTTPClient, nil)
```

in `cvo.go`, so you might want to keep `HTTPClient` in this structure, continue to populate it with `httpClientConstructor`, drop `proxyLister` (because `HTTPClient` will now handle that part for you), and stuff in the custom CA stuff into the client only when the store sets a custom CA.

pkg/cvo/cvo.go
```diff
@@ -295,9 +295,9 @@ func (optr *Operator) InitializeFromPayload(ctx context.Context, restConfig *res
 	}
 
 	customSignatureStore := &customsignaturestore.Store{
-		Lister:     optr.cvLister,
-		Name:       optr.name,
-		HTTPClient: httpClientConstructor.HTTPClient,
+		ClusterVersionLister: optr.cvLister,
+		Name:                 optr.name,
+		ConfigMapLister:      optr.cmConfigLister,
```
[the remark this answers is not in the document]
LalatenduMohanty: This is fixed in https://github.com/openshift/cluster-version-operator/commit/4ead92af5e0927abf842409850329092802b3579

pkg/customsignaturestore/customsignaturestore.go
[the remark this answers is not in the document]
LalatenduMohanty: Fixed this in the https://github.com/openshift/cluster-version-operator/commit/4ead92af5e0927abf842409850329092802b3579

pkg/customsignaturestore/customsignaturestore.go
LalatenduMohanty: Fixed it in https://github.com/openshift/cluster-version-operator/commit/4ead92af5e0927abf842409850329092802b3579

pkg/customsignaturestore/customsignaturestore.go
```diff
@@ -58,44 +67,69 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
...
+	errs = append(errs, errors.New("ClusterVersion spec.signatureStores exhausted without finding a valid signature"))
+	return utilerrors.NewAggregate(errs)
 }
 
-func (s *Store) refreshConfiguration(ctx context.Context) ([]*url.URL, error) {
-	config, err := s.Lister.Get(s.Name)
+// refreshConfiguration retrieves the latest configuration from the ClusterVersionLister
+// and updates the customStores with the URL and CA information from the retrieved configuration.
+// It returns the updated customStores slice and any error encountered during the retrieval process.
+func (s *Store) refreshConfiguration(ctx context.Context) ([]configv1.SignatureStore, error) {
```
petr-muller, line 111: If we change its signature anyway lets remove the `ctx` parameter, it is not used.

petr-muller (commented):
Small code nit, otherwise it looks good (but lets squash the temp commit)

pkg/customsignaturestore/customsignaturestore.go
LalatenduMohanty: Fixed in https://github.com/openshift/cluster-version-operator/commit/50c8a31debcbed5f1495f438d85218d0325b2ba1

wking (approved):
/lgtm

wking:
We don't check signatures in CI, so this is extremely unlikely to destabilize the coming 4.16 fork.

wking:
This one is going to be a pain to QE, and we want to QE once after landing before 4.16 forks off, so let's QE post-merge.

/label no-qe

wking:
[`VpcLimitExceeded`][1] (and [here][2]) are unrelated.  [`the "master" pool should be updated before the CVO reports available at the new version`][3] is weird, but also not related to this sig change.  We want to get in before 4.16 forks off.

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1030/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift/1791209717427605504#1:build-log.txt%3A61
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1030/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift-conformance/1791209720778854400#1:build-log.txt%3A57
[3]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1030/pull-ci-openshift-cluster-version-operator-master-e2e-agnostic-ovn-upgrade-out-of-change/1791209716869763072

PR #1031, opened 2024-01-30
OTA-1159: [1/x] Refactor syncStatus for testability

petr-muller:
updateClusterVersionStatus: add godoc and linewrap signature

---

status: work with `ClusterVersionStatus` instead of `ClusterVersion`
    
Status-computing methods can only read and modify `ClusterVersionStatus`
instead of `ClusterVersion`, resulting in simpler code and lower chance
of making a mistake.

---

status: uncouple `updateClusterVersionStatus` with `Operator`
    
`Operator` is a pretty beefy object with access to half of the universe, but `updateClusterVersionStatus` only uses its member very little. It reads its `release` member and calls its `getAvailableUpdates` method.
    
We can uncouple the long method by passing these two members as parameters. It is slightly awkward for the `getAvailableUpdates` method but I think the benefits outweight this.

---

status: extract status update logic to method
    
I would like to unit-test this logic but `syncStatus` does API operation so such tests would need a fake. Computing new status is enough logic to constitute a method, so we can extract it so that we can test it without a mock server.
    
Extracted with an IDE refactor helper.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -811,18 +811,18 @@ func versionStringFromRelease(release configv1.Release) string {
 // are used as fallbacks for any properties not defined in the release
 // image itself.
 func (optr *Operator) currentVersion() configv1.Release {
-	return optr.mergeReleaseMetadata(optr.release)
+	return mergeReleaseMetadata(optr.release, optr.getAvailableUpdates)
 }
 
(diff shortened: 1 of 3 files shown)


wking (approved):
/lgtm

petr-muller:
/retest
/label no-qe

No pre-merge testing necessary, will get solid amount of testing after merge

PR #1033, opened 2024-02-02
OTA-1159: [3/3] Propagate reconciliation errors via ReconciliationIssues condition

petr-muller:
SyncWorkerStatus: Add Failures field with all failures

Previously, the errors that happened during reconciliation were flattened by `summarizeTaskGraphErrors` method call in `consistentReporter.Errors()` before they were sent to status reconciliation method `syncStatus`, which propagates this flattened error to the ClusterVersion status field.

Rename the `Failure` field to `FailureSummary`, it still caries the flattened error. Add a new `Failures` field to carry the full slice of the original errors.

The error processing is overall a bit weird because the errors get flattened first at the resource reconciling loop side, and then the status reporting loop side uses the flattened error to somehow reconstruct what actually happened, but cleaning up that code would be a much larger effort, so this commit simply makes original errors to be passed to status reporting loop. It is not easy to do all error processing at a single side; it would make sense to do that on status reporting level, but the initial flattening uses some task graph / resource reconciliation data that are not available on the status reporting side.

---

status sync: propagate reconciliation errors via RRI condition

Fulfilling [OTA-1159](https://issues.redhat.com//browse/OTA-1159), this change adds a JSON-serialized form of all reconciliation errors encountered during a single reconciliation loop run to the `Message` field of the `ReconciliationIssues`-type condition, present only when `ReconciliationIssuesCondition` feature gate is enabled (=in tech preview).

### DIFF
--- a/pkg/cvo/sync_worker.go
+++ b/pkg/cvo/sync_worker.go
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
@@ -100,7 +101,8 @@ type CapabilityStatus struct {
 type SyncWorkerStatus struct {
 	Generation int64
 
-	Failure error
+	FailureSummary error
+	Failures       []error
 
 	Done  int
 	Total int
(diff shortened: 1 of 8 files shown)


petr-muller:
```
    --- FAIL: TestRunGraph/mid-task_cancellation_with_work_in_queue_does_not_deadlock
```
Known flake
/test unit

pkg/cvo/sync_worker.go
```diff
@@ -100,7 +101,8 @@ type CapabilityStatus struct {
 type SyncWorkerStatus struct {
 	Generation int64
 
-	Failure error
+	FailureSummary error
+	Failures       []error
```
wking, line 105: I think we can avoid having to pass these in parallel.  I've floated #1041 with a sketch showing `Unwrap()` and `Errors()` to walk the nested errors inside `Failure`.  It probably needs more filling out with the structured information we store for each node in the tree, but it will avoid us having to worry about things like "what if  the actor populating `SyncWorkerStatus` puts something in `FailureSummary` without updating `Failures` to hold the appropriate children?".

petr-muller:
/close

We did https://github.com/openshift/cluster-version-operator/pull/1041 instead

PR #1036, opened 2024-02-22
NO-JIRA: Runbook URLs for ClusterOperatorDown and ClusterOperatorDegraded

petr-muller:
These two are low-hanging fruit because the appropriate runbooks already exist. They are a bit vague which is understandable because each CO is different but some links are better than no links, and having links may increase pressure on improving these runbooks.

This does not fully resolve OCPBUGS-14246 because we are still missing a runbook for `ClusterVersionOperatorDown`


LalatenduMohanty (approved):
/lgtm

petr-muller:
`FAIL: TestOperator_upgradeableSync (30.00s)` :thinking: 

I'm almost certain adding annotations to alerts define in `ServiceMonitor`s should not cause anything bad in unit tests but I don't think I have ever seen this particular flake, weird.

/test unit

petr-muller:
/label no-qe
/retitle NO-JIRA: Runbook URLs for ClusterOperatorDown and ClusterOperatorDegraded

Although this partially addresses OCPBUGS-14246, this PR does not change any code. It can be tested when OCPBUGS-14246 is fully fixed.

PR #1045, opened 2024-04-30
OCPBUGS-33983: vendor: Update openshift/api to pick up zz_generated.crd-manifests

wking:
Catching up with the YAML getting shifted into a subdirectory (openshift/api#1814).  I followed [the new docs][2] to create `pkg/dependencymagnet`, and then:

```console
$ emacs pkg/featuregates/featuregates.go pkg/featuregates/featurechangestopper_test.go # configv1.FeatureGateUpgradeStatus -> features.FeatureGateUpgradeStatus
$ go get github.com/openshift/api@master github.com/openshift/library-go@master
$ go mod tidy
$ go mod vendor
$ sed -i 's|config/v1/0000_00_cluster-version-operator|config/v1/zz_generated.crd-manifests/0000_00_cluster-version-operator|g' $(git grep -l 0000_00_cluster-version-operator)
$ git add -A Dockerfile* hack/test-prerequisites.go pkg/dependencymagnet go.* vendor
```

all using:

```console
$ go version
go version go1.22.2 linux/amd64
```

A few points on the invocation:

* I usually use `go get -u ...`, e.g. in a98232fcca84b8bc829a63d81dcbcc4242d60c4b (#994), but in this case, that would pull in `k8s.io/api@v0.30.0`, and we want to stick with 0.29 until our `release-4.16` branch stops following this development branch.

* I'm explicitly bumping library-go, because if I don't, [builds fail on][3]:

    ```
    Building github.com/openshift/cluster-version-operator (v1.0.0-1203-g258de3f7-dirty)
    # github.com/openshift/library-go/pkg/manifest
    vendor/github.com/openshift/library-go/pkg/manifest/manifest.go:32:35: undefined: configv1.FeatureSets
    # github.com/openshift/cluster-version-operator/pkg/featuregates
    pkg/featuregates/featuregates.go:92:32: undefined: configv1.FeatureGateUpgradeStatus
    pkg/featuregates/featuregates.go:97:33: undefined: configv1.FeatureGateUpgradeStatus
    ```

    If I had been able to use `-u`, library-go would have automatically floated to the tip commit anyway.  But either way, I still needed to  make manual changes to keep up with openshift/api#1821 moving to the new `openshift/api/features` package.

[2]: https://github.com/openshift/api/commit/06baaa444cb8bcfd8abb5cf876223ae727c5ec5e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R211
[...]

PR #1051, opened 2024-06-10
METAL-1053: Remove baremetal dependency on MachineAPI capability

eranco74:
https://github.com/openshift/api/pull/1916 meged.
This PR bump openshift/api to latest master.


Dockerfile.rhel
```diff
@@ -8,6 +8,6 @@ RUN hack/build-go.sh; \
 FROM registry.ci.openshift.org/ocp/4.17:base-rhel9
 COPY --from=builder /tmp/build/cluster-version-operator /usr/bin/
 COPY install /manifests
-COPY vendor/github.com/openshift/api/config/v1/0000_00_cluster-version-operator_* /manifests/
+COPY vendor/github.com/openshift/api/config/v1/zz_generated.crd-manifests/0000_00_cluster-version-operator_* /manifests/
```
petr-muller, line 11: Please do the same also in `Dockerfile`

pkg/payload/render.go
```diff
@@ -10,6 +10,8 @@ import (
 
 	"github.com/pkg/errors"
 
+	_ "github.com/openshift/api/config/v1/zz_generated.crd-manifests"
```
petr-muller, line 13: Can you please include a dependencymagnet file like Trevor did in https://github.com/openshift/cluster-version-operator/pull/1045/files (following the [docs](https://github.com/openshift/api/commit/06baaa444cb8bcfd8abb5cf876223ae727c5ec5e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R211)) instead of pulling it here?

pkg/payload/render.go
```diff
@@ -76,8 +78,8 @@ func renderDir(renderConfig manifestRenderConfig, idir, odir string, skipFiles s
 		if skipFiles.Has(file.Name()) {
 			continue
 		}
-		if strings.Contains(file.Name(), "CustomNoUpgrade") || strings.Contains(file.Name(), "TechPreviewNoUpgrade") {
-			// CustomNoUpgrade and TechPreviewNoUpgrade may add features to manifests like the ClusterVersion CRD,
+		if strings.Contains(file.Name(), "CustomNoUpgrade") || strings.Contains(file.Name(), "TechPreviewNoUpgrade") || strings.Contains(file.Name(), "DevPreviewNoUpgrade") {
+			// CustomNoUpgrade, TechPreviewNoUpgrade and DevPreviewNoUpgrade may add features to manifests like the ClusterVersion CRD,
```
petr-muller, line 82: Thanks for tackling this post-bump installation issue :+1: . We ran into it in https://github.com/openshift/cluster-version-operator/pull/1045 but did not have space to investigate it.

petr-muller (changes requested):
`e2e-agnostic-ovn` and `upgrade-into-change` both seem like unrelated flakes but the problem in `e2e-agnostic-operator` seems like one more place that needs addressing - I think that's `hack/test-prerequisites.go`.

petr-muller:
```
TestRunGraph/mid-task_cancellation_with_work_in_queue_does_not_deadlock
```

That's a known flake
/override ci/prow/unit

petr-muller (approved):
LGTM

Thanks!

Setting `no-qe` based on https://github.com/openshift/api/pull/1916#issuecomment-2188249961 . We would need to bump o/api anyway and the kind of issues this cause is best discovered by the amount of automated testing that happens post-merge, especially this early in 4.17 cycle.

/label no-qe

PR #1074, opened 2024-07-23
[OSD-19661] Escalate CO degraded warning to critical

reedcort:
What type of PR is this?
Update

What this PR does / why we need it?
use case that the degraded operator is blocking the cluster upgrade in https://issues.redhat.com/browse/OHSS-28157

Which Jira/Github issue(s) this PR fixes?
Fixes # https://issues.redhat.com/browse/OSD-19661

### DIFF
--- a/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
+++ b/install/0000_90_cluster-version-operator_02_servicemonitor.yaml
@@ -105,7 +105,7 @@ spec:
         severity: critical
     - alert: ClusterOperatorDegraded
       annotations:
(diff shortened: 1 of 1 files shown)


petr-muller:
/hold

Change like this needs much better rationale than present in the description. This makes things worse by delaying an alert that would have fired after 30m to fire after a *day* (with greater severity, right). I would not be opposed to a new alert that uses stronger language and higher severity to escalate but I really dislike removing the 30m warning signal.

reedcort:
@petr-muller That makes sense, I'll close the PR and create a new alert that uses stronger language and has higher severity

PR #1091, opened 2024-10-01
OTA-1269: Scaffold Update Status Controller (USC)

petr-muller:
Until we have the API merged to openshift/api, we can use the data structures copied from the draft state (PR). File was pulled from

https://github.com/petr-muller/api/tree/update-status-api/update/v1alpha1

Revision: d0af39806bf8d60e44b293fed906ea8856173e41

---

As agreed in [OTA-1269](https://issues.redhat.com//browse/OTA-1269), the controller uses ConfigMap as a pseudo-api until we have a real API committed to openshift/api.

The Update Status Controller is a new Cobra subcommand of the CVO binary. The new command launches two controllers for now:

1. **Control Plane Informer** (CPI) monitors ClusterVersion resources (for now) and produces a respective ClusterVersion status insights. The functionality is limited, because part of the logic necessary to populate all fields on a CV insight requires watching also ClusterOperator monitoring and processing, which is not present yet.
1. **Update Status Controller** (USC) receives insights from the CPI and maintains the ConfigMap that mimicks the eventual Status API.

~~All manifests to deploy the new controller are gated by the `release.openshift.io/feature-set: DevPreviewNoUpgrade` featureset.~~

---

To try the thing:

I decided to pull the deployment manifests from this PR because they were breaking HyperShift, and I did not want to block on solving that. This means the PR can right now only be tested by running the controller locally.

1. Install a cluster via cluster-bot: `launch 4.18 aws,no-spot`
2. Create the namespace: `oc create namespace openshift-update-status-controller`
3. `go run ./cmd update-status-controller -v 4 --kubeconfig $CLUSTER_BOT_KUBECONFIG --namespace openshift-cluster-version`
4. Create a `status-api-cm-prototype` ConfigMap in `openshift-update-status-controller` namespace
5. Check out `.data` of the ConfigMap. The controller should create a `usc-cv-version` item in it, containing a crude YAML-serialized ClusterVersionStatusInsight struct
6. Start updating the cluster somewhere and watch the content of the CM item


petr-muller:
https://github.com/openshift/release/pull/57408 should fix the hypershift jobs

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,165 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: update-status-controller-library
+  namespace: openshift-cluster-version
+  annotations:
+    kubernetes.io/description: Role that allows the USC binary to operate local resources needed by the o/library-go controller code
```
wking, line 7: nit: `o/library-go` -> `openshift/library-go` or drop the library reference entirely, so that cluster admins reading this are less likely to be confused by the `o/` shorthand?

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,165 @@
...
+  namespace: openshift-cluster-version
+  annotations:
+    kubernetes.io/description: Role that allows the USC binary to operate cluster resources needed by the o/library-go controller code
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/feature-set: DevPreviewNoUpgrade
+rules:
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - infrastructures
```
wking, line 51: This and `replicasets` in the Role are weird permissions.  Worth pushing back on the library-go controller maintainers about making it possible to reduce this access in controllers who don't (as far as I can tell) actually need to depend on whatever functionality _does_ need access to these resources?  Or just leave it alone?  Either way, not a blocker for merging this scaffolding; we can always adjust later if library-go does make changes.

pkg/version/version.go
```diff
@@ -18,3 +19,30 @@ var (
...
 	String = fmt.Sprintf("ClusterVersionOperator %s", Raw)
 )
+
+var (
+	// commitFromGit is a constant representing the source version that
+	// generated this build. It should be set during build via -ldflags.
+	commitFromGit string
+	// versionFromGit is a constant representing the version tag that
+	// generated this build. It should be set during build via -ldflags.
+	versionFromGit string
```
wking, line 29: This seems like it's overlapping with out existing variables, which we set via `-ldflags` [here][1].  I'm not against pivoting to a more established approach, or borrowing a bit of tech-debt during this scaffolding phase and supporting both approaches, but I'd like to consolidate around a single approach to ingesting version information in the long term.

[1]: https://github.com/openshift/cluster-version-operator/blob/adf0cf5662854567d1ea7ce1590e0a6fed8fcaaa/hack/build-go.sh#L23-L34

pkg/updatestatus/start.go
```diff
@@ -0,0 +1,49 @@
...
+	uscNamespaceCoreInformers := informers.NewSharedInformerFactoryWithOptions(coreClient, 10*time.Minute, informers.WithNamespace(uscNamespace))
+
+	updateStatusController, sendInsight := newUpdateStatusController(ctx, coreClient, uscNamespaceCoreInformers, cc.EventRecorder)
+	controlPlaneInformerController := newControlPlaneInformerController(configInformers, cc.EventRecorder, sendInsight)
+
+	configInformers.Start(ctx.Done())
+	uscNamespaceCoreInformers.Start(ctx.Done())
+
+	go updateStatusController.Run(ctx, 1)
+	go controlPlaneInformerController.Run(ctx, 1)
```
wking, line 43: The CVO has [a rather elaborate system][1] to track child goroutines, wait some time for them to gracefully terminate, and collect `error` output.  We may not need that kind of graceful shutdown and exit code handling in this controller?  Because we're hopping to kick out `controlPlaneInformerController` soon, after which point we can drop the goroutine for `updateStatusController`?

[1]: https://github.com/openshift/cluster-version-operator/blob/adf0cf5662854567d1ea7ce1590e0a6fed8fcaaa/pkg/cvo/cvo.go#L418-L515

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -0,0 +1,234 @@
...
+
+	cmInformer := coreInformers.Core().V1().ConfigMaps().Informer()
+	controller := factory.New().
+		// call sync every 5 minutes or on CM events in the openshift-cluster-version namespace
+		WithSync(c.sync).ResyncEvery(5*time.Minute).
+		WithFilteredEventsInformersQueueKeysFunc(cmNameKey, nsFilter(uscNamespace), cmInformer).
+		ToController("UpdateStatusController", c.recorder)
+
+	// Spawn the insight collecting goroutine
+	sendInsight := c.startCollectingInsights(ctx)
```
wking, line 89: I'm a bit squeamish about launching a goroutine from the `new*` helper, instead of waiting on a `Run` or `Sync` call in the returned `factory.Controller`.  But I'm also not sure how to restructure this to avoid that yet...

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return controller, sendInsight
+}
+
+// startCollectingInsights starts a goroutine that listens for insights from informers and updates internal status
+// API-mimicking ConfigMap.
+func (c *updateStatusController) startCollectingInsights(ctx context.Context) sendInsightFn {
+	fromInformers := make(chan informerMsg, 100)
```
wking, line 97: do we care about exhaustively processing everything that comes in?  Especially while we're in the "no preserved history" phase of this, I'd expect the sender to populate a locked map holding a single most-recently-pushed entry for that UID.  So the consumer wouldn't have to worry about things like "are there multiple updates for this UID in the queue?", the producer wouldn't have to worry about blocking while the consumer dug out of a hold, and nobody would have to worry about holding obsolete data in memory.

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
petr-muller: Good point

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
petr-muller: I have not investigated this, just added anything that the controller actually failed on startup. The library code contains more controllers that handle some stuff behind the scenes, but I have not looked at the details (I also have not found any docs besides godocs for that library). I will make a note to investigate this - it would be good to have a better understanding what we are pulling in with the library-go code.

But I would not block on it.

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
petr-muller: Resolved in 0d2d67a3866773f64aa7c0b60a4a3ccc287a2621

pkg/version/version.go
petr-muller: Thanks for pointing this out, I added this to make the `controllercmd.NewControllerCommandConfig` [from library-go](https://github.com/openshift/library-go/blob/master/pkg/controller/controllercmd/cmd.go#L79) happy, just needed a [apimachinery version.Info](https://github.com/kubernetes/apimachinery/blob/cfee47580787e57b3cc51782a120cf770a97316e/pkg/version/types.go#L19-L32) so I think I just copied this from some other operator, because I did not want to mess with versions in CVO - that's a downside of adding USC to CVO binary. And then I entirely forgot about it :D

Right now I'm leaning towards creating a little bit of tech debt for this, but let's make sure it is addressed before we make this TechPreview. I'll make this bit even more obvious tech debt by inlining this in USC file, not here. I'll also start sprinkling TODOs for these items (I thinkg https://github.com/openshift/cluster-version-operator/pull/1091#discussion_r1810586011 is a similar thing).

pkg/updatestatus/start.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	uscNamespaceCoreInformers := informers.NewSharedInformerFactoryWithOptions(coreClient, 10*time.Minute, informers.WithNamespace(uscNamespace))
+
+	updateStatusController, sendInsight := newUpdateStatusController(ctx, coreClient, uscNamespaceCoreInformers, cc.EventRecorder)
+	controlPlaneInformerController := newControlPlaneInformerController(configInformers, cc.EventRecorder, sendInsight)
+
+	configInformers.Start(ctx.Done())
+	uscNamespaceCoreInformers.Start(ctx.Done())
```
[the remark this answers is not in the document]
petr-muller: This is done in [library-go controller's Run](https://github.com/openshift/library-go/blob/e59f1b6d8babb69c74eb38921b5b95274532a98b/pkg/controller/factory/base_controller.go#L86). We add all informers with the `WithInformer...` methods ([usc](https://github.com/openshift/cluster-version-operator/blob/0600bfb1afcad7d0203c42420b0cf92395c11ab3/pkg/updatestatus/updatestatuscontroller.go#L85), [cpi](https://github.com/openshift/cluster-version-operator/blob/0600bfb1afcad7d0203c42420b0cf92395c11ab3/pkg/updatestatus/controlplaneinformer.go)) and cache sync is a part of the setup.

pkg/updatestatus/start.go
petr-muller: TBH I'm not entirely sure - I followed some existing operators using library-go code the ones I found used this pattern without any further logic:

- https://github.com/openshift/service-ca-operator/blob/1f7d6472fea332fc307d9c8911f8c3cf57347755/pkg/operator/starter.go#L117-L137
- https://github.com/openshift/cluster-authentication-operator/blob/e401c0f1fbf56c7fa957daf5a8a8c91e568c646f/pkg/operator/starter.go#L88-L92

pkg/updatestatus/updatestatuscontroller.go
petr-muller: Sending locked memory locations makes the controller slightly more coupled than I'd like at this point. My ambition was not to design the final inter-controller communication yet, more like make something that's very easy to understand (both code and while running) and eventually change. Right now I appreciate processing all messages, because it's easy to follow events across the system. Let's optimize later?

petr-muller: But I'll leave a comment about the communication mechanism being intentionally simple.

pkg/updatestatus/updatestatuscontroller.go
petr-muller: /hold 

will address

petr-muller: Reworked in 25061544f3fcd1aa2ad04e50ecd98c2944a9e414 to use `WithPostStartHooks`, this really improved the code structure IMO so thanks for this comment! Besides being cleaner, it also partially addressed your [other comment](https://github.com/openshift/cluster-version-operator/pull/1091#discussion_r1810683168) about exhaustively processing all messages. Launching the receiver in the posthook has the advantage of having access to the controller queue, so the receiver can queue the CM for standard controller sync. The controller queue afaik deduplicates items which means if we receive multiple messages between syncs we will still process all of them and change the controller's internal state for each, but will only sync once because the queued syncs will be deduplicated.

pkg/updatestatus/updatestatuscontroller.go
petr-muller: Partially addressed in faf8310f7f0bc4a61f699821b98ebd132b1639c9 (see https://github.com/openshift/cluster-version-operator/pull/1091#discussion_r1820945098)

go.mod
```diff
@@ -34,19 +34,31 @@ require (
 )
 
 require (
+	github.com/NYTimes/gziphandler v1.1.1 // indirect
```
wking, line 38: Checking why d0569891 added these indirect references, it's via `github.com/openshift/library-go/pkg/controller/controllercmd` (our library-go dependency is old, but the `pkg/controller/controllercmd` dependency is new with this pull). 

```console
$ git --no-pager log --oneline -1
faf8310f (HEAD, origin/pr/1091) USC: Start insight receiver as PostStartHook
$ go version
go version go1.23.1 linux/amd64
$ go mod why github.com/NYTimes/gziphandler
# github.com/NYTimes/gziphandler
github.com/openshift/cluster-version-operator/cmd
github.com/openshift/library-go/pkg/controller/controllercmd
k8s.io/apiserver/pkg/server
k8s.io/kube-openapi/pkg/handler
github.com/NYTimes/gziphandler
```

go.mod
```diff
@@ -58,20 +70,47 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
+	github.com/pkg/profile v1.3.0 // indirect
 	github.com/prometheus/procfs v0.10.1 // indirect
+	github.com/robfig/cron v1.2.0 // indirect
 	github.com/sirupsen/logrus v1.9.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/stoewer/go-strcase v1.2.0 // indirect
+	go.etcd.io/etcd/api/v3 v3.5.10 // indirect
```
wking, line 80: Seems like there's maybe room for improvement in how library-go structures packages, because the new controller cannot possibly depend on etcd APIs.  And yet:

```console
membershippb/ mvccpb/       v3rpc/        version/      
wking@penguin:~/src/cluster-version-operator$ ls vendor/go.etcd.io/etcd/api/v3/authpb/auth.p
auth.pb.go  auth.proto  
wking@penguin:~/src/cluster-version-operator$ go mod why go.etcd.io/etcd/api/v3/version
# go.etcd.io/etcd/api/v3/version
github.com/openshift/cluster-version-operator/cmd
github.com/openshift/library-go/pkg/controller/controllercmd
github.com/openshift/library-go/pkg/config/serving
k8s.io/apiserver/pkg/server/options
k8s.io/apiserver/pkg/storage/storagebackend/factory
go.etcd.io/etcd/client/v3
go.etcd.io/etcd/api/v3/version
```

But they're OCP dependencies, so dragging them over to this repo's vendor directory doesn't extend our product-scoped exposure.

wking (approved):
/lgtm

Very nice!

DavidHurta (commented):
/hold
Going through the install files. Some notes. Feel free to unhold if needed. Nothing worth blocking (eh, maybe the namespace discussion). Going through the code next.

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,171 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
DavidHurta: As mentioned, not worth blocking as the ConfigMap will be dropped by the time we go GA.

pkg/updatestatus/controlplaneinformer_test.go
```diff
@@ -0,0 +1,942 @@
...
+	}{
+		{
+			name: "Progressing=True",
+			progressing: configv1.ClusterOperatorStatusCondition{
+				Status:  configv1.ConditionTrue,
+				Reason:  "ReasonTrue",
+				Message: "MessageTrue",
+			},
+			expectedStatus: metav1.ConditionTrue,
+			// TODO(muller): Extract to API
```
DavidHurta, line 805: Could you please clarify what do we need to extract?

install/0000_00_cluster-version-operator_03_usc_deployment-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,52 @@
...
+          capabilities:
+            drop:
+            - ALL
+          allowPrivilegeEscalation: false
+        resources:
+          requests:
+            cpu: 20m
+            memory: 50Mi
+        terminationMessagePolicy: FallbackToLogsOnError
+      nodeSelector:
```
DavidHurta, line 44: Hmm, the build02 does not have a node with such a role. This led me to https://access.redhat.com/solutions/7028754 and https://issues.redhat.com/browse/OCPBUGS-16964. So applying this on an older cluster with a new version would fail right now, I suppose.

install/0000_00_cluster-version-operator_03_usc_deployment-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,52 @@
...
+      k8s-app: update-status-controller
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      name: update-status-controller
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted-v2
+      labels:
```
DavidHurta, line 22: > The restricted-v2 SCC is the most restrictive of the SCCs that is included by default with the system. [1]

Nice :sunglasses:

[1] https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html

install/0000_00_cluster-version-operator_03_usc_deployment-DevPreviewNoUpgrade.yaml
DavidHurta: My attempt to add a multi-line comment using Goland has failed. `node-role.kubernetes.io/control-plane` is the role.

pkg/updatestatus/controlplaneinformer_test.go
petr-muller: Good catch. Nothing to extract now, I already did it. The API PR was missing these constants.

install/0000_00_cluster-version-operator_03_usc_deployment-DevPreviewNoUpgrade.yaml
petr-muller: Disappointing. I'll revert this to `master`

install/0000_00_cluster-version-operator_01_usc_serviceaccount-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    kubernetes.io/description: SA used by the Update Status Controller.
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/feature-set: DevPreviewNoUpgrade
+  name: update-status-controller
+  namespace: openshift-cluster-version
```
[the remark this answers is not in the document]
petr-muller: Great suggestion. I had to digest the change (I was not too fond of adding another `openshift-` namespace, it's already not easy to find stuff in an openshift cluster) but running in a separate namespace will protect us from inadvertently rely on some CVO-specific bits which is a good thing.

install/0000_00_cluster-version-operator_03_usc_deployment-DevPreviewNoUpgrade.yaml
```diff
@@ -0,0 +1,68 @@
...
+    type: Recreate
+  template:
+    metadata:
+      name: update-status-controller
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+      labels:
+        k8s-app: update-status-controller
+    spec:
+      automountServiceAccountToken: false
```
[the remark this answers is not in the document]
petr-muller: We tested this with David and found out there's probably no good reason, and simplified the deployment as a result.

install/0000_00_cluster-version-operator_02_usc_rbac-DevPreviewNoUpgrade.yaml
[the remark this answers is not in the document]
petr-muller: Good suggestion though because we'll likely start with copying this RBAC when we start using Status API

petr-muller:
/retest

I'm starting to be mildly suspicious about the hard failing HyperShift jobs :monocle_face:

petr-muller:
Adding new (even DevPreviewNoUpgrade-gated) manifests [interacts with HyperShift unfortunately](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1091/pull-ci-openshift-cluster-version-operator-master-e2e-hypershift-conformance/1853764751801192448/artifacts/e2e-hypershift-conformance/dump/artifacts/namespaces/clusters-c01d0e18fc19f1e0757b/core/pods/logs/cluster-version-operator-665c5789d5-8sr59-bootstrap.log):

```
error: error parsing /var/payload/manifests/0000_00_update-status-controller_03_deployment-DevPreviewNoUpgrade.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".ReleaseImage":interface {}(nil)}
```

The [HyperShift code](https://github.com/openshift/hypershift/blob/84fecafa57504139ae6f0623a789369eda05c56f/control-plane-operator/controllers/hostedcontrolplane/cvo/reconcile.go#L306-L331) does not take the featureset tags into account unfortunately. Thanks a lot @DavidHurta for help with the troubleshooting.

I pulled the manifests from this PR for now because that feels like an issue that should be addressed separately, and it is not helpful trying to deal with these issues in the context of a big PR that brings the code scaffolding.

petr-muller:
/label no-qe

This PR only adds code, and the code is not yet used in any form of deployed OCP.

PR #1100, opened 2024-10-31
NO-JIRA: task graph: improve TestRunGraph tests

petr-muller:
The test was a layered amalgam with multiple types of inputs (task
graph, paralellism, `errorOn` which ended up being used for arbirary
callback use cases) and expected outputs (most used by just one or two
testcases).

- document test inputs
- improve invariants to be better readable for individual testcases
- describe desired properties as invariants
- restructure `errorOn` as `callbacks` to be called for individual
  tasks rather than a single big function with `if` matching tasks

### DIFF
--- a/pkg/payload/task_graph_test.go
+++ b/pkg/payload/task_graph_test.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"os"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/payload/task_graph_test.go
```diff
@@ -711,27 +783,34 @@ func TestRunGraph(t *testing.T) {
...
-		errorOn  func(t *testing.T, name string, ctx context.Context, cancelFn func()) error
+		name  string
+		nodes []*TaskNode
+		// maxParallelism is a RunGraph parameter affecting how many nodes are processed in parallel
+		maxParallelism int
+		// sleep is a duration to sleep in each task to simulate work
+		sleep time.Duration
+		// callbacks allows providing a method to be called for a specific task, to simulate work, return error or trigger
+		// cancellations
+		callbacks map[string]callbackFn
```
wking, line 796: possibly doc the `*` special-case here, from `for _, callbackKey := range []string{task.Manifest.OriginalFilename, "*"}`?  An alternative would be to have the keys be regular expressions, but :shrug:, you already have it working your way, and "task name or `*` wildcard" isn't complicated to explain.

wking (approved):
I'm setting `no-qe` for this pure unit-test refactor.

/lgtm
/label no-qe

pkg/payload/task_graph_test.go
wking: /hold in case you want to do this^

Feel free to lift the hold without making changes, if you're fine as things stand.

petr-muller: Good point, will do

petr-muller: Addressed

/hold cancel

petr-muller:
Pretty sure this does not cause 

```
: [bz-etcd][invariant] alert/etcdHighCommitDurations should not be at or above
```

wking (approved):
/lgtm

PR #1106, opened 2024-11-11
OCPBUGS-31462: Set `openshift.io/required-scc: privileged` annotation in `version` pods

DavidHurta:
[pkg/cvo/updatepayload.go: Set openshift.io/required-scc: privileged for version pods](https://github.com/openshift/cluster-version-operator/pull/1106/commits/9761a5b5ca847acece8f7537f78ef8a089faec6e)
Utilize the `openshift.io/required-scc` annotation [1] to pin the required SCC to
`version` pods. This will ensure that any existing custom SCCs in the cluster will
not have an effect on the `version` pods.

The `privileged` default SCC [2] was chosen as the pod accesses and modifies
host `/etc/` files. To do that, a pod must run as root and must also pass SELinux
permission checks. This is currently achieved by the pod running as a privileged
root. For such permission, the `privileged` default SCC is required.

Using the `hostmount-anyuid` default SCC is not sufficient for the existing code
as the pod is not able to pass the SELinux permissions checks.
Additional SELinux, host file system, and/or code changes would be needed.

In the future, we may implement such changes or try to use a local persistent
volume [3] as running the version pod as privileged root is undesirable for
the pod's goal of copying files into another pod.

Some of the other alternatives are modifications to the current architecture
of two separate pods or using a different type of volume.

[1] https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth
[2] https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth
[3] https://kubernetes.io/docs/concepts/storage/volumes/#local

### DIFF
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -227,6 +227,9 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 			Labels: map[string]string{
 				"k8s-app": "retrieve-openshift-release",
 			},
+			Annotations: map[string]string{
+				"openshift.io/required-scc": "privileged",
+			},
 		},
 		Spec: corev1.PodSpec{
 			ActiveDeadlineSeconds: deadline,


pkg/cvo/updatepayload.go
```diff
@@ -225,6 +225,11 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 		Spec: batchv1.JobSpec{
 			ActiveDeadlineSeconds: deadline,
 			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"openshift.io/required-scc": "privileged",
```
DavidHurta, line 230: > 	This is the most relaxed SCC and should be used only for cluster administration. Grant with caution [[1](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth)]

I am not sure whether just using the `hostmount-anyuid` value with some modifications would be enough. However, other docs state that privileged pods must be used for `hostPath` volumes.

> A hostPath volume in an OpenShift Container Platform cluster mounts a file or directory from the host node’s filesystem into your pod. Most pods will not need a hostPath volume, but it does offer a quick option for testing should an application require it.	
> The cluster administrator must configure pods to run as privileged. This grants access to pods in the same node.  [[2](https://docs.openshift.com/container-platform/4.17/storage/persistent_storage/persistent_storage_local/persistent-storage-hostpath.html)]

DavidHurta:
/hold
Exploring https://github.com/openshift/cluster-version-operator/pull/1106#discussion_r1837017719

pkg/cvo/updatepayload.go
DavidHurta: Using the hostmount-anyuid default SCC is not sufficient for the existing code
as the pod is not able to pass the SELinux permissions checks.
Additional SELinux, host file system, and/or code changes would be needed.

DavidHurta:
Failing `ci/prow/e2e-aws-ovn-techpreview` does not seem to be relevant to the PR's changes. I will not override the CI as the PR will most probably merge after the `acknowledge-critical-fixes-only` label requirement is lifted. The tests may heal in the meantime.

petr-muller (approved):
LGTM

DavidHurta:
/override e2e-aws-ovn-techpreview

The PR does not cause the tests to fail. The job itself seems to be permanently failing at the moment.

jiajliu:
cc @xingxingxia @gangwgr

petr-muller (approved):

PR #1132, opened 2025-01-07
WIP: pkg/cvo/availableupdates: Add Recommend:... conditions for each matching or eval-failed risk

wking:
To make it easier for clients to explain individual risks, instead of only having access to a generic `MultipleReasons` `reason` and a long `message` covering all the risks.


pkg/cvo/availableupdates.go
```diff
@@ -503,36 +518,66 @@ func newRecommendedReason(now, want string) string {
...
+			}
+		}
 		if match, err := conditionRegistry.Match(ctx, risk.MatchingRules); err != nil {
 			recommended.Status = newRecommendedStatus(recommended.Status, metav1.ConditionUnknown)
 			recommended.Reason = newRecommendedReason(recommended.Reason, recommendedReasonEvaluationFailed)
-			errorMessages = append(errorMessages, unknownExposureMessage(risk, err))
+			msg := unknownExposureMessage(risk, err)
+			errorMessages = append(errorMessages, msg)
+			conditions = append(conditions, metav1.Condition{
+				Type:    fmt.Sprintf("%s:%s:%s", ConditionalUpdateConditionTypeRecommended, recommendedReasonEvaluationFailed, cleanName),
```
petr-muller, line 548: What is the reason for including the `recommendedReasonEvaluationFailed` in the type of the condition when we fail to evaluate it? To me a `Recommended:RiskName` condition with `Status=Unknown` and `Reason=Something` sounds like it does the job...

Also, it seems that `:` is not a valid character [for condition type](https://github.com/kubernetes/apimachinery/blob/3e8e52d6a1259ada73f63c1c7d1fad39d4ba9fb4/pkg/apis/meta/v1/types.go#L1609-L1613):

```
	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
	// +kubebuilder:validation:MaxLength=316
```

But `(dns1123SubdomainFmt/)?(qualifiedNameFmt)` sounds like we can invent something like `recommended.openshift.io/RiskName` with sounds good to me.

wking: Thanks; I've pivoted to `recommended/...` with a0a3b548 -> c5053bac.  I'm a week minus on `.openshift.io` there; seems unlikely that other folks besides us would be injecting their own conditions, and we don't do that kind of DNS scoping for our core conditions like `Failing` and otherwise :shrug:  But I don't mind adding the scoping if folks think it's worthwhile.

petr-muller: yeah `recommended/risk` sounds good to me :+1:

PR #1143, opened 2025-01-30
OTA-1418: USC: Drop unknown messages from the API

petr-muller:
Assume the informers, in each message, include a list of all active insight IDs that infomer knows about. The Status API can then drop the unknown insights formerly reported by that informer (=insights present in the API but not known to be active by the responsible informer)

This is only a part of [OTA-1418](https://issues.redhat.com//browse/OTA-1418). The remaining changes will be as follows:

- Do not drop the unknown insights right away but make them expire after some some time of not being renewed (handle the case where informer starts up and starts sending messages about insights as it discovers them)
- Actually make the informers remember their insights and populate the `knownInsights` field of the message
- Add additional `deletedInsights` field to the message, to allow informers explicitly tell the API some insight went away
- Make the USC respect `deletedInsights`
- Make informers fill `deletedInsight`
- Potentially add a out-of-API memory cache for insights to be expired / deleted but allow them to reappear with formerly known content

Pair-coded with @DavidHurta

### DIFF
--- a/pkg/updatestatus/updatestatuscontroller.go
+++ b/pkg/updatestatus/updatestatuscontroller.go
@@ -30,9 +30,9 @@ import (
 // we have the Status API we need to serialize ourselves anyway.
 type informerMsg struct {
 	informer string
-	// knownInsight contains the UIDs of insights known by the informer, so the controller can remove insights formerly
+	// knownInsights contains the UIDs of insights known by the informer, so the controller can remove insights formerly
 	// reported by the informer but no longer known to it (e.g. because the informer was restarted and the culprit
-	// condition ceased to exist in the meantime).
+	// condition ceased to exist in the meantime). The `uid` of the insight in the message payload is always assumed to be known, and is not required to be included in `knownInsights` by the informers (but informers can do so).
 	knownInsights []string
 
 	uid     string
@@ -222,7 +222,7 @@ func (c *updateStatusController) updateInsightInStatusApi(msg informerMsg) {
 }
 
 // removeUnknownInsights removes insights from the status API that are no longer reported as known to the informer
-// that originally reported them
+// that originally reported them.
 // Assumes the statusApi field is locked.
 func (c *updateStatusController) removeUnknownInsights(message informerMsg) {
 	known := sets.New(message.knownInsights...)
--- a/pkg/updatestatus/updatestatuscontroller.go
+++ b/pkg/updatestatus/updatestatuscontroller.go
@@ -32,7 +32,8 @@ type informerMsg struct {
 	informer string
 	// knownInsights contains the UIDs of insights known by the informer, so the controller can remove insights formerly
 	// reported by the informer but no longer known to it (e.g. because the informer was restarted and the culprit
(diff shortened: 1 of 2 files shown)


petr-muller:
/label no-qe
Increment

Changed code only applied in DevPreview

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -213,7 +219,21 @@ func (c *updateStatusController) updateInsightInStatusApi(msg informerMsg) {
 			klog.Infof("USC :: Collector :: Insight (uid=%s) content did not change (len=%d)", msg.uid, len(updatedContent))
 		}
 	}
+}
 
+// removeUnknownInsights removes insights from the status API that are no longer reported as known to the informer
+// that originally reported them
```
DavidHurta, line 225: nit:
```suggestion
// that originally reported them.
```

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -29,6 +30,10 @@ import (
 // we have the Status API we need to serialize ourselves anyway.
 type informerMsg struct {
 	informer string
+	// knownInsight contains the UIDs of insights known by the informer, so the controller can remove insights formerly
+	// reported by the informer but no longer known to it (e.g. because the informer was restarted and the culprit
+	// condition ceased to exist in the meantime).
+	knownInsights []string
```
DavidHurta, line 36: We should add a comment that the `knownInsights` of a `message` do not contain the `uid` of the message. Something along these lines.

DavidHurta (changes requested):

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -29,6 +30,10 @@ import (
 // we have the Status API we need to serialize ourselves anyway.
 type informerMsg struct {
 	informer string
+	// knownInsight contains the UIDs of insights known by the informer, so the controller can remove insights formerly
+	// reported by the informer but no longer known to it (e.g. because the informer was restarted and the culprit
+	// condition ceased to exist in the meantime).
```
petr-muller, line 35:
```suggestion
	// condition ceased to exist in the meantime). The `uid` of the insight in the message payload should be assumed to be known, and may or may not be included `knownInsights` by the informers.
```

pkg/updatestatus/updatestatuscontroller.go
petr-muller: Addressed in 0dfbb5cc7dfac9165d93bf9e38e3e6f8bab06530

DavidHurta (approved):

petr-muller:
There is an ongoing problem with `azure-deprovision-sp-and-custom-role` step, already [discussed on Slack](https://redhat-internal.slack.com/archives/CBN38N3MW/p1739368208849509)

```
ERROR: No subscriptions found for 979a5836-9a52-4e1f-9fec-39af8798ddfe.
```

This is a deprovision step so if it is the only thing that failed, tests have already ran and hence succeeded, so it is fine to override

Holding to prevent the retester from acting, to give me space to issue overrides when appropriate
/hold

PR #1150, opened 2025-02-07
OTA-1418: USC: Drop unknown insights after grace period 

petr-muller:
When we first see an insight in the API that is no longer known by the informer that originally reported it, mark it for expiration instead of dropping it right away. If it becomes known again, unmark the expiration. If it does not become known until expired, drop it from the API.

### DIFF
--- a/pkg/updatestatus/updatestatuscontroller.go
+++ b/pkg/updatestatus/updatestatuscontroller.go
@@ -25,6 +25,10 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
+const (
+	unknownInsightGracePeriod = 60 * time.Second
+)
+
 // informerMsg is the communication structure between informers and the update status controller. It contains the UID of
 // the insight and the insight itself, serialized as YAML. Passing serialized avoids shared data access problems. Until
 // we have the Status API we need to serialize ourselves anyway.
@@ -72,6 +76,9 @@ func isStatusInsightKey(k string) bool {
 	return strings.HasPrefix(k, "usc.")
 }
 
+// insightExpirations is UID -> expiration time map
+type insightExpirations map[string]time.Time
+
 // updateStatusController is a controller that collects insights from informers and maintains a ConfigMap with the insights
 // until we have a proper UpdateStatus API. The controller maintains an internal desired content of the ConfigMap (even
 // if it does not exist in the cluster) and updates it in the cluster when new insights are received, or when the ConfigMap
@@ -98,11 +105,20 @@ type updateStatusController struct {
 		sync.Mutex
 		cm *corev1.ConfigMap
 
+		// unknownInsightExpirations is a map of informer -> map of UID -> expiration time. It is used to track insights
+		// that were reported by informers but are no longer known to them. The API keeps unknown insights until they
+		// expire. If an insight is reported as known again before it expires, it is removed from the map.
+		// TODO (muller): Needs to periodically rebuilt to avoid leaking memory
+		unknownInsightExpirations map[string]insightExpirations
+
 		// processed is the number of insights processed, used for testing
 		processed int
 	}
 
 	recorder events.Recorder
+	// TODO: Get rid of this and use `clock.Clock` in all controllers, passed from start.go main function's
+	// controllercmd.ControllerContext
+	now func() time.Time
 }
 
 // newUpdateStatusController creates a new update status controller and returns it. The second return value is a function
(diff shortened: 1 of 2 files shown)


pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -117,6 +128,7 @@ func newUpdateStatusController(
 	c := &updateStatusController{
 		configMaps: coreClient.CoreV1().ConfigMaps(uscNamespace),
 		recorder:   uscRecorder,
+		now:        time.Now,
```
DavidHurta, line 131: The `newUpdateStatusController` function is called from the `Run` function https://github.com/openshift/cluster-version-operator/blob/3af45bc532e17e6ef5abb3eee9c28317714a348d/pkg/updatestatus/start.go#L25

The `Run` function has access to a clock using `cc.clock`. Thus we can use that value instead of using `time.Now` in multiple places. Meaning that we could pass the clock as an argument to the  `newUpdateStatusController` function and assign its `Now` method.

Something like:

```go
func newUpdateStatusController(
	coreClient kubeclient.Interface,
	coreInformers kubeinformers.SharedInformerFactory,
	recorder events.Recorder,
	clock clock.Clock,
) (factory.Controller, sendInsightFn) {
	uscRecorder := recorder.WithComponentSuffix("update-status-controller")

	c := &updateStatusController{
		configMaps: coreClient.CoreV1().ConfigMaps(uscNamespace),
		recorder:   uscRecorder,
		now:        clock.Now,
	}
```

We can then have one source of time for all the controllers. This may be extra handy when testing time-based operations across multiple parallel controllers (e.g., by using the [`k8s.io/utils/clock/testing`](https://pkg.go.dev/k8s.io/utils@v0.0.0-20241210054802-24370beab758/clock/testing#section-documentation) package).

pkg/updatestatus/updatestatuscontroller.go
```diff
@@ -98,11 +102,18 @@ type updateStatusController struct {
 		sync.Mutex
 		cm *corev1.ConfigMap
 
+		// unknownInsightExpirations is a map of informer -> map of UID -> expiration time. It is used to track insights
+		// that were reported by informers but are no longer known to them. The API keeps unknown insights until they
+		// expire. If an insight is reported as known again before it expires, it is removed from the map.
+		// TODO (muller): Needs to periodically rebuilt to avoid leaking memory
+		unknownInsightExpirations map[string]map[string]time.Time
```
DavidHurta, line 109: I can see how the `map[string]map[string]time.Time` type simplifies a lot of the written logic in the code as opposed to having a type like:

```go
unknownInsightExpirationsV2 []struct {
	informer string
	uids     []struct {
		expiration time.Time
	}
}
```

However, I would not be surprised if we would have to refactor the code in case the logic gets a little bit more complex. Keeping a mental model of maps of maps of strings is more complex than using simple structures. However, I am okay with the current approach given the circumstances and the current state.

pkg/updatestatus/updatestatuscontroller.go
petr-muller: This is a good suggestion but I am not too keen on addressing it immediately in this PR. The other controllers we have already use this `now func() metav1.Time` struct member pattern and if we want to get rid of it then its better to do everywhere at once. I'll add a `TODO` to port them to `clock.Clock`.

pkg/updatestatus/updatestatuscontroller.go
petr-muller: Not a fan of  this actually. If the logic to manipulate the data becomes complex  then we just couple with the data structure itself (=make it a method, like `Insert`, `Expire` etc) . Map is actually a more native structure than implementing a map as a slice of `key, value` structure - adding a map semantics to a slice IMO adds mental overhead than reducing it. Plus it has complexity consequences: access is no longer O(constant) but O(n), and similar for removal (which we'll do often).

That said, this should not be thought of as map of strings to maps of strings to times. It is a map of informer names to informer expiration records, where an expiration record is a map of insight names to times when they expire. We'll probably eventually abstract it, like introduce an alias for the internal map, and maybe even for the external one, and add methods to encapsulate desired behavior but because I think these data structures and operations we need from them may still change, I avoided that for now.

petr-muller: But I guess introducing an alias for the internal map helps a bit even now so lets add it

DavidHurta: My point was mainly about simplicity rather than the exact implementation; I agree that my example of using slices was not the best idea, and my wording could have been better. Thanks for the detailed arguments.

Yes, type aliases and potential usage of methods after the code becomes more stable/complex. That sounds good!

Just trying to keep the code simple :stars:

Even things such as the new `insightExpirations` type make the code easier to read, IMO.

pkg/updatestatus/updatestatuscontroller.go
DavidHurta: Sounds good.

DavidHurta (approved):

PR #1162, opened 2025-02-19
OTA-1418: Decribe informer->usc communication

petr-muller:
Going from USC-scrapes-all-insights periodically to unidirectional
message passing makes insight lifecycle a bit harder (if an informer is
not saying anyting about an insight Status API knows about, is it just
unchanged, or did the informer lose track of it?).

List of known insight UIDs is lightweight-enough payload to be included
in every message sent to USC and the USC can (eventually) drop the
insights that the informer stops reporting as known, and can drop
the insights the informer requests to be dropped.

Unsure whether this mechanism survives until the final version but it is
reasonably robust for the current state.

### DIFF
--- a/pkg/updatestatus/updatestatuscontroller.go
+++ b/pkg/updatestatus/updatestatuscontroller.go
@@ -31,6 +31,53 @@ const (
 	unknownInsightGracePeriod = 60 * time.Second
 )
 
+// High-level description of the informers -> USC communication protocol:
+// ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

We might need a way to recover from USC losing track of insights known to the informer.  Something like a Kube-informer's list/watch cycle, where we periodically push insights (not just the ID) even if they haven't changed.  If it's been over 10m or something since that Insight was pushed in full?  But no worries from me deferring that kind of situation to future work, and what you have here makes lots of sense for the other situations you're trying to cover.

PR #1161, opened 2025-02-19
OTA-209: Add `ClusterVersionOperator` manifests

DavidHurta:
https://github.com/openshift/cluster-version-operator/pull/1137 will be broken down into smaller PRs.

This the initial PR.

For more information regarding the resources, see https://github.com/openshift/enhancements/blob/master/enhancements/update/cvo-log-level-api.md.

This pull request references https://issues.redhat.com//browse/OTA-209.

PR #1163, opened 2025-02-24
OTA-209: Add CVOConfiguration controller

DavidHurta:
The controller has an empty reconciliation logic as of now.
The logic will be implemented in a follow-up PR. The goal of this PR is to introduce a new CVO controller, which can optionally (depending on the state of the `CVOConfiguration` feature gate) create a new informer.


DavidHurta:
The PR contains the manifests as of now as well. However, the manifests will be merged in https://github.com/openshift/cluster-version-operator/pull/1161. Right now the PR contains them for testing purposes. I will drop the commits once the PR merges (or I suppose that the GitHub will start to ignore the commits).

pkg/cvo/configuration/configuration.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	config.started = true
+	return nil
+}
+
+func (config *ClusterVersionOperatorConfiguration) Sync(ctx context.Context, key string) error {
+	if !config.started {
+		panic("ClusterVersionOperatorConfiguration instance was not properly started before its synchronization.")
+	}
```
petr-muller, line 91: So we recently had a case in USC where we used https://pkg.go.dev/sync#Once for a similar kind of initialization - first call to `sync` performs an initialization through `sync.Once` and then you do not need to do a check like this. Could we use a similar pattern here?

pkg/start/start.go
```diff
@@ -220,6 +229,7 @@ func (o *Options) run(ctx context.Context, controllerCtx *Context, lock resource
 	controllerCtx.OpenshiftConfigInformerFactory.Start(informersDone)
 	controllerCtx.OpenshiftConfigManagedInformerFactory.Start(informersDone)
 	controllerCtx.InformerFactory.Start(informersDone)
+	controllerCtx.OperatorInformerFactory.Start(informersDone)
```
petr-muller, line 232: Will this work on cluster without the CRD? :thinking:

pkg/cvo/configuration/configuration.go
DavidHurta: Thanks, I'll have a look.

pkg/start/start.go
DavidHurta: [`Start`](https://github.com/openshift/cluster-version-operator/blob/d09c388bbfd709cf226f0d0e33bab2a4dcdeaa98/vendor/github.com/openshift/client-go/operator/informers/externalversions/factory.go#L213-L216) does not block and does not return anything. So the operation should be a no-op in such a cluster.

If informers are created but no CRD for the informer is present on the cluster:

- Calling `WaitForCacheSync` on the factory will result in a permanently blocking operation as the caches are not able to sync.
- Informers will start logging errors such as (see my previous [comment](https://github.com/openshift/cluster-version-operator/pull/1163#issuecomment-2678981915))

```
E0224 16:25:40.333321  166921 reflector.go:166] "Unhandled Error" err="github.com/openshift/client-go/operator/informers/externalversions/factory.go:125: Failed to watch *v1alpha1.ClusterVersionOperator: failed to list *v1alpha1.ClusterVersionOperator: the server could not find the requested resource (get clusterversionoperators.operator.openshift.io)" logger="UnhandledError"
```

A good question as I am not sure how the code functions in case the `operator` group does not exist yet (during an install).

pkg/cvo/configuration/configuration.go
DavidHurta: I had a look at the structure (thanks for the suggestion, I did not know about the type). It looks like error handling is the obstacle of using the `sync.Once` type here. The USC only [logs](https://github.com/openshift/cluster-version-operator/blob/ff3778c368351aea28f7245743966745b14d40a8/pkg/updatestatus/nodeinformer.go#L98) errors that have occurred while executing the `once` method and continues. The CVO on the other hand tends to [shutdown on initialization errors](https://github.com/openshift/cluster-version-operator/blob/fecf267ec5d282b9dedc12ca2b6e8bc63b26187b/pkg/cvo/cvo.go#L254-L265) and is thus much stricter. The `sync.Once` does not provide simple error handling in this case.

The configuration controller here has two major methods: `Start` and `Sync`. `Start` initializes and starts informers. If an error occurs, it fails the CVO (this is consistent with other similar logic in the CVO [[1](https://github.com/openshift/cluster-version-operator/blob/fecf267ec5d282b9dedc12ca2b6e8bc63b26187b/pkg/cvo/cvo.go#L437-L439), [2](https://github.com/openshift/cluster-version-operator/blob/fecf267ec5d282b9dedc12ca2b6e8bc63b26187b/pkg/cvo/cvo.go#L254-L265)]).

We could maybe modify the `Start` method to return a wrapper that has the exportable `Sync` method. Thus, a developer must call `Start` to get a variable to call the `Sync` method. What do you think? Just an idea.

```go
// Start initializes and starts the configuration's informers. Must be run before Sync is called.
// Blocks until informers caches are synchronized or the context is cancelled.
func (config *ClusterVersionOperatorConfiguration) Start(ctx context.Context) (*ClusterVersionOperatorConfigurationWrapper, error) {
        // ...
	return &ClusterVersionOperatorConfigurationWrapper{config: config}, nil
}

type ClusterVersionOperatorConfigurationWrapper struct {
	config *ClusterVersionOperatorConfiguration
}

func (w *ClusterVersionOperatorConfigurationWrapper) Sync(ctx context.Context, key string) error {
	startTime := time.Now()
	klog.V(2).Infof("Started syncing CVO configuration %q", key)
	defer func() {
		klog.V(2).Infof("Finished syncing CVO configuration (%v)", time.Since(startTime))
	}()

	desiredConfig, err := w.config.lister.Get(ClusterVersionOperatorConfigurationName)
	if apierrors.IsNotFound(err) {
		return nil
	}
	if err != nil {
		return err
	}
	return w.config.sync(ctx, desiredConfig)
}
```

I am open to suggestions.

pkg/start/start.go
DavidHurta: I have just realized that a cluster without the CRD is a normal Default cluster at the moment :sweat_smile: The CI seems to be happy, and as mentioned, the operation should be a no-op in such a cluster. To check how the PR may affect a cluster during an install, I have created the https://github.com/openshift/release/pull/62184.

pkg/cvo/configuration/configuration.go
```diff
@@ -0,0 +1,111 @@
...
+// Callers should use this with an informer.
+func (config *ClusterVersionOperatorConfiguration) clusterVersionOperatorEventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(_ interface{}) {
+			config.queue.Add(config.queueKey)
+		},
+		UpdateFunc: func(_, _ interface{}) {
+			config.queue.Add(config.queueKey)
+		},
+		DeleteFunc: func(_ interface{}) {},
```
petr-muller, line 50: If there is an instance that configured us, and the instance got deleted, then we should `sync` and reset to some default. But maybe we can address this case later - can we add a todo?

pkg/cvo/configuration/configuration.go
petr-muller: Yeah, I skimmed the code and I think it's good as it is. It's annoying how hard it is to write controllers without the library-go lib.

petr-muller (approved):
/hold

I'd be more comfortable if Trevor took a look but I do not want to block this

pkg/cvo/configuration/configuration.go
DavidHurta: Could we make the logic to set default values when the resource is not found? To not make distinctions between `not found` vs. `got deleted`.  The motivation is to simplify the logic. 

The `DeleteFunc` would add the key to queue, the `sync` would set default values when `err == isNotFound`.

petr-muller: Yeah, that sounds reasonable :+1:

DavidHurta: Updated the `DeleteFunc` and added a TODO.

petr-muller (approved):

DavidHurta:
/label no-qe

As per internal discussion, labeling as no-qe. The OTA-209 will be tested via post-merge testing. The card consists of multiple smaller PRs. There is nothing extra to be verified by the QE on the PR. Most of the functionalities are gated and any significant regression should be discoverable via CI.

PR #1166, opened 2025-03-03
OCPBUGS-52258: pkg/cvo/updatepayload: Context around ValidateDirectory calls

wking:
These usually either succeed or give `os.IsNotExist`, but in [some cases can return errors like][1]:

    Unable to download and prepare the update: stat /etc/cvo/updatepayloads/7WNaprXJNWTsPAepCHJ00Q/release-manifests/release-metadata: permission denied.

To make it easier to debug those situations, this commit adds context to errors near `ValidateDirectory` calls, so it's more clear that the issue is the CVO's attempts to read `/etc/cvo/...` failing, and not an error with the `version-...` Pod, registry access, etc.

Also while I'm in this space, I've replaced the old `errors.Wrap` in `ImageForShortName` with `fmt.Errorf(...%w, ...)`, now that the Go standard libraries support error-wrapping like that.  One fewer 3rd-party dependency call-site.

[1]: https://issues.redhat.com/browse/OCPBUGS-52258


jiajliu:
cc @JianLi-RH

petr-muller (approved):

PR #1169, opened 2025-03-13
OTA-1403: Move the CVO CLI Application to a Separate Folder

DavidHurta:
So that we can have multiple CLI applications in the `cmd` folder.

For example, `cmd/cluster-version-operator/`, `cmd/cvo-tests/`, `cmd/cluster-version-operator-v2/`.

Being done to tidy the folder for https://github.com/openshift/cluster-version-operator/pull/1168.


DavidHurta:
I do not have a strong opinion regarding naming. Feel free to propose a different alternative. Maybe `cmd/cvo` is enough?

DavidHurta:
>  I do not have a strong opinion regarding naming. Feel free to propose a different alternative. Maybe cmd/cvo is enough?



And then we can call the new extension `cvo-tests` instead of `cluster-version-operator-tests`, `openshift-tests-extension`, `tests-extension`, or `tests`.

petr-muller (approved):

DavidHurta:
\label no-qe

The change is minimal. The only instance of referencing the files is the `hack/build-go.sh` script, which was updated. Any issues would result in compilation/build errors.

DavidHurta:
/label no-qe

Typo in https://github.com/openshift/cluster-version-operator/pull/1169#issuecomment-2722055142.

PR #1184, opened 2025-04-29
OTA-1531: [1/x] cvo: refactor option processing

petr-muller:
Extract option validation to a new `ValidateAndComplete()` method, and run it separate from `Run()` through cobra `PreRunE` feature.

Refactor the `AlwaysEnabledCapabilities` processing to be consistent with how the other options are processed. The `NewControllerContext` is a method on `Options` struct, so it does not make sense to pass it some options through the struct, and some options through method parameters. I went further and separated the validation (performed with all other validation) from conversion, resulting in two smaller methods instead one larger one. The result is slightly less efficient but more readable.

The reason to do this refactor is that I need to clean up the `NewControllerContext` constructor. I will need to do some changes around, and removing the explicit pass of one option-originated argument (while all others are passed by being `Options` members) will make the resulting code less messy. The reasoning for the changes:

1. I need `NewControllerContext` to consume AlwaysEnabledCapabilities from the struct
2. To consume it from the struct, it must be validated beforehands
3. If it is validated beforehands, we do not need to track unknown and valid capabilities separately and validation can be simpler
4. It would be useful if `Run` could generally rely on running with "all options were validated" assumption

The reason for this refactor is that I need to clean up the `NewControllerContext` constructor and the `Context` structure so that they do not become messy after the changes needed for [OTA-1531](https://issues.redhat.com//browse/OTA-1531). Check out https://github.com/openshift/cluster-version-operator/pull/1182 for the probe coarse implementation of the full thing.

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


petr-muller:
/hold

Should wait until branching at

pkg/start/start.go
```diff
@@ -511,7 +517,7 @@ func (o *Options) NewControllerContext(cb *ClientBuilder, alwaysEnableCapabiliti
 		o.PromQLTarget,
 		o.InjectClusterIdIntoPromQL,
 		o.UpdateService,
-		alwaysEnableCapabilities,
+		stringsToCapabilities(o.AlwaysEnableCapabilities),
```
petr-muller, line 520: This is the main point of the change - all other `o.Members` are passed into `NewControllerContext` via `Options`, and the method relies on earlier validation. Treating one `Options` member in a different way makes the interface messy.

wking (approved):
/lgtm

PR #1186, opened 2025-05-06
EXAMPLE: Motivating example for OTA-1531

petr-muller:
Illustrates the caveat of the current approach to determining which gates are enabled and disabled in the cluster. The current approach performs this initialization late in the execution (once the initial payyload is fully loaded) so there is an awkward period on CVO startup where enabled gates cannot be checked and the programmer cannot gate behavior.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -315,6 +315,10 @@ func (optr *Operator) LoadInitialPayload(ctx context.Context, startingRequiredFe
 		return nil, fmt.Errorf("unable to create a configuration client: %v", err)
 	}
 
+	if optr.enabledFeatureGates.SomethingInPayloadLoading() {
+		fmt.Println("SomethingInPayloadLoading feature gate is enabled")
+	}
+
 	customSignatureStore := &customsignaturestore.Store{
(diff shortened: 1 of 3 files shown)


petr-muller:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1186/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change/1919795510286225408

There you go:

```
: [sig-architecture] platform pods in ns/openshift-cluster-version should not exit an excessive amount of times
{  1 containers with multiple restarts

namespace/openshift-cluster-version node/ci-op-9sxh8i8f-c736d-6jxdj-master-0 pod/cluster-version-operator-86654d7c4c-lcsn6 uid/fb7cf998-2a31-4f9e-9f3c-7c7a7b060701 container/cluster-version-operator restarted 24 times at:
non-zero exit at 2025-05-06 18:06:57.252676209 +0000 UTC m=+332.601532538: cause/Error code/2 reason/ContainerExit penshift.io/feature-set=TechPreviewNoUpgrade
...
panic: CVO feature flags were used before they were initialized
```

PR #1203, opened 2025-06-06
NO-JIRA: tests: fix `make integration-test` target

petr-muller:
Fix the name of the script called, so that I can pass rehearsals in https://github.com/openshift/release/pull/65742 and then eventually remove the script in https://github.com/openshift/cluster-version-operator/pull/1202

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ test:
 .PHONY: test
 
 integration-test:
-	./hack/integration-test.sh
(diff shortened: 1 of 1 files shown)


petr-muller:
/label no-qe

No change necessary, will be exercised in https://github.com/openshift/release/pull/65742 and later https://github.com/openshift/cluster-version-operator/pull/1203 CI jobs.

PR #1205, opened 2025-06-20
NO-JIRA: resourceapply/rbac: fix resource diff logs

petr-muller:
Previously the logs were wrong and misleading because they were created by diffing a resource _after modification_ which means the actual change is never shown in the diff. These diffs should be created by comparing the resource before and after it is modified.

Partially related to OCPBUGS-36246 but is just a prep for an eventual fix, so marking as NO-JIRA.


petr-muller:
/hold

Need to fix some copy-pasta bugs

PR #1206, opened 2025-06-20
OTA-1531: Rework error handling in FeatureGate processing

petr-muller:
- The `FeatureGate` GET is now made over an informer client which is backed by a cache, so polling to smooth over non-isNotFound errors no longer makes sense. The GET runs after a cache sync succeeded earlier so no errors are expected (and if an error happens, it is unlikely to be resolved by a retry)
- Waiting for cache sync was not bounded; impose a 30s timeout on startup (this is similar to the deadline imposed by the poll before we started to use the informer client)

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


pkg/start/start.go
```diff
@@ -244,48 +243,42 @@ func (o *Options) getOpenShiftVersion() string {
...
 	featureGates := configInformerFactory.Config().V1().FeatureGates().Lister()
 	configInformerFactory.Start(ctx.Done())
-	configInformerFactory.WaitForCacheSync(ctx.Done())
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	for key, synced := range configInformerFactory.WaitForCacheSync(ctx.Done()) {
+		if !synced {
+			return startingFeatureSet, cvoGates, fmt.Errorf("failed to sync %s informer cache: %v", key.String(), ctx.Err())
```
hongkailiu, line 257: nit:

```suggestion
			return startingFeatureSet, cvoGates, fmt.Errorf("failed to sync %s informer cache: %w", key.String(), ctx.Err())
```

pkg/start/start.go
```diff
@@ -244,48 +243,42 @@ func (o *Options) getOpenShiftVersion() string {
...
 
 func (o *Options) processInitialFeatureGate(ctx context.Context, configInformerFactory configinformers.SharedInformerFactory) (configv1.FeatureSet, featuregates.CvoGates, error) {
+	var startingFeatureSet configv1.FeatureSet
+	var cvoGates featuregates.CvoGates
+
 	featureGates := configInformerFactory.Config().V1().FeatureGates().Lister()
 	configInformerFactory.Start(ctx.Done())
-	configInformerFactory.WaitForCacheSync(ctx.Done())
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
hongkailiu, line 252: Just want to understand your reasoning on the number:
`30s` here are `WaitForCacheSync` (maybe `5s`?) + poll timeout (`25s`)?

pkg/start/start.go
```diff
@@ -244,48 +243,42 @@ func (o *Options) getOpenShiftVersion() string {
...
-			return true, nil
-		}
-	}); err != nil {
-		if lastError != nil {
-			return "", cvoGates, lastError
-		}
-		return "", cvoGates, err
+	gate, err := featureGates.Get("cluster")
+	switch {
+	case err != nil && apierrors.IsNotFound(err):
```
hongkailiu, line 268: nit: non-nil checking seems redundant:

https://github.com/openshift/cluster-version-operator/blob/25f02d1c2b32020b20ef9a6e1e6c0b3cd4f8496a/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L526-L528

```suggestion
	case apierrors.IsNotFound(err):
```

pkg/start/start.go
petr-muller: Well the whole functionality is "get the feature gate from the cluster". Previously we had 30s-ish timeout for the `Get` operation to succeed (previously the client was not informer-based, it was a get directly to apiserver), either through client-go internal retries or the previous poll timeout.

Now when the client is informer-based, the operation that can fail / timeout is the cache sync, not the get, so we apply the same timeout on the cache sync like we previously did for the get.

petr-muller:
/label no-qe

I do not think this needs to be explicitly tested, I'm not even sure how we would approach it; we'd need a broken apiserver on CVO startup.

PR #1208, opened 2025-07-01
OTA-1480: Add readonlyRootFilesystem

dusk125:


pkg/cvo/updatepayload.go
```diff
@@ -264,6 +264,9 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 				setContainerDefaults(corev1.Container{
 					Name:    "rename-to-final-location",
 					Command: []string{"mv", tmpDir, dir},
+					SecurityContext: &corev1.SecurityContext{
+						ReadOnlyRootFilesystem: ptr.To(true),
+					},
```
petr-muller, line 269: Should the init containers (L236-L262) have this too? If not, can the commit + pr description contain reasoning for that?. 

Or even better, change the `setContainerDefaults` to default to `true` so that the right thing is default and whatever needs a `false` would need to override?

jiajliu:
> /label qe-approved

/label qe-approved

dusk125:
@petr-muller if you have a chance to review again, please see my comments about the "move" containers.

pkg/cvo/updatepayload.go
```diff
@@ -246,15 +246,15 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
...
 					Name: "move-operator-manifests-to-temporary-directory",
 					Command: []string{
-						"mv",
+						"cp",
 						filepath.Join(payload.DefaultPayloadDir, payload.CVOManifestDir),
 						filepath.Join(tmpDir, payload.CVOManifestDir),
 					},
 				}),
 				setContainerDefaults(corev1.Container{
 					Name: "move-release-manifests-to-temporary-directory",
```
wking, line 255: nit: rename `move-...` here and above to `copy-...` to keep up with the `mv` -> `cp` implementation change?

PR #1218, opened 2025-08-01
Draft: test changes not the final outputs

dis016:
testing out test case execution with OTE framework not to be merged

### DIFF
--- a/test/cvo/cvo.go
+++ b/test/cvo/cvo.go
@@ -1,12 +1,26 @@
 package cvo
 
 import (
-	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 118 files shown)


JianLi-RH:
hi @dis016 can we start with a very simple CVO case? I mean let's only copy necessary files from openshift-tests-private to cluster-version-operator? If you don't mind, I want to create another PR for same work. I will also avoid to use OC client (please refer https://redhat-internal.slack.com/archives/CJ1J9C3V4/p1755771964418519?thread_ts=1755602807.015039&cid=CJ1J9C3V4)

PR #660, opened 2021-09-20
Bug 2005581: install/0000_00_cluster-version-operator_03_deployment: Explicit kube-api-access

wking:
[This content][2] is [injected by an admission webhook][1].  When we started removing not-in-manifest volumes in 83faa6e716 (#654), the cluster-version operator started removing the webhook-injected volume, leading to the cluster-version operator crash-looping on updates from 4.8 to 4.9 with messages [like][3]:

    F0920 13:23:23.565439       1 start.go:24] error: error creating clients: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable

With this commit, we follow the precedent of the Kubernetes API server's own manifest: openshift/cluster-kube-apiserver-operator#1142.

[1]: https://github.com/kubernetes/kubernetes/blob/2f68346fbb6246961ce0a3176418630950aea500/plugin/pkg/admission/serviceaccount/admission.go#L53-L54
[2]: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=2005581

PR #607, opened 2021-06-14
Revert "prevent pod deployment deadlock due to custom SA projected volume injection"

wking:
This reverts commit 9bde7bb89677fe166303c040d220b763c3291234, #585.

It was a temporary workaround in 4.8 to bridge a gap between Kube-API-server and Kube-controller-manager during 4.7 -> 4.8 updates. Once the cluster has reconciled 4.8, we will no longer need the crutch.  And 4.8 has now forked off master, so we can revert now, and that revert will go out in 4.9.

PR #1225, opened 2025-08-18
OCPBUGS-60579: Suppuort automountServiceAccountToken modify

lance5890:
1. I noticed that @wking has tried to remove the `automountServiceAccountToken` in https://github.com/openshift/cluster-version-operator/pull/607 when upgrade to 4.9(this is not suppuorted when upgrade, because cvo do not suppuort remvoing the `automountServiceAccountToken`), and finally revert this in https://github.com/openshift/cluster-version-operator/pull/660
2. removing the  `automountServiceAccountToken` not only needed by cvo, but also by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1896, and https://github.com/openshift/cluster-kube-controller-manager-operator/pull/858, and https://github.com/openshift/cluster-kube-scheduler-operator/pull/571, and https://github.com/openshift/library-go/pull/2001
3. removing the `automountServiceAccountToken` is ok when install a new cluster, but will encounter a `error: error creating clients: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable` issue because the cvo do not suppuort setting the new  `automountServiceAccountToken`  value (which will use the existing false value in the existing obj)

4. maybe we should merge this PR first ,and then the related cluster operators
5. and finally remove the  `automountServiceAccountToken` for cvo in the next upgrade version

### DIFF
--- a/install/0000_00_cluster-version-operator_03_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_03_deployment.yaml
@@ -22,7 +22,7 @@ spec:
       labels:
         k8s-app: cluster-version-operator
     spec:
-      automountServiceAccountToken: false
+      automountServiceAccountToken: false  # removed in the next version.
       containers:
       - name: cluster-version-operator
         image: '{{.ReleaseImage}}'
--- a/lib/resourcemerge/core_test.go
+++ b/lib/resourcemerge/core_test.go
@@ -62,7 +62,7 @@ func TestEnsurePodSpec(t *testing.T) {
 				AutomountServiceAccountToken: boolPtr(false)},
 
 			expectedModified: true,
-			expected:         corev1.PodSpec{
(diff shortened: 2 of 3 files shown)


install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       labels:
         k8s-app: cluster-version-operator
     spec:
-      automountServiceAccountToken: false
+      automountServiceAccountToken: false  # removed in the next version.
```
lance5890, line 25: the `automountServiceAccountToken` should be deleted in the next version

tchap:
Taking over this change in https://github.com/openshift/cluster-version-operator/pull/1264

/close

PR #1235, opened 2025-09-25
OTA-1585: test: Rename an existing test to comply with the OTE integration guide

DavidHurta:
Per the integration guide for OTE, each test must have a clearly defined owner (by using a `[Jira:Component]` tag) or an entry in the ci-test-mapping repository. Naming the owner as `"Cluster Version Operator"` with spaces, as this is the format used to distinguish our component in the mapping repository [[1]].

Renaming tests should not be done without a proper setup, as this can lose a test's history. However, this test was never executed, and it acts as a signal of the OTE integration.

[1]: https://github.com/openshift-eng/ci-test-mapping/blob/b2ca5844ead12ede1cb72cae34321f3215beb6e3/pkg/components/clusterversionoperator/component.go#L14

### DIFF
--- a/.openshift-tests-extension/openshift_payload_cluster-version-operator.json
+++ b/.openshift-tests-extension/openshift_payload_cluster-version-operator.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "[cvo-testing] cluster-version-operator-tests should support passing tests",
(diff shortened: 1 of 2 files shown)


DavidHurta:
/verified bypass

Renaming a test. The test is not executed anywhere.

PR #1249, opened 2025-10-22
NO-ISSUE: OTA-1605 Create first e2e test OCP-42543

JianLi-RH:
In this PR, I finished 2 things:
1. MustGetKubeClient() function helps us create an Clientset instance which can be used to connect to cluster.
2. Update the README.md file, we can follow it to create our e2e test cases.

/cc @wking @DavidHurta @PratikMahajan @hongkailiu @jhou1 @dis016 @jiajliu

### DIFF
--- a/test/cvo/cvo.go
+++ b/test/cvo/cvo.go
@@ -2,17 +2,85 @@ package cvo
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"log"
+	"time"
 
 	g "github.com/onsi/ginkgo/v2"
 	o "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientmetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
+	applyconfigurationspolicyv1 "k8s.io/client-go/applyconfigurations/policy/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
 
 	v1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/cluster-version-operator/test/utilities"
 )
+
+func CreateServiceAccount(client *kubernetes.Clientset, accountName string, clusterRole string, namespace string) (token string, err error) {
+
+	_, err = client.CoreV1().ServiceAccounts(namespace).Get(context.TODO(), accountName, metav1.GetOptions{})
+
+	if err == nil {
+		token, err := client.CoreV1().ServiceAccounts(namespace).CreateToken(context.TODO(), accountName, &authenticationv1.TokenRequest{}, metav1.CreateOptions{})
+		return token.String(), err
+	}
+
+	account := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      accountName,
+			Namespace: namespace,
+		},
+	}
+	_, err = client.CoreV1().ServiceAccounts(namespace).Create(context.TODO(), account, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	rb := &rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s:%s:%s", namespace, clusterRole, accountName),
+			Namespace: namespace,
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     "cluster-admin",
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      accountName,
+				Namespace: namespace,
+			},
+		},
+	}
+	_, err = client.RbacV1().ClusterRoleBindings().Create(context.TODO(), rb, metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	newToken, err := client.CoreV1().ServiceAccounts(namespace).CreateToken(context.TODO(), accountName, &authenticationv1.TokenRequest{}, metav1.CreateOptions{})
+	return newToken.String(), err
+}
+
+func DeleteServiceAccount(client *kubernetes.Clientset, accountName string, clusterRole string, namespace string) {
+	name := fmt.Sprintf("%s:%s:%s", namespace, clusterRole, accountName)
+	err := client.RbacV1().ClusterRoleBindings().Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
+		panic("failed to delete ClusterRoleBindings")
+	}
+
+	err = client.CoreV1().ServiceAccounts(namespace).Delete(context.TODO(), accountName, metav1.DeleteOptions{})
+	if err != nil {
+		panic("failed to delete ServiceAccount")
+	}
+}
 
 var _ = g.Describe(`[Jira:"Cluster Version Operator"] cluster-version-operator-tests`, func() {
 	g.It("should support passing tests", func() {
@@ -80,5 +148,107 @@ var _ = g.Describe("[Jira:Cluster Version Operator] The cluster version operator
 
 			fmt.Println("success")
 		}
+	})
+
+	g.It(`Precheck with oc adm upgrade recommend`, g.Label("Conformance", "Low", "70980"), func() {
+
+		g.By("create a namespace")
+		ns := "ns-70980"
+		tmpNs := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}}
+		kubeclient.CoreV1().Namespaces().Create(context.TODO(), tmpNs, metav1.CreateOptions{})
+
+		defer func() {
+			kubeclient.CoreV1().Namespaces().Delete(context.TODO(), ns, metav1.DeleteOptions{})
+		}()
+
+		g.By("create a deployment")
+		deploymentName := "hello-openshift"
+		containerName := "hello-openshift"
+		containerImage := "openshift/hello-openshift:invaid"
+		deployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: deploymentName,
+			},
+			Spec: appsv1.DeploymentSpec{
+				Replicas: ptr.To(int32(2)), // Number of desired replicas
+				Selector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"app": containerName,
+					},
+				},
+				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: map[string]string{
+							"app": containerName,
+						},
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:  containerName,
+								Image: containerImage,
+								Ports: []corev1.ContainerPort{
+									{
+										ContainerPort: 80,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+		kubeclient.AppsV1().Deployments(ns).Create(context.TODO(), deployment, metav1.CreateOptions{})
+
+		defer func() {
+			kubeclient.AppsV1().Deployments(ns).Delete(context.TODO(), deploymentName, metav1.DeleteOptions{})
+		}()
+
+		err := wait.Poll(1*time.Minute, 3*time.Minute, func() (bool, error) {
+			allPods, err := kubeclient.CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{})
+			if err != nil {
+				log.Fatalf("Error listing pods: %v", err)
+			}
+			for _, pod := range allPods.Items {
+				if pod.Status.Phase == corev1.PodRunning {
+					return true, errors.New("there are pods running: " + pod.Name)
+				}
+			}
+			return true, nil
+		})
+		allPods, _ := kubeclient.CoreV1().Pods(ns).List(context.TODO(), metav1.ListOptions{})
+		fmt.Printf("there are %v pods\n", len(allPods.Items))
+		for _, pod := range allPods.Items {
+			fmt.Printf("	- Pod: %s - %s\n", pod.Name, pod.Status.Phase)
+		}
+		o.Expect(kerrors.IsNotFound(err)).To(o.BeFalse(), "The NotFound error should not occur")
+
+		g.By("create a PodDisruptionBudget")
+		pdbName := "my-pdb"
+		pdb := &applyconfigurationspolicyv1.PodDisruptionBudgetApplyConfiguration{
+			ObjectMetaApplyConfiguration: &clientmetav1.ObjectMetaApplyConfiguration{
+				Name:      &pdbName,
+				Namespace: &ns,
+			},
+			Spec: &applyconfigurationspolicyv1.PodDisruptionBudgetSpecApplyConfiguration{
+				MaxUnavailable: &intstr.IntOrString{
+					Type:   intstr.Int,
+					IntVal: 1,
+				},
+			},
+		}
+		kubeclient.PolicyV1().PodDisruptionBudgets(ns).Apply(context.TODO(), pdb, metav1.ApplyOptions{})
+
+		defer func() {
+			kubeclient.PolicyV1().PodDisruptionBudgets(ns).Delete(context.TODO(), pdbName, metav1.DeleteOptions{})
+		}()
+
+		g.By("wait some minutes, there is a critical issue for PDB")
+		token, _ := CreateServiceAccount(kubeclient, "monitorer", "cluster-admin", ns)
+		defer func() {
+			DeleteServiceAccount(kubeclient, "monitorer", "cluster-admin", ns)
+		}()
+		// TODO: get alert
+		fmt.Println(token)
 	})
 })


JianLi-RH:
Tested on my local machine:
```
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$ make update
hack/build-go.sh
Using version from git...
Building binaries into _output/linux/amd64
Building github.com/openshift/cluster-version-operator cluster-version-operator-tests binary (v1.0.0-1482-gd9440005-dirty)
Compressing the cluster-version-operator-tests binary
Building github.com/openshift/cluster-version-operator cluster-version-operator binary (v1.0.0-1482-gd9440005-dirty)
hack/update-test-metadata.sh
Using version from git...
successfully updated metadata
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$ _output/linux/amd64/cluster-version-operator-tests list
[
  {
    "name": "[Jira:Cluster Version Operator] cluster-version-operator-tests Author:jianl-High-42543-the removed resources are not created in a fresh installed cluster",
    "labels": {
      "42543": {},
      "High": {},
      "cvo": {}
    },
    "resources": {
      "isolation": {}
    },
    "source": "openshift:payload:cluster-version-operator",
    "lifecycle": "blocking",
    "environmentSelector": {}
  }
]
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$
```
Run the case:
```
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$ _output/linux/amd64/cluster-version-operator-tests list | jq '.[] | select(.name | contains("42543")) | .name' | xargs _output/linux/amd64/cluster-version-operator-tests run-test
  Running Suite:  - /home/jianl/1_code/cluster-version-operator
  =============================================================
  Random Seed: 1761292866 - will randomize all specs

  Will run 1 of 1 specs
  ------------------------------
  [Jira:Cluster Version Operator] cluster-version-operator-tests Author:jianl-High-42543-the removed resources are not created in a fresh installed cluster [cvo, High, 42543]
  /home/jianl/1_code/cluster-version-operator/test/cvo/cvo.go:22
    STEP: Validate resource with 'release.openshift.io/delete: "true"' annotation is not installed @ 10/24/25 16:01:06.862
  • [1.428 seconds]
  ------------------------------

  Ran 1 of 1 Specs in 1.428 seconds
  SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
[
  {
    "name": "[Jira:Cluster Version Operator] cluster-version-operator-tests Author:jianl-High-42543-the removed resources are not created in a fresh installed cluster",
    "lifecycle": "blocking",
    "duration": 1428,
    "startTime": "2025-10-24 08:01:06.858981 UTC",
    "endTime": "2025-10-24 08:01:08.287644 UTC",
    "result": "passed",
    "output": "  STEP: Validate resource with 'release.openshift.io/delete: \"true\"' annotation is not installed @ 10/24/25 16:01:06.862\n"
  }
]
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$ 
```

test/README.md
hongkailiu: We have https://github.com/openshift/cluster-version-operator/blob/main/cmd/cluster-version-operator-tests/README.md already.

That folder is the entry of the test extension. I feel it fits better to hold the README.md there.

test/README.md
```diff
@@ -0,0 +1,82 @@
+# Cluster Version Operator Tests
+
+We hold all e2e tests and related common methods in this folder.
+Test script are using client-go, here is the document: https://pkg.go.dev/k8s.io/client-go
```
hongkailiu: I do not think we need to mention the dependency of the test, unless there are something special for this particular one: `client-go`.

test/utilities/connection.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// getClient
+// @Description	create a kubernetes.Clientset object
+// @Create 		jianl Oct 23 2025
+// @Return 		(Instance of kubernetes.Clientset, error)
```
hongkailiu: Are those are not valid format for GoDocs?
See https://go.dev/doc/comment#func for ref.

test/cvo/cvo.go
```diff
@@ -1,12 +1,34 @@
...
+	defer g.GinkgoRecover()
+	var oc *kubernetes.Clientset
+
+	g.BeforeAll(func() {
+		oc, _ = utilities.GetClient()
+		o.Expect(oc).NotTo(o.BeNil())
+	})
+
+	g.It("Author:jianl-High-42543-the removed resources are not created in a fresh installed cluster", g.Label("High", "42543"), func() {
+		g.By("Validate resource with 'release.openshift.io/delete: \"true\"' annotation is not installed")
```
hongkailiu:
```suggestion
		g.By("checking if service openshift-cloud-credential-operator/controller-manager-service does not exist")
```

.gitignore
```diff
@@ -2,3 +2,4 @@
 bin/
 .DS_Store
 _output
+.openshift-tests-extension
```
hongkailiu: I am wondering why the file showed up.

.openshift-tests-extension/openshift_payload_cluster-version-operator.json
hongkailiu: Why is it deleted?

From https://github.com/openshift/cluster-version-operator/pull/1235/commits/cfab71da3b522300cbbf48aeab1b58644634d8ed i feel it needs an update tho.

test/README.md
JianLi-RH: This description is helpful for new contributors who are unfamiliar with client-go.

test/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Prerequisites
+- Install golang, the version should match it in go.mod. Here is the how-to doc https://go.dev/doc/install
+- Have the environment variable `KUBECONFIG` set pointing to your cluster.
+
+## Compile the executable binary
+In root folder, run below command to build executable binary:
+```
+$ make update
```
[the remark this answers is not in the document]
JianLi-RH: Because I see `make update` do more work than `make build`, and sometimes it will cause issues when there is no .openshift-tests-extension/openshift_payload_cluster-version-operator.json

```
update: build
	hack/update-test-metadata.sh
.PHONY: update
```
I will replace then command.

test/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+The executable binary can be found in ${BIN_PATH} folder, BIN_PATH is set in hack/build-info.sh, in linux it is _output/linux/amd64/ folder.
+
+## Run test case
+
+- Run below command to list all available tests:
```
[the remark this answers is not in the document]
JianLi-RH: Simplified the description, but I still want to keep the examples.

.gitignore
JianLi-RH: I added this line otherwise we need to update the file every time when we add/remove tests

cmd/cluster-version-operator-tests/README.md
```diff
@@ -4,6 +4,16 @@ It integrates [openshift-tests-extension](https://github.com/openshift-eng/opens
 cluster-version-operator which allows openshift components to contribute tests to openshift-tests' suites with
 extension binaries.
 
+## Prerequisites
```
hongkailiu: I will continue the discussion https://github.com/openshift/cluster-version-operator/pull/1249#discussion_r2469998683 here.

My point is if any of those Prerequisites is not satisfied, the commands below will error out and then the user can act accordingly.

* if Go is not install, `make build` will fail.
* if `$KUBECONFIG` is not set, then `cluster-version-operator-tests run-suite` will fail if it is required. Moreover, perhaps not all of the tests need a `$KUBECONFIG`.

cmd/cluster-version-operator-tests/README.md
```diff
@@ -21,4 +31,10 @@ After [installing-ginkgo](https://onsi.github.io/ginkgo/#installing-ginkgo):
 $ ginkgo ./test/...
 ```
 
+## Using binary 
+```console
+$ _output/linux/amd64/cluster-version-operator-tests run-test <test case name>
+```
+Note: `test case name` can be found in .openshift-tests-extension/openshift_payload_cluster-version-operator.json
```
hongkailiu: This will be replaced by the above `cluster-version-operator-tests list` cmd.

test/utilities/connection.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"os"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// MustGetKubeClient will create a kubernetes.Clientset object instance.
+//
+// The function returns an instance of kubernetes.Clientset when it success
```
hongkailiu:
```suggestion
```

.gitignore
JianLi-RH: Removed this line

test/cvo/cvo.go
```diff
@@ -1,12 +1,37 @@
...
+var _ = g.Describe("[Jira:Cluster Version Operator] The cluster version operator", g.Ordered, g.Label("cvo"), func() {
+	defer g.GinkgoRecover()
+	var client *kubernetes.Clientset
+
+	g.BeforeAll(func() {
+		client = utilities.MustGetKubeClient()
+	})
+
+	//author: jianl@redhat.com
+	g.It("the removed resources are not created in a fresh installed cluster", g.Label("High", "42543"), func() {
```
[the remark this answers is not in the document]
JianLi-RH: Updated the case title in polarion

JianLi-RH:
Case passed in my local machine:
```
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$ ginkgo ./test/...
Running Suite: CVO Suite - /home/jianl/1_code/cluster-version-operator/test/cvo
===============================================================================
Random Seed: 1762243336

Will run 1 of 1 specs
•

Ran 1 of 1 Specs in 1.085 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2.313942766s
Test Suite Passed
[jianl@jianl-thinkpadt14gen4 cluster-version-operator]$
```

test/cvo/cvo.go
```diff
@@ -1,12 +1,33 @@
...
+	})
+
+	g.It("should not install resources annotated with release.openshift.io/delete=true", g.Label("Conformance", "High", "42543"), func() {
+		g.By("checking if Service openshift-cloud-credential-operator/controller-manager-service does not exist")
+		_, err = client.CoreV1().Services("openshift-cloud-credential-operator").Get(context.TODO(), "controller-manager-service", metav1.GetOptions{})
+		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue(), "The NotFound error should occur")
+
+		g.By("checking if ClusterRoleBindings default-account-openshift-machine-config-operator does not exist")
+		_, err = client.RbacV1().ClusterRoleBindings().Get(context.TODO(), "default-account-openshift-machine-config-operator", metav1.GetOptions{})
+		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue(), "The NotFound error should occur")
```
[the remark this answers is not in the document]
JianLi-RH: Yes, I checked OCP 420's manifest, the 3rd one is not applicable, it applies to OCP 4.21.

test/cvo/cvo.go
```diff
@@ -1,12 +1,33 @@
...
+	defer g.GinkgoRecover()
+	var err error
+	var client *kubernetes.Clientset
+
+	g.BeforeAll(func() {
+		client = utilities.MustGetKubeClient()
+	})
+
+	g.It("should not install resources annotated with release.openshift.io/delete=true", g.Label("Conformance", "High", "42543"), func() {
+		g.By("checking if Service openshift-cloud-credential-operator/controller-manager-service does not exist")
```
jiajliu: Original test [coverage](https://github.com/openshift/openshift-tests-private/blob/master/test/extended/ota/cvo/cvo.go#L716-L759) is to check all resources with "release.openshift.io/delete=true" should not be created, shall we keep the test coverage of the case? Only checking two of them sounds like a risk of missing bugs because it will not expose the issue caused by other resources.

hongkailiu: I have to admit that it is better to be thorough.

With that said, this case is not my turning point to start to depend on oc-cli (`oc admin extract release` seems useful in [the original implementation](https://github.com/openshift/openshift-tests-private/blob/b735913caa47aa358bf8df0750d1048940e815ce/test/extended/ota/cvo/utils.go#L538).

As far as I know, the code of `oc admin extract release` is not that exportable and thus I would not go that direction either.

Choosing two manifests that we know as a fact are with the annotation (and thus are not supposed to be installed on the cluster) to check sounds sufficient to reach the coverage on the CVO code (this is my blind guess without checking tho).

If you know a case that you think we would not find a _simple_ enough solution without oc-cli, let us work on that now and see how it goes.

I totally understand that the existing extended e2e tests use `oc` a lot. To me, it is an chance to review alternatives. But I can also understand it if we stay with `oc`.

hongkailiu:
/lgtm

Since this is the first test to migrate, it is better to start with consensus. 😄 

Let me see if I convinced @jiajliu in https://github.com/openshift/cluster-version-operator/pull/1249#discussion_r2492514934

/hold

test/cvo/cvo.go
jiajliu: The oc adm release output is used in various places, so this issue extends beyond the current situation. We'll collect and share them as we find during migration, for now, the team will likely prioritize the ones that don't rely on the command. If low risk is seen here, proceeding is acceptable. Leave the final decision to the owner.

test/cvo/cvo.go
```diff
@@ -1,12 +1,33 @@
...
+	o "github.com/onsi/gomega"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/cluster-version-operator/test/utilities"
 )
 
-var _ = Describe("[Jira:Cluster Version Operator] cluster-version-operator-tests", func() {
-	It("should support passing tests", func() {
```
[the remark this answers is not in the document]
DavidHurta: Its future usefulness is questionable. However, for the moment, I am using it to validate that the https://github.com/openshift/origin/pull/30316 PR is functional. Would it be okay for me rename to the component in the `[Jira:Cluster Version Operator] cluster-version-operator-tests should support passing tests` test case in another PR to get it merged quickly, then merge the origin PR, and for this PR to just ignore the test case? The test case is not resource-intensive, and it provides a useful "if you can't see this test, then something is wrong with the integration" signal. 

/cc @hongkailiu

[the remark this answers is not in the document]
DavidHurta: I don't mind just keeping the test forever.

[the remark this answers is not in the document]
JianLi-RH: @DavidHurta I just added back the case and rename it as you mentioned. It is a independent test suite.

hongkailiu:
/lgtm

@JianLi-RH 

Free to cancel the hold after 24h or so (giving @DavidHurta the chance of final check).

hongkailiu:
/testwith openshift/cluster-version-operator/main/e2e-agnostic-ovn http://openshift/origin#30316

hongkailiu:
/testwith openshift/cluster-version-operator/main/e2e-agnostic-ovn openshift/origin#30316

test/cvo/cvo.go
```diff
@@ -1,12 +1,39 @@
 package cvo
```
DavidHurta: **Comment not related to the code**

TBD, commits, commit titles, commit descriptions, PR title, JIRA reference

test/cvo/cvo.go
```diff
@@ -1,12 +1,39 @@
...
+	var err error
+	var client *kubernetes.Clientset
+
+	g.BeforeAll(func() {
+		client = utilities.MustGetKubeClient()
+	})
+
+	g.It(`should not install resources annotated with release.openshift.io/delete=true`, g.Label("Conformance", "High", "42543"), func() {
+		g.By("checking if Service openshift-cloud-credential-operator/controller-manager-service does not exist")
+		_, err = client.CoreV1().Services("openshift-cloud-credential-operator").Get(context.TODO(), "controller-manager-service", metav1.GetOptions{})
```
DavidHurta: TODO context

.openshift-tests-extension/openshift_payload_cluster-version-operator.json
```diff
@@ -1,12 +1,27 @@
 [
   {
-    "name": "[Jira:Cluster Version Operator] cluster-version-operator-tests should support passing tests",
+    "name": "[Jira:\"Cluster Version Operator\"] cluster-version-operator-tests should support passing tests should support passing tests",
```
DavidHurta: Please note when deleting and recreating the metadata file in a separate commit. The file should only be updated via the `make update` normally. We are intentionally bypassing this as of the moment, but that should be clearly communicated in a commit description as in "we know what we are doing, in the future, do not do that, do it properly or else history is lost"

test/cvo/cvo.go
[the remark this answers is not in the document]
DavidHurta: Thanks. It helps with the integration. It also provides a signal of the integration working in the future. Such tests are also in several other OpenShift repositories.

test/cvo/cvo.go
```diff
@@ -1,12 +1,39 @@
...
+		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue(), "The NotFound error should occur")
+
+		g.By("checking if ClusterRoleBindings default-account-openshift-machine-config-operator does not exist")
+		_, err = client.RbacV1().ClusterRoleBindings().Get(context.TODO(), "default-account-openshift-machine-config-operator", metav1.GetOptions{})
+		o.Expect(kerrors.IsNotFound(err)).To(o.BeTrue(), "The NotFound error should occur")
```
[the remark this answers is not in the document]
DavidHurta: Testing for existence of objects we expect have a manifest annotation `release.openshift.io/delete: "true"` may work for the current minor version; however, there is no guarantee that such test would cover the expected functionality in the next minor release or any future minor release.

.openshift-tests-extension/openshift_payload_cluster-version-operator.json
DavidHurta: I will be updating the README file to clearly state this for the future.

test/cvo/cvo.go
```diff
@@ -1,12 +1,39 @@
...
+var _ = g.Describe(`[Jira:"Cluster Version Operator"] The cluster version operator`, g.Ordered, g.Label("cvo"), func() {
+	defer g.GinkgoRecover()
+	var err error
+	var client *kubernetes.Clientset
+
+	g.BeforeAll(func() {
+		client = utilities.MustGetKubeClient()
+	})
+
+	g.It(`should not install resources annotated with release.openshift.io/delete=true`, g.Label("Conformance", "High", "42543"), func() {
```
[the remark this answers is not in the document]
JianLi-RH: Yes, it is polarion ID, I have no idea if it is common but I think it is helpful for us when we want to read the case from polarion.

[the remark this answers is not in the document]
JianLi-RH: `High` is a property of the case on polarion, with it we can create test suites with same level, then we can run test suite by case level.
For example, we want to run all `High` level test cases on a special version, then we can follow below steps:
```
	ext.AddSuite(extension.Suite{
		Name:       "high level test cases",
		Parents:    []string{"openshift/conformance"},
		Qualifiers: []string{`"High" in labels`},
	})
```
Then run it:
```
$ _output/linux/amd64/cluster-version-operator-tests run-suite "high level test cases"
```

test/cvo/cvo.go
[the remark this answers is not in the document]
DavidHurta: > we should update the manifests to check in the CURRENT minor version.

The manifests can also be added and removed via backports, thus affecting all existing minor versions.

> If we find that the annotation comes and goes very often on the manifests over minor versions

Who would be monitoring this?

> When the first build of each minor version is available, we update the list. 

Again, the manifests can also be removed and added via backports.

> I do not see we need to invest in that direction right now.

Me as well, but for other reasons.

---

This means we would need to monitor all minor releases periodically via automation, as our attention is better utilized on more important efforts. This, in conclusion, means at minimum an initial development cost for the automation tooling and its maintenance.

Otherwise, the current solution may theoretically stop covering the expected functionality as soon as the day after merging. As a result, we, the OTA team, cannot rely on the test to cover the expected functionality. Then we would have a passing signal for a functionality, and we cannot depend on the signal being useful. That’s a rather large regression in the migrated test. I would personally rather drop the test case than go this route.

---

However, we still have the option to loop through all manifests and thus not change the covered functionality. This could be done by implementing the original logic, not using the `oc` CLI, or by importing origin packages. This could be done using a Go library for processing container images to fetch and extract the release image. We can then loop through the manifests as done originally. This would be done in collaboration with the developer and quality engineers. No regression in covered functionality, no extreme imports hopefully, no implementation of automation for a single test case.

What do you think?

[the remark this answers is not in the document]
JianLi-RH: @DavidHurta @hongkailiu Let's not waste any more time on this case, I will find another case to automate.

JianLi-RH:
@hongkailiu @DavidHurta I just removed the case, right now the PR only finished one thing: MustGetKubeClient().
Then we can start creating test cases in new PRs.

JianLi-RH:
/hold

I want to investigate more about client-go.

test/cvo/cvo.go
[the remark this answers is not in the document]
hongkailiu: > The manifests can also be added and removed via backports, thus affecting all existing minor versions.

Interesting. I thought it was a thing that takes place only when a new minor is cut.
Now it sounds like a manifest can come and go (and so is the annotation `release.openshift.io/delete=true` on any manifest).

> Otherwise, the current solution may theoretically stop covering the expected functionality as soon as the day after merging.

Very true. What a headache!

We need to ensure that the annotation on the chosen manifest to check is still there (I still think it can be achieved by a periodic job. I can bring more details if you like this direction).

> This could be done by implementing the original logic, not using the oc CLI, or by importing origin packages.

The implementation is not trivial at all (to me).
I have worked on a few PRs in `oc adm release extract` and the code there is nasty.
Even if we cooked up our own thing for this, how to deal with the next case where it uses another `oc` cmd. I want to avoid re-implementing our own `oc` all over.

I am trying to find a tradeoff for this case. But "no implementation of automation for a single test case." sounds not acceptable to you which I totally understand as it is not perfect.

I am wondering within the same minor version, how often does it happen?
The annotation is removed, or the manifest with the annotation is removed.
If it does not happen often (or even not at all), a simple automation to check this is fine with me. It will ensure our CI signal is safe to trust.
I understand also that this way has disadvantage too. We find a tradeoff for this case, but may not find one for the next.

But if you really want to implement our own extract manifests from a release image, I am fine with it too.
Maybe I overestimate the difficulty.

Or ... just saying ... perhaps spawning `oc` process in the test as what origin does is not a bad idea. 😉 

I believe it may start a vote:

Opt1: Choose 1 or 2 manifests with the annotation and ensure they are not installed. Make a job to signal us if the chosen manifests are no longer valid.

Opt2: Implement manifests extraction in CVO with GO. Then loop over them.

Opt3: Utilize `oc` to extract the manifests.

Opt4: others?

And then go with it until the next revisit as a team.

What do you think?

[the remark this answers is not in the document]
hongkailiu: @JianLi-RH 

We will have to face this issue, whenever you migrate an existing case it uses `oc` in the implementation and you do not find an easy workaround for it. I honestly do not know how many such cases there are.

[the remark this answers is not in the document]
JianLi-RH: > the manifest with the annotation is removed.

This has a very low frequency, last time we hit the issue is at 4.20, the manifest with "release.openshift.io/delete" was introduced in 4.7 or 4.8, and the file was removed in 4.20.

PR #1262, opened 2025-11-19
OPNET-678: Update openshift/api vendoring

cybertron:
Specifically looking to pull in the dnsRecordsType field from https://github.com/openshift/api/pull/2460


cybertron:
/retest-required

API throttling again.

cybertron:
Looks like this will be pulled in by https://github.com/openshift/cluster-version-operator/pull/1269 which is already approved.

PR #1266, opened 2025-11-28
OCPBUGS-65621: add dedicated service account to crb, cvo and version pod

ehearne-redhat:
### What

- Add dedicated service account to CVO and version pods in `openshift-cluster-version` namespace. 
- Add cluster role binding and attach dedicated service account to it.

### Why

- Default service account should not be used on OpenShift components.


install/0000_00_cluster-version-operator_02_roles.yaml
```diff
@@ -11,4 +11,17 @@ roleRef:
...
+metadata:
+  name: cvo-dedicated-sa-scc-privileged-access
+subjects:
+- kind: ServiceAccount
+  name: update-payload-dedicated-sa
+  namespace: openshift-cluster-version
+roleRef:
+  kind: ClusterRole
+  name: system:openshift:scc:privileged
+  apiGroup: rbac.authorization.k8s.io
```
ehearne-redhat: I'm unsure whether this is needed or not... I initially added it because I was looking at the wrong test `e2e-upgrade-into-change`, and did not see the `update-payload-dedicated-sa` service account on `version-` pod in must gather files. However, I did see them in `e2e-upgrade-out-of-change` . 

I'm thinking it might be necessary as the updatepayload pod contains the `openshift.io/required-scc: privileged` annotation . So, I'll leave it for the reviewer to decide. I can test without too if that suits.

pkg/cvo/updatepayload.go
```diff
@@ -232,6 +232,7 @@ func (r *payloadRetriever) fetchUpdatePayloadToDir(ctx context.Context, dir stri
 			},
 		},
 		Spec: corev1.PodSpec{
+			ServiceAccountName:    "update-payload-dedicated-sa",
```
wking: > Default service account should not be used on OpenShift components

I think "Don't attach privileges to the default service account in OpenShift namespaces" makes a lot of sense, but I'm less clear on the downsides of Pods using the default service account if that account comes with no privileges.  This version Pod does not need any Kube API privileges.  In fact, it doesn't need any network communication at all, it's just shoveling bits around between the local container filesystem and a volume-mounted host directory.  Can we leave it running in the default service account, or is there a way to request no-service-account-at-all?  Maybe that's effectively what [`automountServiceAccountToken: false`][1] does?

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting

install/0000_00_cluster-version-operator_02_service_account.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cvo-dedicated-sa
```
wking: nit: can we use `name: cluster-version-operator`?  That avoids the `cvo` acronym, avoids the `sa` echo of the `kind`, and relies on `kubernetes.io/description` to express the dedicated-ness.  The cluster-version-operator-ness is the only aspect we need to keep in the `name` to help avoid colliding with other ServiceAccounts in this namespaces, and the other attributes can be covered in those non-name locations.

install/0000_00_cluster-version-operator_02_service_account.yaml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: openshift-cluster-version
+  annotations:
+    kubernetes.io/description: Dedicated Service Account for the Cluster Version Operator.
+    include.release.openshift.io/self-managed-high-availability: "true"
```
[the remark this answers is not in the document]
ehearne-redhat: They appear to be using their own service account and setup --> https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/v2/assets/cluster-version-operator/deployment.yaml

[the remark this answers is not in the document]
ehearne-redhat: If I can't get my changes to work it might not be a bad idea to replicate their own setup. Thanks for pointing me in the right direction! :)

ehearne-redhat:
Will address failing tests tomorrow.

install/0000_00_cluster-version-operator_02_ZZ_roles.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
wking: I'd rather avoid the `ZZ` just to sort these after the new ServiceAccount file.  It's fine if you want to renumber sibling files to make more room.  Ideally leave some space too, so other folks can add additional stuff later without having to renumber siling files again as soon.

install/0000_00_cluster-version-operator_02_ZZ_roles.yaml
```diff
@@ -0,0 +1,44 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: cluster-version-operator-1
```
wking: Redundant vs. the `cluster-version-operator` you declare down at the end of this file?

install/0000_00_cluster-version-operator_02_ZZ_roles.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: cluster-version-operator-payload
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+subjects:
+- kind: ServiceAccount
+  name: update-payload-dedicated-sa
+  namespace: openshift-cluster-version
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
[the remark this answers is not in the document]
ehearne-redhat: I've commented it out to test this. If proves true in testing, I'll remove entirely.

install/0000_00_cluster-version-operator_03_deployment.yaml
```diff
@@ -23,6 +23,7 @@ spec:
         k8s-app: cluster-version-operator
     spec:
       automountServiceAccountToken: false
+      serviceAccountName: cluster-version-operator
```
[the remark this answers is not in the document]
ehearne-redhat: I've added it to the bootstrap manifest. I'll wait to see how tests behave before squashing the commits into a narrative commit, or a collection of them depending.

install/0000_90_cluster-version-operator_02_roles-cluster-version-operator.yaml
```diff
@@ -5,9 +5,11 @@ metadata:
   annotations:
     kubernetes.io/description: Grant the cluster-version operator permission to perform cluster-admin actions while managing the OpenShift core.
     include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/delete: "true"
 roleRef:
   kind: ClusterRole
   name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
```
lance5890: nit: with `release.openshift.io/delete: "true"`, there is no need to add this new line here, but it doesn't hurt

ehearne-redhat:
@wking so I have all tests passing now, besides the broken `e2e-hypershift-conformance` . 

1. I was unable to add the service account to the bootstrap pod, and ordering of the manifests couldn't be changed too much. Moving the service accounts down to `90` run level meant they were applied too late and caused the deployments to fail.
2. I was unable to add the annotation for deleting the default service account CRB, as it is needed for the `upgrade-into-change` test to pass. 

I'm going to clean up the commit history now to reflect all the changes + justifications in the commit description. I will re-ping once that's done for review. :)

install/0000_00_cluster-version-operator_02_service_account-cluster-version-operator.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
DavidHurta: **Comment not relevant to the selected line**

---

The commit message mentions:

> add dedicated service account to cluster-version-operator- , version- , and update-payload- pods
This change adds a dedicated service account to the pods mentioned above.

However, there are only two pods if I am not mistaken. The cluster-version-operator via a deployment and a version pod defined in the `` file.

install/0000_00_cluster-version-operator_02b_roles-cluster-version-operator.yaml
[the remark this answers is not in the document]
DavidHurta: I believe we can do [something that was proposed by Trevor](https://github.com/openshift/cluster-version-operator/pull/1266#discussion_r2579084461) back in December. I was thinking the same although I am unable to see the exact changes, which the comment reffenced.

> With these references to ServiceAccounts defined in 0000_00_cluster-version-operator_02_service_account.yaml, we probably want to rename this file to 0000_00_cluster-version-operator_03_roles.yaml (and possibly also shift some later manifests), to get the ServiceAccounts in place before the ClusterRoleBindings that reference them are added.

That is, move `roles` to `03`, and move the siblings up. That is, move the `03_deployment` to `30_deployment` and the `04_service` to `40_service`. This should leave enough place for the future.

DavidHurta:
### The Problem

I was ready to verify the PR; however, there seems to be an issue. Not with the PR per se, but rather the verification has highlighted an existing issue with the delete annotation and the way the CVO acts during a cluster installation.

---

### Verification Process

The CI Jobs verify the approved changes.

* e2e-agnostic-ovn verifies that the CVO has the needed permissions
* e2e-agnostic-ovn-upgrade-into-change verifies the into transition, which is critical
* e2e-agnostic-ovn-upgrade-out-of-change verifies installation, the rollback transition, and verifies the update-payload Pod functionality 



Manually verifying whether the default cluster role binding is deleted in the `into-change` job:

```
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1266/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change/2036773283176648704/artifacts/e2e-agnostic-ovn-upgrade-into-change/gather-extra/artifacts/clusterrolebindings.json | jq '.items[] | select(.metadata.name | startswith("cluster-version"))'
{
  "apiVersion": "rbac.authorization.k8s.io/v1",
  "kind": "ClusterRoleBinding",
  "metadata": {
    "annotations": {
      "include.release.openshift.io/self-managed-high-availability": "true"
    },
    "creationTimestamp": "2026-03-25T12:55:51Z",
    "name": "cluster-version-operator-1",
    "ownerReferences": [
      {
        "apiVersion": "config.openshift.io/v1",
        "controller": true,
        "kind": "ClusterVersion",
        "name": "version",
        "uid": "2f4409cc-4829-4225-8b19-8b19624203f8"
      }
    ],
    "resourceVersion": "39102",
    "uid": "20dbbc44-3928-46c1-952f-6d4d55d12cee"
  },
  "roleRef": {
    "apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole",
    "name": "cluster-admin"
  },
  "subjects": [
    {
      "kind": "ServiceAccount",
      "name": "cluster-version-operator",
      "namespace": "openshift-cluster-version"
    }
  ]
}
```

It is deleted.



Checking `out-of-change`, should be present:

```
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1266/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-out-of-change/2036773283419918336/artifacts/e2e-agnostic-ovn-upgrade-out-of-change/gather-extra/artifacts/clusterrolebindings.json | jq '.items[] | select(.metadata.name | startswith("cluster-version"))'
{
  "apiVersion": "rbac.authorization.k8s.io/v1",
  "kind": "ClusterRoleBinding",
  "metadata": {
    "annotations": {
      "include.release.openshift.io/self-managed-high-availability": "true",
      "kubernetes.io/description": "Grant the cluster-version operator permission to perform cluster-admin actions while managing the OpenShift core.",
      "release.openshift.io/delete": "true"
    },
    "creationTimestamp": "2026-03-25T12:14:44Z",
    "name": "cluster-version-operator",
    "ownerReferences": [
      {
        "apiVersion": "config.openshift.io/v1",
        "controller": true,
        "kind": "ClusterVersion",
        "name": "version",
        "uid": "9eeb3192-8d2a-4e32-a8e9-4e1cbef27087"
      }
    ],
    "resourceVersion": "41927",
    "uid": "9d375490-1200-4463-80cc-81bf12af669d"
  },
  "roleRef": {
    "apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole",
    "name": "cluster-admin"
  },
  "subjects": [
    {
      "kind": "ServiceAccount",
      "name": "default",
      "namespace": "openshift-cluster-version"
    }
  ]
}
{
  "apiVersion": "rbac.authorization.k8s.io/v1",
  "kind": "ClusterRoleBinding",
  "metadata": {
    "annotations": {
      "include.release.openshift.io/self-managed-high-availability": "true"
    },
    "creationTimestamp": "2026-03-25T12:14:34Z",
    "name": "cluster-version-operator-1",
    "ownerReferences": [
      {
        "apiVersion": "config.openshift.io/v1",
        "controller": true,
        "kind": "ClusterVersion",
        "name": "version",
        "uid": "9eeb3192-8d2a-4e32-a8e9-4e1cbef27087"
      }
    ],
    "resourceVersion": "2844",
    "uid": "f91251d3-cd0f-46bd-ab7f-170641330554"
  },
  "roleRef": {
    "apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole",
    "name": "cluster-admin"
  },
  "subjects": [
    {
      "kind": "ServiceAccount",
      "name": "cluster-version-operator",
      "namespace": "openshift-cluster-version"
    }
  ]
}
```

It is present... but with the delete annotation?



Checking installation only:

```
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1266/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn/2036773282895630336/artifacts/e2e-agnostic-ovn/gather-extra/artifacts/clusterrolebindings.json | jq '.items[] | select(.metadata.name | startswith("cluster-version"))'
{
  "apiVersion": "rbac.authorization.k8s.io/v1",
  "kind": "ClusterRoleBinding",
  "metadata": {
    "annotations": {
      "include.release.openshift.io/self-managed-high-availability": "true",
      "kubernetes.io/description": "Grant the cluster-version operator permission to perform cluster-admin actions while managing the OpenShift core.",
      "release.openshift.io/delete": "true"
    },
    "creationTimestamp": "2026-03-25T12:15:59Z",
    "name": "cluster-version-operator",
    "resourceVersion": "241",
    "uid": "f2c60a58-a77d-416d-bbb0-cf4db4801d09"
  },
  "roleRef": {
    "apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole",
    "name": "cluster-admin"
  },
  "subjects": [
    {
      "kind": "ServiceAccount",
      "name": "default",
      "namespace": "openshift-cluster-version"
    }
  ]
}
{
  "apiVersion": "rbac.authorization.k8s.io/v1",
  "kind": "ClusterRoleBinding",
  "metadata": {
    "annotations": {
      "include.release.openshift.io/self-managed-high-availability": "true"
    },
    "creationTimestamp": "2026-03-25T12:15:49Z",
    "name": "cluster-version-operator-1",
    "ownerReferences": [
      {
        "apiVersion": "config.openshift.io/v1",
        "controller": true,
        "kind": "ClusterVersion",
        "name": "version",
        "uid": "55875f29-285a-439d-a34d-b93e46af1dbb"
      }
    ],
    "resourceVersion": "2829",
    "uid": "566d0391-c697-4f7c-b97f-c7bc9e4080fa"
  },
  "roleRef": {
    "apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole",
    "name": "cluster-admin"
  },
  "subjects": [
    {
      "kind": "ServiceAccount",
      "name": "cluster-version-operator",
      "namespace": "openshift-cluster-version"
    }
  ]
}
```

Is the CRB with the delete annotation present? It is. It shouldn’t be.

---

### The most likely cause

This led me to a deep rabbit hole, tbh. I believe this is due to the [CVO render command](https://github.com/openshift/cluster-version-operator/blob/main/cmd/cluster-version-operator/render.go). It does not respect the delete annotation, nor does it have a special logic for its rendering command to bypass this, and it creates the CRB during installation-bootstrap time. The delete annotation deletes stuff only during cluster upgrades. Thus, the CRB gets left in the cluster after an installation. This is not acceptable. The delete annotation's purpose is to enable developers to stage the deletion of manifests to ensure no leftover objects are left in the cluster after a period of time. This present state does not allow the CVO to utilize the annotation in such a way for its manifests. Nor the CVO has smarter rendering for its manifests for the bootstrap stage. The CRB would get applied for all freshly installed clusters, and we could never delete the CRB manifest file.

[This does not affect other similar components, to my understanding. This is strictly related to the very early bootstrapping of the cluster, where individual bootstrap components use their own render commands to render their bootstrap manifests](https://github.com/openshift/installer/blob/3a35f919f5f22700746e74831265b2ccc42ce20a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template), [which are then applied.](https://github.com/openshift/cluster-bootstrap/blob/b23c6ce3df43aed15158e999239694ec75371f18/pkg/start/start.go#L142)

---

### Proposal

My proposal is to keep the role binding for the default service account, remove the proposed delete annotation, move it to the `0000_00_cluster-version-operator_03_` runlevel to ensure correctness, and create a new separate issue for the problem: the default service account in the CVO namespace should not have cluster-admin permissions. Are you fine with such an outcome to your PR and to your assigned bug?

The bug's title and the PR's title describe the goal of having dedicated service accounts. We would have achieved that. Unfortunately, this newly found bug makes the next step of removing the extra permission from the default SA more difficult.

PR #1271, opened 2025-12-10
OCPBUGS-66898: Implement mTLS authentication and authorization for CVO metrics endpoint

DavidHurta:
This pull request proposes to migrate the CVO metrics endpoint from bearer token authentication to mutual TLS (mTLS) to comply with [OpenShift metrics conventions](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#metrics).

**Key changes:**

* Replaced custom certificate watching logic with the `k8s.io/apiserver/pkg/server/dynamiccertificates` package to simplify logic
* Authenticate clients using mutual TLS (mTLS)
* Authorize clients by verifying CN
* Add configurable authentication/authorization options for HyperShift compatibility
* Update ServiceMonitor to use mTLS instead of bearer tokens

**Breaking Changes:**

- Prometheus must now authenticate using client certificates signed by a trusted cluster CA. Clients without valid certificates are rejected during the TLS handshake.

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -3,6 +3,7 @@ package cvo
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"net"
@@ -125,52 +126,101 @@ type asyncResult struct {
 	error error
 }
 
-func createHttpServer(disableAuth bool) *http.Server {
-	if disableAuth {
+func createHttpServer(options MetricsOptions, clientCA dynamiccertificates.CAContentProvider) *http.Server {
+	if options.DisableAuthentication && options.DisableAuthorization {
 		handler := http.NewServeMux()
 		handler.Handle("/metrics", promhttp.Handler())
-		server := &http.Server{
-			Handler: handler,
-		}
-		return server
+		return &http.Server{Handler: handler}
 	}
 
-	auth := authHandler{downstream: promhttp.Handler()}
+	auth := authHandler{
+		downstream:           promhttp.Handler(),
+		clientCA:             clientCA,
+		enableAuthentication: !options.DisableAuthentication,
+		enableAuthorization:  !options.DisableAuthorization,
+	}
 	handler := http.NewServeMux()
 	handler.Handle("/metrics", &auth)
-	server := &http.Server{
-		Handler: handler,
-	}
-	return server
+	return &http.Server{Handler: handler}
 }
 
 type authHandler struct {
-	downstream http.Handler
+	downstream           http.Handler
+	clientCA             dynamiccertificates.CAContentProvider
+	enableAuthentication bool
+	enableAuthorization  bool
 }
 
+// ServeHTTP performs application-level authentication and authorization.
 func (a *authHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if !a.enableAuthentication && !a.enableAuthorization {
+		a.downstream.ServeHTTP(w, r)
+		return
+	}
+
+	// Both authentication and authorization require a client certificate
 	if r.TLS == nil || len(r.TLS.PeerCertificates) == 0 {
 		klog.V(4).Info("Client certificate required but not provided")
 		http.Error(w, "client certificate required", http.StatusUnauthorized)
 		return
 	}
 
+	if a.enableAuthentication && !a.authenticate(w, r) {
+		return
+	}
+
+	if a.enableAuthorization && !a.authorize(w, r) {
+		return
+	}
+
+	a.downstream.ServeHTTP(w, r)
+}
+
+// authenticate verifies the client certificate chain against the configured CA.
+// Returns true if authenticated successfully, false otherwise (and writes HTTP error).
+func (a *authHandler) authenticate(w http.ResponseWriter, r *http.Request) bool {
+	opts, ok := a.clientCA.VerifyOptions()
+	if !ok {
+		klog.Error("verify options from client CA provider could not be loaded")
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return false
+	}
+
+	if len(r.TLS.PeerCertificates) > 1 {
+		intermediates := x509.NewCertPool()
+		for _, cert := range r.TLS.PeerCertificates[1:] {
+			intermediates.AddCert(cert)
+		}
+		opts.Intermediates = intermediates
+	}
+
+	if _, err := r.TLS.PeerCertificates[0].Verify(opts); err != nil {
+		klog.V(4).Infof("Client certificate verification failed: %v", err)
+		http.Error(w, "client certificate not trusted", http.StatusUnauthorized)
+		return false
+	}
+
+	return true
+}
+
+// authorize verifies the client certificate CN against the allowed CN.
+// Returns true if authorized, false otherwise (and writes HTTP error).
+func (a *authHandler) authorize(w http.ResponseWriter, r *http.Request) bool {
 	// metricsAllowedClientCommonName is the Common Name (CN) of the client certificate
 	// that is authorized to access the metrics endpoint. This corresponds to the
 	// well-known Prometheus service account in OpenShift monitoring.
 	// See: https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#metrics
 	metricsAllowedClientCommonName := "system:serviceaccount:openshift-monitoring:prometheus-k8s"
 
-	// The first element is the leaf certificate that the connection is verified against
 	commonName := r.TLS.PeerCertificates[0].Subject.CommonName
 	if commonName != metricsAllowedClientCommonName {
 		klog.V(4).Infof("Access denied for common name: %s", commonName)
 		http.Error(w, fmt.Sprintf("unauthorized common name: %s", commonName), http.StatusForbidden)
-		return
+		return false
 	}
 
 	klog.V(5).Infof("Access granted for common name: %s", commonName)
-	a.downstream.ServeHTTP(w, r)
+	return true
 }
 
 func startListening(svr *http.Server, tlsConfig *tls.Config, lAddr string, resultChannel chan asyncResult) {
@@ -288,8 +338,10 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, res
 		// Assign to interface variable to ensure proper nil handling
 		clientCA = clientCAController
 
-		// Enforce mTLS
-		clientAuth = tls.RequireAndVerifyClientCert
+		// Request client certificates but don't verify at TLS layer.
+		// Verification happens in the HTTP handler, which allows returning
+		// HTTP error codes that are expected by the origin test suite, instead of TLS errors.
+		clientAuth = tls.RequestClientCert
 	}
 
 	// Log certificate controller events to stdout because the controller is reported to generate invalid events,
@@ -327,7 +379,7 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, res
 		resultChannel <- asyncResult{name: "serving certification controller"}
 	}()
 
-	server := createHttpServer(options.DisableAuthorization)
+	server := createHttpServer(options, clientCA)
 	tlsConfig := crypto.SecureTLSConfig(&tls.Config{
 		GetConfigForClient: func(clientHello *tls.ClientHelloInfo) (*tls.Config, error) {
 			config, err := servingCertController.GetConfigForClient(clientHello)
--- a/pkg/cvo/metrics_test.go
+++ b/pkg/cvo/metrics_test.go
@@ -3,7 +3,6 @@ package cvo
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"crypto/x509/pkix"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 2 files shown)


DavidHurta:
@coderabbitai review

pkg/cvo/metrics.go
```diff
@@ -247,70 +168,155 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+
+func verifyPeerCertificate(_ [][]byte, verifiedChains [][]*x509.Certificate) error {
+	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
+		return errors.New("no verified client certificate")
+	}
+	// The first element of a chain is the leaf certificate
+	cn := verifiedChains[0][0].Subject.CommonName
+	if cn != metricsAllowedClientCN {
+		klog.V(4).Infof("Access denied for CN: %s", cn)
+		return errors.New("unauthorized CN")
```
wking: nit: I don't mind as much for logs, but for error messages, which might eventually reach users, can we avoid accronyms with `unauthorized common name` or similar instead of using `CN`?

Also, not clear to me why we'd log the rejected CN in the previous `Infof` but not include it in the error message here.  The client has the certificate, and they can go look at their CN there, but if we include the CN in the error message, maybe that saves them a bit of work?

pkg/cvo/metrics.go
```diff
@@ -247,70 +168,155 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
 
+	// Register listeners so servingCertController is notified when certificates change.
+	if clientCAController != nil {
+		clientCAController.AddListener(servingCertController)
+	}
+	servingContentController.AddListener(servingCertController)
+
+	resultChannelCount++
+	go func() {
+		servingCertController.Run(1, metricsContext.Done())
```
wking: Wondering about  that `workers=1` argument, [😂][1]:

> doesn't matter what workers say, only start one.

[1]: https://github.com/kubernetes/apiserver/blame/v0.35.0/pkg/server/dynamiccertificates/tlsconfig.go#L249

pkg/cvo/metrics.go
```diff
@@ -322,63 +328,24 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, lis
...
 			shutdown = true
 			shutdownError := server.Shutdown(shutdownContext)
+			if shutdownError != nil { // log the error we are discarding
+				klog.Errorf("Failed to gracefully shut down metrics server: %v", shutdownError)
+			}
+
 			if loopError == nil {
 				loopError = shutdownError
-			} else if shutdownError != nil { // log the error we are discarding
-				klog.Errorf("Failed to gracefully shut down metrics server: %v", shutdownError)
```
wking: I think you want to keep the logging down here, instead of shifting it to log unconditionally, otherwise you'll double up on the `loopError` message (which is getting logged below at the `Infof` level, and then _again_ (presumably at the `Errorf` level) in whatever is at the top of the bubbled-back `return loopError` pipeline.

pkg/cvo/metrics.go
DavidHurta: Let me double-check where exactly does the returned error bubble up to.

pkg/cvo/metrics.go
```diff
@@ -247,70 +168,155 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
-		tlsConfig, err = makeTLSConfig(certFile, keyFile)
-		if err != nil {
-			return fmt.Errorf("Failed to create TLS config: %w", err)
-		}
-	} else {
+// The TLS configuration automatically reloads certificates when
+// they change on disk using dynamiccertificates.
+func RunMetrics(runContext context.Context, shutdownContext context.Context, listenAddress, certFile, keyFile string, restConfig *rest.Config, metricsOptions MetricsOptions) error {
+	if listenAddress == "" {
 		return errors.New("TLS configuration is required to serve metrics")
```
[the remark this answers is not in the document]
DavidHurta: Agreed :sweat_smile:

pkg/cvo/metrics.go
DavidHurta: The error value bubbles up to the server's logger, that is, the value is logged in CVO logs (see a [net/http](https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/net/http/server.go;drc=c0e149b6b1aa2daca64c00804809bc2279e21eee;l=1979) line). This can be verified by verifying the CVO's logs. We currently use the default server's logger.

```
2026/01/08 17:08:57 http: TLS handshake error from 10.0.110.147:42700: unauthorized CN
```

The user observers only a [TLS alert message](https://www.rfc-editor.org/rfc/rfc5878.html#section-4). See a [crypto/tls code section](https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/crypto/tls/handshake_client.go;drc=3a3c006ac07886aa923a8aad0a4b3ed954640973;l=1197):

```go
	if c.config.VerifyPeerCertificate != nil && !echRejected {
		if err := c.config.VerifyPeerCertificate(certificates, c.verifiedChains); err != nil {
			c.sendAlert(alertBadCertificate)
			return err
		}
	}
```

This is unfortunate to some degree, as there is a specific `access_denied` TLS alert message that I would probably want to return instead of `bad_certificate` for failed authorizations.

DavidHurta: I could highlight the security aspects in more detail in the code comments.

The optimization was not my main goal with the commit. It was rather to move the authorization into the TLS layer and to make the mechanism's nature more explicit. The optimization was a nice plus.
However, the trade-off is more difficult debugging and technically wrong return codes to clients. This is an internal service, so I don't presume to cause it too many disruptions. However, I am inclined into reverting the commit to keep things correct.

pkg/cvo/metrics.go
DavidHurta: This change deserved its own commit.

It seems to be possible for the previous logic:

```go
			shutdown = true
			shutdownError := server.Shutdown(shutdownContext)

			if loopError == nil {
				loopError = shutdownError
			} else if shutdownError != nil { // log the error we are discarding
				klog.Errorf("Failed to gracefully shut down metrics server: %v", shutdownError)
			}
```

to not print an error from the http server not shutting down correctly. This was not previously possible as only one go routine existed.

Imagine a scenario:

1. `loopError` is nil
2. `metricsContext` is cancelled
3. http server shuts down incorrectly; the `shutdownError` holds the error
4. not printed immediately; the value gets saved to `loopError`
5. `loopError` may get overridden while other go routines are shutting down
6. The `Graceful shutdown complete for metrics server: %s` outputs the last value saved in `loopError`
7. the `shutdownError` gets forgotten

---

This does not resolve your comment regarding the even triple _same_ logs. Maybe it's worth revisiting the purpose of the `loopError` and my usage of it, as the logic seems to be a bit too tricky at first glance. My proposed changes should log every failure and return the last loop's error. Without the change, it is still possible to even triple log failures (`handleServerResult`, the `Graceful shutdown complete for metrics server`, and the parent function), but we can also forget the http server's failures all together.

pkg/cvo/metrics.go
```diff
@@ -247,70 +168,155 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
-	var tlsConfig *tls.Config
-	if listenAddress != "" {
-		var err error
-		tlsConfig, err = makeTLSConfig(certFile, keyFile)
-		if err != nil {
-			return fmt.Errorf("Failed to create TLS config: %w", err)
-		}
-	} else {
+// The TLS configuration automatically reloads certificates when
+// they change on disk using dynamiccertificates.
```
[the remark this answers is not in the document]
DavidHurta: Done in https://github.com/openshift/cluster-version-operator/pull/1271/commits/9ad1d4e10dc016992ba9274a2a99dd4cae9b928d.

pkg/cvo/metrics.go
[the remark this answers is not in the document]
DavidHurta: Done in https://github.com/openshift/cluster-version-operator/pull/1271/commits/021e46fbbed7d18e86c64426225b9a17c4bfb0f2.

pkg/cvo/metrics.go
```diff
@@ -247,70 +168,155 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+	// Create a dynamic serving cert/key controller to watch for serving certificate changes from files.
+	servingContentController, err := dynamiccertificates.NewDynamicServingContentFromFiles("metrics-serving-cert", certFile, keyFile)
 	if err != nil {
-		return fmt.Errorf("Failed to initialize certificate file checksum: %w", err)
+		return fmt.Errorf("failed to create serving certificate controller: %w", err)
 	}
-	origKeyChecksum, err := checksumFile(keyFile)
-	if err != nil {
-		return fmt.Errorf("Failed to initialize key file checksum: %w", err)
+	if err := servingContentController.RunOnce(metricsContext); err != nil {
```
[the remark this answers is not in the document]
DavidHurta: It is not currently set up in the PR.

pkg/cvo/metrics.go
```diff
@@ -247,70 +210,144 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+
+	// Log certificate controller events to stdout because the controller is reported to generate invalid events,
+	// which are rejected by the Kubernetes API server when used with DynamicServingContentFromFiles.
+	eventBroadcaster := record.NewBroadcaster(record.WithContext(metricsContext))
+	eventBroadcaster.StartLogging(klog.Infof)
+	defer eventBroadcaster.Shutdown()
+
+	// baseTlSConfig is a template passed to servingCertController,
+	// which generates updated configs via GetConfigForClient callback on each TLS handshake.
+	// This enables automatic certificate rotation without server restarts.
```
hongkailiu: This is lovely.

We do not need to use `fsnotify` directly which I know is very subtle under K8S environment where the files are mounted from ConfigMaps/Secrets. They are refreshed via Syslinks, and so on and so on.

pkg/cvo/metrics.go
```diff
@@ -151,72 +152,34 @@ func createHttpServer(ctx context.Context, client *authenticationclientsetv1.Aut
...
 
-	authorized, err := a.authorize(token)
-	if err != nil {
-		klog.Warningf("Failed to authorize token: %v", err)
-		http.Error(w, "failed to authorize due to an internal error", http.StatusInternalServerError)
+	// The first element is the leaf certificate that the connection is verified against
+	commonName := r.TLS.PeerCertificates[0].Subject.CommonName
+	if commonName != metricsAllowedClientCommonName {
+		klog.V(4).Infof("Access denied for common name: %s", commonName)
+		http.Error(w, fmt.Sprintf("unauthorized common name: %s", commonName), http.StatusForbidden)
```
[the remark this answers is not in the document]
hongkailiu: > I don't like providing unnecessary information

This is what I would do too.

> maybe that saves them a bit of work?

Trevor was so kind. Let us leave it then. The question was to raise the awareness anyway.

DavidHurta:
@dis016 :wave: The PR is ready for a final verification :bow:

DavidHurta:
Interesting, the new serial Job is failing on:

```
{  fail [github.com/openshift/origin/test/extended/prometheus/prometheus.go:184]: Expected
    <[]error | len:1, cap:1>: [
        <*fmt.wrapError | 0xc002794040>{
            msg: "failed to ensure scraping target https://10.0.0.6:9099/metrics of pod openshift-cluster-version/cluster-version-operator/cluster-version-operator-6d8985944-947gh requires auth: context deadline exceeded",
            err: <context.deadlineExceededError>{},
        },
    ]
to be empty}
```

I will investigate this on Monday.

hongkailiu (commented):
Only a question about a0b08df6917d16cabb0f442919b64b1e26654042.
Others are just NITs and remarks.

dis016:
Test Scenario's:
Prometheus can scrape using mTLS
```
dinesh@Dineshs-MacBook-Pro ~ % oc get clusterversion 
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.22.0-0-2026-02-20-041325-test-ci-ln-z63ctmb-latest   True        False         59m     Cluster version is 4.22.0-0-2026-02-20-041325-test-ci-ln-z63ctmb-latest
dinesh@Dineshs-MacBook-Pro ~ % 
```
Prometheus should not scrape with bearer token.
```
 
dinesh@Dineshs-MacBook-Pro ~ % token=$(oc create token prometheus-k8s  -n openshift-monitoring)
dinesh@Dineshs-MacBook-Pro ~ % oc -n openshift-monitoring exec -c prometheus pod/prometheus-k8s-0 -- curl -i -H "Authorization:Bearer $token"  https://cluster-version-operator.openshift-cluster-version.svc.cluster.local:9099/metrics
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (60) SSL certificate problem: self-signed certificate in certificate chain
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
command terminated with exit code 60
dinesh@Dineshs-MacBook-Pro ~ % 
```
Client key is required with CA bundle
```

dinesh@Dineshs-MacBook-Pro ~ % oc -n openshift-monitoring exec -c prometheus pod/prometheus-k8s-0 -- curl -i --cacert /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt --cert /etc/prometheus/secrets/metrics-client-certs/tls.crt https://cluster-version-operator.openshift-cluster-version.svc.cluster.local:9099/metrics
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (58) unable to set private key file: '/etc/prometheus/secrets/metrics-client-certs/tls.crt' type PEM
command terminated with exit code 58
dinesh@Dineshs-MacBook-Pro ~ % 
```

PR #1276, opened 2025-12-18
OTA-1787: pkg: Update Upgradeable condition to explicitly mention major version updates

wking:
The `Upgradeable` condition already blocks both minor and major version updates when set to `False`, but all the error messages and comments only mentioned `minor versions` or `minor level` updates. This was misleading, especially in the context of enabling updates to version 5.0.

Changes:
- Update error messages in `pkg/cvo/upgradeable.go` to say `minor or major versions`
- Update comments in `pkg/payload/precondition/clusterversion/upgradeable.go` to clarify that major version updates are also blocked
- Update all test expectations to match the new messages
- Standardize terminology from `minor level` to `minor version` for consistency

My prompts were:

> Change the Upgradeable ClusterVersion condition handling so that it applies to major version updates as well as the current minor version updates.

and then:

> commit these changes, and set yourself as the commit author

### DIFF
--- a/pkg/payload/precondition/clusterversion/upgradeable.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable.go
@@ -3,6 +3,7 @@ package clusterversion
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/blang/semver/v4"
@@ -108,11 +109,11 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 				Name:    pf.Name(),
 			}
 		} else {
-			if completedVersion := minorUpdateFrom(cv.Status, currentVersion); completedVersion != "" && patchOnly {
+			if completedVersion, majorOrMinor := majorOrMinorUpdateFrom(cv.Status, currentVersion); completedVersion != "" && patchOnly {
 				// This is to generate an accepted risk for the accepting case 4.y.z -> 4.y+1.z' -> 4.y+1.z''
 				return &precondition.Error{
-					Reason:             "MinorVersionClusterUpdateInProgress",
-					Message:            fmt.Sprintf("Retarget to %s while a minor version update from %s to %s is in progress", targetVersion, completedVersion, currentVersion),
+					Reason:             fmt.Sprintf("%sVersionClusterUpdateInProgress", majorOrMinor),
+					Message:            fmt.Sprintf("Retarget to %s while a %s version update from %s to %s is in progress", targetVersion, strings.ToLower(majorOrMinor), completedVersion, currentVersion),
 					Name:               pf.Name(),
 					NonBlockingWarning: true,
 				}
@@ -130,24 +131,29 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 	}
 }
 
-// minorUpdateFrom returns the version that was installed completed if a minor version upgrade is in progress
-// and the empty string otherwise
-func minorUpdateFrom(status configv1.ClusterVersionStatus, currentVersion semver.Version) string {
+// majorOrMinorUpdateFrom returns the version that was installed
+// completed if a minor or major version upgrade is in progress and the
+// empty string otherwise.  It also returns "Major", "Minor", or "" to name
+// the transition.
+func majorOrMinorUpdateFrom(status configv1.ClusterVersionStatus, currentVersion semver.Version) (string, string) {
 	completedVersion := GetCurrentVersion(status.History)
 	if completedVersion == "" {
-		return ""
+		return "", ""
 	}
 	v, err := semver.Parse(completedVersion)
 	if err != nil {
-		return ""
+		return "", ""
 	}
 	if cond := resourcemerge.FindOperatorStatusCondition(status.Conditions, configv1.OperatorProgressing); cond != nil &&
-		cond.Status == configv1.ConditionTrue &&
-		v.Major == currentVersion.Major &&
-		v.Minor < currentVersion.Minor {
-		return completedVersion
+		cond.Status == configv1.ConditionTrue {
+		if v.Major < currentVersion.Major {
+			return completedVersion, "Major"
+		}
+		if v.Major == currentVersion.Major && v.Minor < currentVersion.Minor {
+			return completedVersion, "Minor"
+		}
 	}
-	return ""
+	return "", ""
 }
 
 // Name returns the name of the precondition.
--- a/pkg/payload/precondition/clusterversion/upgradeable_test.go
+++ b/pkg/payload/precondition/clusterversion/upgradeable_test.go
@@ -141,6 +141,19 @@ func TestUpgradeableRun(t *testing.T) {
 			},
 		},
 		{
+			name:             "move-x-then-z, with false condition",
+			upgradeable:      ptr(configv1.ConditionFalse),
+			completedVersion: "4.1.1",
+			currVersion:      "5.0.1",
+			desiredVersion:   "5.0.3",
+			expected: &precondition.Error{
+				NonBlockingWarning: true,
+				Name:               "ClusterVersionUpgradeable",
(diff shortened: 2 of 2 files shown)


pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -97,7 +97,8 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
 	klog.V(4).Infof("The current version is %s parsed from %s and the target version is %s parsed from %s", currentVersion.String(), cv.Status.Desired.Version, targetVersion.String(), releaseContext.DesiredVersion)
 	patchOnly := targetVersion.Major == currentVersion.Major && targetVersion.Minor == currentVersion.Minor
 	if targetVersion.LTE(currentVersion) || patchOnly {
-		// When Upgradeable==False, a patch level update with the same minor level is allowed unless overrides are set
+		// When Upgradeable==False, a patch level update with the same minor version is allowed unless overrides are set.
+		// However, minor or major version updates are blocked when Upgradeable==False (handled below at line 124).
```
DavidHurta:
```suggestion
		// However, minor or major version updates are blocked when Upgradeable==False.
```

pkg/payload/precondition/clusterversion/upgradeable_test.go
```diff
@@ -136,7 +136,7 @@ func TestUpgradeableRun(t *testing.T) {
 			expected: &precondition.Error{
 				NonBlockingWarning: true,
 				Name:               "ClusterVersionUpgradeable",
-				Message:            "Retarget to 4.2.3 while a minor level update from 4.1.1 to 4.2.1 is in progress",
+				Message:            "Retarget to 4.2.3 while a minor version update from 4.1.1 to 4.2.1 is in progress",
 				Reason:             "MinorVersionClusterUpdateInProgress",
 			},
 		},
```
DavidHurta: A comment not relevant to the selected line.

---

I did ask Claude whether the commit missed anything, and it came up with a few places, pretty neat.


```text
● Yes, the commit did miss a few places:

  Missed Updates

  1. pkg/internal/constants.go:59-68

  The comments for the Upgradeable condition constants still say "minor version":

  // UpgradeableAdminAckRequired is False if there is API removed from the Kubernetes API server which requires admin
  // consideration, and thus update to the next minor version is blocked.
  UpgradeableAdminAckRequired configv1.ClusterStatusConditionType = "UpgradeableAdminAckRequired"

  // UpgradeableDeletesInProgress is False if deleting resources is in progress, and thus update to the next minor
  // version is blocked.
  UpgradeableDeletesInProgress configv1.ClusterStatusConditionType = "UpgradeableDeletesInProgress"

  // UpgradeableClusterOperators is False if something is wrong with Cluster Operators, and thus update to the next minor
  // version is blocked.
  UpgradeableClusterOperators configv1.ClusterStatusConditionType = "UpgradeableClusterOperators"

  // UpgradeableClusterVersionOverrides is False if there are overrides in the Cluster Version, and thus update to the next minor
  // version is blocked.
  UpgradeableClusterVersionOverrides configv1.ClusterStatusConditionType = "UpgradeableClusterVersionOverrides"

  These should say "minor or major version" to be consistent with the rest of the changes.

  2. install/0000_90_cluster-version-operator_02_servicemonitor.yaml:90

  The Prometheus alert summary says:
  summary: One or more cluster operators have been blocking minor version cluster upgrades for at least an hour.

  This should say "minor or major version cluster upgrades" for consistency.

  ---
  These are documentation/messaging issues in the same spirit as the original commit. Would you like me to fix these remaining inconsistencies?

```

DavidHurta: Manipulating Claude into finding more places found:

```text
● Found it! Line 196 in the vendor file says:

  // When set on ClusterOperator, False will block minor
  // OpenShift updates.

  This should say "False will block minor or major" for consistency.

  However, this is in the vendor directory (vendor/github.com/openshift/api/config/v1/types_cluster_operator.go), which is a vendored dependency from the upstream openshift/api repository.

  This file cannot be edited directly in this repository - it needs to be fixed in the upstream openshift/api repository first, then vendored back into this project.
```

That is [this comment description](https://github.com/openshift/api/blob/d4b2fef98af257b6303059706077cc3c130f83e3/config/v1/types_cluster_operator.go#L191-L202) in the o/api repository. Worth pointing out where we have:

>  ...
    // When set on
	// ClusterVersion, the message will explain which updates (minor or patch)
	// are impacted. When set on ClusterOperator, False will block minor
	// OpenShift updates.
   ...

wking: > ... `pkg/internal/constants.go` ...

I've updated this and the alert in d01b53a4 -> 1bc95136.  I don't think we want to update the API at this point, but if we are still using `Upgradeable` by the time a 5->6 rolls around, we would want to update the API Godocs at that point.  I'll mark this thread resolved.

pkg/payload/precondition/clusterversion/upgradeable.go
```diff
@@ -129,24 +131,29 @@ func (pf *Upgradeable) Run(ctx context.Context, releaseContext precondition.Rele
...
+		if v.Major < currentVersion.Major {
+			return completedVersoin, "Major"
+		}
+		if v.Major == currentVersion.Major && v.Minor < currentVersion.Minor {
+			return completedVersion, "Minor"
+		}
 	}
-	return ""
+	return "", ""
 }
```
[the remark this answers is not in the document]
wking: Fixed in 6c5311e8 -> 07c1bb84

DavidHurta (approved):

JianLi-RH:
Do we need to update the error message when there is ClusterVersionGiantHop error:

```
Precondition "ClusterVersionGiantHop" failed because of "MultipleMinorVersionsUpdate": 4.21.0-rc.1 is more than one minor version beyond the current target 4.19.0-0.nightly-2026-01-06-203205 (4.21 > 4.(19+1)), and only updates within the current minor version or to the next minor version are supported.
```
in

```
[jianl@jianl-thinkpadt14gen4 419]$ oc adm upgrade
Cluster version is 4.19.0-0.nightly-2026-01-06-203205

Upgradeable=False

  Reason: AdminAckRequired
  Message: The admissionregistration.k8s.io/v1beta1 group version is deprecated in 4.19 and will be removed in 4.20. Any clients using the v1beta1 version of these resources must be updated to use the corresponding v1 version instead. See https://access.redhat.com/articles/7130599 for more information.

ReleaseAccepted=False

  Reason: PreconditionChecks
  Message: Preconditions failed for payload loaded version="4.21.0-rc.1" image="quay.io/openshift-release-dev/ocp-release@sha256:884e1ff5effeaa04467fab9725900e7f0ed1daa89a7734644f14783014cebdee": Multiple precondition checks failed:
  * Precondition "ClusterVersionGiantHop" failed because of "MultipleMinorVersionsUpdate": 4.21.0-rc.1 is more than one minor version beyond the current target 4.19.0-0.nightly-2026-01-06-203205 (4.21 > 4.(19+1)), and only updates within the current minor version or to the next minor version are supported.
  * Precondition "ClusterVersionUpgradeable" failed because of "AdminAckRequired": The admissionregistration.k8s.io/v1beta1 group version is deprecated in 4.19 and will be removed in 4.20. Any clients using the v1beta1 version of these resources must be updated to use the corresponding v1 version instead. See https://access.redhat.com/articles/7130599 for more information.
  * Precondition "ClusterVersionRecommendedUpdate" failed because of "UnknownUpdate": RetrievedUpdates=False (VersionNotFound), so the recommended status of updating from 4.19.0-0.nightly-2026-01-06-203205 to 4.21.0-rc.1 is unknown.

Upstream is unset, so the cluster will use an appropriate default.
Channel: stable-4.19
  Reason: VersionNotFound
  Message: Unable to retrieve available updates: currently reconciling cluster version 4.19.0-0.nightly-2026-01-06-203205 not found in the "stable-4.19" channel

[jianl@jianl-thinkpadt14gen4 419]$
```

PR #1297, opened 2026-01-18
Revert "OTA-1604: migrate ocp-46922 from otp to cvo repo"

neisw:
Reverts openshift/cluster-version-operator#1279

Tracked in [TRT-2511](https://issues.redhat.com/browse/TRT-2511)

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

Multiple payload failures due to microshift failures for cluster-version-operator should have correct runlevel and scc`

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Payload testing microshift may not be possible in your PR please review the failure and address the issue.

CC: @jiajliu @hongkailiu


jiajliu:
The case was reverted to unblock the CI tests because we hadn't accounted for it running on microshift. This particular pr is a clean revert of a recently migrated case from OTP to OTE. I will investigate the microshift issue further and address it in [TRT-2511](https://issues.redhat.com/browse/TRT-2511). For now, we are bypassing the test to proceed.

/lgtm
/verified bypass

PR #1311, opened 2026-02-02
NO-JIRA: More flexible timeout option for oc-cli

hongkailiu:
Follow up https://github.com/openshift/cluster-version-operator/pull/1309#discussion_r2752275622

We made an `option` type for oc cmd.

Currently, it is mostly `timeout` for running the command. The current implementation use it as the timeout for the context when spawning the process.

The `option` also gives the flexibility to introduce other options for oc.

/cc @JianLi-RH

### DIFF
--- a/test/cvo/cvo.go
+++ b/test/cvo/cvo.go
@@ -25,7 +25,7 @@ var _ = g.Describe(`[Jira:"Cluster Version Operator"] cluster-version-operator-t
 	})
 
 	g.It("can use oc to get the version information", func() {
-		ocClient, err := oc.NewOC(logger)
+		ocClient, err := oc.NewOC(ocapi.Options{Logger: logger})
 		o.Expect(err).NotTo(o.HaveOccurred())
 		o.Expect(ocClient).NotTo(o.BeNil())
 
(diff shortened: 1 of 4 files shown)


JianLi-RH:
/hold

Let's discuss it together, may be @jhou1 wants to join?

hongkailiu:
> that means we have to instantiate oc client for each case.

I do not see why for each case.

The oc with default timeout (30s). If you do not want to use it because the timeout is too short, you may create a new one for it. There are only [a few cases](https://github.com/openshift/cluster-version-operator/pull/1267/changes#r2583336139) like it, right?

Do you want a timeout for each cmd?
We could do it that way but I do not see the use cases.

hongkailiu:
> I'd like to put the oc client instantiation scripts into g.BeforeEach()

I am fine with it but that is not really a blocker of this pull, right?
You can do it with a commit in https://github.com/openshift/cluster-version-operator/pull/1309 after this one gets in.
I did not do it here because there is only one case using oc at the moment.
Moving it now inside `BeforeEach` looks odd to me.

What I did here is only making the API more easier to use: Parameters in an option vs Environment variable.

PR #1338, opened 2026-03-07
OTA-1836: Honor the centralized TLS configuration

DavidHurta:
Honor the [centralized TLS configuration](https://github.com/openshift/enhancements/blob/master/enhancements/security/centralized-tls-config.md).

A follow-up PR https://github.com/openshift/hypershift/pull/8013 after this one merges.

## Summary by CodeRabbit

* **New Features**
  * Central TLS profile manager for metrics: APIServer-driven TLS settings are applied at runtime and can be overridden via CLI. Startup validates overrides and waits for APIServer cache sync.

* **Tests**
  * Extensive tests for profile selection, override precedence, informer-driven updates, error recovery, and adherence policy behavior.

* **Chores**
  * Updated test frameworks and refreshed multiple indirect dependencies.


pkg/cvo/metrics.go
```diff
@@ -276,7 +447,7 @@ type MetricsOptions struct {
 // Continues serving until runContext.Done() and then attempts a clean
 // shutdown limited by shutdownContext.Done(). Assumes runContext.Done()
 // occurs before or simultaneously with shutdownContext.Done().
-func RunMetrics(runContext context.Context, shutdownContext context.Context, restConfig *rest.Config, options MetricsOptions) error {
+func RunMetrics(runContext context.Context, shutdownContext context.Context, restConfig *rest.Config, apiServerInformer configinformersv1.APIServerInformer, options MetricsOptions) error {
```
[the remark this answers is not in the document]
DavidHurta: We do not check `restConfig` for nil as well. This is consistent and places the responsibility on the caller. Share your thoughts.

DavidHurta:
@coderabbitai resume

DavidHurta:
@coderabbitai summary

pkg/cvo/metrics.go
```diff
@@ -259,6 +264,135 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+		return err
+	}
+
+	// By storing the apply function rather than extracted values, we automatically pick up any
+	// new fields that the tls profile package might set in future versions.
+	m.mu.Lock()
+	m.applyProfile = applyFunc
+	m.mu.Unlock()
+
+	klog.V(2).Info("Updated cached TLS profile")
```
DavidHurta: TODO: The wording suggests that the internal profile was indeed updated / changed. However, the event handler is also called every `resync` period. We may want to update the wording to something like "Synced cached TLS profile".

pkg/cvo/metrics.go
```diff
@@ -259,6 +264,135 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+	}
+
+	klog.V(4).Info("Resolved central TLS profile apply function")
+	return applyFunc, nil
+}
+
+// applySettings applies the TLS configuration to the provided config.
+// Applies: crypto defaults → central profile → overrides
+func (m *tlsProfileManager) applySettings(config *tls.Config) {
+	crypto.SecureTLSConfig(config)
```
DavidHurta: nit:

```suggestion
config = crypto.SecureTLSConfig(config)
```

pkg/cvo/metrics_test.go
```diff
@@ -1428,3 +1435,724 @@ func (m *mockCAProvider) VerifyOptions() (x509.VerifyOptions, bool) {
...
+	config := &tls.Config{}
+	mgr.applySettings(config)
+
+	if config.MinVersion != tls.VersionTLS12 {
+		t.Errorf("expected TLS 1.2, got %d", config.MinVersion)
+	}
+	if len(config.CipherSuites) != 1 {
+		t.Errorf("expected 1 cipher suite, got %d", len(config.CipherSuites))
+	}
+}
```
DavidHurta: De facto already covered by a different test:

```suggestion
```

DavidHurta:
Default Profile, Non-Strict Compliance: [Relevant tests are passing.](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608)

Modern Profile, Strict Compliance: [Relevant tests are passing.](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-modern-profile/2052138024979927040)

DavidHurta:
# Initial Verification

## Default Profile, Non-Strict Compliance

[Relevant tests are passing.](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608)

## Modern Profile, Strict Compliance

[Relevant tests are passing.](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-modern-profile/2052138024979927040)

## Updating the Profile in a Live cluster

### Prerequisites

1. Enable `TechPreviewNoUpgrade` feature set (the TLS adherence is feature gated)
2. Wait for the cluster to stabilize
3. Set modern TLS profile, set strict TLS adherence
4. Wait for the cluster to stabilize

### Metrics can be scraped

Screenshot From 2026-05-12 19-01-37

### CVO metrics logs

```
$ oc logs -n openshift-cluster-version deploy/cluster-version-operator | grep metrics.go | grep -v "skipping metrics for"
I0512 16:37:27.159291       1 metrics.go:315] Updated cached TLS profile
I0512 16:37:27.159350       1 metrics.go:614] Waiting on 4 outstanding goroutines.
I0512 16:37:27.159661       1 metrics.go:249] Metrics port listening for HTTPS on 0.0.0.0:9099
I0512 16:37:27.163438       1 metrics.go:315] Updated cached TLS profile
I0512 16:43:04.587875       1 metrics.go:315] Updated cached TLS profile
I0512 16:45:53.510898       1 metrics.go:315] Updated cached TLS profile
I0512 16:48:42.433600       1 metrics.go:315] Updated cached TLS profile
I0512 16:51:31.356648       1 metrics.go:315] Updated cached TLS profile
I0512 16:54:20.279379       1 metrics.go:315] Updated cached TLS profile
I0512 16:57:09.201934       1 metrics.go:315] Updated cached TLS profile
I0512 16:59:58.124773       1 metrics.go:315] Updated cached TLS profile
I0512 17:02:47.047004       1 metrics.go:315] Updated cached TLS profile
I0512 17:05:35.969635       1 metrics.go:315] Updated cached TLS profile
I0512 17:08:24.892142       1 metrics.go:315] Updated cached TLS profile
```

hongkailiu:
/test e2e-hypershift-conformance

Please ignore/forgive me.

cmd/cluster-version-operator/start.go
```diff
@@ -37,6 +39,8 @@ func init() {
...
 	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.ServingCertFile, "serving-cert-file", opts.MetricsOptions.ServingCertFile, "The X.509 certificate file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file unless you set --listen empty.")
 	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.ServingKeyFile, "serving-key-file", opts.MetricsOptions.ServingKeyFile, "The X.509 key file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file unless you set --listen empty.")
+	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.TLSMinVersionOverride, "tls-min-version", opts.MetricsOptions.TLSMinVersionOverride, "Minimum TLS version supported. When set, overrides the value from the central TLS profile. Possible values: "+strings.Join(cliflag.TLSPossibleVersions(), ", "))
+	cmd.PersistentFlags().StringSliceVar(&opts.MetricsOptions.TLSCipherSuitesOverride, "tls-cipher-suites", opts.MetricsOptions.TLSCipherSuitesOverride, "Comma-separated list of cipher suites for the server. When set, overrides the value from the central TLS profile. Accepts the cipher suite names defined by Go's crypto/tls package.")
```
hongkailiu:
```suggestion
	cmd.PersistentFlags().StringSliceVar(&opts.MetricsOptions.TLSCipherSuitesOverride, "tls-cipher-suites", opts.MetricsOptions.TLSCipherSuitesOverride, "Comma-separated list of TLS cipher suites. When set, overrides the value from the central TLS profile used to configure the communication with Cluster Version Operator, such as scraping metrics. Accepts the cipher suite names defined by Go's crypto/tls package.")
```

cmd/cluster-version-operator/start.go
```diff
@@ -37,6 +39,8 @@ func init() {
 	cmd.PersistentFlags().StringVar(&opts.ReleaseImage, "release-image", opts.ReleaseImage, "The Openshift release image url.")
 	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.ServingCertFile, "serving-cert-file", opts.MetricsOptions.ServingCertFile, "The X.509 certificate file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file unless you set --listen empty.")
 	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.ServingKeyFile, "serving-key-file", opts.MetricsOptions.ServingKeyFile, "The X.509 key file for serving metrics over HTTPS.  You must set both --serving-cert-file and --serving-key-file unless you set --listen empty.")
+	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.TLSMinVersionOverride, "tls-min-version", opts.MetricsOptions.TLSMinVersionOverride, "Minimum TLS version supported. When set, overrides the value from the central TLS profile. Possible values: "+strings.Join(cliflag.TLSPossibleVersions(), ", "))
```
hongkailiu:
```suggestion
	cmd.PersistentFlags().StringVar(&opts.MetricsOptions.TLSMinVersionOverride, "tls-min-version", opts.MetricsOptions.TLSMinVersionOverride, "Minimum TLS version supported. When set, overrides the value from the central TLS profile used to configure the communication with Cluster Version Operator, such as scraping metrics. Possible values: "+strings.Join(cliflag.TLSPossibleVersions(), ", ")+".")
```

pkg/cvo/metrics.go
```diff
@@ -285,6 +456,21 @@ func RunMetrics(runContext context.Context, shutdownContext context.Context, res
...
+	if err != nil {
+		return fmt.Errorf("invalid TLS configuration: %w", err)
+	}
+
+	if overrides != nil {
+		if overrides.minVersion != 0 {
+			klog.V(2).Infof("TLS min version override: %d (will override central TLS profile)", overrides.minVersion)
+		}
+		if len(overrides.cipherSuites) > 0 {
+			klog.V(2).Infof("TLS cipher suites override applied (will override central TLS profile)")
```
hongkailiu:
```suggestion
			klog.V(2).Infof("TLS cipher suites override: %v (will override central TLS profile)", overrides.cipherSuites)
```

pkg/cvo/metrics.go
```diff
@@ -259,6 +264,135 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+
+// resolveTLSProfile resolves the TLS profile apply function from the APIServer resource.
+func (m *tlsProfileManager) resolveTLSProfile(apiServer *configv1.APIServer) (func(*tls.Config), error) {
+	// No APIServer or TLS adherence disabled
+	if apiServer == nil || !crypto.ShouldHonorClusterTLSProfile(apiServer.Spec.TLSAdherence) {
+		return nil, nil
+	}
+
+	// Get the TLS profile spec
+	profile, err := tlsprofile.GetTLSProfileSpec(apiServer.Spec.TLSSecurityProfile)
```
hongkailiu: I am wondering how you figure out this API?
I do not see it in https://github.com/openshift/enhancements/blob/master/enhancements/security/centralized-tls-config.md.

hongkailiu (commented):
Only a few minors on the description of the two new flags.
And a few questions as always.
Otherwise LGTM.

I got an idea about the refactoring: TLS is not bound to Metrics and it should work as a separate pkg. Besides it is easy to read and test, it could be re-used if other parts of CVO need a TLSProfileMgr.

I tried the idea with https://github.com/openshift/cluster-version-operator/pull/1392. The goal there is to get the following `RunMetrics`. It adds an `applySettings` function and nothing else is needed.

```
func RunMetrics(runContext context.Context, shutdownContext context.Context, restConfig *rest.Config, applySettings func(config *tls.Config), options MetricsOptions) error {
```

I even believe that it might be useful to push the tls pkg into `o/lib-go`. Consumers of the lib do not need to know the details of `applySettings` like `ShouldHonorClusterTLSProfile`. 

Feel free to ignore the idea. By carrying out the idea, I understood better the pull.

pkg/cvo/metrics.go
```diff
@@ -259,6 +264,135 @@ func handleServerResult(result asyncResult, lastLoopError error) error {
...
+	_, err := apiServerInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if apiServer, ok := obj.(*configv1.APIServer); ok {
+				if err := m.updateSettings(apiServer); err != nil {
+					klog.Errorf("Failed to apply TLS settings on APIServer add: %v", err)
+				}
+			}
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			if apiServer, ok := newObj.(*configv1.APIServer); ok {
```
hongkailiu: Is it possible that the TLS profile is changed without modifying of `apiServer.Spec.TLSSecurityProfile`?

If not possible, we could compare its value in `oldObj` and `newObj` and call `m.updateSettings(apiServer)` only when the values is changed.

hongkailiu:
https://github.com/openshift/cluster-version-operator/pull/1338#issuecomment-4433294657

I am wondering how to use those jobs to verify?

https://github.com/openshift/release/pull/76044/changes/0d3be70d9ce8fed1e520bb37ad5291722235de88

says

```
A failing test of the tls-scanner-run step [1] does not
fail the whole job [2]. At least, as of the moment.
Thus, make the job optional. An always passing job has no significant value.
```

Or the above commit msg is stale?

> Wait for the cluster to stabilize

Do we have a signal of "stabilize"? Or it is a fixed time duration from experience?

I assume it is the latter.

I am thinking of automating the test as an e2e (SERIAL as it is modifying apiserver which might impact other tests). But I need to know more about how you did it manually.

> Enable TechPreviewNoUpgrade feature set (the TLS adherence is feature gated)

I think all we need to verify is that whenever `TLSAdherence` is modified, CVO applies the TLS profile to its TLS config.

0. if TechPreviewNoUpgrade is not enabled, skip the test.
1. get apiServer/cluster with the client
2. if apiServer == nil, skip the test (or even fail the test?)
3. save the current `TLSAdherence`  value (if not set, then save [the default](https://github.com/openshift/cluster-version-operator/blob/28e62a8ec3b884d18cbe639bbd8c3d3bdc6daf87/vendor/github.com/openshift/api/config/v1/types_apiserver.go#L93): LegacyAdheringComponentsOnly) and update with a new value
4. check if CVO has a log "Updated cached TLS profile" after the update time (in 5 mins? or a signal from the above question)
5. Prometheus has an active target from CVO after the change above (similar to [this](https://github.com/openshift/origin/blob/f4d1c208855b7216452041276a7f909c3cf477ce/test/extended/prometheus/prometheus.go#L723)).
6. recover the saved value (after the test) from Step 3.

I think Step 5 is optional because `resolveTLSProfile` is mainly about calling lib functions to get `applyFunc`.
If `applyFunc` makes the TLS config not working and Prometheus fail to scape.
It is more likely a bug from the lib functions.
I would be more confident to claim this if we could push `resolveTLSProfile` itself into `o/lib-go`. I can work on it if I can convince you that is the right thing to do.

pkg/cvo/metrics.go
DavidHurta: I don't remember exactly the source as places are being updated, but currently some of the code examples are provided here: https://docs.google.com/document/d/1cMc9E8psHfnoK06ntR8kHSWB8d3rMtmldhnmM4nImjs/edit?tab=t.4cxmujrb3zyn

The https://pkg.go.dev/github.com/openshift/controller-runtime-common/pkg/tls package is also referenced there, so it was probably me checking internal docs and the structure of relevant packages.

pkg/cvo/metrics.go
DavidHurta: Sure. We can optimize it a bit further.

DavidHurta:
> Or the above commit msg is stale?

Not to my knowledge. The job outputs JUnit files, but the CI step does not propagate the exit code value to fail the overall job. You can still view JUnit files, results, logs from the TLS scanner, and anything that is collected to verify the functionality.



> Do we have a signal of "stabilize"? Or it is a fixed time duration from experience?

For the CVO, I was only interested in the CVO being stabilized, and the Prometheus being functional. Changing the feature set and the cluster profile disrupts the whole cluster. Thus, to be technically correct, one would watch all required services and the CVO itself.


> I think all we need to verify is that whenever TLSAdherence is modified, CVO applies the TLS profile to its TLS config.

There are a number of things to check. My approach was to check:

- The CVO should not adhere to the central TLS configuration when it should not
- The CVO should adhere to the central TLS configuration when it should
- Updating the central configuration results in the CVO updating its TLS configuration
- Metrics should be scrapable in all cases

There can be a number of approaches to verify this. From manual to automated. There are also periodics running of the TLS scanner somewhere in the OpenShift org.

I went with extensive unit testing, manual testing for updating APIServer API, verifying TLS configurations using the TLS scanner, and relying for periodic jobs for regressions. This is enough for me personally.

Also consider that it is an OpenShift-wide effort. There may be more jobs running and such. Furthermore, adding such a test case to the serial suite may be too disruptive; that is, it may not complete and return the cluster to the original state in a mere five minutes. Longer tests in the serial test suite need approving, or maybe some engineers are working on some extensive testing to verify this for the whole cluster? Ideally and hopefully.

hongkailiu:
I checked [output.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608/artifacts/tls-scanner-default-profile/tls-scanner-run/artifacts/tls-scanner/output.log) and [junit_tls_scan.xml](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608/artifacts/tls-scanner-default-profile/tls-scanner-run/artifacts/tls-scanner/junit_tls_scan.xml) from [tls-scanner-default-profile](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608):

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608/artifacts/tls-scanner-default-profile/tls-scanner-run/artifacts/tls-scanner/output.log | grep 'Exit code:'
Scan complete. Exit code: 0

$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2052138024946372608/artifacts/tls-scanner-default-profile/tls-scanner-run/artifacts/tls-scanner/junit_tls_scan.xml | grep failures
<testsuite name="TLSSecurityScan" tests="1" failures="0" time="0">
```

and the same from [tls-scanner-modern-profile](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-modern-profile/2052138024979927040).

Is it enough to do the verification?

> Do we have a signal of "stabilize"?

I found `oc adm wait-for-stable-cluster` from [tls-13-commands.sh](https://github.com/openshift/release/blob/5606072a068769a7965b72ed6529a710cc344b30/ci-operator/step-registry/tls-13/tls-13-commands.sh#L146). Probably we could do the same (either via oc-cli or lib-go) if we do the e2e.

pkg/cvo/metrics_test.go
DavidHurta: https://github.com/openshift/cluster-version-operator/pull/1338/commits/6ed9b90634a8a33e70d6de28ec72582490bcd321

pkg/cvo/metrics.go
DavidHurta: https://github.com/openshift/cluster-version-operator/pull/1338/commits/3b6a868c9a30a9d20ed1bf3b2d91e4b1c75b0dd8

pkg/cvo/metrics.go
DavidHurta: https://github.com/openshift/cluster-version-operator/pull/1338/commits/1975895fe7f6904cf19b03016cd6044efb08cbf9

cmd/cluster-version-operator/start.go
DavidHurta: https://github.com/openshift/cluster-version-operator/pull/1338/commits/d33326ef656c89f0cc9e67c914c781681997441d

cmd/cluster-version-operator/start.go
DavidHurta: https://github.com/openshift/cluster-version-operator/pull/1338/commits/d33326ef656c89f0cc9e67c914c781681997441d

pkg/cvo/metrics.go
DavidHurta: Done by your cherry-picked changes.

DavidHurta:
> Is it enough to do the verification?

It checks several things. What is important is that it checks whether we do support PQC and that we do adhere to the cluster TLS profile. 

That is, our CVO supports the required cipher suites and TLS minimum version as expected.

The job should be enough to verify **these** aspects. That is, for example, it does not check whether metrics are available, it solely check the available TLS profile and whether it adheres to the cluster profile.

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,731 @@
...
+	fakeClient := configfake.NewClientset(objects...)
+	informerFactory := configinformers.NewSharedInformerFactory(fakeClient, 0)
+	apiServerInformer := informerFactory.Config().V1().APIServers()
+	apiServerInformer.Lister() // we have to call this before Start
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+
+	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
+
```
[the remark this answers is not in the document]
DavidHurta: true

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,731 @@
...
+		}
+
+		_, err := fakeClient.ConfigV1().APIServers().Update(ctx, modernAPIServer, metav1.UpdateOptions{})
+		if err != nil {
+			t.Fatalf("failed to update APIServer: %v", err)
+		}
+
+		// Give the informer a moment to process the update
+		time.Sleep(100 * time.Millisecond)
+
```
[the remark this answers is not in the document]
DavidHurta: true, 100ms should be enough, but maybe we can use 1s for the meantime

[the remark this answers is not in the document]
DavidHurta: 100ms should be enough, imo

pkg/tls/tls_test.go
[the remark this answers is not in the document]
DavidHurta: no need, imo

pkg/cvo/metrics.go
DavidHurta: Optimized in the latest changes.

DavidHurta:
I have brought a majority of changes from the https://github.com/openshift/cluster-version-operator/pull/1392 branch.

I have rebased and squashed some of the commits for clarity. In case changes to the attribution or source code are to be made, please let me know, and I am happy to modify them.

I have not brought over the changes from the `Move things around` commit. Its movement of the NewProfileManager does make more sense; however, it does not consider the informer cache and its synchronization, I believe.

For iteration purposes, I am proposing to do any further larger adjustments in a potential future PR to unblock the functionality.

pkg/cvo/metrics.go
DavidHurta: Wait a minute.

DavidHurta: The last commit tried to optimize the logic; however, it introduced a potential bug.

We would need to have the ability to start tracking the status of the profile evaluation to not get stuck on a stale local profile in case there was a transient error. Previously, this was addressed by the last successful profile being applied. Failures would be addressed on periodic resyncs. The downside was that we also executed the `updateSettings` logic when we did not have to.

I am open to using the previous logic, as it should not be that costly.

Otherwise, we need to introduce some sort of retrying logic and status tracking. What do you suggest?

DavidHurta: If we can guarantee that failures of profile evaluation solely depend and will depend on the APIServer, then the proposed solution should suffice, though.

hongkailiu:
> Its movement of the NewProfileManager does make more sense; however, it does not consider the informer cache and its synchronization, I believe.

I am not sure if I follow.

`apiServerInformer` is the detail of `profileMgr` that `RunMetrics` does not need to know.
https://github.com/openshift/cluster-version-operator/pull/1392/changes/43c87d3e0628b4ae4a04793dad2d06b9109094e9 does not change how `profileMgr`.
It just moves the instance creation of `profileMgr` out of `metrics.go`.
The cache and the synchronization stay the same as before. Or I missed something obvious here?

The same for [the overrides logic](https://github.com/openshift/cluster-version-operator/pull/1338/changes#diff-338242793b0013e033bc403cbd313e2cccfe74d03c9fcc16fa117ac2a735b464R293-R306). `metrics` does not need to know it.

`RunMetrics` needs only a `ApplySettings(config)` function which is provided by `profileMgr`. This is my [motivation](https://github.com/openshift/cluster-version-operator/pull/1338#pullrequestreview-4330700152) to do the refactoring. ^_^

The advantages are:
-  Give only the need part. `RunMetrics`/`metrics.go` is complicated already. I want to avoid making it more.
- If there is another place CVO that uses `profileMgr`, the same instance can be shared.
- The `TLSOptions` becomes [in parallel with](https://github.com/openshift/cluster-version-operator/pull/1392/changes/43c87d3e0628b4ae4a04793dad2d06b9109094e9#diff-c8a70a647926b2d4315935186c18bb43afcf0ad4334a6a768f34d679f6688ed7R63) `MetricsOptions`. It reflects the real relation between them. TLS options are details for  `profileMgr`, not `metrics`. 


> For iteration purposes, I am proposing to do any further larger adjustments in a potential future PR to unblock the functionality.

Tomorrow, I will do the review again for the current version which I did not get chance to do today.
I can follow up with "move things around" which i feel is a very nice thing to have. I would do it in this pull if I could.

pkg/tls/tls.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	klog.V(4).Info("Resolved central TLS profile apply function")
+	return applyFunc, nil
+}
+
+// ApplySettings applies the TLS configuration to the provided config.
+// Applies: crypto defaults → central profile → overrides
+func (m *ProfileManager) ApplySettings(config *tls.Config) {
+	config = crypto.SecureTLSConfig(config)
```
hongkailiu, line 139: It came from https://github.com/openshift/cluster-version-operator/pull/1338#discussion_r3203251866

https://github.com/openshift/cluster-version-operator/blob/59c9a0a66c88d7335fbf99e14a91498d4dc40698/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L312-L322

To me `config =` is redundant. Do you mind telling me the intension?

DavidHurta: The description for the function is not clear on whether it modifies in place the argument or whether the return value is a pointer to a new struct. I am just going around being defensive and using the return value of the function and keeping the same logic, which is used on main. That is using the returned value of the function.

Thus, it's redundant for the current implementation, but rather to be safe than sorry in the future, IMO.

DavidHurta:
I have tried to make your review more straightforward. I have directly cherry-picked your commits (only minor merge conflict resolution needed for the `Move things around` commit).

Furthermore, I have added a new https://github.com/openshift/cluster-version-operator/pull/1338/commits/97a5bd23f73a649f1ca8d0364eca57318f3ed274 commit to resolve that one minor mentioned issue. PTAL. Please let me know if the description is not clear enough or is incorrect. I am happy to follow up.

pkg/tls/tls.go
hongkailiu: :resolved

Thanks for sharing.

In the ideal world, we should fix the "the description for the function".

> keeping the same logic, which is used on main. 

I have to admit that I did not notice it.

> rather to be safe than sorry in the future

Strictly speaking, you took it granted the returned value is the modified one. But as you said, it is not clear what is the returned value about from its description.

DavidHurta: > In the ideal world, we should fix the "the description for the function".

True. 

> Strictly speaking, you took it granted the returned value is the modified one. But as you said, it is not clear what is the returned value about from its description.

Yeah, I am assuming that the return value has a purpose; otherwise, its purpose would be a bit questionable.

hongkailiu: Let us see how https://github.com/openshift/library-go/pull/2250 goes.
It is not a blocker for this pull.

hongkailiu:
> I have added a new https://github.com/openshift/cluster-version-operator/commit/97a5bd23f73a649f1ca8d0364eca57318f3ed274 commit to resolve that one minor mentioned issue.

Thanks for catching it up. It saved us lots of troubles of fixing it.

I should not have made that mistake in the first place given that it took me 2 hours to fix the same issue in test.
https://github.com/openshift/cluster-version-operator/pull/1338/changes#diff-92c83cc60428f2d2d8fc8055c7de208fc9fc40de60ed134df861ba0ee01af18eR39

```go
apiServerInformer.Lister() // we have to call this before Start
```

hongkailiu:
> Initializing ProfileManager before the APIServer informer syncs causes
the lister to be empty

The above from https://github.com/openshift/cluster-version-operator/pull/1338/changes/97a5bd23f73a649f1ca8d0364eca57318f3ed274

Yesterday I thought about it and led to https://github.com/openshift/cluster-version-operator/pull/1338#issuecomment-4578305559 but now I am thinking maybe the claim "the lister to be empty"  is not true.

Did you get the conclusion by testing it on a cluster?

I am going to do some testing otherwise.

This is NOT a blocker to merge the current pull. I can bring this up with a follow up pull if needed.

hongkailiu:
share some e2e testing result: https://github.com/openshift/cluster-version-operator/pull/1393#issuecomment-4587813381

It shows the e2e running against the current pull.
The testing result has nothing to do with https://github.com/openshift/cluster-version-operator/pull/1338#issuecomment-4582472199.

DavidHurta:
> Did you get the conclusion by testing it on a cluster?

No testing. Static analysis (prone to misunderstanding the code, of course).

The wording of the commit could be more technically correct. It is a data race issue. It may or may not occur depending on when informers are started and when the caches sync.

The CVO can sometimes make these things (initialization, passing down only the needed structs, etc.) a bit awkward due to its architecture. Thus, why I always double-check my code in this area.

Previously, the profile manager wanted to fetch the APIServer before the underlying informer was even started.

Start is described as:

```
	// Start initializes all requested informers. They are handled in goroutines
	// which run until the stop channel gets closed.
	// Warning: Start does not block. When run in a go-routine, it will race with a later WaitForCacheSync.
	Start(stopCh <-chan struct{})
```

Thus, here we should get a clear error. I have checked the past CI, and we can see such CVO logs in https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1393/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-operator/2058900598811529216/artifacts/e2e-agnostic-operator/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-867dc65658-nxxj2_cluster-version-operator.log:

```
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1393/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-operator/2058900598811529216/artifacts/e2e-agnostic-operator/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-867dc65658-nxxj2_cluster-version-operator.log | grep tls.go | head
W0525 13:56:44.993892       1 tls.go:41] APIServer resource not available at startup: apiserver.config.openshift.io "cluster" not found, using fallback defaults
I0525 13:56:44.993935       1 tls.go:92] Updated cached TLS profile
I0525 13:56:45.002808       1 tls.go:92] Updated cached TLS profile
I0525 13:59:03.710753       1 tls.go:92] Updated cached TLS profile
I0525 14:01:22.418936       1 tls.go:92] Updated cached TLS profile
I0525 14:03:41.126518       1 tls.go:92] Updated cached TLS profile
I0525 14:05:59.834533       1 tls.go:92] Updated cached TLS profile
I0525 14:08:18.543529       1 tls.go:92] Updated cached TLS profile
I0525 14:09:43.475549       1 tls.go:92] Updated cached TLS profile
I0525 14:12:02.183322       1 tls.go:92] Updated cached TLS profile
```

That's a CI run for https://github.com/openshift/cluster-version-operator/commit/6a8c6072b2f4e45fe5ab9269b4efa96d4608b0d0 from https://github.com/openshift/cluster-version-operator/pull/1393, which at the time contained the CVO changes, if I understand correctly.

hongkailiu:
> $ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1393/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-operator/2058900598811529216/artifacts/e2e-agnostic-operator/gather-extra/artifacts/pods/openshift-cluster-version_cluster-version-operator-867dc65658-nxxj2_cluster-version-operator.log | grep tls.go | head
W0525 13:56:44.993892       1 tls.go:41] APIServer resource not available at startup: apiserver.config.openshift.io "cluster" not found, using fallback defaults

Great. You have a solid proof from CI.
Thanks.

I will dig a bit about it (still no blockers for this pull).

DavidHurta:
/unhold

Intentionally not marking as verified at the moment.

hongkailiu:
Some infra issue occurred in the previous attempt. Let us retry.

/test tls-scanner-modern-profile
/test tls-scanner-default-profile

DavidHurta:
> Some infra issue occurred in the previous attempt. Let us retry.

A fix is needed. Something changed in the image availability :shrug: 

@hongkailiu, please review https://github.com/openshift/release/pull/79977

DavidHurta:
# Verification

## Centralized TLS Configuration

### The CVO should adhere to the TLS profile when adherence is strict

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-modern-profile/2062290834996334592

The CVO passes the tls-scanner check.

The CVO [uses](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-modern-profile/2062290834996334592/artifacts/tls-scanner-modern-profile/tls-scanner-run/artifacts/tls-scanner/output.log) only the TLS 1.3 as expected:

```
 Testing ciphers per protocol via OpenSSL plus sockets against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.   Encryption  Bits     Cipher Suite Name (IANA/RFC)
-----------------------------------------------------------------------------------------------------------------------------
SSLv2
 - 
SSLv3
 - 
TLS 1
 - 
TLS 1.1
 - 
TLS 1.2
 - 
TLS 1.3
 x1302   TLS_AES_256_GCM_SHA384            ECDH/MLKEM AESGCM      256      TLS_AES_256_GCM_SHA384                             
 x1303   TLS_CHACHA20_POLY1305_SHA256      ECDH/MLKEM ChaCha20    256      TLS_CHACHA20_POLY1305_SHA256                       
 x1301   TLS_AES_128_GCM_SHA256            ECDH/MLKEM AESGCM      128      TLS_AES_128_GCM_SHA256                             

 Done 2026-06-03 23:50:00 [  22s] -->> 10.0.48.31:9099 (10.0.48.31) <<--
```

### The CVO should not adhere to the TLS profile when adherence is not strict

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2062206087536316416

The CVO passes the tls-scanner check.

The CVO [uses](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1338/pull-ci-openshift-cluster-version-operator-main-tls-scanner-default-profile/2062206087536316416/artifacts/tls-scanner-default-profile/tls-scanner-run/artifacts/tls-scanner/output.log) the TLS 1.2 as expected:

```
 Testing ciphers per protocol via OpenSSL plus sockets against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.   Encryption  Bits     Cipher Suite Name (IANA/RFC)
-----------------------------------------------------------------------------------------------------------------------------
SSLv2
 - 
SSLv3
 - 
TLS 1
 - 
TLS 1.1
 - 
TLS 1.2
 xc030   ECDHE-RSA-AES256-GCM-SHA384       ECDH 521   AESGCM      256      TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384              
 xcca8   ECDHE-RSA-CHACHA20-POLY1305       ECDH 521   ChaCha20    256      TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256        
 xc02f   ECDHE-RSA-AES128-GCM-SHA256       ECDH 521   AESGCM      128      TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256              
TLS 1.3
 x1302   TLS_AES_256_GCM_SHA384            ECDH/MLKEM AESGCM      256      TLS_AES_256_GCM_SHA384                             
 x1303   TLS_CHACHA20_POLY1305_SHA256      ECDH/MLKEM ChaCha20    256      TLS_CHACHA20_POLY1305_SHA256                       
 x1301   TLS_AES_128_GCM_SHA256            ECDH/MLKEM AESGCM      128      TLS_AES_128_GCM_SHA256                             

 Done 2026-06-03 17:13:54 [  23s] -->> 10.0.55.82:9099 (10.0.55.82) <<--

```

### Updating the TLS profile causes the CVO to sync its TLS profile

Running `launch 4.23,https://github.com/openshift/cluster-version-operator/pull/1338 aws`

Updating the TLS profile and its adherence causes the TLS profile to be synced immediately:

```
$ grep "tls.go" cvo.log 
I0602 15:26:08.306039       1 tls.go:109] Synced cached TLS profile
I0602 15:26:08.306459       1 tls.go:109] Synced cached TLS profile
I0602 16:47:27.242573       1 tls.go:109] Synced cached TLS profile
```

## Regression

Manually checking whether metrics are still available via Prometheus when explicitly setting the intermediate and modern profiles:

Screenshot From 2026-06-04 17-45-57

The CVO metrics are being scraped by Prometheus.

Checking logs when setting the intermediate profile:

```
$ grep "tls.go"  cvo.log 
I0604 15:14:05.826792       1 tls.go:109] Synced cached TLS profile
I0604 15:14:05.827420       1 tls.go:109] Synced cached TLS profile
I0604 15:27:38.323335       1 tls.go:109] Synced cached TLS profile
I0604 15:31:41.844427       1 tls.go:109] Synced cached TLS profile
```

No failures, panics or restarts were observed in the CVO due to CVO handling the TLS profile. However, the modern profile does cause nodes to restart due to new machineconfigpools, which causes the CVO pod to reschedule in a graceful manner.

PR #1351, opened 2026-03-16
OTA-1892: pkg/cvo/metrics: Serve cluster_version_available_updates when channel is set

wking:
Since 8b9118992d (#45), `cluster_version_available_updates` is only served when there are unconditionally-recommended updates and retrieval is succeeding.  This makes it hard to understand at the fleet level when:

```
cluster_operator_conditions{name="version", condition="RetrievedUpdates", reason="VersionNotFound"} == 0
```

is because of a misconfigured `channel`, or a misbehaving Update Service, or otherwise.  We should always export
`cluster_version_available_updates` whenever a `channel` is set, to make it easier to isolate the “because the cluster-admin has somehow selected a channel not compatible with their current version” (which we can’t do much about other than keep serving our [existing `CannotRetrieveUpdates` alert][1]) from the other possibilities (which we might be able to do something about).

[1]: https://github.com/openshift/cluster-version-operator/blame/6d31e54fc074493d15eb0847d82ac1636afbbc27/install/0000_90_cluster-version-operator_02_servicemonitor.yaml#L55-L69

### DIFF
--- a/pkg/cvo/metrics.go
+++ b/pkg/cvo/metrics.go
@@ -629,7 +629,7 @@ func (m *operatorMetrics) Collect(ch chan<- prometheus.Metric) {
 			ch <- g
 		}
 
-		if len(cv.Spec.Upstream) > 0 || len(cv.Status.AvailableUpdates) > 0 || resourcemerge.IsOperatorStatusConditionTrue(cv.Status.Conditions, configv1.RetrievedUpdates) {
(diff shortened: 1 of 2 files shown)


DavidHurta (approved):
Nice, thanks!

DavidHurta:
/test all

Has been some time since the initial run.

PR #1314, opened 2026-02-09
OTA-1860: Stop blocking patch updates when cluster version overrides are set

harche:
## Summary

- When `cv.spec.overrides` has any entry with `unmanaged: true`, CVO sets `Upgradeable=False` and blocks **all** upgrades, including z-stream/patch updates. This is overly restrictive — patch updates deliver security fixes and bug fixes that should not be blocked just because an override is set.
- This change makes the override precondition check a **non-blocking warning** for patch-level updates instead of a hard block, effectively reverting the patch-blocking behavior introduced in PR #364.
- Minor/major version upgrades remain blocked by `Upgradeable=False` when overrides are set — that behavior is unchanged.
- The warning message is updated to reflect the new scope: "Disabling ownership via cluster version overrides prevents minor/major version upgrades."

### Behavior after this change

| Update type | Overrides set | Result |
|---|---|---|
| Patch (4.21.1 → 4.21.2) | Yes | **Allowed** with accepted risk recorded |
| Minor (4.21.x → 4.22.0) | Yes | **Blocked** by `Upgradeable=False` |
| Major | Yes | **Blocked** by `Upgradeable=False` |

## Test plan

- [x] All existing unit tests pass with updated message strings
- [x] New test: patch update with overrides produces non-blocking warning
- [x] New test: minor update with overrides is still blocking
- [ ] Integration test: cluster with overrides set can complete a z-stream update (accepted risk recorded in history)
- [ ] Integration test: cluster with overrides set is still blocked from minor version update

Jira: https://issues.redhat.com/browse/OTA-1860

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1368, opened 2026-04-06
OTA-1935: pkg: Use risk.Source framework to feed Upgradeable

wking:
Building on #1367, this simplifies the complex `cvo` package by shifting some logic into small, narrowly scoped helper packages.  That makes for easier unit testing of that logic, and less complication when reading the remaining `cvo` package code.

Using #1367's `risk.Source` framework with the new aggregate implementation allows `Upgradeable` risks to feed into the `conditionalUpdateRisks` framework.  This should help reduce confusion some users experience when the see `Upgradeable=False` and think it applies to all updates (when in reality it applies to major and minor bumps, but does not
apply to patch updates, especially since 6f8f984547, #1314).  It will also allow structured access to individual risks that are reported via this pipeline, while `Upgradeable` had an aggregated message string only consumable by humans.

I'm also removing `internal.UpgradeableAdminAckRequired` and similar from ClusterVersion `status.conditions`.  While these were visible by end users, they were mostly a mechanism for passing data from one part of the CVO to another.  There are no known consumers outside of the CVO.

With the refactor providing clearer abstraction layers, I could drop the entire `upgradeableCheckIntervals` structure.  Major touches in the history of that structure were bd05174b43 (#808) and cc94c95cb5 (#882).  The new approach is event-driven via the informer handlers for almost all of the new `risk.Source` implementations.  The one source that still
relies on polling is the `deletion` source, where it's watching a CVO-side global variable, and not a Kube-side informer.  It will still sync whenever ClusterVersion status is updated though, which should be often enough for this usually-transient situation.  If folks want lower latency, follow up work could add informer-style handler callbacks to the deletion global.


pkg/cvo/status.go
```diff
@@ -606,6 +606,69 @@ func setReleaseAcceptedCondition(cvStatus *configv1.ClusterVersionStatus, status
...
+		risks, _, err = upgradeable.Risks(ctx, updateVersions)
+		if err != nil {
+			klog.Errorf("errors while calculating risks for the %s condition: %v", configv1.OperatorUpgradeable, err)
+		}
+
+		for i := len(risks) - 1; i >= 0; i-- {
+			if c := meta.FindStatusCondition(risks[i].Conditions, internal.ConditionalUpdateRiskConditionTypeApplies); c != nil && c.Status == metav1.ConditionFalse {
+				risks = append(risks[:i], risks[i+1:]...) // exclude non-applicable risks
+			}
+		}
```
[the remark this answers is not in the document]
wking: > ... so a missing admin ack will still force `Upgradeable=False` on a patch-only cluster, which is the opposite of the new semantics.

That's fine.  `Upgradeable` currently talks to folks about general things they need to address before updating to later major/minor, even if they don't have any next-hop updates to later major/minor available in their current channel today.  I'm trying to preserve that functionality with this pull.

But looking at this block some more, I think we could have risks coming back from the `upgradeable` `risk.Source` wihtout an `Applies` condition set yet, so I've pushed 149126fc1 -> 98bc10b84, rerolling to wash all the risks through the condition registry to try and get matching assessed.  I'm dropping any risks we know don't apply to the cluster, and keeping (via `continue`s) any risks we know do apply, or where we failed to evaluate whether they apply.  I expect most risks coming back from `upgradeable` to be `Always` risks, so evaluating whether or not they apply should be both cheap and reliable.

pkg/risk/overrides/overrides.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	majorAndMinorUpdates := upgradeable.MajorAndMinorUpdates(version, versions)
+
+	var versionMap map[string][]string
+	if len(majorAndMinorUpdates) > 0 {
+		versionMap = map[string][]string{risk.Name: majorAndMinorUpdates}
+	}
+
+	return []configv1.ConditionalUpdateRisk{risk}, versionMap, nil
```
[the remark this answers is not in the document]
wking: The URL depends on a parsed version, so we can link to that version's docs.  And we can't drop the `URL` property, because [it's required][1].  If we return an `error` here on a failed version parse, folks are likely offroading already, and they can do what they want, without us attempting to map a safe path through uncharted desert.

[1]: https://github.com/openshift/api/blob/f50e695cb194278edf2aa4cdc1a8c82dd5102d50/config/v1/types_cluster_version.go#L912

wking:
Rebased onto `main` with 782b77f97 -> f5453e7ed, now that #1367 has merged.

pkg/risk/adminack/adminack.go
```diff
@@ -0,0 +1,202 @@
...
+
+const (
+	adminAckGateFmt string = "^ack-([4-5][.][0-9]{1,})-[^-]"
+)
+
+var adminAckGateRegexp = regexp.MustCompile(adminAckGateFmt)
+
+type adminAck struct {
+	name             string
+	currentVersion   func() configv1.Release
```
hongkailiu: nit: we need only the string of `configv1.Release.Version`, right?

pkg/risk/upgradeable/upgradeable.go
```diff
@@ -0,0 +1,136 @@
...
+		return []configv1.ConditionalUpdateRisk{{
+			Name:          fmt.Sprintf("%sUnknown", u.name),
+			Message:       fmt.Sprintf("Failed to parse cluster version to check Upgradeable conditions: %v", err),
+			URL:           "https://docs.redhat.com/en/documentation/openshift_container_platform/",
+			MatchingRules: []configv1.ClusterCondition{{Type: "Always"}},
+		}}, version, err
+	}
+
+	ops, err := u.coLister.List(labels.Everything())
+	if meta.IsNoMatchError(err) {
```
hongkailiu, line 91: (I know this is moved from existing code.)
Is it possible that no COs are found on any cluster where CVO is running?

pkg/risk/upgradeable/upgradeable.go
```diff
@@ -0,0 +1,136 @@
...
+	name           string
+	currentVersion func() configv1.Release
+	coLister       configlistersv1.ClusterOperatorLister
+	lastSeen       []configv1.ConditionalUpdateRisk
+}
+
+// New returns a new update-risk source, tracking ClusterOperator Upgradeable conditions.
+func New(name string, currentVersion func() configv1.Release, coInformer configinformersv1.ClusterOperatorInformer, changeCallback func()) risk.Source {
+	coLister := coInformer.Lister()
+	source := &upgradeable{name: name, currentVersion: currentVersion, coLister: coLister}
```
hongkailiu, line 37: nit: I would call it `upgradeable` or something like it.
The name `source` is misleading when `source.eventHandler` is used later, such as line 39-41, because `eventHandler` is not a general method for `source`.

pkg/risk/upgradeable/upgradeable.go
```diff
@@ -0,0 +1,136 @@
...
+	coInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(_ interface{}) { source.eventHandler(changeCallback) },
```
hongkailiu, line 39: nit: we may save `changeCallback` as a field of `upgradeable` to simplify the code a bit.

pkg/risk/upgradeable/util.go
```diff
@@ -0,0 +1,24 @@
+package upgradeable
+
+import (
+	"github.com/blang/semver/v4"
+)
+
+// MajorAndMinorUpdates returns all versions that are a major or minor
+// version beyond the current version.  Versions that cannot be parsed
```
hongkailiu: nit: Remove an extra space?

```suggestion
// version beyond the current version. Versions that cannot be parsed
```

pkg/risk/adminack/adminack.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if len(risks) == 0 {
+		a.lastSeen = nil
+		return nil, nil, err
+	}
+
+	a.lastSeen = risks
+	majorAndMinorUpdates := upgradeable.MajorAndMinorUpdates(version, versions)
+	if len(majorAndMinorUpdates) == 0 {
+		return risks, nil, err
```
hongkailiu, line 85: Does returning risks that apply to no versions (the 2nd returned value is `nil`) make sense?

- `risks, nil, err`
- risks, map, err` where `map(riskName)=nil or empty slice` for every `riskName` in `risks`, 
- `nil, nil, err`

I feel all of the above 3 should be equivalent. If my understanding is valid, I would prefer returning `nil, nil, err` instead.

pkg/risk/upgradeable/upgradeable_test.go
```diff
@@ -0,0 +1,197 @@
...
+		resourcemerge.SetOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.OperatorUpgradeable,
+			Status:  configv1.ConditionFalse,
+			Reason:  "TestReason",
+			Message: "Cluster is not upgradeable for testing.",
+		})
+
+		select {
+		case <-changeChannel:
+			t.Fatalf("updating the local pointer triggered a change notification, when it should have taken until the fake client update")
```
hongkailiu: Is it the intension?

```suggestion
			t.Fatalf("updating the local pointer triggered a change notification, when it should have not taken place until the fake client update")
```

pkg/risk/upgradeable/upgradeable_test.go
```diff
@@ -0,0 +1,197 @@
...
+	changeChannel := make(chan struct{})
+	changeCallback := func() {
+		t.Logf("%s sending change notification", time.Now())
+		changeChannel <- struct{}{}
+	}
+
+	currentVersion := func() configv1.Release {
+		return configv1.Release{Version: "4.21.0"}
+	}
+	versions := []string{"4.21.1", "4.22.0", "5.0.0"}
```
hongkailiu, line 52: nit: It could be moved to right before line 119 where it is the first used.

pkg/risk/risk.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // validName enforces the non-empty CamelCase risk-naming convention.
-var validName = regexp.MustCompile(`^[A-Z][A-Za-z0-9]*$`)
+var validName = regexp.MustCompile(`^[A-Z][A-Za-z0-9.-]*$`)
```
hongkailiu, line 26: Why do we allow for `.-` in name?

`Risk-` `Risk.` will become valid?

I googled [the naming convention](https://builtin.com/articles/pascal-case-vs-camel-case): by "CamelCase", we actually mean `PascalCase`, right?

Maybe sync it to [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/update/targeted-update-edge-blocking.md#proposal) as well if we want to modify?

pkg/risk/updating/updating.go
```diff
@@ -0,0 +1,111 @@
+// Package update declares a risk of updating to a later major or
```
hongkailiu:
```suggestion
// Package updating declares a risk of updating to a later major or
```

pkg/risk/updating/updating.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// New returns a new updating-risk source, tracking whether an update is in progress.
+func New(name string, cvName string, cvInformer configinformersv1.ClusterVersionInformer, changeCallback func()) (risk.Source, error) {
+	cvLister := cvInformer.Lister()
+	source := &updating{name: name, cvName: cvName, cvLister: cvLister}
+	_, err := cvInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(_ interface{}) { source.eventHandler(changeCallback) },
+		UpdateFunc: func(_, _ interface{}) { source.eventHandler(changeCallback) },
+	})
```
hongkailiu, line 39: We do not reconcile on deleting CV? Sounds reasonable to me. Just want to confirm if this is intensional.

pkg/risk/deletion/deletion.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Risks returns the current set of risks the source is aware of.
+func (d *deletion) Risks(_ context.Context, versions []string) ([]configv1.ConditionalUpdateRisk, map[string][]string, error) {
+	currentRelease := d.currentVersion()
+	version, err := semver.Parse(currentRelease.Version)
+	if err != nil {
+		return []configv1.ConditionalUpdateRisk{{
+			Name:          fmt.Sprintf("%sUnknown", d.name),
+			Message:       fmt.Sprintf("Failed to parse cluster version to check resource deletion: %v", err),
+			URL:           "https://docs.redhat.com/en/documentation/openshift_container_platform/",
```
hongkailiu, line 43: The `Message` helps but I do not know what I am looking for in `URL`, e.g., https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/updating_clusters/understanding-openshift-updates-1#update-manifest-application_how-updates-work. It does not mention resource deletion at all?

pkg/risk/deletion/deletion.go
```diff
@@ -0,0 +1,76 @@
...
+		})
+	}
+
+	if len(risks) == 0 {
+		return nil, nil, nil
+	}
+
+	majorAndMinorUpdates := upgradeable.MajorAndMinorUpdates(version, versions)
+	if len(majorAndMinorUpdates) == 0 {
+		return risks, nil, err
```
hongkailiu: nit: `err` is `nil`


```suggestion
		return risks, nil, nil
```

Same for line 74.

pkg/risk/aggregate/aggregate_test.go
```diff
@@ -0,0 +1,133 @@
...
+			if diff := cmp.Diff(tt.expectedVersions, versions); diff != "" {
+				t.Errorf("versions mismatch (-want +got):\n%s", diff)
+			}
+			var actualError string
+			if err != nil {
+				actualError = strings.ReplaceAll(err.Error(), "\u00a0", " ")
+			}
+			if diff := cmp.Diff(tt.expectedError, actualError); diff != "" {
+				t.Errorf("error mismatch (-want +got):\n%s", diff)
+			}
```
hongkailiu, line 176: `expectedError` could be removed in this test.

```suggestion
			if err != nil {
				t.Errorf("unexpected error: %v", err)
			}
```

pkg/risk/adminack/adminack.go
```diff
@@ -0,0 +1,203 @@
...
+			Message:       fmt.Sprintf("%s configmap gate %s must contain a non-empty value.", internal.AdminGatesConfigMap, gateName),
+			URL:           "https://docs.redhat.com/en/documentation/openshift_container_platform/",
+			MatchingRules: []configv1.ClusterCondition{{Type: "Always"}},
+		}
+	}
+	if val, ok := ackConfigmap.Data[gateName]; !ok || val != "true" {
+		return &configv1.ConditionalUpdateRisk{
+			Name:          riskName,
+			Message:       gateValue,
+			URL:           "https://example.com/FIXME-look-for-a-URI-in-the-message",
```
hongkailiu: nit: IDE like GoLand can filter those lines with TODO marker easily.

```suggestion
            // TODO: Fix URL
			URL:           "https://example.com/FIXME-look-for-a-URI-in-the-message",
```

pkg/risk/adminack/adminack.go
```diff
@@ -0,0 +1,203 @@
...
+		if risk := checkAdminGate(k, v, fmt.Sprintf("%s-", a.name), version, ackCm); risk != nil {
+			risks = append(risks, *risk)
+		}
+	}
+
+	slices.SortFunc(risks, func(a, b configv1.ConditionalUpdateRisk) int {
+		return strings.Compare(a.Name, b.Name)
+	})
+
+	return risks, version, err
```
hongkailiu: nit: a bit more readable.


```suggestion
	return risks, version, nil
```

pkg/risk/adminack/adminack_test.go
```diff
@@ -0,0 +1,463 @@
...
+	kfake "k8s.io/client-go/kubernetes/fake"
+	ktesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/cluster-version-operator/pkg/internal"
+)
+
+func TestOperator_upgradeableSync(t *testing.T) {
```
hongkailiu: Noticed this is copied from existing code.
Renaming to keep aligned with other Sources' test.

```suggestion
func Test_New(t *testing.T) {
```

pkg/cvo/status.go
```diff
@@ -606,6 +608,78 @@ func setReleaseAcceptedCondition(cvStatus *configv1.ClusterVersionStatus, status
...
+					continue // keep the risks that failed to eval or matched
+				}
+			} else if c.Status != metav1.ConditionFalse {
+				continue // keep the risks that failed to eval or matched
+			}
+			risks = append(risks[:i], risks[i+1:]...) // exclude non-applicable risks
+		}
+	}
+
+	// remove obsolete subconditions
```
hongkailiu, line 640: Maybe also deprecate them on the constants?

pkg/cvo/status.go
```diff
@@ -606,6 +608,78 @@ func setReleaseAcceptedCondition(cvStatus *configv1.ClusterVersionStatus, status
...
+			klog.Errorf("errors while calculating risks for the %s condition: %v", configv1.OperatorUpgradeable, err)
+		}
+
+		for i := len(risks) - 1; i >= 0; i-- {
+			c := meta.FindStatusCondition(risks[i].Conditions, internal.ConditionalUpdateRiskConditionTypeApplies)
+			if c == nil {
+				if match, err := conditionRegistry.Match(ctx, risks[i].MatchingRules); err != nil || match {
+					continue // keep the risks that failed to eval or matched
+				}
+			} else if c.Status != metav1.ConditionFalse {
```
hongkailiu, line 633: In my mind, this branch should never happen because `upgradeable.Risks(ctx, updateVersions)` in line 622 does not generate risks with conditions.

pkg/cvo/status.go
```diff
@@ -227,8 +227,8 @@ func updateClusterVersionStatus(
 
 	var riskNamesForDesiredImage []string
 	if shouldReconcileAcceptRisks() {
-		updates := getAvailableUpdates()
 		var riskConditions map[string][]metav1.Condition
+		updates := getAvailableUpdates()
```
hongkailiu, line 230: Is this change (shuffling) only for better readability?

pkg/cvo/cvo.go
```diff
@@ -296,7 +296,9 @@ func New(
 
 	optr.proxyLister = proxyInformer.Lister()
 	optr.cmConfigLister = cmConfigInformer.Lister().ConfigMaps(internal.ConfigNamespace)
+	optr.cacheSynced = append(optr.cacheSynced, cmConfigInformer.Informer().HasSynced)
```
hongkailiu, line 306: Thanks for the "why" in the commit msg.

pkg/cvo/status.go
wking: Hmm, I actually don't remember why I did this.  Maybe it was just variables-before-function-calls convention?

pkg/risk/risk.go
wking: I've added this to the commit message:

> The addition of `.` and `-` to `validName` in `pkg/risk` allows for names like `ClusterOperatorUpgradeable-test-operator` in the `Upgradeable` test, composed with the ClusterOperator `name`.  That requires us to extend the regular expression to allow [the characters supported by ClusterOperator names][2].  We could alternatively build a system that CamelCased ClusterOperator names, but that seemed difficult, and not worth the effort.


[2]: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

pkg/risk/updating/updating.go
wking: yeah, if we delete the CV, there's nowhere to write our results.

pkg/risk/deletion/deletion.go
wking: not yet ;)  Seems like a good place to link, unless you have alternative suggestions?  And we can extend the docs to cover the bits we aren't covering today?

pkg/cvo/status.go
wking: Dropped in f5453e7ed -> 8c6c114da

pkg/risk/deletion/deletion.go
wking: Fixed in f5453e7ed -> 8c6c114da

pkg/risk/updating/updating.go
wking: Fixed in f5453e7ed -> 8c6c114da

pkg/risk/upgradeable/upgradeable_test.go
wking: it could, but this keeps it close to the current-version, so all the strings can get shifted together to more modern versions in the future, if we ever need to shift strings.

pkg/risk/upgradeable/upgradeable_test.go
wking: f5453e7ed -> 8c6c114da moved this to `t.Fatal` (no arguments that needed `Fatalf`) and reworded to:

> updating the local pointer triggered a change notification, when there should not have been change triggeres until the fake client update

pkg/risk/upgradeable/util.go
wking: Fixed in f5453e7ed -> 8c6c114da

pkg/risk/upgradeable/upgradeable.go
wking: Unlikely, but I guess theoretically possible.  Maybe _very_ early in bootstrapping?

pkg/risk/adminack/adminack.go
wking: Done with f5453e7ed -> 8c6c114da

pkg/risk/upgradeable/upgradeable_test.go
```diff
@@ -0,0 +1,197 @@
...
+		resourcemerge.SetOperatorStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+			Type:    configv1.OperatorUpgradeable,
+			Status:  configv1.ConditionFalse,
+			Reason:  "TestReason",
+			Message: "Cluster is not upgradeable for testing.",
+		})
+
+		select {
+		case <-changeChannel:
+			t.Fatal("updating the local pointer triggered a change notification, when there should not have been change triggeres until the fake client update")
```
hongkailiu:
```suggestion
			t.Fatal("updating the local pointer triggered a change notification, when there should not have been change triggered until the fake client update")
```

pkg/risk/upgradeable/upgradeable_test.go
hongkailiu: :resolved

pkg/risk/upgradeable/upgradeable_test.go
```diff
@@ -0,0 +1,197 @@
...
+			risks, versionsMap, err := source.Risks(ctx, versions)
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+			if diff := cmp.Diff(expectedRisks, risks); diff != "" {
+				t.Errorf("risk mismatch (-want +got):\n%s", diff)
+			}
+			if diff := cmp.Diff(expectedVersions, versionsMap); diff != "" {
+				t.Errorf("versions mismatch (-want +got):\n%s", diff)
+			}
```
hongkailiu, line 129:
```suggestion
			}
			if _, _, mergeErr := risk.Merge(nil, nil, expectedRisks, nil); mergeErr != nil {
				t.Errorf("unexpected error: %v", mergeErr)
			}
```

This is to cover the change on the validName regex.

pkg/risk/risk.go
hongkailiu: We indeed have a test case with `risk.Name=="ClusterOperatorUpgradeable-test-operator"`. But that case does not cover this change. 
I have another comment to add the coverage.

Let me drop the suggestion on the naming convention. https://en.wikipedia.org/wiki/Camel_case says Camel case could be camelCase or CamelCase. 

* Pascal case or upper camel case means CamelCase.
* Dromedary case or lower camel case means camelCase.

We use "the non-empty CamelCase" in the comment Line 25 which seems accurate enough.

:resolved

pkg/risk/deletion/deletion.go
hongkailiu: If we do not plan to update the doc, we could use `https://docs.redhat.com/en/documentation/openshift_container_platform/no-docs-on-this-topic` to tell our use not to dig into the url.

Otherwise, we could use soemthing like
`https://docs.redhat.com/en/documentation/openshift_container_platform/to-be-updated`
(and add a TODO comment in the code too?)

You can ignore this comment completely too. My goal was to check if you have forgot to do it in the code.

:resolved

pkg/risk/upgradeable/upgradeable_test.go
wking: Thanks; applied with 8c6c114da -> 092e83b65, although I used `err` instead of `mergeErr`.

pkg/risk/upgradeable/upgradeable_test.go
wking: Thanks; fixed in 8c6c114da -> 092e83b65,

pkg/cvo/status.go
wking: Maybe not today.  But we sometimes hear folks asking for "I checked `$THING`, and we're good there :+1: " encouragement, and this guard would keep us from complaining about those happy checks in `Upgradeable` as if they were a problem.

pkg/risk/adminack/adminack_test.go
wking: Thanks; fixed in 092e83b65 -> 13caf244d.

pkg/risk/adminack/adminack.go
wking: Thanks; fixed in 092e83b65 -> 13caf244d.

pkg/risk/adminack/adminack.go
wking: it does make sense, because `setUpgradeableCondition` needs to be able to set `Upgradeable=False`, even if the cluster is currently in a channel that includes no updates to the next major or minor version.  It currently delivers on that with a:

```golang
risks, _, err = upgradeable.Risks(ctx, updateVersions)
```

call that ignores the returned map.

pkg/risk/aggregate/aggregate_test.go
wking: or it's a sign that I forgot to add a test case to excercise the error-generation logic :sweat:.  `divergent definitions` added in 13caf244d -> a684bff8c

pkg/risk/adminack/adminack.go
wking: I've pushed 7764ea6c6 to replace this FIXME with the message-URI-extraction the FIXME was calling for.

wking:
Testing with ClusterBot `launch 5.0.0-0.ci,openshift/cluster-version-operator#1368 gcp` ([logs][1]):

```console
$ cat cluster-operator.yaml 
kind: ClusterOperator
metadata:
  name: test-operator
spec: {}
$ oc apply -f cluster-operator.yaml 
clusteroperator.config.openshift.io/test-operator created
$ oc patch --subresource status clusteroperator test-operator --type json -p '[{"op": "add", "path": "/status", "value": {"conditions": [{"lastTransitionTime":"2026-05-18T23:38:52Z","type":"Upgradeable","status":"False","reason":"TestReason","message":"Test-operator is not upgradeable: test issue."}]}}]'
clusteroperator.config.openshift.io/test-operator patched
$ oc -n openshift-cluster-version logs -l k8s-app=cluster-version-operator --tail -1 | grep -A9 'ClusterOperator Upgradeable' | tail
I0519 00:22:16.024065       1 upgradeable.go:131] ClusterOperator Upgradeable changed (-old +new):
  []v1.ConditionalUpdateRisk(
- 	nil,
+ 	{
+ 		{
+ 			URL:           "https://docs.redhat.com/en/documentation/openshift_container_pla"...,
+ 			Name:          "ClusterOperatorUpgradeable-test-operator",
+ 			Message:       "Test-operator is not upgradeable: test issue.",
+ 			MatchingRules: []v1.ClusterCondition{{...}},
+ 		},
$ oc adm upgrade
Cluster version is 5.0.0-0-2026-05-18-230404-test-ci-ln-j17gvft-latest

Upgradeable=False

  Reason: MultipleReasons
  Message: Cluster should not be upgraded between minor or major versions for multiple reasons:
  * Disabling ownership via cluster version overrides prevents updates between minor or major versions. Please remove overrides before requesting a minor or major version update.
  * Test-operator is not upgradeable: test issue.

  Reason: NoChannel
  Message: The update channel has not been configured.
$ oc -n openshift-config-managed set data configmap admin-gates --from-literal 'ack-5.0-test-ack=Important stuff https://example.com/test-ack'
configmap/admin-gates data updated
$ oc adm upgrade
Cluster version is 5.0.0-0-2026-05-18-230404-test-ci-ln-j17gvft-latest

Upgradeable=False

  Reason: MultipleReasons
  Message: Cluster should not be upgraded between minor or major versions for multiple reasons:
  * Disabling ownership via cluster version overrides prevents updates between minor or major versions. Please remove overrides before requesting a minor or major version update.
  * Important stuff https://example.com/test-ack
  * Test-operator is not upgradeable: test issue.

  Reason: NoChannel
  Message: The update channel has not been configured.
```

All of that looks good to me :)

/verified by @wking

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-gcp-modern/2056508652931518464

hongkailiu:
/lgtm


/hold

In case https://github.com/openshift/cluster-version-operator/pull/1368#discussion_r3204344300 is missed unintentionally. 

Free to cancel otherwise.

wking:
I left [the constant-deprecations][1] off, because I know the Go linter sometimes grumbles at us for using deprecated APIs, and I didn't want to delay this further in additional rounds of CI checking on that.  We can drop the constants once `release-5.0` forks off and we enter 5.1 development.  Not sure how to track that kind of minor tech-debt though.  Completely fine with me if we open a new follow-up pull adding the deprecation notes to the constants' Godocs and see how it plays vs. the linter, but I don't think we need to hold this pull on that.

/hold cancel

[1]: https://github.com/openshift/cluster-version-operator/pull/1368#discussion_r3204344300

pkg/cvo/status.go
wking: I've opened #1391 to see how this plays in CI.

PR #1370, opened 2026-04-10
OCPBUGS-81675: pkg/cvo/egress: trust user-ca-bundle in HyperShift mode

dpateriya:
## Problem

In a HyperShift hosted cluster, `HostedCluster.spec.additionalTrustBundle` is synced by the Hosted Cluster Config Operator (HCCO) into `openshift-config/user-ca-bundle`. However, that ConfigMap is **not** automatically merged into `openshift-config-managed/trusted-ca-bundle` — which is the only source `getTLSConfig()` currently reads.

As a result, when `spec.updateService` points to an update service whose TLS certificate is signed by an internal CA provided exclusively via `additionalTrustBundle`, the CVO Cincinnati client fails with:

```
x509: certificate signed by unknown authority
```

This affects customers running an internal Cincinnati-compatible update service (e.g. OpenShift Update Service / OSUS) behind a wildcard certificate signed by an internal CA.

## Fix

When running with `--hypershift`, also read `openshift-config/user-ca-bundle` (via the already-wired `cmConfigLister`) and merge its contents into the cert pool alongside `trusted-ca-bundle`.

The new code path is strictly gated on `optr.hypershift`, so standard OCP clusters are completely unaffected.

Both listers (`cmConfigLister` for `openshift-config` and `cmConfigManagedLister` for `openshift-config-managed`) are already wired in `cvo.New()` — no new informers or flags are required.

## Changes

- `pkg/cvo/egress.go`: extend `getTLSConfig()` to merge `openshift-config/user-ca-bundle` into the cert pool when `--hypershift` is set.
- `pkg/cvo/egress_test.go`: new `TestGetTLSConfig` with 8 table-driven cases covering all combinations of bundle presence and the `hypershift` flag.

## Testing

```
go test ./pkg/cvo/ -run TestGetTLSConfig -v
```

All 8 cases pass:
- non-hypershift, no bundles → nil config
- non-hypershift, `trusted-ca-bundle` only → CA trusted
- non-hypershift, `user-ca-bundle` present but `hypershift=false` → ignored
- non-hypershift, both present but `hypershift=false` → only `trusted-ca-bundle` used
- hypershift, no bundles → nil config
- hypershift, `trusted-ca-bundle` only → CA trusted
- hypershift, `user-ca-bundle` only → CA trusted
- hypershift, both bundles → both CAs merged and trusted

## Summary by CodeRabbit

* **New Features**
  * HyperShift-aware certificate bundle handling to include user CA bundles when applicable.

* **Bug Fixes**
  * More robust certificate loading that augments system trust with custom CAs and clearer error reporting for managed/user bundle failures.
  * Return nil when no CA material is available to avoid using an empty cert pool.

* **Tests**
  * Added table-driven tests covering HyperShift mode, presence/absence of managed and user bundles, and error scenarios.

### DIFF
--- a/pkg/cvo/egress.go
+++ b/pkg/cvo/egress.go
@@ -72,27 +72,43 @@ func (optr *Operator) getProxyConfig() (*httpproxy.Config, error) {
 }
 
 func (optr *Operator) getTLSConfig() (*tls.Config, error) {
+	certPool := x509.NewCertPool()
+	found := false
+
+	// Always try the managed trusted CA bundle (proxy-merged bundle written by
+	// the platform into openshift-config-managed/trusted-ca-bundle).
 	cm, err := optr.cmConfigManagedLister.Get("trusted-ca-bundle")
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	}
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
-
-	certPool := x509.NewCertPool()
-
-	if cm.Data["ca-bundle.crt"] != "" {
+	if err == nil && cm.Data["ca-bundle.crt"] != "" {
 		if ok := certPool.AppendCertsFromPEM([]byte(cm.Data["ca-bundle.crt"])); !ok {
-			return nil, fmt.Errorf("unable to add ca-bundle.crt certificates")
+			return nil, fmt.Errorf("unable to add trusted-ca-bundle certificates")
 		}
-	} else {
+		found = true
+	}
+
+	// In HyperShift the platform does not automatically merge additionalTrustBundle
+	// into trusted-ca-bundle, so also load openshift-config/user-ca-bundle which
+	// HCCO syncs from HostedCluster.spec.additionalTrustBundle.  This allows CVO
+	// to verify TLS to a custom spec.updateService whose certificate is signed by
+	// an internal CA that is present only in additionalTrustBundle.
+	if optr.hypershift {
+		ucm, err := optr.cmConfigLister.Get("user-ca-bundle")
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+		if err == nil && ucm.Data["ca-bundle.crt"] != "" {
+			if ok := certPool.AppendCertsFromPEM([]byte(ucm.Data["ca-bundle.crt"])); !ok {
+				return nil, fmt.Errorf("unable to add user-ca-bundle certificates")
+			}
+			found = true
+		}
+	}
+
+	if !found {
 		return nil, nil
 	}
 
-	config := &tls.Config{
-		RootCAs: certPool,
-	}
-
-	return config, nil
+	return &tls.Config{RootCAs: certPool}, nil
 }
--- a/pkg/cvo/egress.go
+++ b/pkg/cvo/egress.go
@@ -72,7 +72,14 @@ func (optr *Operator) getProxyConfig() (*httpproxy.Config, error) {
 }
 
 func (optr *Operator) getTLSConfig() (*tls.Config, error) {
-	certPool := x509.NewCertPool()
+	// Seed from the OS system cert pool so that any custom CAs we append
+	// augment rather than replace system trust.  SystemCertPool may fail on
+	// some platforms (e.g. no system root store); fall back to an empty pool
+	// so the function degrades gracefully instead of hard-failing.
+	certPool, err := x509.SystemCertPool()
+	if err != nil {
+		certPool = x509.NewCertPool()
+	}
 	found := false
 
 	// Always try the managed trusted CA bundle (proxy-merged bundle written by
--- a/pkg/cvo/egress.go
+++ b/pkg/cvo/egress.go
@@ -110,5 +110,8 @@ func (optr *Operator) getTLSConfig() (*tls.Config, error) {
 		return nil, nil
 	}
 
+	// MinVersion is intentionally omitted; Go defaults to TLS 1.2 for clients,
+	// which is acceptable for Cincinnati/OSUS connectivity and avoids breaking
+	// compatibility with update services that do not yet support TLS 1.3.
 	return &tls.Config{RootCAs: certPool}, nil
 }
--- a/pkg/cvo/egress_test.go
+++ b/pkg/cvo/egress_test.go
@@ -3,6 +3,7 @@ package cvo
 import (
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"testing"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


dpateriya:
@PratikMahajan can you please review this PR?

dpateriya:
@coderabbitai review

enxebre:
>However, that ConfigMap is not automatically merged into openshift-config-managed/trusted-ca-bundle 

There might be other consumers of openshift-config-managed/trusted-ca-bundle with the same broken expectation. There's also additional trusted ca input coming from proxy.Spec.TrustedCA.Name.
I think this should be fixed in hypershift by:
- Ensuring the secret pushed down to the guest cluster by the HCCO and picked by CNO for reconciliation into openshift-config-managed/trusted-ca-bundle, contains both inputs hc.proxy.Spec.TrustedCA and hc.spec.additionalTrustBundle in their data.
- Also additionalTrustBundle could be arguably deprecated.

pkg/cvo/egress.go
```diff
@@ -72,27 +72,53 @@ func (optr *Operator) getProxyConfig() (*httpproxy.Config, error) {
...
 func (optr *Operator) getTLSConfig() (*tls.Config, error) {
-	cm, err := optr.cmConfigManagedLister.Get("trusted-ca-bundle")
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	}
+	// Seed from the OS system cert pool so that any custom CAs we append
+	// augment rather than replace system trust.  SystemCertPool may fail on
+	// some platforms (e.g. no system root store); fall back to an empty pool
+	// so the function degrades gracefully instead of hard-failing.
+	certPool, err := x509.SystemCertPool()
```
vincentdephily, line 79: Adding system certs seems like a big change. Previously users had complete control over the trusted CAs, now some of that is handed over to the OS vendor. This should probably be configurable (opt-in, for backward-compatibility reasons) ?

pkg/cvo/egress.go
```diff
@@ -72,27 +72,53 @@ func (optr *Operator) getProxyConfig() (*httpproxy.Config, error) {
...
+		}
+		found = true
+	}
+
+	// In HyperShift the platform does not automatically merge additionalTrustBundle
+	// into trusted-ca-bundle, so also load openshift-config/user-ca-bundle which
+	// HCCO syncs from HostedCluster.spec.additionalTrustBundle.  This allows CVO
+	// to verify TLS to a custom spec.updateService whose certificate is signed by
+	// an internal CA that is present only in additionalTrustBundle.
+	if optr.hypershift {
```
vincentdephily, line 103: I'd be tempted to not special-case hypershift here, and load either way ? AppendCertFromPEM ignores duplicates. This simplifies the logic, and you could factorize trusted-ca-bundle and user-ca-bundle together.

pkg/cvo/egress.go
```diff
@@ -72,27 +72,53 @@ func (optr *Operator) getProxyConfig() (*httpproxy.Config, error) {
...
+			found = true
 		}
-	} else {
-		return nil, nil
 	}
 
-	config := &tls.Config{
-		RootCAs: certPool,
+	if !found {
+		return nil, nil
```
vincentdephily, line 117: We likely do have the system CAs in the pool at this stage. If it's intentionally ignored, please document the rationale in the function doc. If the system CAs are good to take, you could replace `if !found` by `if certPool.len() < 1`.

pkg/cvo/egress_test.go
```diff
@@ -0,0 +1,245 @@
...
+			name:            "hypershift: both bundles → both CAs trusted",
+			hypershift:      true,
+			trustedCABundle: newCMForTest("trusted-ca-bundle", testCABundle1),
+			userCABundle:    newCMForTest("user-ca-bundle", testCABundle2),
+			wantCA1Present:  true,
+			wantCA2Present:  true,
+		},
+		// Error path: non-NotFound errors from listers must be propagated.
+		{
+			name:             "managed lister non-NotFound error is propagated",
```
vincentdephily, line 172: Might be worth adding an actual err=NotFound case.

pkg/cvo/egress.go
vincentdephily: Ah, misunderstanding on my part: the previous behavior was already to use the system certs by default, it was just done later, by the caller. There must be a cleaner way to express this than returning nil ?

dpateriya:
@enxebre Thank you for the architectural insight — you're absolutely right that fixing this in HyperShift (HCCO) would be the more comprehensive approach and would benefit all consumers of `openshift-config-managed/trusted-ca-bundle`, not just CVO. That is a worthwhile follow-up.

The rationale for this CVO-side fix was:

1. **Targeted relief for the customer** — The HCCO change requires reconciliation logic changes in the HyperShift project, which has its own release cadence. This CVO fix can ship faster as a targeted patch.

2. **CVO already has the right hook** — The `--hypershift` flag and `cmConfigLister` (watching `openshift-config`) are already wired. No new informers, flags, or cross-component coordination is needed.

3. **Strictly additive, no regression risk** — The change is gated on `optr.hypershift`, so standard OCP is completely unaffected.

I agree that the HyperShift approach (ensuring HCCO merges `additionalTrustBundle` into the secret consumed by CNO → `openshift-config-managed/trusted-ca-bundle`) is the right long-term fix and would make this CVO change unnecessary. Happy to open a follow-up issue/PR in the HyperShift repo for that. Would you prefer we pursue that first and close this PR, or treat this as an immediate fix while the HyperShift fix is tracked separately?

enxebre:
>Targeted relief for the customer — The HCCO change requires reconciliation logic changes in the HyperShift project, which has its own release cadence. This CVO fix can ship faster as a targeted patch.

This is no sense. Hypershift has two components with different release cadences the HO and the CPO. The HO is continuous delivered and the CPO belongs to the OCP payload just like CVO. So a change in o/hypershift can only be released be at least as fast as here.

>CVO already has the right hook — The --hypershift flag and cmConfigLister (watching openshift-config) are already wired. No new informers, flags, or cross-component coordination is needed.
>Strictly additive, no regression risk — The change is gated on optr.hypershift, so standard OCP is completely unaffected.

These are no good reasons to add tech debt.

>Would you prefer we pursue that first and close this PR, or treat this as an immediate fix while the HyperShift fix is tracked separately?

Yes please, let's close this and fix it in the right place. Thank you!!

dpateriya:
Closing this PR as per reviewer feedback from @enxebre.

The correct fix is in HyperShift's Hosted Cluster Config Operator (HCCO), not in CVO. The CPO ships with the OCP payload at the same cadence as CVO, so there is no release-speed advantage to fixing it here, and adding a HyperShift special-case in CVO is unnecessary tech debt.

**The root cause**: HCCO reconciles `HostedCluster.spec.additionalTrustBundle` → `openshift-config/user-ca-bundle` in the guest cluster, but this ConfigMap is never referenced by the guest cluster's `proxy.config.openshift.io/cluster` object. CNO only merges the ConfigMap referenced by `proxy.spec.trustedCA.Name` into `openshift-config-managed/trusted-ca-bundle`. So any component reading `trusted-ca-bundle` (including CVO) does not see the `additionalTrustBundle` CA.

**The fix** will be in `openshift/hypershift` (HCCO `reconcileProxyTrustedCAConfigMap`): merge both `proxy.trustedCA` and `additionalTrustBundle` into the ConfigMap that CNO reads, so `trusted-ca-bundle` contains both CAs. This benefits all consumers, not just CVO.

Tracking: OCPBUGS-81675

dpateriya:
Closing in favour of a fix in openshift/hypershift (HCCO) as suggested by @enxebre. See comment above.

PR #1377, opened 2026-04-16
OTA-1836: chore(deps): Bump general OpenShift dependencies

DavidHurta:
The API update is needed for the new TLS adherence API and other efforts.

Other general openshift packages are updated as well for periodic maintenance. We might as well update those while we are at it. De facto an updated clone of the https://github.com/openshift/cluster-version-operator/pull/1346 PR.

Generated by:

```
$ go get github.com/openshift/api@master
$ go get github.com/openshift/client-go@master
$ go get github.com/openshift/library-go@master
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

using:

```
$ go version
go version go1.25.0 linux/amd64
```

---

The PR also adds additional polling to one of the integration tests. See the commits for more information.

The agnostic job was fairly confidently failing due to the specific testing not being robust enough. The PR aims to add simple polling to remedy this.

Looking at https://search.dptools.openshift.org/?search=no+leader+election+events+found+in&maxAge=336h&context=1&type=all&name=.*e2e-agnostic-operator.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job:

image

Not that many failures, and most of them come from the dependency bump. The PR aims to make the testing more robust, as polling is a normal and expected method in testing.

---

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift Go dependencies to newer versions for improved compatibility and stability.
* **Tests**
  * Improved an integration test to be more resilient by waiting for leader-election events with retries, reducing intermittent failures.


DavidHurta:

This is blocking everything, which may need an updated Openshift API. The branch cut has occurred, so I am okay with updating these dependencies if the CI passes.

DavidHurta:
Let me investigate the agnostic-operator.

/hold

DavidHurta:
The agnostic-operator failures could also be just bad luck. Nonetheless, the fix makes sense even without them, IMO.

DavidHurta:
/unhold

https://github.com/openshift/cluster-version-operator/pull/1377#issuecomment-4282470204 is addressed.

pkg/start/start_integration_test.go
```diff
@@ -382,12 +382,20 @@ func TestIntegrationCVO_gracefulStepDown(t *testing.T) {
 	}
 
 	t.Logf("verify the controller writes a leadership change event")
-	events, err := kc.CoreV1().Events(ns).List(ctx, metav1.ListOptions{})
+	var lastEvents []v1.Event
+	err = wait.PollUntilContextTimeout(ctx, 200*time.Millisecond, 10*time.Second, true, func(localCtx context.Context) (bool, error) {
```
hongkailiu, line 386: nit: `200*time.Millisecond` looks like very diligent polling.
I may do it 1 or 2 sec instead.

/lgtm

/hold
in case you want to change.
Free to ignore.

DavidHurta: You are correct. I was being consistent with the above polling, but there is no need to be **that** diligent. I believe it's okay enough for the testing functionality, but I appreciate your comment. We may review the entire integration testing for the polling values in the future if priority allows for it. 

/unhold

DavidHurta:
it's aliveeee

PR #1386, opened 2026-05-07
CNTRLPLANE-3429: Respect configured cipher order for inject-tls

vincentdephily:
Cipher order is meaningful, don't sort alphabetically.

* Remove unwanted string sort
* Adjust unittests
* Tweak logs

## Summary by CodeRabbit

* **Bug Fixes**
  * TLS injection now preserves the observed cipher-suite order, validates both operator and controller ConfigMap entries, skips unsupported entries, and improves logging with the detected config kind and accepted kinds.
  * TLS observation no longer sorts cipher suites before producing the resulting TLS configuration.

* **Tests**
  * Updated fixtures and test helpers to treat cipher order as significant, verify injection/update behavior (including reversed-order scenarios), and cover both operator and controller config kinds.

### DIFF
--- a/lib/resourcebuilder/core.go
+++ b/lib/resourcebuilder/core.go
@@ -40,13 +40,15 @@ type tlsConfig struct {
 	cipherSuites  optional[[]string]
 }
 
+// modifyConfigMap sets/clears minTLSVersion and cipherSuites in the CM's data entries if
+// * ConfigMapInjectTLSAnnotation is "true"
+// * Data entry kind is GenericOperatorConfig or GenericControllerConfig
 func (b *builder) modifyConfigMap(ctx context.Context, cm *corev1.ConfigMap) error {
 	// Check for TLS injection annotation
 	if value, ok := cm.Annotations[ConfigMapInjectTLSAnnotation]; !ok || value != "true" {
 		return nil
 	}
-
-	klog.V(2).Infof("ConfigMap %s/%s has %s annotation set to true", cm.Namespace, cm.Name, ConfigMapInjectTLSAnnotation)
+	klog.V(2).Infof("ConfigMap %s/%s has annotation %s: true", cm.Namespace, cm.Name, ConfigMapInjectTLSAnnotation)
 
 	// Empty data, nothing to inject into
 	if cm.Data == nil {
@@ -68,10 +70,10 @@ func (b *builder) modifyConfigMap(ctx context.Context, cm *corev1.ConfigMap) err
 	if tlsConf.cipherSuites.found {
 		cipherSuitesLog = fmt.Sprintf("%v", tlsConf.cipherSuites.value)
 	}
-	klog.V(4).Infof("ConfigMap %s/%s: observed minTLSVersion=%v, cipherSuites=%v",
+	klog.V(4).Infof("ConfigMap %s/%s will apply observed minTLSVersion=%v, cipherSuites=%v",
 		cm.Namespace, cm.Name, minTLSLog, cipherSuitesLog)
 
-	// Process each data entry that contains GenericOperatorConfig
+	// Process each data key
 	for key, value := range cm.Data {
 		klog.V(4).Infof("Processing %q key", key)
 		// Parse YAML into RNode to preserve formatting and field order
@@ -82,18 +84,19 @@ func (b *builder) modifyConfigMap(ctx context.Context, cm *corev1.ConfigMap) err
 			continue
 		}
 
-		// Check if this is a supported config kind
+		// Check if this is a supported data node kind
+		rnodeKind := rnode.GetKind()
 		switch {
-		case rnode.GetKind() == "GenericOperatorConfig" && rnode.GetApiVersion() == operatorv1alpha1.GroupVersion.String():
-		case rnode.GetKind() == "GenericControllerConfig" && rnode.GetApiVersion() == configv1.GroupVersion.String():
+		case rnodeKind == "GenericOperatorConfig" && rnode.GetApiVersion() == operatorv1alpha1.GroupVersion.String():
+		case rnodeKind == "GenericControllerConfig" && rnode.GetApiVersion() == configv1.GroupVersion.String():
 		default:
 			klog.V(4).Infof("ConfigMap's %q entry is not a supported config type. Only GenericOperatorConfig (%v) and GenericControllerConfig (%v) are. Skipping this entry", key, operatorv1alpha1.GroupVersion.String(), configv1.GroupVersion.String())
 			continue
 		}
 
-		klog.V(2).Infof("ConfigMap %s/%s processing GenericOperatorConfig in key %s", cm.Namespace, cm.Name, key)
+		klog.V(2).Infof("ConfigMap %s/%s processing %s in key %s", cm.Namespace, cm.Name, rnodeKind, key)
 
-		// Inject TLS settings into the GenericOperatorConfig while preserving structure
+		// Inject TLS settings into the data node while preserving structure
 		if err := updateRNodeWithTLSSettings(rnode, tlsConf); err != nil {
 			return fmt.Errorf("failed to inject the TLS configuration: %v", err)
 		}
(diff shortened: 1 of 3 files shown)


lib/resourcebuilder/core_test.go
```diff
@@ -77,10 +76,10 @@ func makeGenericControllerConfigYAML(cipherSuites []string, minTLSVersion string
 	return makeGenericConfigYAML(configv1.GroupVersion.String(), "GenericControllerConfig", cipherSuites, minTLSVersion)
 }
 
-func getDefaultCipherSuitesSorted() []string {
-	cipherSuites := crypto.CipherSuitesToNamesOrDie(crypto.DefaultCiphers())
```
ingvagabund, line 81: Why not just drop the sort.Strings statement?

lib/resourcebuilder/core_test.go
```diff
@@ -99,35 +98,37 @@ const (
...
 		"ECDHE-RSA-AES128-GCM-SHA256",
 	}
-
+	// Same as previous, but in a different order
 	testOpenSSLCipherSuitesReversed = []string{
 		"ECDHE-RSA-AES128-GCM-SHA256",
 		"ECDHE-ECDSA-AES128-GCM-SHA256",
 	}
 
+	// Alternate cipher suite
```
ingvagabund: Can this change be made on the original lines? To simplify the diff.

lib/resourcebuilder/core_test.go
```diff
@@ -99,35 +98,37 @@ const (
...
+	// Same as previous, but using OpenSSL ids
 	testOpenSSLCipherSuites2 = []string{
+		"AES128-SHA256",
 		"AES128-GCM-SHA256",
 		"AES256-GCM-SHA384",
-		"AES128-SHA256",
 	}
 
-	defaultCipherSuites = getDefaultCipherSuitesSorted()
+	// Default cipher suite returned bu library-go
```
ingvagabund: nit: bu -> by

lib/resourcebuilder/core_test.go
vincentdephily: Because `crypto.DefaultCiphers()` and `configv1.TLSProfiles[crypto.DefaultTLSProfileType]` return the same ciphers but in a different order. The new TLS1.3 ciphers have a higher precedence in `os library-go` than in `golang crypto`.

Ideally I would have taken the cipher suite by calling `apiserver.ObserveTLSSecurityProfile()`, but the context args made that akward. So instead I extracted the core logic for the 'default case', and added a comment. If the apiserver code drifts, we'll have to update this test.

lib/resourcebuilder/core_test.go
vincentdephily: I changed the order because having related suites together felt more readable to me (in particular to notice same ciphers with different IDs), but if you feel otherwise I can keep the old order and reword the comments ?

lib/resourcebuilder/core.go
DavidHurta: nit:

Thank you for the fix! Could we also update the various existing comments referencing `GenericOperatorConfig` in the file to something more generic, perhaps? There are a few of them.

lib/resourcebuilder/core.go
```diff
@@ -145,8 +144,6 @@ func (b *builder) observeTLSConfiguration(ctx context.Context, cm *corev1.Config
 	if cipherSuites, ciphersFound, err := unstructured.NestedStringSlice(observedConfig, "servingInfo", "cipherSuites"); err != nil {
 		return nil, err
 	} else if ciphersFound {
-		// Sort cipher suites for consistent ordering
-		sort.Strings(cipherSuites)
```
DavidHurta, line 149: This is very interesting.

Have you reached out to the relevant centralized-tls-profile folks about whether the API description should be updated?

Currently, the relevant APIServer API is described as [[1](https://github.com/openshift/api/blob/9b69e5fa27e94ac3e4f61a94b56e8a0034ea4aee/config/v1/types_tlssecurityprofile.go#L145-L165)]:

```go
// TLSProfileSpec is the desired behavior of a TLSSecurityProfile.
type TLSProfileSpec struct {
	// ciphers is used to specify the cipher algorithms that are negotiated
	// during the TLS handshake. Operators may remove entries that their operands
	// do not support. For example, to use only ECDHE-RSA-AES128-GCM-SHA256 (yaml):
	//
	//   ciphers:
	//     - ECDHE-RSA-AES128-GCM-SHA256
	//
	// TLS 1.3 cipher suites (e.g. TLS_AES_128_GCM_SHA256) are not configurable
	// and are always enabled when TLS 1.3 is negotiated.
	// +listType=atomic
	Ciphers []string `json:"ciphers"`
```

No mention of preferred ordering. Maybe there should be?

Ideally, the majority of components use library functions, which hopefully respect the ordering. Hopefully. However, at a minimum, some components will not respect this.

Also, the tls-scanner compliance check does not seem to check the ordering [[2](https://github.com/openshift/tls-scanner/blob/eee0a3a8dc371d60170e7a4bf4c73b9ce3498eac/internal/scanner/compliance.go#L95-L123)] as well. Maybe it should?

ingvagabund: Just a note: this will be less a problem from TLS v1.3 given the list of ciphers will be ignored/rejected.

lib/resourcebuilder/core_test.go
ingvagabund: > The new TLS1.3 ciphers have a higher precedence in os library-go than in golang crypto.

That does not play well with the cipher ordering either. Both [config/v1/types_tlssecurityprofile.go](https://github.com/openshift/api/blob/9ee7e57fc8cb331eff64e17285db5741c587e38e/config/v1/types_tlssecurityprofile.go#L229-L237) and [DefaultCiphers()](https://github.com/openshift/library-go/blob/d918b32c6dbf3f6715792a4164960b968a06d2b1/pkg/crypto/crypto.go#L296-L307) need to respect the same ordering then. Maybe noone has though of it before?

lib/resourcebuilder/core_test.go
ingvagabund: Ok. Let's keep it as it is then.

lib/resourcebuilder/core.go
vincentdephily: Having [investigated this properly](https://github.com/vincentdephily/tls_cipher_choice), it turns out that Go actually does ignore the configured cipher order (and cipher list, with TLS1.3). This is Go-specific behaviour (and new since 1.17), so I don't feel too bad for having assumed that order mattered :sweat_smile: 

I'll update this PR with clarifying comments.

There's a future-proofing/least-surprise argument to be made for keeping the user's order, but I think that the config-reconcilation-minimizing argument wins out, so I'll put the sort back in place unless you have other opinions.

ingvagabund: Nice exploration. This is quite surprising.

Yeah, dropping down a comment makes sense here.

DavidHurta: Yes, a nice exploration indeed :bow: 

> I think that the config-reconcilation-minimizing argument wins out

Me as well :+1:

vincentdephily:
I've identified that unittest race condition with @ricardomaraschini but I'll treat it in another PR, just retrying here.

/retest

vincentdephily:
PR updated: I've addressed the comment nitpicks, added/removed/tweaked some testcases, refactored the test code... And kept the "no sorting" behavior despite my earlier opinion about  minimizing config reconciliation.

The rationale is in the commit message and code comments, but the main arguments are that we don't want the CVO to inject a different config than what the operator can get via the APIServer, and that we don't want to make an assumption that might not be valid for some operators.

I've spent a lot of time trying to get the best of both worlds by keeping the order when injecting, but only injecting when it makes a semantic difference (eg ignoring order, old-style names, etc). But the deeper I dug, the more opinionated quirks I found in Go and library-go. Too many opportunities for bitrot and false negatives, so I deleted those attempts and kept things simple, stupid.

vincentdephily:
[e2e-agnostic-ovn-upgrade-into-change flaky](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change)

/retest

DavidHurta (approved):
Thank you for the changes! 🙇

vincentdephily:
/retest

Clearly unrelated failures

PR #1382, opened 2026-04-24
OTA-1966: Manage proposals

hongkailiu:
Will rebase after https://github.com/openshift/cluster-version-operator/pull/1381 gets in: Done.

This pull implements the following logic in the `Sync` function of proposal lifecycle controller:

- Create proposals for each direct update path.
- Clean up the proposals if needed.

The image used in Proposals is `quay.io/openshift/ci:ocp_5.0_agentic-skills`. It is built in CI. Prow jobs have the pull credentials to access it. For manual testing, https://docs.ci.openshift.org/how-tos/use-registries-in-build-farm/#how-do-i-gain-access-to-qci provides how to retrieve it.

ReadinessJson will be implemented with another pull.

~[OTA-1971](https://redhat.atlassian.net/browse/OTA-1971): Implement OlsWorkflow and OlsAgent management for the cluster-version operator~

~Each proposal uses the OLS `Workflow/ota-advisory` which uses the OLS `Agent/ota-advisor` to do analysis. The latter uses  a [SystemPrompt](https://developers.cloudflare.com/ai-search/configuration/retrieval/system-prompt/) stored in `ConfigMpag/ota-advisory-prompt`. Those static manifests, taken from https://github.com/openshift/cluster-version-operator/pull/1379, get installed on a tech-preview cluster after this pull.~

[OTA-1971](https://redhat.atlassian.net/browse/OTA-1971): Add ConfigMap/ota-advisory-prompt

* The manifest is taken from [1]. It is used to build the request for
the proposals.

* The Workflow is removed from the implementation. There is no CRD defined
for it.

* CVO does not maintain its own Agent. For testing purpose, it uses a default
agent `smart` maintained by a cluster admin.

In additional, the manifest for `namespace/openshift-lightspeed` is added because the fore-mentioned ConfigMap is there and it is also the default namespace for proposals. 

## Summary by CodeRabbit

* **New Features**
  * Added Proposal, Workflow, Agent, LlmProvider and OLSConfig APIs, feature-gated proposal controller, and install manifests (namespace, prompt ConfigMap, Agent, Workflow).

* **Bug Fixes & Improvements**
  * Proposal reconciliation hooks into cluster-version sync; semver-based update classification added; proposals enqueued when status history is pruned.

* **Tests**
  * New unit and integration tests covering proposal generation, deletion, sync behavior, and end-to-end proposal creation.

PR #1394, opened 2026-05-26
OTA-1966: Install namespace/openshift-lightspeed only when TechPreview is enabled

hongkailiu:
I forgot to do this in https://github.com/openshift/cluster-version-operator/pull/1382.

The feature of creating proposals is [enabled only on TP clusters](https://github.com/openshift/cluster-version-operator/blob/fa129da3b3cba865873715ec88a0fafde98f6585/pkg/cvo/cvo.go#L1215-L1219). The namespace is only for testing that feature. Installing it only on TP clusters is to keep the damage away from any production clusters.

The renaming of ConfigMap from `ota-advisory-prompt` to `cluster-update-advisory-prompt` is to avoid using an obsolete team name.

## Summary by CodeRabbit

* **Chores**
  * Added a TechPreviewNoUpgrade feature-set annotation to cluster metadata.
  * Renamed the cluster advisory prompt resource to a clearer name for update notices.

* **Documentation**
  * Clarified namespace description to specify removal timeframe ("before 5.0 GA") and added guidance for clusters using the Lightspeed operator, including installing the advisory prompt when applicable.

### DIFF
--- a/install/0000_00_cluster-version-operator_45_openshift-lightspeed_namespace.yaml
+++ b/install/0000_00_cluster-version-operator_45_openshift-lightspeed_namespace.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   name: openshift-lightspeed
   annotations:
-    kubernetes.io/description: This manifest is only for testing purpose and will be removed when its own operator becomes available on the cluster.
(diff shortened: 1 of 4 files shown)


wking (approved):
/lgtm

wking (approved):
/lgtm

hongkailiu:
/verified by @hongkailiu 

CI is green.

PR #1403, opened 2026-06-11
OCPCLOUD-3368: bumping with o/api changes to reflect new capabilities (crdcompatibilitychecker+capi)

miyadav:
bumping changes from my fork to add new [capabilities](https://github.com/openshift/api/pull/2884) .
WIP - [OCPCLOUD-3368](https://redhat.atlassian.net/browse/OCPCLOUD-3368)
/hold

Generated by - claude-opus-4-6(2.1.169)

## Summary by CodeRabbit

* **Bug Fixes**
  * Cluster version capability reporting now respects feature gates, ensuring gated capabilities are excluded from reported results when their gates are not enabled.
* **Chores**
  * Updated module dependency resolution via an API dependency redirect.
* **Tests**
  * Refreshed CVO scenario expectations and expanded capability unit tests to validate feature-gate filtering and excluded-capability behavior.


lib/capability/capability.go
```diff
@@ -9,6 +9,54 @@ import (
...
+// annotations on ClusterVersionCapability in the openshift/api types.
+var featureGatedCapabilities = map[configv1.ClusterVersionCapability][]string{
+	configv1.ClusterVersionCapabilityCompatibilityRequirements: {
+		"CRDCompatibilityRequirementOperator",
+		"ClusterAPIMachineManagement",
+	},
+	configv1.ClusterVersionCapabilityClusterAPI: {
+		"ClusterAPIMachineManagement",
+	},
+}
```
hongkailiu: Could we reuse `configv1.FeatureGateName` instead of string? like:

```suggestion
var featureGatedCapabilities = map[configv1.ClusterVersionCapability][]configv1.FeatureGateName{
	configv1.ClusterVersionCapabilityCompatibilityRequirements: {
		features.FeatureGateCRDCompatibilityRequirementOperator,
		features.FeatureGateClusterAPIMachineManagement,
	},
	configv1.ClusterVersionCapabilityClusterAPI: {
		features.FeatureGateClusterAPIMachineManagement,
	},
}
```

lib/capability/capability.go
```diff
@@ -9,6 +9,54 @@ import (
...
+		if !enabled {
+			excluded.Insert(cap)
+		}
+	}
+	return excluded
+}
+
+// FilterByFeatureGates returns the subset of capabilities whose required
+// feature gates are enabled.
+func FilterByFeatureGates(capabilities []configv1.ClusterVersionCapability,
```
hongkailiu: Is it used only by unit tests?

lib/capability/capability.go
```diff
@@ -9,6 +9,54 @@ import (
...
+	excluded := sets.New[configv1.ClusterVersionCapability]()
+	for cap, gates := range featureGatedCapabilities {
+		enabled := false
+		for _, gate := range gates {
+			if enabledFeatureGates.Has(gate) {
+				enabled = true
+				break
+			}
+		}
+		if !enabled {
```
hongkailiu: nit: feels like we could use the syntactical sugur from `sets`:

```suggestion
		if enabledFeatureGates.Intersection(sets.New[configv1.FeatureGateName](gates...)).Len() > 0 {
```

lib/capability/capability.go
```diff
@@ -9,6 +9,54 @@ import (
...
+		"ClusterAPIMachineManagement",
+	},
+	configv1.ClusterVersionCapabilityClusterAPI: {
+		"ClusterAPIMachineManagement",
+	},
+}
+
+// gatedCapabilities returns the set of capabilities that should be excluded
+// because none of their enabling feature gates are in enabledFeatureGates.
+func gatedCapabilities(enabledFeatureGates sets.Set[string]) sets.Set[configv1.ClusterVersionCapability] {
```
hongkailiu: nit: `excludedCapabilities` sounds simpler to me.

```suggestion
func excludedCapabilities(enabledFeatureGates sets.Set[string]) sets.Set[configv1.ClusterVersionCapability] {
```

lib/capability/capability.go
```diff
@@ -35,15 +83,19 @@ func SortedList(s sets.Set[configv1.ClusterVersionCapability]) []configv1.Cluste
 // SetCapabilities populates and returns cluster capabilities from ClusterVersion's capabilities specification and a
 // collection of capabilities that are enabled including implicitly enabled.
 // It ensures that each capability in the collection is still enabled in the returned cluster capabilities
-// and recognizes all implicitly enabled ones.
+// and recognizes all implicitly enabled ones. Capabilities gated behind feature gates that are not in
+// enabledFeatureGates are excluded from both Known and Enabled sets.
```
hongkailiu: Do I understand it correctly?
If a cap is in `excluded`, it is retreated as not Known, not Enabled, and not implicitlyEnabled?

If that is the case, we could do on the `clusterCapabilities` object before returning and keep the rest of the function intact.
Something like:

```go
	clusterCapabilities.Known=clusterCapabilities.Known.Difference(excluded)
	clusterCapabilities.Enabled=clusterCapabilities.Enabled.Difference(excluded)
	clusterCapabilities.ImplicitlyEnabled=clusterCapabilities.ImplicitlyEnabled.Difference(excluded)
```

Another question: what is the difference between a unknown cap and an excluded cap?

To me, all the [caps defined the current version of the cluster](https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/installation_overview/cluster-capabilities) are known, both to the cluster and the admins.

The excluded ones are kind of implicitly disabled but they are known.

hongkailiu (commented):
I do not know much about cluster-api and some of my comments are only to clear up my own confusion.

lib/capability/capability.go
miyadav: yes (cvo scenario and capability test outside hence exported)

lib/capability/capability.go
miyadav: unknown is  API server rejects the request and excluded is when code exists in the binary and its constants exist in the API types, but it's hidden until its feature gate is enabled .

hongkailiu: Would a user be confused, if they see a cap is listed in the OpenShift docs, but it is not visible (because it is hidden) from the known API?
Are we going to talk about this in the docs?

lib/capability/capability.go
hongkailiu: I am sure if I follow.
Is `FilterByFeatureGates` going to imported by other repos?

- If yes, I think `openshift/library-go` is a better place for the function.
- If no, we do not see why we define a function in the core code but not used in the core code.

lib/capability/capability.go
miyadav: There is a doc [card](OSDOCS-17389) , We can add this to it. Will check further , thanks .

lib/capability/capability.go
miyadav: it might be useful to oc , but at this moment not sure for lib-go , it should have more use cases I think .

hongkailiu: I would strongly suggest to move it to the testing code now, and add the code back to core ONLY when we really need it, or have a clear use case.
From my experience, otherwise, we will forget why it exists very soon. ^_^

lib/capability/capability.go
```diff
@@ -35,16 +81,28 @@ func SortedList(s sets.Set[configv1.ClusterVersionCapability]) []configv1.Cluste
...
 	clusterCapabilities.Enabled, clusterCapabilities.ImplicitlyEnabled =
 		categorizeEnabledCapabilities(config.Spec.Capabilities, capabilities)
 
+	clusterCapabilities.Enabled = clusterCapabilities.Enabled.Difference(excluded)
+	if clusterCapabilities.ImplicitlyEnabled != nil {
+		clusterCapabilities.ImplicitlyEnabled = clusterCapabilities.ImplicitlyEnabled.Difference(excluded)
+		if clusterCapabilities.ImplicitlyEnabled.Len() == 0 {
+			clusterCapabilities.ImplicitlyEnabled = nil
+		}
+	}
```
hongkailiu: Sorry. I forgot to handle the `nil` in the last round:

```suggestion
	clusterCapabilities.Enabled = exclude(clusterCapabilities.Enabled, excluded)
	clusterCapabilities.ImplicitlyEnabled = exclude(clusterCapabilities.ImplicitlyEnabled, excluded)
```

with a syntax sugar function:

```go
func exclude(current, excluded sets.Set[configv1.ClusterVersionCapability]) sets.Set[configv1.ClusterVersionCapability] {
	if current == nil {
		return nil
	}
	ret := current.Difference(excluded)
	if ret.Len() == 0 {
		return ret
	}
	return ret
}
```

It will handle the `nil` on `clusterCapabilities.Enabled` as well.

lib/capability/capability.go
```diff
@@ -99,7 +157,9 @@ func categorizeEnabledCapabilities(capabilitiesSpec *configv1.ClusterVersionCapa
 	enabled := sets.New[configv1.ClusterVersionCapability](configv1.ClusterVersionCapabilitySets[capSet]...)
 
 	if capabilitiesSpec != nil {
-		enabled.Insert(capabilitiesSpec.AdditionalEnabledCapabilities...)
+		for _, c := range capabilitiesSpec.AdditionalEnabledCapabilities {
+			enabled.Insert(c)
+		}
```
hongkailiu: Do we still need this change?

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -38,12 +39,20 @@ import (
...
+var sortedCaps []configv1.ClusterVersionCapability
+var sortedKnownCaps []configv1.ClusterVersionCapability
 
 func init() {
 	architecture = runtime.GOARCH
 
+	// The scenario tests run without feature gates enabled, so filter out
+	// capabilities gated behind disabled feature gates to match runtime behavior.
+	noGates := sets.New[string]()
+	sortedCaps = capability.FilterByFeatureGates(
```
hongkailiu: Is `FilterByFeatureGates` sorting the result?

lib/capability/capability_test.go
```diff
@@ -286,3 +381,27 @@ func TestSetFromImplicitlyEnabledCapabilities(t *testing.T) {
 		})
 	}
 }
+
+func TestFilterByFeatureGates(t *testing.T) {
```
hongkailiu: We could remove the test if the target function is not in the core code.

lib/capability/capability.go
hongkailiu: After reading the usage of `FilterByFeatureGates` again, I think we can just remove the function completely. If it is useful to `oc`, we can implement and test it in `oc`.

For `cvo`, we can just use the result of `FilterByFeatureGates` directly.

pkg/cvo/cvo_scenarios_test.go
```diff
@@ -38,18 +39,19 @@ import (
...
-		return sortedCaps[i] < sortedCaps[j]
-	})
-	sort.Slice(sortedKnownCaps, func(i, j int) bool {
-		return sortedKnownCaps[i] < sortedKnownCaps[j]
-	})
+	// The scenario tests run without feature gates enabled, so derive the
+	// expected capability lists from SetCapabilities directly.
+	// SortedList returns capabilities in sorted order.
+	noGates := sets.New[string]()
+	caps := capability.SetCapabilities(&configv1.ClusterVersion{}, nil, noGates)
```
hongkailiu: nit: a bit simpler:


```suggestion
	caps := capability.SetCapabilities(&configv1.ClusterVersion{}, nil, nil)
```

lib/capability/capability_test.go
```diff
@@ -11,37 +11,67 @@ import (
...
+// allFeatureGates returns a feature gate set with all gates that gate capabilities enabled.
+func allFeatureGates() sets.Set[string] {
+	gates := sets.New[string]()
+	for _, capGates := range featureGatedCapabilities {
+		for _, g := range capGates {
+			gates.Insert(string(g))
+		}
+	}
+	return gates
+}
```
hongkailiu: nit:

I would use something like the following directly.


```suggestion
var allFeatureGatesA = sets.New[string](string(features.FeatureGateCRDCompatibilityRequirementOperator),
	string(features.FeatureGateClusterAPIMachineManagement))
```

lib/capability/capability_test.go
```diff
@@ -11,37 +11,67 @@ import (
...
+
+func filterExcluded(capabilities []configv1.ClusterVersionCapability, enabledGates sets.Set[string]) []configv1.ClusterVersionCapability {
```
hongkailiu, line 20: nit: the same here. I would define the list directly in the unit tests.

The function is used to manipulate the expected results. It is not easy to see what is the calculated results. I have to admit that I learned this from the hard way.

But this is in the testing code now and i do not want to block the merge on this.

lib/capability/capability_test.go
```diff
@@ -147,13 +237,29 @@ func TestSetCapabilitiesWithImplicitlyEnabled(t *testing.T) {
...
 		},
+		{name: "prior-enabled gated capability with gate disabled",
+			config: &configv1.ClusterVersion{
+				Spec: configv1.ClusterVersionSpec{
+					Capabilities: &configv1.ClusterVersionCapabilitiesSpec{
+						BaselineCapabilitySet: configv1.ClusterVersionCapabilitySetCurrent,
+					},
+				},
+			},
+			enabledGates: sets.New[string](),
```
hongkailiu: nit: we do not have too set up the zero value.

```suggestion
```

lib/capability/capability_test.go
```diff
@@ -102,32 +136,87 @@ func TestSetCapabilities(t *testing.T) {
...
+			// Verify gated capabilities are absent when gate is disabled
+			excluded := excludedCapabilities(test.enabledGates)
+			for cap := range excluded {
+				if caps.Known.Has(cap) {
+					t.Errorf("Gated capability %q should not be in Known when gate is disabled", cap)
+				}
+				if caps.Enabled.Has(cap) {
+					t.Errorf("Gated capability %q should not be in Enabled when gate is disabled", cap)
+				}
+			}
```
hongkailiu, line 217: Do we verify `if caps.ImplicitlyEnabled.Has(cap) ` as well?

hongkailiu (commented):
Only a few NITs.
Free to ignore tho.

/lgtm

damdo:
@hongkailiu @miyadav who are we expecting to put the verified on this? TY!

miyadav:
/verified by miyadav

premerged test mentioned in Jira and [here](https://gitlab.cee.redhat.com/miyadav/promptdetails/-/blob/main/ocpcloud3368tpresultscomparison?ref_type=heads)

miyadav:
> @hongkailiu @miyadav who are we expecting to put the verified on this? TY!

I have verified it .

PR #1415, opened 2026-06-26
OTA-1966: pkg/readiness/crd_compat: Drop unnecessary CustomResourceDefinition check

wking:
We grew this check recently in b0e4c90a3d (#1395).  But I don't think we need it.  I agree with the guard, having a CRD with storage versions that were not served would be weird, and probably not what the CRD maintainers intended.  But what would a cluster-admin be expected to do about it?  There's nothing in the code I'm removing here, or in
https://github.com/openshift/agentic-skills as far as I can see, that turns `stored version no longer served` into an actionable plan.

In the CRD-evolution space, OpenShift packages the Kube storage-version migrator.  But [we don't run the trigger logic][1]. And, as far as I can tell, the migration operator isn't checking to see if any [StorageVersionMigrations are completed][2].  I expect that, if OpenShift ever decides to remove support for a previously-stored version, we'll set up something to run (and require the completion of) a StorageVersionMigration to ensure storage is off the outgoing version before we leave the last release where it was served.  So that would protect OpenShift users from OpenShift-managed CRD evolution, without needing the guard I'm deleting.

In addition to OpenShift-managed CRDs, there might be many CRDs on an OpenShift cluster that are not part of the OpenShift release payload.  They may have been installed by OLM-installed operators.  They may have been installed by individual cluster users running `oc apply ...`.  Lots of other folks who could be maintaining CRDs on the cluster.  We don't want to confuse admins who are considering an OpenShift version update by any issues that might be going on with those non-OpenShift CRDs.

[1]: https://github.com/openshift/cluster-kube-storage-version-migrator-operator/pull/90/changes/807c909d19d0fcde043d2d54550c34b3c5053c7e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R7
[2]: https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/USER_GUIDE.md#check-if-migration-has-completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated readiness reporting to omit CRD compatibility details, reducing the total number of checks reported.
  * Adjusted readiness output expectations to match the reduced check set.

* **Tests**
  * Updated readiness test fixtures and assertions to remove CRD compatibility scenarios.
  * Revised expected total and successful check counts from 9 to 8 and removed related check-name/verifications.


pkg/proposal/controller_test.go
```diff
@@ -1372,7 +1359,7 @@ func TestGetProposals_WithReadinessData(t *testing.T) {
 	for _, name := range []string{
 		"cluster_conditions", "operator_health", "api_deprecations",
 		"node_capacity", "pdb_drain", "etcd_health", "network",
-		"crd_compat", "olm_operator_lifecycle",
+		"olm_operator_lifecycle",
```
harche, line 1362: Anchoring here since the lines are just above. The two count assertions at 1346 and 1349 still hardcode 9. With `crd_compat` gone there are 8 checks so this test fails.

```
controller_test.go:1347: readiness total_checks = 8, want 9
controller_test.go:1350: readiness checks_ok = 8, want 9
```

Both need to be `float64(8)`.

harche (commented):
Some context on why this existed since I was involved in adding it. The check came in with the readiness set in #1395. The intent was to catch CRDs where a stored version is no longer served, because that state can leave objects sitting in etcd that nothing can decode anymore, and the thought was to surface that to the admin before an upgrade so they could make an informed go or no-go call.

Having sat with it, I think it should go, but not because the signal is useless. The problem is it answers the wrong question. It scans the cluster as it is right now and flags CRDs that are already in the stranded state. Those objects are already unreadable today, with or without the upgrade, so it isn't really telling you whether this upgrade is safe, it's telling you something on your cluster is already broken and unrelated to the upgrade. It also never read `current` or `target`, it ignored both, which is another way of seeing that it was a present state hygiene scan rather than an upgrade readiness check.

The genuinely upgrade relevant question is the forward looking one. Will this upgrade strand any objects, which happens when the target release stops serving a version you still have objects stored in. Answering that means diffing the target release's CRD served versions against the cluster's current storedVersions. The deleted check did none of that. So if we want this signal back, that target aware version is the one worth building, and it would actually support the before hand decision this one was meant to.

One thing to fix first. This breaks `TestGetProposals_WithReadinessData`. Most of the 9 to 8 updates landed but the two count assertions in `controller_test.go` still expect 9, so `go test ./pkg/proposal/...` fails.

pkg/proposal/controller_test.go
wking: Thanks!  Fixed with f1868ade2 -> 2377aaa62.  Marking this thread resolved.

harche: Thanks. 

/lgtm

wking:
[`blocks manual creation of Endpoints pointing to the cluster or service network` 403][1] and [`Netpol NetworkPolicy between server and client should enforce policy to allow ingress traffic from pods in all namespaces`][2] timeout are unrelated:

/override ci/prow/e2e-aws-ovn-techpreview

No consumers yet, without having OLS installed, and the unit-test coverage of our Proposal handling is enough to make me comfortable that's safe enough.

/verified by presubmits passing

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1415/pull-ci-openshift-cluster-version-operator-main-e2e-aws-ovn-techpreview/2071727203787214848
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1415/pull-ci-openshift-cluster-version-operator-main-e2e-hypershift-conformance/2071727203829157888

wking:
[`HTPasswd IDP should successfully configure htpasswd and be responsive` EOFs][1] are unrelated:

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1415/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn/2072395420620820480

PR #1410, opened 2026-06-22
OTA-1956: pkg/payload: Add Images map and tolerate unknown template fields during upgrades

jhadvig:
## Summary

Two changes to the CVO manifest template engine:

**1. Add `Images` map to `manifestRenderConfig`**

Populated from the release payload's `image-references` ImageStream. CVO manifests in `/manifests/` can now reference component images by short name using Go template syntax: `{{index .Images "component-name"}}`.

**2. Tolerate unknown template fields during `LoadUpdate`**

When an older CVO binary loads a newer payload that uses template fields the older binary does not know about (e.g. `.Images`), the manifest is skipped with a warning instead of failing the entire payload load. The new CVO binary re-loads the full payload after replacing the old one at run-level 0.

This is needed for upgrade safety. Without it, adding any new template field to a CVO manifest would break upgrades — the old CVO would fail to load the new payload before it can replace itself.

## How it works during upgrade

```
Old CVO (no .Images) loads new payload:
  ✅ CVO deployment.yaml     → {{.ReleaseImage}} renders fine
  ⚠️  console-plugin.yaml    → {{.Images}} unknown → skipped with warning

Old CVO applies run-level 00 → replaces itself

New CVO (has .Images) re-loads payload:
  ✅ CVO deployment.yaml     → {{.ReleaseImage}} renders fine
  ✅ console-plugin.yaml     → {{.Images}} renders fine → applied
```

## Related

- [openshift/oc#2287](https://github.com/openshift/oc/pull/2287) — keeps CVO's image-references entries in the release
- [openshift/cluster-version-operator#1398](https://github.com/openshift/cluster-version-operator/pull/1398) — console plugin manifests (follow-up, merges after this lands in a nightly)

## Release Notes

* **Bug Fixes**
  * Improved error handling during manifest preprocessing/template rendering: failed manifests are now skipped with a warning instead of impacting the overall result.

* **Improvements**
  * Enhanced manifest rendering by resolving release image references and injecting derived image mappings into template inputs (including OpenShift image stream support).

* **Tests**
  * Added coverage to ensure manifests with unknown template fields are skipped.
  * Added unit tests for image-reference-to-template mapping behavior.

PR #1421, opened 2026-07-09
OCPBUGS-98572: Tolerate unknown template fields during payload loading

jhadvig:
## Summary

- Make template rendering errors non-fatal during `LoadUpdate`
- When the CVO encounters a manifest with unknown template fields, skip it with a warning instead of failing the entire payload load

## Why

Backport of the tolerance logic from [PR #1410](https://github.com/openshift/cluster-version-operator/pull/1410) (5.0). Without this, **4.22→5.0 major upgrades are blocked**: the 4.22 CVO tries to render 5.0 manifests that use `{{index .Images "..."}}`, fails because `.Images` doesn't exist in the 4.22 `manifestRenderConfig`, and sets `ReleaseAccepted=False`.

With this fix, the 4.22 CVO skips unrenderable manifests with a warning. The new 5.0 CVO re-loads the full payload after it replaces the old one at run-level 0, correctly rendering all manifests.

This was one of two issues that caused the [revert of CVO#1398](https://github.com/openshift/cluster-version-operator/pull/1420) and blocked 5.0 nightly payloads for 69+ hours ([TRT-2776](https://redhat.atlassian.net/browse/TRT-2776)).

## Test plan

- [ ] New unit test `TestLoadUpdateSkipsUnknownTemplateFields` verifies manifests with unknown template fields are skipped while valid manifests are loaded
- [ ] CI passes

### DIFF
--- a/pkg/payload/payload.go
+++ b/pkg/payload/payload.go
@@ -195,7 +195,13 @@ func LoadUpdate(dir, releaseImage, excludeIdentifier string, requiredFeatureSet 
 			if task.preprocess != nil {
 				raw, err = task.preprocess(raw)
 				if err != nil {
-					errs = append(errs, fmt.Errorf("preprocess %s: %w", file.Name(), err))
+					// Template rendering may fail when an older CVO binary
+					// loads a newer payload that uses template fields the
+					// older binary does not know about (e.g. .Images). Skip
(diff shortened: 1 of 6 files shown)


wking (approved):
/lgtm

PR #1425, opened 2026-07-20
OTA-2084: Conditionally deploy console plugin when AgenticRun CRD is present

jhadvig:
## Summary

- Move console plugin manifests from `install/` into embedded assets (`pkg/agenticrun/bindata/`)
- Deploy the plugin conditionally based on AgenticRun CRD presence, not feature-set
- Clean up plugin resources when OLS is uninstalled

## Why

The [OCPSTRAT-2241 descoping decision](https://redhat.atlassian.net/browse/OTA-2084) requires the feature to be TP but **not** `TechPreviewNoUpgrade`. The plugin should install on any cluster where the lightspeed-agentic-operator provides the AgenticRun CRD, regardless of feature set.

Keeping manifests in `install/` without the TP annotation would deploy the plugin on every cluster, wasting resources. This PR follows the `console-operator` bindata pattern: embed manifests via `//go:embed`, apply them conditionally from the agenticrun controller.

## Changes

**Bindata structure** (`pkg/agenticrun/bindata/`):
- `assets.go` — `//go:embed assets/*` wrapper
- `assets/*.yaml` — 7 plugin manifests (namespace, serviceaccount, networkpolicy, configmap, deployment, service, consoleplugin) with gating annotations stripped and `${IMAGE}` placeholder for the deployment image

**Controller** (`pkg/agenticrun/controller.go`):
- `crdAvailable()` — checks for `agenticruns.agentic.openshift.io` via dynamic client with 5-minute TTL cache
- `ensureConsolePlugin()` — loads embedded manifests, substitutes image, applies via controller-runtime client
- `cleanupConsolePlugin()` — deletes plugin resources when CRD is absent (OLS uninstalled)
- `consolePluginEnsured` flag — skips redundant applies on subsequent syncs
- `SetConsolePluginImage()` — receives image from `image-references`, resets ensured flag on change

**CVO wiring** (`pkg/cvo/cvo.go`):
- Removed `shouldEnableAgenticRunController()` — CRD check is now inside the controller's `Sync()`
- Controller goroutine always starts (self-gating)
- Plugin image passed from `image-references` in `InitializeFromPayload()`

**Queue management** (`availableupdates.go`, `status.go`):
- Removed `shouldEnableAgenticRunController()` guards — controller queue always receives items

## Sync() flow

```
1. if !crdAvailable():
     cleanupConsolePlugin()
     return nil
2. if !consolePluginEnsured:
     ensureConsolePlugin()
     consolePluginEnsured = true
3. (existing AgenticRun lifecycle logic)
```

## Test plan

- [ ] Unit tests pass (`go test ./pkg/agenticrun/ ./pkg/cvo/`)
- [ ] Build passes (`make build`)
- [ ] TechPreview cluster with OLS: plugin deploys automatically
- [ ] Default-feature-set cluster without OLS: no plugin deployed
- [ ] OLS installed after CVO boot: plugin deploys within 5 minutes
- [ ] OLS uninstalled: plugin resources cleaned up

## Summary by CodeRabbit

* **New Features**
  * Automatically ensures console plugin manifests are applied or removed based on console-plugin support (CRD availability), using the plugin image from the release payload.
  * Added a NetworkPolicy to allow console ingress to the plugin on TCP port 9001.
* **Improvements**
  * Console plugin resources are re-applied when the configured plugin image changes.
  * The agentic-run controller is started and queued consistently, including after status history pruning; when support is unavailable, manifests are cleaned up.
  * Updated console plugin manifest naming/annotations and simplified OpenShift-specific annotations.

### DIFF
--- a/pkg/agenticrun/bindata/assets/consoleplugin.yaml
+++ b/pkg/agenticrun/bindata/assets/consoleplugin.yaml
@@ -1,13 +1,9 @@
 apiVersion: console.openshift.io/v1
 kind: ConsolePlugin
 metadata:
-  name: openshift-cluster-update-console-plugin
+  name: cluster-update-console-plugin
   annotations:
     kubernetes.io/description: The OpenShift cluster-update console plugin provides a web-console interface for managing ClusterVersion updates.
-    capability.openshift.io/name: Console
-    release.openshift.io/feature-set: TechPreviewNoUpgrade
-    exclude.release.openshift.io/internal-openshift-hosted: "true"
-    include.release.openshift.io/self-managed-high-availability: "true"
 spec:
   displayName: Cluster Updates
   i18n:
@@ -15,7 +11,7 @@ spec:
   backend:
     type: Service
     service:
-      name: openshift-cluster-update-console-plugin
+      name: cluster-update-console-plugin
       namespace: openshift-cluster-update-console-plugin
       port: 9001
       basePath: /
--- a/pkg/agenticrun/consoleplugin.go
+++ b/pkg/agenticrun/consoleplugin.go
@@ -1,0 +1,101 @@
+package agenticrun
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/cluster-version-operator/pkg/agenticrun/bindata"
+	i "github.com/openshift/cluster-version-operator/pkg/internal"
+)
+
+var consolePluginAssets = []string{
+	"assets/namespace.yaml",
+	"assets/serviceaccount.yaml",
+	"assets/networkpolicy.yaml",
+	"assets/networkpolicy-allow-console.yaml",
+	"assets/configmap.yaml",
+	"assets/deployment.yaml",
+	"assets/service.yaml",
+	"assets/consoleplugin.yaml",
+}
+
+func applyConsolePluginManifests(ctx context.Context, client ctrlruntimeclient.Client, image string) error {
+	for _, asset := range consolePluginAssets {
+		raw := bindata.MustAsset(asset)
+
+		if asset == "assets/deployment.yaml" {
+			raw = []byte(strings.ReplaceAll(string(raw), "${IMAGE}", image))
+		}
+
+		obj := &unstructured.Unstructured{}
+		if err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(string(raw)), len(raw)).Decode(obj); err != nil {
+			return fmt.Errorf("decoding %s: %w", asset, err)
+		}
+
+		existing := &unstructured.Unstructured{}
+		existing.SetGroupVersionKind(obj.GroupVersionKind())
+		err := client.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(obj), existing)
+		if err != nil {
+			if ctrlruntimeclient.IgnoreNotFound(err) != nil {
+				return fmt.Errorf("getting %s %s: %w", obj.GetKind(), obj.GetName(), err)
+			}
+			if err := client.Create(ctx, obj); err != nil {
+				return fmt.Errorf("creating %s %s: %w", obj.GetKind(), obj.GetName(), err)
+			}
+			klog.V(i.Normal).Infof("Created console plugin %s %s", obj.GetKind(), obj.GetName())
+			continue
+		}
+
+		if !needsUpdate(existing, obj) {
+			klog.V(i.Debug).Infof("Console plugin %s %s is up to date", obj.GetKind(), obj.GetName())
+			continue
+		}
+		obj.SetResourceVersion(existing.GetResourceVersion())
+		if err := client.Update(ctx, obj); err != nil {
+			return fmt.Errorf("updating %s %s: %w", obj.GetKind(), obj.GetName(), err)
+		}
+		klog.V(i.Normal).Infof("Updated console plugin %s %s", obj.GetKind(), obj.GetName())
+	}
+	return nil
+}
+
+func needsUpdate(existing, desired *unstructured.Unstructured) bool {
+	if !reflect.DeepEqual(existing.Object["spec"], desired.Object["spec"]) {
+		return true
+	}
+	return !reflect.DeepEqual(existing.Object["data"], desired.Object["data"])
+}
+
+func cleanupConsolePluginManifests(ctx context.Context, client ctrlruntimeclient.Client) error {
+	for idx := len(consolePluginAssets) - 1; idx >= 0; idx-- {
+		raw := bindata.MustAsset(consolePluginAssets[idx])
+
+		obj := &unstructured.Unstructured{}
+		if err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(string(raw)), len(raw)).Decode(obj); err != nil {
+			return fmt.Errorf("decoding %s: %w", consolePluginAssets[idx], err)
+		}
+
+		existing := &unstructured.Unstructured{}
+		existing.SetGroupVersionKind(obj.GroupVersionKind())
+		existing.SetName(obj.GetName())
+		existing.SetNamespace(obj.GetNamespace())
+
+		if err := client.Delete(ctx, existing); err != nil {
+			if !kerrors.IsNotFound(err) {
+				return fmt.Errorf("deleting %s %s: %w", obj.GetKind(), obj.GetName(), err)
+			}
+		} else {
+			klog.V(i.Normal).Infof("Deleted console plugin %s %s", obj.GetKind(), obj.GetName())
+		}
+	}
+	return nil
+}
--- a/pkg/agenticrun/consoleplugin.go
+++ b/pkg/agenticrun/consoleplugin.go
@@ -9,6 +9,7 @@ import (
 
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	appsv1 "k8s.io/api/apps/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -104,7 +105,21 @@ func cleanupConsolePluginManifests(ctx context.Context, client ctrlruntimeclient
 	return nil
 }
 
-const consolePluginName = "cluster-update-console-plugin"
+const (
+	consolePluginName      = "cluster-update-console-plugin"
+	consolePluginNamespace = "openshift-cluster-update-console-plugin"
+)
+
+func waitForPluginReady(ctx context.Context, client ctrlruntimeclient.Client) error {
+	deployment := &appsv1.Deployment{}
+	if err := client.Get(ctx, types.NamespacedName{Name: consolePluginName, Namespace: consolePluginNamespace}, deployment); err != nil {
+		return fmt.Errorf("getting deployment: %w", err)
+	}
+	if deployment.Status.AvailableReplicas < 1 {
+		return fmt.Errorf("deployment %s has no available replicas", consolePluginName)
+	}
+	return nil
+}
 
 func enableConsolePlugin(ctx context.Context, client ctrlruntimeclient.Client) error {
 	console := &operatorv1.Console{}
--- a/pkg/agenticrun/consoleplugin.go
+++ b/pkg/agenticrun/consoleplugin.go
@@ -2,6 +2,7 @@ package agenticrun
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -10,8 +11,11 @@ import (
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/klog/v2"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
 
 	"github.com/openshift/cluster-version-operator/pkg/agenticrun/bindata"
 	i "github.com/openshift/cluster-version-operator/pkg/internal"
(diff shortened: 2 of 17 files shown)


jhadvig:
Manually validated on a cluster-bot cluster (5.0.0-0-2026-07-20-121922):

| Scenario | Result |
|---|---|
| Cluster without OLS (no AgenticRun CRD) → no plugin deployed | **Pass** |
| Apply AgenticRun CRD → plugin auto-deploys | **Pass** — all 7 resources created (namespace, SA, networkpolicy, configmap, deployment, service, consoleplugin), pod Running 1/1 |
| Remove AgenticRun CRD → plugin resources cleaned up | **Pass** — all resources deleted after cache TTL (5 min) |
| Re-apply AgenticRun CRD → plugin re-deploys | Testing now |

pkg/agenticrun/bindata/assets/consoleplugin.yaml
```diff
@@ -0,0 +1,17 @@
...
+    kubernetes.io/description: The OpenShift cluster-update console plugin provides a web-console interface for managing ClusterVersion updates.
+spec:
+  displayName: Cluster Updates
+  i18n:
+    loadType: Preload
+  backend:
+    type: Service
+    service:
+      name: cluster-update-console-plugin
+      namespace: cluster-update-console-plugin
```
wking: I am ok with the `name` changing if necessary, but I think we need to keep the `openshift-*` prefix for the `namespace` to stay within [our reserved Namespaces][1] and [platform monitoring][2].

[1]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/building_applications/projects#working-with-projects
[2]: https://github.com/openshift/cluster-monitoring-operator/blob/e42163ebd689725a0c0f6f30ff70fb5aeb692100/assets/control-plane/prometheus-rule.yaml#L22

pkg/agenticrun/consoleplugin.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+			if err := client.Create(ctx, obj); err != nil {
+				return fmt.Errorf("creating %s %s: %w", obj.GetKind(), obj.GetName(), err)
+			}
+			klog.V(i.Normal).Infof("Created console plugin %s %s", obj.GetKind(), obj.GetName())
+			continue
+		}
+
+		obj.SetResourceVersion(existing.GetResourceVersion())
+		if err := client.Update(ctx, obj); err != nil {
```
wking, line 68: This looks like it might hot-loop on updates?  Without AgenticRun in the CI runs, I guess we don't have Kube API audit logs to know, but I'd have expected us to be using `lib/resourceapply` client-side existing-vs.-required diff logic to compare the `existing` we pulled from the cluster with our local-manifest expectations, and to only update when we saw changes we thought needed to be pushed into the cluster.

pkg/agenticrun/controller.go
```diff
@@ -124,13 +129,64 @@ func (c *Controller) QueueKey() string {
...
+		return c.crdAvailableCache
+	}
+	c.crdLastChecked = time.Now()
+	if c.dynamicClient == nil {
+		c.crdAvailableCache = false
+		return false
+	}
+	agenticRunGVR := schema.GroupVersionResource{Group: "agentic.openshift.io", Version: "v1alpha1", Resource: "agenticruns"}
+	_, err := c.dynamicClient.Resource(agenticRunGVR).List(context.Background(), metav1.ListOptions{Limit: 1})
+	c.crdAvailableCache = err == nil || !kerrors.IsNotFound(err) && !isNoMatchError(err)
```
wking: we probably want to actually check the CRD itself, with a Go analog to:

```console
$ oc get customresourcedefinitions.apiextensions.k8s.io agenticruns.agentic.openshift.io
```

That way we can simplify this `|| !kerrors.IsNotFound(err) && !isNoMatchError(err)` handling.

pkg/agenticrun/controller.go
```diff
@@ -124,13 +129,64 @@ func (c *Controller) QueueKey() string {
...
+	if !c.crdAvailable() {
+		if err := cleanupConsolePluginManifests(ctx, c.client); err != nil {
+			klog.V(i.Normal).Infof("Failed to clean up console plugin: %v", err)
+		}
+		c.consolePluginEnsured = false
+		return nil
+	}
+
+	if !c.consolePluginEnsured {
+		if err := c.ensureConsolePlugin(ctx); err != nil {
```
wking, line 200: The outgoing manifests included annotations for `capability.openshift.io/name: Console` and `exclude.release.openshift.io/internal-openshift-hosted: "true"`.  To represent the latter in Go here, we probably want to switch on [`Operator.hypershift`][1], which we'll need to plumb down to this controller.  To represent the former in Go, we'll probably need to plumb through [the sync-worker status][2], and then check for `ClusterVersionCapabilityConsole` in the enabled caps.  [Capabilities cannot be uninstalled][3], so we don't have to worry about switching `cleanupConsolePluginManifests` on either HyperShift or the `Console` capability.  But we don't want to be installing the plugin components, even if we see the AgenticRun CRDs, on clusters that are HyperShift and/or lack the `Console` capability.

[1]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/cvo/cvo.go#L158-L159
[2]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/cvo/sync_worker.go#L135
[3]: https://github.com/openshift/enhancements/blob/23373e194aa3596c0e4922cf571ab59789b51d1b/enhancements/installer/component-selection.md#capabilities-cannot-be-uninstalled

pkg/cvo/status.go
```diff
@@ -179,11 +179,9 @@ func (optr *Operator) syncStatus(ctx context.Context, original, config *configv1
...
 		klog.Infof("Apply config: %s", cmp.Diff(original, config))
 	}
-	if optr.shouldEnableAgenticRunController() {
-		if original != nil && len(config.Status.History) < len(original.Status.History) {
-			klog.V(internal.Normal).Infof("Reconciling agentic runs because ClusterVersion.status.history got pruned")
-			optr.agenticRunController.Queue().Add(optr.agenticRunController.QueueKey())
-		}
+	if original != nil && len(config.Status.History) < len(original.Status.History) {
+		klog.V(internal.Normal).Infof("Reconciling agentic runs because ClusterVersion.status.history got pruned")
+		optr.agenticRunController.Queue().Add(optr.agenticRunController.QueueKey())
```
wking, line 184: Hmm, I see this logic dates back to 6de7f5a5035fe1186ea490e79f4892fba42d61ff (#1382), but I'm personally not clear on why a history prune would need an agentic-run controller poke.  @hongkailiu, do you remember why you'd needed to do that?

hongkailiu: https://github.com/openshift/cluster-version-operator/blob/de1fddbf5fb7287c5f0212db549f05ef33e09eda/pkg/proposal/controller.go#L131

IIRC, CVO has the logic to prune `CV.history` when it grows to some length.
The proposals on the cluster are related to the versions in `CV.history`.
Thus, in my mind, CVO pruning CV.history leads to reconciliation on `proposals` .

jhadvig: ok so if I read it correctly the logic should stay as is ...

wking: > The proposals on the cluster are related to the versions in `CV.history`.

Are they?  The AgenticRuns themselves are [based on forward-looking available and conditional updates][1], not history.  `dynamicClient` gives the ability to pull in whatever data though, so we also need to [drop][2] into `runReadinessJSON`.  Looking at the current readiness checks, there is a history mention:

```console
$ git grep -il history pkg/readiness
pkg/readiness/checks_test.go
pkg/readiness/cluster_conditions.go
```

That's including [the most recent 5 history entries in `recent_history`][3].  But I don't see anything in either the CVO repo or https://github.com/openshift/agentic-skills trying to consume `recent_history`.  Maybe we can drop it?  And we're [currently strongly preferring to preserve the 5 most recent history entries when pruning][4] anyway, so it seems unlikely to me that history pruning would need fresh AgenticRun prompt generation, even if we do need to keep `recent_history` in the prompt for some reason.

[1]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/agenticrun/controller.go#L306-L345
[2]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/agenticrun/controller.go#L324
[3]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/readiness/cluster_conditions.go#L44-L62
[4]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/cvo/status_history.go#L13-L14

pkg/agenticrun/bindata/assets/consoleplugin.yaml
wking: `openshift-...` namespace restored as of 27b4b992af08276ccef718378a0b83a6a7c834d0 ; I'm marking this thread resolved.

pkg/agenticrun/controller.go
wking: As of 27b4b992af08276ccef718378a0b83a6a7c834d0 , we're using `c.client.Get(context.Background(), ctrlruntimeclient.ObjectKey{Name: "agenticruns.agentic.openshift.io"}, crd)`.  Marking this thread resolved.

pkg/agenticrun/controller.go
```diff
@@ -124,13 +130,83 @@ func (c *Controller) QueueKey() string {
...
 func (c *Controller) Sync(ctx context.Context, key string) error {
 	startTime := time.Now()
 	klog.V(i.Normal).Infof("Started syncing CVO configuration %q", key)
 	defer func() {
 		klog.V(i.Normal).Infof("Finished syncing CVO configuration (%v)", time.Since(startTime))
 	}()
 
+	if !c.crdAvailable() {
+		if err := disableConsolePlugin(ctx, c.client); err != nil {
+			klog.V(i.Normal).Infof("Failed to disable console plugin: %v", err)
```
wking: If we fail to disable the console plugin, we want to avoid cleaning up those manifests, right?  I'm not clear on all the conditions where disabling the console plugin would fail, but cleaning out the backing resources while the console plugin was still requested in the console configuration seems like it might cause console-loading problems.

pkg/agenticrun/controller.go
```diff
@@ -124,13 +130,83 @@ func (c *Controller) QueueKey() string {
...
+			klog.V(i.Normal).Infof("Failed to clean up console plugin: %v", err)
+		}
+		c.consolePluginEnsured = false
+		return nil
+	}
+
+	if c.shouldDeployConsolePlugin() && !c.consolePluginEnsured {
+		if err := c.ensureConsolePlugin(ctx); err != nil {
+			klog.V(i.Normal).Infof("Failed to ensure console plugin: %v", err)
+		} else if err := enableConsolePlugin(ctx, c.client); err != nil {
```
wking, line 204: [The docs include][1]:

> Make sure the Operator finishes installing the dynamic plugin before you run the following patch command.

It's not entirely clear what "finishes installing" means.  Is it "wait until at least one Pod that backs the Service is `Ready=True`"?  Or how gracefully does the console handle the enabled ConsolePlugin being down?  I'm noticing now that we don't set `replicas` in the Deployment manifest, so [it defaults to 1][2].  And we don't have a covering PodDisruptionBudget (yet!).  So I imagine we're often down for a bit (and possibly for an extended period) while Nodes are being drained and whatnot, and if that is handled fairly smoothly (just our cluster-update page 503ing for a bit, but the console otherwise working?), maybe we don't need anything more sophisticated here to smooth out initial-enablement time?

[1]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/web_console/dynamic-plugins#enabling-a-dynamic-plugin-by-using-the-cli_deploy-plugin-cluster
[2]: https://github.com/kubernetes/api/blame/bfcbbbe7918a67eaef358b452a3416764c34375a/apps/v1/types.go#L418

pkg/agenticrun/controller.go
```diff
@@ -124,13 +130,83 @@ func (c *Controller) QueueKey() string {
...
+			klog.V(i.Normal).Infof("Failed to disable console plugin: %v", err)
+		}
+		if err := cleanupConsolePluginManifests(ctx, c.client); err != nil {
+			klog.V(i.Normal).Infof("Failed to clean up console plugin: %v", err)
+		}
+		c.consolePluginEnsured = false
+		return nil
+	}
+
+	if c.shouldDeployConsolePlugin() && !c.consolePluginEnsured {
```
wking, line 199: `!c.consolePluginEnsured` here assumes our local-memory-cached `consolePluginEnsured` is accurate.  Possibly fine for a tech-preview offering.  But before we hit production, we'll probably want something that more actively reconciles the owned resources (like [the sync worker manifest reconciliation][1] this pull is moving us away from]) that will notice and stomp any concerning changes that leak into these managed resources (e.g. from careless admins).  Hmm, and we'll also want to look at [the other knobs that drive manifest inclusion][2] like [ClusterVersion's `spec.overrides`][3], to allow careful admins to wave off the CVO when we're doing a poor job of managing a particular resource and the cluster admin wants to take over until they can update to pick up a fix.

[1]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/cvo/sync_worker.go#L1062
[2]: https://github.com/openshift/cluster-version-operator/blob/f232fe121a512465f5f6a58afaf6bf311c1b8a60/pkg/cvo/sync_worker.go#L1096
[3]: https://github.com/openshift/api/blob/ef67063fadc9bf0494a84f45449e7e2cc55d863b/config/v1/types_cluster_version.go#L120-L129

pkg/cvo/status.go
jhadvig: You're right that AgenticRuns are forward-looking (based on available/conditional updates, not history). But `deleteAgenticRuns` also uses history as a *keep-list* — line 344 preserves runs whose target version appears in `cv.Status.History`, so they don't get deleted as "irrelevant" just because the version is no longer an available update target (e.g. after the cluster already upgraded to it).

When history is pruned (>100 entries), a version drops from that keep-list. Without the trigger, the associated run would linger until the next non-cached `syncAvailableUpdates` poke, which could be 2-4 minutes away (the controller has no self-requeue — it only runs when explicitly poked).

That said, it's a belt-and-suspenders measure.. the 2-4 minute delay is inconsequential for stale metadata, and history pruning only fires on clusters with 100+ update history entries. But the trigger is 4 lines, negligible cost, and makes the intent explicit... thats why Id lean toward keeping it, but honestly I ok with both approches.

wking: > ...so they don't get deleted as "irrelevant" ...

I thought we needed to never delete any AgenticRuns in the CVO, with [OLS-3278](https://redhat.atlassian.net/browse/OLS-3278) adding OLS-side garbage-collection, and [OLS-3328](https://redhat.atlassian.net/browse/OLS-3328) tracking log-preservation.  Am I stale?

pkg/agenticrun/controller.go
wking: This is handled as of 27b4b992af082; marking this thread resolved.

wking (approved):
There are some outstanding nits, but nothing here that's so serious that I see a reason to block this tech-preview feature work.  The [`oc adm upgrade recommend`][1] output divergence is also unrelated to this pull.

/override ci/prow/e2e-agnostic-ovn-techpreview-serial-1of3
/lgtm
/verified by CI

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1425/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial-1of3/2079925560699523072

wking:
[e2e-agnostic-ovn][1] failed to install, but [a previous run off this same commit][2] had passed, so:

[serial-2of3 had the unrelated `oc adm upgrade recommend` output issue again][3]:

/override ci/prow/e2e-agnostic-ovn-techpreview-serial-2of3

[serial-3of3 had unrelated authless-access to netobserv pods][4] :

/override ci/prow/e2e-agnostic-ovn-techpreview-serial-3of3

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1425/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn/2080024944581808128
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1425/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn/2079677366266761216
[3]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1425/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial-2of3/2080024959677108224
[4]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1425/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial-3of3/2080024968053133312

PR #1426, opened 2026-07-21
OCPBUGS-77042: use NewFromConfig instead of NewFromConfigOrDie in builder

Cali0707:
This should fix the bug we are seeing with panics when there is a missing CA cert temporarily. By moving from `NewFromConfigOrDie` to `NewFromConfig` we remove the possibility for a panic and instead this will enter an exponential backoff retry before surfacing as a status error

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented abrupt termination during client setup by switching initialization to return errors instead of panicking.
  - Updated resource builder initialization and factory wiring to propagate client-creation failures to callers and reuse initialized clients safely across runs.
- **Tests**
  - Adjusted test builder setup to use the updated error-returning initialization behavior.

### DIFF
--- a/lib/resourcebuilder/resourcebuilder.go
+++ b/lib/resourcebuilder/resourcebuilder.go
@@ -56,22 +56,66 @@ type builder struct {
 	securityClientv1              securityclientv1.SecurityV1Interface
 }
 
-func newBuilder(config *rest.Config, m manifest.Manifest) Interface {
+func newBuilder(config *rest.Config, m manifest.Manifest) (Interface, error) {
+	admissionregistrationClientv1, err := admissionregistrationclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	apiextensionsClientv1, err := apiextensionsclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	apiregistrationClientv1, err := apiregistrationclientv1.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	appsClientv1, err := appsclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	batchClientv1, err := batchclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	configClientv1, err := configclientv1.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	coreClientv1, err := coreclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	imageClientv1, err := imageclientv1.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	operatorsClientv1, err := operatorsclientv1.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+	rbacClientv1, err := rbacclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
+	}
+	securityClientv1, err := securityclientv1.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
 	return &builder{
 		raw: m.Raw,
 
-		admissionregistrationClientv1: admissionregistrationclientv1.NewForConfigOrDie(withProtobuf(config)),
-		apiextensionsClientv1:         apiextensionsclientv1.NewForConfigOrDie(withProtobuf(config)),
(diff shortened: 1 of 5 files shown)


lib/resourcebuilder/resourcebuilder.go
```diff
@@ -118,6 +129,42 @@ func newBuilder(config *rest.Config, m manifest.Manifest) (Interface, error) {
...
+
+func getOrCreateClientSet(config *rest.Config) (*clientSet, error) {
+	if cs, ok := clientSetCache.Load(config); ok {
+		return cs.(*clientSet), nil
+	}
+	cs, err := newClientSet(config)
+	if err != nil {
+		return nil, err
+	}
+	actual, _ := clientSetCache.LoadOrStore(config, cs)
```
wking: we'll want to [clear the map](https://pkg.go.dev/sync#Map.Clear) here too, right?  To avoid a slow memory leak as the REST config slowly rotates over time?

lib/resourcebuilder/resourcebuilder.go
```diff
@@ -56,22 +56,66 @@ type builder struct {
 	securityClientv1              securityclientv1.SecurityV1Interface
 }
 
-func newBuilder(config *rest.Config, m manifest.Manifest) Interface {
+func newBuilder(config *rest.Config, m manifest.Manifest) (Interface, error) {
+	admissionregistrationClientv1, err := admissionregistrationclientv1.NewForConfig(withProtobuf(config))
+	if err != nil {
+		return nil, err
```
wking, line 78: Say there's an error here.  Can the commit message talk through what happens in the CVO then?  Are we using the old, cached client, and hoping that it still works, even though the REST config has evolved?  Are we giving up on reconciling resources for a bit while we try to create fresh clients, and once we have successfully created all these clients, we pick up reconciling again?  Other?

lib/resourcebuilder/resourcebuilder.go
Cali0707: yes, just forgot about that initially - will add

Cali0707: Looking through this, we would not in practice get a memory leak here, as:
1. cache is keyed by the pointer, not the value
2. There are only two rest.Config pointers, which are only ever created at startup

So in practice this map will only ever have two entries

lib/resourcebuilder/resourcebuilder.go
```diff
@@ -56,22 +57,115 @@ type builder struct {
...
+	if cs, ok := clientSetCache.Load(config); ok {
+		return cs.(*clientSet), nil
+	}
+	cs, err := newClientSet(config)
+	if err != nil {
+		return nil, err
+	}
+	actual, _ := clientSetCache.LoadOrStore(config, cs)
+	return actual.(*clientSet), nil
+}
```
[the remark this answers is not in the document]
Cali0707: only 2 rest.Config pointers exist for the process lifetime (created once in pkg/start/start.go). The cache is bounded at 2 entries by construction. Eviction would add complexity with no benefit

lib/resourcebuilder/resourcebuilder.go
Cali0707: Updated, good point. The TLDR is that this is retried through an exponential backoff, retrying making the clients. If that fails to resolve in time, the error propagates to the status

lib/resourcebuilder/resourcebuilder.go
wking: > 1. cache is keyed by the pointer, not the value
> 2. There are only two rest.Config pointers, which are only ever created at startup

This makes me a bit nervous.  Say the value changes (cred rotation, whatever) but the pointer does not.  Will the old client automatically update to use the new creds?  Or do we need to create a new client in that situation?

If we are confident we'll only ever have two entries, maybe we can simplify by dropping the map, and instead tracking the two clients with per-client properties?  That way it's very clear that there are only two slots, and it will also be very clear if someone makes a future change that adds an additional slot.

Cali0707: The two `rest.Config` pointers are created once at startup in `pkg/start/start.go` via `cb.RestConfig(defaultQPS)` and `cb.RestConfig(highQPS)`. They are never recreated, the same two pointers are held by the resourceBuilder in `pkg/cvo/cvo.go` for the lifetime of the process..

Credential rotation is handled transparently by client-go, so we do not need to worry about recreating the rest.Config` or the clients built from it.

That said, I like your suggestion to replace the map with a fixed size array so that the two slot constraint is clear

PR #1427, opened 2026-07-23
OTA-1997: Allow the CVO to use the agentic-skills payload image when creating proposals.

racheljpg:
Hello. This is a draft PR to get some feedback on this PR for [OTA-1997](https://redhat.atlassian.net/browse/OTA-1997). Thanks!

## Summary by CodeRabbit

* **Improvements**
  * Updated the default Agentic Skills image to `registry.redhat.io/openshift-lightspeed/agentic-skills:latest`.
  * Deployment now uses the configured `agentic-skills` image reference for the Cluster Version Operator.
  * Added an `agentic-skills` tag to the ImageStream to align image selection.
* **Bug Fixes**
  * Improved consistency between configured, rendered, and deployed Agentic Skills image references.

### DIFF
--- a/install/0000_00_cluster-version-operator_30_deployment.yaml
+++ b/install/0000_00_cluster-version-operator_30_deployment.yaml
@@ -74,7 +74,7 @@ spec:
         - name: CLUSTER_PROFILE
           value: '{{ .ClusterProfile }}'
         - name: LIGHTSPEED_SKILLS_IMAGE
-          value: "quay.io/openshift/ci:ocp_5.0_agentic-skills"
+          value: '{{ index .Images "agentic-skills" }}'
       # this pod is hostNetwork and uses the internal LB DNS name when possible, which the kubelet also uses.
       # this dnsPolicy allows us to use the same dnsConfig as the kubelet, without access to read it ourselves.
       dnsPolicy: Default
--- a/pkg/agenticrun/controller.go
+++ b/pkg/agenticrun/controller.go
@@ -111,7 +111,7 @@ func DefaultConfig() Config {
 	return Config{
 		Namespace:       envOrDefault("LIGHTSPEED_AGENTIC_RUN_NAMESPACE", "openshift-lightspeed"),
 		PromptConfigMap: envOrDefault("LIGHTSPEED_PROMPT_CONFIGMAP", "cluster-update-advisory-prompt"),
-		SkillsImage:     envOrDefault("LIGHTSPEED_SKILLS_IMAGE", "quay.io/openshift/ci:ocp_5.0_agentic-skills"),
+		SkillsImage:     os.Getenv("LIGHTSPEED_SKILLS_IMAGE"),
 	}
 }
 
@@ -206,6 +206,11 @@ func (c *Controller) Sync(ctx context.Context, key string) error {
 		} else {
 			c.consolePluginEnsured = true
 		}
+	}
+
+	if c.config.SkillsImage == "" {
+		klog.V(i.Normal).Infof("Skipping agentic run creation: LIGHTSPEED_SKILLS_IMAGE is not set")
+		return nil
 	}
 
 	updates, conditionalUpdates, err := c.updatesGetterFunc()
--- a/pkg/agenticrun/controller_test.go
+++ b/pkg/agenticrun/controller_test.go
@@ -103,7 +103,7 @@ Update path: Recommended
 								Tools: agenticrunv1alpha1.ToolsSpec{
 									Skills: []agenticrunv1alpha1.SkillsSource{
 										{
-											Image: "quay.io/openshift/ci:ocp_5.0_agentic-skills",
+											Image: "registry.example.com/agentic-skills:latest",
 											Paths: []string{
 												"/skills/cluster-update/update-advisor",
 												"/skills/cluster-update/product-lifecycle",
@@ -140,6 +140,7 @@ Update path: Recommended
 			}, func() string {
 				return "4.22.1"
 			})
+			c.config.SkillsImage = "registry.example.com/agentic-skills:latest"
 			c.crdAvailableCache = true
 			c.crdLastChecked = time.Now()
 			actual := c.Sync(context.Background(), tt.name)
@@ -786,7 +787,7 @@ Other recommended versions available:
 						Tools: agenticrunv1alpha1.ToolsSpec{
 							Skills: []agenticrunv1alpha1.SkillsSource{
 								{
-									Image: "quay.io/openshift/ci:ocp_5.0_agentic-skills",
+									Image: "registry.example.com/agentic-skills:latest",
 									Paths: []string{
 										"/skills/cluster-update/update-advisor",
 										"/skills/cluster-update/product-lifecycle",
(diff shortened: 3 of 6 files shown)


pkg/agenticrun/controller_test.go
```diff
@@ -103,7 +103,7 @@ Update path: Recommended
 								Tools: agenticrunv1alpha1.ToolsSpec{
 									Skills: []agenticrunv1alpha1.SkillsSource{
 										{
-											Image: "quay.io/openshift/ci:ocp_5.0_agentic-skills",
+											Image: "registry.redhat.io/openshift-lightspeed/agentic-skills:latest",
```
racheljpg: The changing of the image in the tests is maybe unnecessary - what do we think?

pkg/agenticrun/controller.go
```diff
@@ -111,7 +111,7 @@ func DefaultConfig() Config {
 	return Config{
 		Namespace:       envOrDefault("LIGHTSPEED_AGENTIC_RUN_NAMESPACE", "openshift-lightspeed"),
 		PromptConfigMap: envOrDefault("LIGHTSPEED_PROMPT_CONFIGMAP", "cluster-update-advisory-prompt"),
-		SkillsImage:     envOrDefault("LIGHTSPEED_SKILLS_IMAGE", "quay.io/openshift/ci:ocp_5.0_agentic-skills"),
+		SkillsImage:     envOrDefault("LIGHTSPEED_SKILLS_IMAGE", "registry.redhat.io/openshift-lightspeed/agentic-skills:latest"),
```
wking: I don't think we want a default pulling in some random image, because pulling surprising images into production clusters is risky.  I'd rather the controller just log complaints if the environment variable isn't set.

pkg/agenticrun/controller_test.go
wking: For tests, I'd prefer using [example.com](https://www.rfc-editor.org/info/rfc6761/), like `registry.example.com/agentic-skills:latest` or some such.

racheljpg: Done :)

pkg/agenticrun/controller.go
racheljpg: Thanks for pointing this out, I think my eyes glossed over here updating the test images. I've updated it so that if the env var isn't set, SkillsImage will be an empty string, and no AgenticRuns will get created. Is that something more like what you were thinking?

pkg/agenticrun/controller.go
```diff
@@ -242,6 +242,11 @@ func (c *Controller) Sync(ctx context.Context, key string) error {
 		return kutilerrors.NewAggregate(errs)
 	}
 
+	if c.config.SkillsImage == "" {
+		klog.V(i.Normal).Infof("Skipping agentic run creation: LIGHTSPEED_SKILLS_IMAGE is not set")
```
wking: This is a CVO-config issue that's unlikely to be resolved between one poll round and the next (resolving it would take a completely new CVO Pod with a different env. var. configuration).  Can we shift the guard up to run before the `updatesGetterFunc` call, to save us from wasting time doing work that we then throw out if the image is unset?

pkg/agenticrun/controller.go
wking: b94dfafea33 pivots to `os.Getenv("LIGHTSPEED_SKILLS_IMAGE")` (with no default) and a `c.config.SkillsImage == ""` guard.  I'm marking this thread resolved.

pkg/agenticrun/controller_test.go
wking: b94dfafea33 pivots us `registry.example.com/agentic-skills:latest` and leaves us with no lingering `ocp_5.0_agentic` references.  I'm marking this thread resolved.

pkg/agenticrun/controller.go
racheljpg: Moved now!

wking: 47bfd33038aab shifts this guard, marking the thread resolved.

wking (approved):
/lgtm

install/0000_00_cluster-version-operator_30_deployment.yaml
```diff
@@ -74,7 +74,7 @@ spec:
         - name: CLUSTER_PROFILE
           value: '{{ .ClusterProfile }}'
         - name: LIGHTSPEED_SKILLS_IMAGE
-          value: "quay.io/openshift/ci:ocp_5.0_agentic-skills"
+          value: '{{ index .Images "agentic-skills" }}'
```
wking, line 77: [upgrade-into-change](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change/2080703201174097920) -> [must-gather artifacts][1]:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change/2080703201174097920/artifacts/e2e-agnostic-ovn-upgrade-into-change/gather-extra/artifacts/deployments.json | jq -c '.items[] | select(.metadata.name == "cluster-version-operator").spec.template.spec.containers[0].env[]' | grep LIGHTSPEED_SKILLS_IMAGE
{"name":"LIGHTSPEED_SKILLS_IMAGE","value":"quay-proxy.ci.openshift.org/openshift/ci@sha256:9d42beab76b788795db5b6ebe365664b13a0d6322691a3bef2ad3f6abc4a2eb0"}
```

So that looks good to me.

[1]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-upgrade-into-change/2080703201174097920/artifacts/e2e-agnostic-ovn-upgrade-into-change/gather-extra/artifacts/

wking:
/verified by https://github.com/openshift/cluster-version-operator/pull/1427#discussion_r3647737409

wking:
[Bootstrap `AuthorizationPermissionMismatch`][1] [are][2] [all][3] unrelated to this pull, and while they leave us blind to how those suites would have run, I'm confident from the  passing CI, and how decoupled this change is from tested functionality, that this change won't have broken them.

/override ci/prow/e2e-agnostic-ovn-techpreview-serial-2of3
/override ci/prow/e2e-agnostic-ovn-techpreview-serial-3of3

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn/2080703192785489920
[2]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial-2of3/2080703196140933120
[3]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-version-operator/1427/pull-ci-openshift-cluster-version-operator-main-e2e-agnostic-ovn-techpreview-serial-3of3/2080703198657515520

PR #1466, opened 2026-08-31
OKD-194: Add OKD cincinnati as the default update service for OKD

jatinsu:
This PR setups up the newly created OKD cincinnati as the default update service for OKD. After this PR has been merged, https://github.com/openshift/installer/pull/10834 will be merged to update the installer repo

## Summary by CodeRabbit

- **New Features**
  - Automatically selects the OKD update service for OKD releases when no custom update source is configured.
  - Continues using the standard update service for other release types.
  - Improves reuse of cached update data across default update services.
  - Adds an informational alert when a cluster uses a legacy update service.

- **Bug Fixes**
  - More reliably identifies OKD pre-release versions.
  - Handles invalid version strings safely without classifying them as OKD.
  - Updates legacy-service status when the configured update source changes.

### DIFF
--- a/pkg/cvo/availableupdates.go
+++ b/pkg/cvo/availableupdates.go
@@ -32,6 +32,7 @@ import (
 const noArchitecture string = "NoArchitecture"
 const noChannel string = "NoChannel"
 const defaultUpdateService string = "https://api.openshift.com/api/upgrades_info/v1/graph"
+const defaultOKDUpdateService string = "https://updates.okd.io/api/updates/graph"
 
 // syncAvailableUpdates attempts to retrieve the latest updates and update the status of the ClusterVersion
 // object. It will set the RetrievedUpdates condition. Updates are only checked if it has been more than
@@ -48,8 +49,13 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 		updateServiceSource = "ClusterVersion spec.upstream"
 	} else {
 		usedDefaultUpdateService = true
-		updateService = defaultUpdateService
-		updateServiceSource = "the operator's default update service"
+		if isOKDRelease(optr.release.Version) {
+			updateService = defaultOKDUpdateService
+			updateServiceSource = "the operator's default OKD update service"
+		} else {
+			updateService = defaultUpdateService
+			updateServiceSource = "the operator's default update service"
+		}
 	}
 
 	channel := config.Spec.Channel
@@ -85,7 +91,7 @@ func (optr *Operator) syncAvailableUpdates(ctx context.Context, config *configv1
 	} else if !optrAvailableUpdates.RecentlyAttempted(optr.minimumUpdateCheckInterval) {
 		klog.V(2).Infof("Retrieving available updates again, because more than %s has elapsed since last attempt at %s", optr.minimumUpdateCheckInterval, optrAvailableUpdates.LastAttempt.Format(time.RFC3339))
 		preserveCacheOnFailure = true
-	} else if updateService == optrAvailableUpdates.UpdateService || (updateService == defaultUpdateService && optrAvailableUpdates.UpdateService == "") {
+	} else if updateService == optrAvailableUpdates.UpdateService || (usedDefaultUpdateService && optrAvailableUpdates.UpdateService == "") {
 		needsConditionalUpdateEval := false
 		preserveCacheOnFailure = true
 		for _, conditionalUpdate := range optrAvailableUpdates.ConditionalUpdates {
@@ -450,6 +456,25 @@ func loadRiskVersions(conditionalUpdates []configv1.ConditionalUpdate) map[strin
 		return nil
 	}
 	return riskVersions
+}
+
+// isOKDRelease returns true when the given release version string identifies an
+// OKD release. OKD releases embed an "okd" identifier in the semantic version
+// pre-release segment (for example "4.19.0-0.okd-2024-01-06-084517" or
+// "4.22.0-0.okd-scos-nightly-2025-..."), while OCP releases do not embed this
+// identifier. It is used to select the appropriate default update service.
+func isOKDRelease(version string) bool {
+	v, err := semver.Parse(version)
+	if err != nil {
+		klog.V(2).Infof("Unable to parse release version %q to determine whether this is an OKD cluster: %v", version, err)
+		return false
+	}
+	for _, pre := range v.Pre {
+		if strings.HasPrefix(pre.VersionStr, "okd") {
+			return true
+		}
+	}
+	return false
 }
 
 func (optr *Operator) getDesiredArchitecture(update *configv1.Update) string {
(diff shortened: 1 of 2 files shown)


pkg/cvo/availableupdates.go
```diff
@@ -452,6 +458,25 @@ func loadRiskVersions(conditionalUpdates []configv1.ConditionalUpdate) map[strin
 	return riskVersions
 }
 
+// isOKDRelease returns true when the given release version string identifies an
+// OKD release. OKD releases embed an "okd" identifier in the semantic version
+// pre-release segment (for example "4.19.0-0.okd-2024-01-06-084517" or
+// "4.22.0-0.okd-scos-nightly-2025-..."), while OCP releases
+// identifier. It is used to select the appropriate default update service.
+func isOKDRelease(version string) bool {
+	v, err := semver.Parse(version)
```
Prashanth684, line 467: along with this check i would also make sure this operator itself has been built for OKD - we pass in `TAGS=SCOS` to build OKD. we can check if that is present like [other](https://github.com/openshift/installer/blob/c8d299965fc1527648ce7aeb368c198a7b28e85d/pkg/types/installconfig_scos.go#L2) components have

pkg/cvo/availableupdates.go
```diff
@@ -32,6 +32,7 @@ import (
 const noArchitecture string = "NoArchitecture"
 const noChannel string = "NoChannel"
 const defaultUpdateService string = "https://api.openshift.com/api/upgrades_info/v1/graph"
+const defaultOKDUpdateService string = "https://updates.okd.io/api/updates/v1/graph"
```
wking: I'd recommed not using the `/v1/` URI-based versioning.  See openshift/cincinnati#545 teaching Cincinnati to use media-type based versioning.  [OTA-529](https://redhat.atlassian.net/browse/OTA-529) is in the backlog for us updating the OCP handling here in the CVO to use that new approach.  But it's tricky because maybe folks have very specific firewall configs that would break cluster access if we pivoted, and we don't have a v2 expected soon to make the pivot worthwhile.  But for OKD, this is all a net-new thing, so you could use the media-type versioning from the start, and not have to worry about a URI-versioning -> media-type-versioning transition.

pkg/cvo/availableupdates.go
wking: With that tag in place, why check the release version at all?  Can't you just switch on the tag to figure out which default URI to use?

jatinsu: This tag was never put in place for the CVO. This was an intentional decision since we didn't want to diverge OKD CVO too much from OCP CVO, hence why the release version check is put in place

pkg/cvo/legacy_update_service.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/cluster-version-operator/lib/resourcemerge"
+	"github.com/openshift/cluster-version-operator/pkg/internal"
+)
+
+const legacyOKDUpdateService = "https://amd64.origin.releases.ci.openshift.org/graph"
+
+func setLegacyUpdateServiceCondition(status *configv1.ClusterVersionStatus, upstream configv1.URL) {
```
Prashanth684: rather than calling it set, can we call it `UpdateOKDLegacyUpdateServiceCondition`?

pkg/internal/constants.go
```diff
@@ -70,6 +70,10 @@ const (
 	// enabled capabilities.
 	ImplicitlyEnabledCapabilities configv1.ClusterStatusConditionType = "ImplicitlyEnabledCapabilities"
 
+	// ClusterVersionLegacyUpdateService is True when spec.upstream selects a
+	// legacy update service which the cluster administrator should replace or clear.
+	ClusterVersionLegacyUpdateService configv1.ClusterStatusConditionType = "LegacyUpdateService"
```
Prashanth684: have OKD in the wording so that it is clear that this is OKD specific

jatinsu:
/hold 
still needs review from ota team

sarthakpurohit:
@jatinsu got it!

wking:
No need to [hold waiting on approvers][1].  The Prow configuration is [already waiting on their `approved` label][2], and this won't merge without it, even if we lift the hold.

/hold cancel

[1]: https://github.com/openshift/cluster-version-operator/pull/1466#issuecomment-5623288956
[2]: https://github.com/openshift/cluster-version-operator/pull/1466#issuecomment-5623280159

pkg/cvo/availableupdates.go
wking: Can we put it in place for the CVO?  I understand "we don't want to diverge", but now that we've chosen to diverge on this point, a build-time, compiled in divergence seems more reliable than an "inspect my name and guess in dynamic Go" divergence.

install/0000_90_cluster-version-operator_02_servicemonitor.yaml
```diff
@@ -74,6 +74,17 @@ spec:
         sum by (channel, namespace, upstream) (cluster_version_available_updates) > 0
       labels:
         severity: info
+    - alert: ClusterVersionOperatorLegacyUpdateService
+      annotations:
+        summary: ClusterVersion is configured to use a legacy update service.
+        description: ClusterVersion spec.upstream points to a legacy update service. At an appropriate time, clear spec.upstream to use the CVO's default update service, or configure another update service.
```
wking, line 80: nit: we use "CVO" familiarly, but this alert might be the first time a responding admin hears about this controller.  Can we spell out "the cluster-version operator's" here, to give them more of a handle to onboard with?  Or maybe drop "CVO's" and just say "the default update service" or "the default OKD update service" or whatever you think seems most accessible to a hypothetical new OKD cluster-admin.

pkg/cvo/availableupdates.go
jatinsu: @Prashanth684 WDYT? We already attach the tag [during building](https://github.com/openshift/release/blob/6dbdd9ca6a9e246246de3451e05d63a0d5c405e2/ci-operator/config/openshift/cluster-version-operator/openshift-cluster-version-operator-main__okd-scos.yaml#L17-L19), so would only need a change here.

Prashanth684: > compiled in divergence seems more reliable than an "inspect my name and guess in dynamic Go" divergence.

Yes agree with this - our tests use the `okd-scos` string check to identify clusters, but it's necessary there (well, we could claim if the test binary is built for OKD, it must be running on an OKD cluster), whereas here, the compiled in variable asserting that it is OKD is much better (the assumption being if it is built for OKD, you are running on an OKD cluster)

jatinsu:
/verified by @jatinsu 
image

wking (approved):
/lgtm

PR #1467, opened 2026-09-04
OCPBUGS-8252: pkg/cvo report unavailable multi-arch transitions

ericahinkleRH:
## Summary

- When a single-architecture cluster requests `--to-multi-arch` while CVO has `RetrievedUpdates=False` and no usable multi-architecture payload in `status.availableUpdates`, CVO now sets `ReleaseAccepted=False` with reason `MultiArchTransitionUnavailable` and a message explaining the transition cannot start.
- The existing current-release fallback is preserved so the worker never receives an empty or fabricated payload.
- All existing behavior for nil desired updates, explicit images, ordinary version lookups, successful single-to-multi migrations, and already-multi clusters is unchanged.

## Changes

| File | What |
|------|------|
| `pkg/cvo/updatepayload.go` | `findUpdateFromConfig` returns a typed `*payload.UpdateError` when a single-to-multi transition cannot resolve the current release from `availableUpdates` |
| `pkg/cvo/cvo.go` | `Operator.sync` carries the selection error through `LoadPayloadStatus` so `setReleaseAcceptedCondition` produces `ReleaseAccepted=False` |
| `pkg/cvo/updatepayload_test.go` | 8 table-driven tests for `findUpdateFromConfig` covering all desired-update paths |
| `pkg/cvo/cvo_test.go` | Controller-level regression test proving the fallback, condition, and spec preservation |

## Test plan

- [x] `go test ./pkg/cvo/ -count=1` — all tests pass
- [x] `make format` — no formatting changes
- [x] `make build` — binary builds cleanly
- [x] New `Test_findUpdateFromConfig` covers: nil desired, explicit image, normal found/missing, single-to-multi success/failure, already-multi paths
- [x] New `TestOperator_sync_multiArchTransitionUnavailable` verifies: worker gets current release fallback, `ReleaseAccepted=False` with correct reason/message, `Failing` not set to True, `spec.desiredUpdate` preserved

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-8252

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling when a requested release or architecture transition is unavailable.
  - The cluster now falls back to its current release while preserving the requested update.
  - Status reports the update as not accepted with a clear `DesiredReleaseUnavailable` reason and message, without incorrectly marking the cluster as failing.

- **Tests**
  - Added coverage for unavailable releases, historical versions, and single- to multi-architecture transitions.

### DIFF
--- a/pkg/cvo/cvo.go
+++ b/pkg/cvo/cvo.go
@@ -837,7 +837,7 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
 	config := validation.ClearInvalidFields(original, errs)
 
 	// identify the desired next version
-	desired, found := findUpdateFromConfig(config, optr.release.Architecture)
+	desired, found, selectionErr := findUpdateFromConfig(config, optr.release.Architecture)
 	initialized := optr.configSync.Initialized()
 	if found && initialized {
 		klog.V(2).Infof("Desired version from spec is %#v after initialization", desired)
@@ -881,6 +881,19 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
 
 	// inform the config sync loop about our desired state
 	status := optr.configSync.Update(ctx, config.Generation, desired, config, state, optr.getEnabledFeatureGates())
+
+	// surface the multi-arch transition failure through the existing status/condition path
+	if selectionErr != nil && initialized {
+		updateErr, _ := selectionErr.(*payload.UpdateError)
+		if updateErr != nil {
+			status.loadPayloadStatus = LoadPayloadStatus{
+				Step:    updateErr.Reason,
+				Message: updateErr.Message,
+				Failure: updateErr,
+				Update:  *config.Spec.DesiredUpdate,
+			}
+		}
+	}
 
 	// write cluster version status
 	return optr.syncStatus(ctx, original, config, status, errs)
--- a/pkg/cvo/updatepayload.go
+++ b/pkg/cvo/updatepayload.go
@@ -438,22 +438,31 @@ func (r *payloadRetriever) prunePods(ctx context.Context) error {
 // the desired architecture is changed to multi it resolves the payload using the current
 // version since that's the only valid available update. Otherwise it attempts to resolve
 // the payload using the specified desired version.
-func findUpdateFromConfig(config *configv1.ClusterVersion, currentArch configv1.ClusterVersionArchitecture) (configv1.Update, bool) {
+func findUpdateFromConfig(config *configv1.ClusterVersion, currentArch configv1.ClusterVersionArchitecture) (configv1.Update, bool, error) {
 	update := config.Spec.DesiredUpdate
 	if update == nil {
-		return configv1.Update{}, false
+		return configv1.Update{}, false, nil
 	}
 	if len(update.Image) == 0 {
 		version := update.Version
 
+		isMultiArchTransition := update.Architecture == configv1.ClusterVersionArchitectureMulti &&
+			currentArch != configv1.ClusterVersionArchitectureMulti
+
 		// Architecture changed to multi so only valid update is the multi arch version of current version
-		if update.Architecture == configv1.ClusterVersionArchitectureMulti &&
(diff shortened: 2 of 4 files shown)


pkg/cvo/updatepayload.go
```diff
@@ -438,22 +438,31 @@ func (r *payloadRetriever) prunePods(ctx context.Context) error {
...
+
 		// Architecture changed to multi so only valid update is the multi arch version of current version
-		if update.Architecture == configv1.ClusterVersionArchitectureMulti &&
-			currentArch != configv1.ClusterVersionArchitectureMulti {
+		if isMultiArchTransition {
 			version = config.Status.Desired.Version
 		}
-		return findUpdateFromConfigVersion(config, version, update.Force)
+		resolved, found := findUpdateFromConfigVersion(config, version, update.Force)
+		if !found && isMultiArchTransition {
```
wking: I'm not clear on why we need handling to be multi-arch specific here.  Aren't all `!found` situations similarly problematic, and worth reporting clearly in `status`?

ericahinkleRH: Fair point, I generalized it. The error now covers every `!found` case instead of just the multi-arch one. A nil desired update still stays quiet, since that's the normal no-op.

Writing the tests turned up a second case I hadn't thought about. A version that only exists in `status.history` passes `countPayloadsForVersion`, so validation lets it through, but it still fails to resolve against `availableUpdates`. That one and the multi-arch case both have tests now.

On naming, the reason you'll see on `ReleaseAccepted` is `ResolveRelease`, since every other reason on that condition names the stage that failed rather than the error itself. `DesiredReleaseUnavailable` is still the reason carried on the selection error.

ericahinkleRH:
@coderabbitai review

pkg/cvo/cvo.go
```diff
@@ -882,6 +882,21 @@ func (optr *Operator) sync(ctx context.Context, key string) error {
...
+	// describe the payload the worker actually loaded, which status reporting relies on.
+	// Step names the stage that failed, like the values the sync worker reports, because
+	// setReleaseAcceptedCondition uses it as the ReleaseAccepted reason. The error keeps
+	// its own reason for callers that report errors rather than stages.
+	if selectionErr != nil && initialized {
+		updateErr, _ := selectionErr.(*payload.UpdateError)
+		if updateErr != nil {
+			status.loadPayloadStatus.Step = "ResolveRelease"
+			status.loadPayloadStatus.Message = updateErr.Message
+			status.loadPayloadStatus.Failure = updateErr
```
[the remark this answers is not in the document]
ericahinkleRH: Good catch — fixed, though not with the suggested patch.

Confirmed the path. `SyncWorker.Update` loads payloads synchronously (`sync_worker.go:567` and `:601`) and returns with `loadPayloadStatus.Failure` set when that load fails (`:612`). By then `desired` has already been replaced by the current release, so that really can be a failure to load the fallback rather than a stale error about some other target. And `updateLoadStatus` deliberately leaves the top-level `Failure` alone (`sync_worker.go:965`), so `ReleaseAccepted` is the only condition carrying it — overwriting it would hide the failure completely.

Two problems with the suggested diff, so I went a different way:

- `apierrors` in `cvo.go` is `k8s.io/apimachinery/pkg/api/errors`, which has no `NewAggregate` (that is `util/errors`), so it would not compile as written.
- Aggregating discards the `*payload.UpdateError` and its `DesiredReleaseUnavailable` reason.

Instead the selection error simply yields to a failure the worker already reported:

```go
if selectionErr != nil && initialized && status.loadPayloadStatus.Failure == nil {
```

A failure to load the release the cluster is actually running is the more actionable problem, and the selection error is not lost: the worker keeps reconciling the fallback, so once that load succeeds `ReleaseAccepted` picks up `ResolveRelease` on the next sync. Either way the condition is `False`, so `ClusterReleaseNotAccepted` still fires.

Covered by a new case in `TestOperator_sync_desiredReleaseUnavailable`, `payload load failure is not replaced by the selection error`.

ericahinkleRH:
@coderabbitai review

